PR #86, opened 2015-12-17
Rather than computing stock in the tpl, let's use the pre-computed value

borkweb:
We have an `original_stock()` method. Let's use that instead.

Related: https://github.com/moderntribe/event-tickets-plus/pull/51

See: https://central.tri.be/issues/42355


bordoni:
:100:

PR #94, opened 2016-01-05
Message: Tickets are no longer or not yet on sale

masteradhoc:
My PullRequest comes from here:
https://theeventscalendar.com/support/forums/topic/no-message-end-date-for-sales-reached/

I think the user should be informed when tickets are no longer on sale. I would love to also code this for "woocommerce" and "easy digital downloads". May you allow me to see those files? i'll add the code asap.

Thanks

PR #95, opened 2016-01-06
Add Message to inform user that there is no ticket available

masteradhoc:
Add Message "There are no Tickets available for this event!"

### DIFF
--- a/src/views/tickets/rsvp.php
+++ b/src/views/tickets/rsvp.php
@@ -115,4 +115,8 @@ ob_start();
 $content = ob_get_clean();
 if ( $is_there_any_product ) {
 	echo $content;
+} else {
+	?>
+	<span class="tickets_nostock"><?php esc_html_e( 'There are no Tickets available for this event!', 'event-tickets' ); ?></span>
+	<?php
 }


bordoni:
Looks good. :+1:

barryhughes:
Hi @masteradhoc,

Thanks for this PR!

I'm just a little worried that the message would mislead/confuse users in situations where tickets _are_ available, but just aren't available "for sale" yet. Thinking of [a related PR you submitted](https://github.com/moderntribe/event-tickets/pull/94) if both are merged then that might mean users get two messages which basically restate the same information.

I definitely see that there is still a use case for this message though - ie, when all tickets have sold out - but we'd probably need to tighten it up just a bit to ensure it only renders in that situation and not when tickets are in stock but just not on sale yet.

masteradhoc:
Thanks @barryhughes ! May you point me to the right direction, ill update the PR if you want. Or will you take a look at the case? I think it has to be added asap because user may be more confused now ;)

borkweb:
@masteradhoc - Like @barryhughes said, thanks for this and the other PR (#94)!

There's definitely some merit to the ideas behind both of those PRs that we'd love to comment on in a single location. Since both PRs deal with messaging for tickets - it _feels_ justifiable (in my brain) to combine both of the changesets. :smile: 

Would it be too troublesome to put the changes from this PR and #94 into a single Pull Request and then PR it against our `develop` branch rather than `release/122`? Thank you!

masteradhoc:
Thanks @borkweb for the message. I'll do that tomorrow. Alright?

borkweb:
That sounds great to me! Thank you, sir!

PR #110, opened 2016-02-04
Prevent RSVP recursion on RSVPing for an event with custom meta

borkweb:
Turns out this was causing the following reported issue: https://central.tri.be/issues/43666

Related: https://github.com/moderntribe/event-tickets-plus/pull/82


barryhughes:
:ear_of_rice:

PR #111, opened 2016-02-08
Don't swap out the name attribute on ticket meta fields

borkweb:
We don't have to worry about duplicate ticket meta fields when dealing with custom meta.

See: https://central.tri.be/issues/43780

Also resolving the issue where custom meta fields are cleared out if fields are altered after the initial save (this was cause by not properly resetting the advanced fields).

See: https://central.tri.be/issues/43776


barryhughes:
...:rabbit: ...
o-:necktie:-o
.`_/\_`.

PR #112, opened 2016-02-08
Fix fatal that prevents the email modal from functioning on attendees

borkweb:
See: https://central.tri.be/issues/43781


barryhughes:
Short and sweet :candy:

PR #113, opened 2016-02-08
Basis for global stock

barryhughes:
Adds facilities to work with global stock (but doesn't really swing into action unless a provider declares global stock support).

### DIFF
--- a/src/Tribe/Ticket_Object.php
+++ b/src/Tribe/Ticket_Object.php
@@ -80,6 +80,22 @@ if ( ! class_exists( 'Tribe__Tickets__Ticket_Object' ) ) {
 		protected $stock = 0;
 
 		/**
+		 * The mode of stock handling to be used for the ticket when global stock
+		 * is enabled for the event.
+		 *
+		 * @var string
+		 */
+		protected $global_stock_mode = Tribe__Tickets__Global_Stock::GLOBAL_STOCK_MODE;
+
+		/**
+		 * The maximum permitted number of sales for this ticket when global stock
+		 * is enabled for the event and CAPPED_STOCK_MODE is in effect.
+		 *
+		 * @var int
+		 */
+		protected $global_stock_cap = 0;
+
+		/**
 		 * Amount of tickets of this kind sold
 		 * Use $this->qty_sold( value ) to set manage and get the value
 		 *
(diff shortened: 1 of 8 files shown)


src/Tribe/Ticket_Object.php
```diff
@@ -272,6 +298,43 @@ public function stock( $value = null ) {
 		}
 
 		/**
+		 * Sets of gets the current global stock mode in effect for the ticket.
```
borkweb: `Sets or gets` vs `Sets of gets`?

src/Tribe/Ticket_Object.php
```diff
@@ -272,6 +298,43 @@ public function stock( $value = null ) {
...
+		 *
+		 * @param string $mode
+		 *
+		 * @return string
+		 */
+		public function global_stock_mode( $mode = null ) {
+			if ( is_string( $mode ) ) {
+				$this->global_stock_mode = $mode;
+			} elseif ( null === $mode ) {
+				return $this->global_stock_mode;
```
borkweb: Is there any reason to avoid always returning this regardless of what the value of `$mode` is?

src/Tribe/Ticket_Object.php
```diff
@@ -272,6 +298,43 @@ public function stock( $value = null ) {
...
+		 *
+		 * @param int $cap
+		 *
+		 * @return int
+		 */
+		public function global_stock_cap( $cap = null ) {
+			if ( is_numeric( $cap ) ) {
+				$this->global_stock_cap = (int) $cap;
+			} elseif ( null === $cap ) {
+				return (int) $this->global_stock_cap;
```
borkweb: Essentially the same question as [the one](https://github.com/moderntribe/event-tickets/pull/113/files#r52367872) on `global_stock_mode()`.

src/resources/js/tickets.js
```diff
@@ -160,8 +163,57 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
+				var $global_stock_fields = $provider_fields.filter( '.global-stock-mode' );
+				var $sales_cap_field     = $global_stock_fields.filter( '.sales-cap-field' );
+
+				var mode     = $( '#ticket_global_stock' ).val();
+				var enabled  = global_stock_enabled();
+
+				// Show or hide global (and normal, "per-ticket") stock settings as appropriate
+				$global_stock_level.toggle(enabled);
+				$global_stock_fields.toggle(global_stock_enabled());
+				$normal_stock_field.toggle(!enabled);
```
borkweb: These method calls could use spaces inside the parens and after the `!`.

src/resources/js/frontend-ticket-form.js
```diff
@@ -0,0 +1,258 @@
+/**
+ * Provides global stock handling for frontend ticket forms.
+ *
+ * @var object tribe_tickets_stock_data
+ */
+jQuery( function( $ ) {
```
borkweb: Since this is a new file, should we take the opportunity to expose the functions and vars globally (adding them as properties of a globally accessible object)? That way we can do a bit more testing in the console and allow plugin authors to override/modify the functionality provided by this file.

The [RSVP js file](https://github.com/moderntribe/event-tickets/blob/release/122/src/resources/js/rsvp.js) is an example of what I mean.

barryhughes:
Good thoughts and catches - changes en route.

barryhughes:
Updated!

src/resources/js/frontend-ticket-form.js
```diff
@@ -0,0 +1,260 @@
+var tribe_tickets_ticket_form = {};
+
+/**
+ * Provides global stock handling for frontend ticket forms.
+ *
+ * @var object tribe_tickets_stock_data
+ */
+( function( $, my ) {
+	var $tickets_lists   = $( '.tribe-events-tickets' );
+	var $quantity_fields = $tickets_lists.find( '.quantity' ).find( 'input' );
```
borkweb: Wrap these two lines (along with [this one](https://github.com/moderntribe/event-tickets/pull/113/files#diff-0f98cb8f6ef0e2fe89036c6177442ce9R258)) in a `my.init()` to make sure they aren't hunted for in the DOM before they exist on the page (if someone sets a this JS file as a dependency to another that loads in the header or something).

Then you can call it on document ready: `$( function() { my.init(); } );`

barryhughes: Makes sense

barryhughes: Done.

borkweb:
Woo woo! :bullettrain_front:

PR #114, opened 2016-02-10
Feature/18656 global stock js fixes

barryhughes:
Tweaks our handling of capped inventories (if tickets A and B both draw on a global stock of 200 units and B is capped at 50, the available stock for B should reduce when A has a quantity above 150). Also ensures we compile this new asset.

https://central.tri.be/issues/18656


src/resources/js/frontend-ticket-form.js
```diff
@@ -109,8 +109,9 @@ var tribe_tickets_ticket_form = {};
 			}
 
 			if ( 'capped' === ticket.mode ) {
-				remaining = ticket.cap - $( '[data-product-id=' + ticket_id + ']' ).find( 'input' ).val();
-				$tickets_lists.find( '.available-stock[data-product-id=' + ticket_id + ']').html( remaining );
+				var effective_cap = Math.min( remaining, ticket.cap );
+				var remaining_capped_stock = effective_cap - $( '[data-product-id=' + ticket_id + ']' ).find( 'input' ).val();
```
borkweb: With some providers, this _can_ find a hidden input value that contains an id attributed. I'd suggest the following:

```js
$( '[data-product-id=' + ticket_id + ']' ).find( 'input:not([type="hidden"])' ).val();
```

borkweb: Oh, wait. Ignore me. I'm thinking of the `.quantity` element.

borkweb:
:turtle:

PR #115, opened 2016-02-11
Help stop global stock setting changes from being lost on unload | 18656

barryhughes:
Global stock settings apply to the whole event (a bit like the ticket header img). Since updates for these things currently don't function via ajax, whereas tickets do, we need to warn users if changes to global stock settings are in risk of being lost should they navigate away too early.


borkweb:
:boom:

PR #117, opened 2016-02-12
Tickets should use their own 'independent' stock by default

barryhughes:
Better defaults for global stock data.


borkweb:
:ear_of_rice:

PR #120, opened 2016-02-15
Fix/43020 ticket stock mods

barryhughes:
Simplifies stock reporting: we now report the actual stock remaining rather than performing a calculation of sold + remaining, etc, which can fall out of sync in some conditions [(see here for an example)](https://central.tri.be/issues/43020#note-5). Also updates global stock facilities to jive with these changes.

[C#43020](https://central.tri.be/issues/43020) | [C#18656](https://central.tri.be/issues/18656)


src/template-tags/tickets.php
```diff
@@ -194,24 +194,38 @@ function tribe_events_ticket_is_on_sale( Tribe__Tickets__Ticket_Object $ticket )
...
-		if ( ! $stock ) {
-			$message = sprintf( esc_html__( 'Sold all %1$d %2$s', 'event-tickets' ), esc_html( $sold ), esc_html( $pending_message ) );
-		} else {
-			$message = sprintf( esc_html__( 'Sold %1$d of %2$d %3$s', 'event-tickets' ), esc_html( $sold ), esc_html( $stock ), esc_html( $pending_message ) );
+		// If we do have a fixed stock then we can provide more information
+		else {
+			$cancelled_count = empty( $cancelled ) ? '' : esc_html( sprintf(
+				_x( ' %1$s %2$d', 'ticket stock message (cancelled stock)', 'event-tickets' ),
+				_x( 'cancelled:', 'ticket stock message (cancelled stock)', 'event-tickets' ),
+				(int) $cancelled
```
borkweb: I'm guessing I'm not seeing something obvious, but out of curiosity why do this:

```php
sprintf(
  _x( ' %1$s %2$d', 'ticket stock message (cancelled stock)', 'event-tickets' ),
  _x( 'cancelled:', 'ticket stock message (cancelled stock)', 'event-tickets' ),
  (int) $cancelled
)
```

Rather than this:

```php
sprintf(
  _x( ' cancelled: %1$d', 'ticket stock message (cancelled stock)', 'event-tickets' ),
  (int) $cancelled
)
```

barryhughes: No you're not missing anything - that was re-worked more than once and went all awry in the process I guess; will simplify like you suggested :+1:

barryhughes: Modified for the above and for the same issue re `$pending_count` a couple lines down.

borkweb:
:beauty, eh: :lumberjack: :plaid:

barryhughes:
Those emoji are banned in several provinces.

PR #124, opened 2016-02-17
Only show the confirmation field error message when it is appropriate

borkweb:
Before, it was showing the message any time any other messages were displayed. Which sucked.

See: https://central.tri.be/issues/44153


bordoni:
:100:

PR #126, opened 2016-02-19
Global sales count helper | 43020

barryhughes:
Implementing a count of global stock sales - simple helper function plus linking it to the global stock field in the event meta box.

[C#43020](https://central.tri.be/issues/43020)


borkweb:
:+1:

PR #127, opened 2016-02-23
Add an Action to add fields on Tickets Metabox

bordoni:
_Ref:_ [C#40065](https://central.tri.be/issues/40065)


src/Tribe/RSVP.php
```diff
@@ -61,6 +61,13 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 	public $security_code = '_tribe_rsvp_security_code';
 
 	/**
+	 * Meta key that if this attendee wants to show on the attendee list
+	 *
+	 * @var string
+	 */
+	public $optout = '_tribe_rsvp_attendee_optout';
```
barryhughes: We've got two separate words here - maybe add in an underscore them so we have `$opt_out` instead of `$optout`? Though the meaning is obvious now I'm writing this comment, it did give me a moment of confusion on first reading.

barryhughes: ... Should it be a constant?

barryhughes: Sorry one more - if we change optout to opt_out here, let's do it everywhere! (Unless there's a back compat concern of course)

src/admin-views/meta-box.php
```diff
@@ -90,6 +90,15 @@
 			</td>
 		</tr>
 	<?php endif; ?>
+
+	<?php
+	/**
+	 * Fired to allow for the insertion of additional content into the ticket admin form before the tickets listing
+	 *
+	 * @var Post ID
```
barryhughes: Probably should use `@param` instead of `$var` here

src/Tribe/RSVP.php
bordoni: It won't cause any Back compatibility for users, but on QA anyone with data, will loose it, might make the QA a little bit funky.
I have no reasoning to have it be `optout` together, besides that whenever I see Opt-In in variables I normally see it as a `optin`. We could change it

barryhughes: Hmm ok, it _is_ a bit subjective i guess. Do as you think best, but I do think this would be better as a class constant than a public property :pray:

bordoni:
Can you re-review it?

barryhughes:
:+1:

PR #132, opened 2016-03-09
Twenty Fifteen compatibility (frontend ticket form)

barryhughes:
Fixes minor compatibility issue with Twenty Fifteen (re the frontend ticket form quantity input), also rolls in a spotfix that ensures grunt actually minifies our CSS assets.

https://central.tri.be/issues/44567


bordoni:
:+1:

PR #136, opened 2016-03-11
Fix capability check problems #135

### LINKED ISSUE #135: Capability problems on CPT Tickets
Hi,

I’m working on a site where I need to sell tickets for a Custom Post Type (CPT).
I set up the plugin but it’s not saving the tickets because of a php error.

I’m getting an error about:

> Notice: Undefined offset: 0 in /srv/www/domain.com/current/web/wp/wp-includes/capabilities.php on line 115

This happens because `Tribe__Tickets__Tickets->has_permission( )` is checking the capability in a bad way. (I guess)

```
    $cap = "edit_{$post->post_type}";
    if ( 'post' === $post->post_type || 'page' === $post->post_type ) {
        $cap .= 's';
    }
```

So my post type has registered with a singular name `course` (how it should be by default) and you check the `edit_{$post->post_type}` which is `edit_course` in this case. But to check this permission (`edit_post => edit_course`) you have to add another argument which is the post id.

[http://wordpress.stackexchange.com/a/84205/81135](http://wordpress.stackexchange.com/a/84205/81135)

I would suggest to check the capability based on the post type object.

bokorir:
Fixes #135

### DIFF
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -320,12 +320,7 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 				$post = get_post( $post );
 			}
 
-			$cap = "edit_{$post->post_type}";
-			if ( 'post' === $post->post_type || 'page' === $post->post_type ) {
-				$cap .= 's';
-			}
-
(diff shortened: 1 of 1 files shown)


barryhughes:
Ticketed: [C#45609](https://central.tri.be/issues/45609)

barryhughes:
Thanks for submitting this pull request, @bokorir - looks great! We'll see about merging it in shortly.

PR #138, opened 2016-03-14
Trigger selection of a provider as soon as advanced fields added

borkweb:
When editing a ticket, we were not triggering the selection of the ticket provider early enough for the advanced ticket fields to be initialized.

See: https://central.tri.be/issues/44794


barryhughes:
:+1:

PR #139, opened 2016-03-15
Maintain the original checkin() method signature

borkweb:
Rolling back the signature change that we attempted in the last MR in favor of an uglier backward compatible change.

Related: https://github.com/moderntribe/event-tickets-plus/pull/130

See: https://central.tri.be/issues/44831


barryhughes:
:+1:

PR #141, opened 2016-03-16
Messages to inform user of the availability of Tickets

masteradhoc:
My PullRequest comes originally comes from here:
https://theeventscalendar.com/support/forums/topic/no-message-end-date-for-sales-reached/

I think the user should be informed when tickets are no longer or not yet on sale. Also i added a message when there are no Tickets available for this event!. As this only comes up when Event Tickets is installed it should work out for the users of the plugin.

OLD PR:
https://github.com/moderntribe/event-tickets/pull/95
https://github.com/moderntribe/event-tickets/pull/94

Internal ticket reference: https://central.tri.be/issues/45467

### DIFF
--- a/src/views/tickets/rsvp.php
+++ b/src/views/tickets/rsvp.php
@@ -87,14 +87,6 @@ $messages_class = $messages ? 'tribe-rsvp-message-display' : '';
 				 */
 				do_action( 'event_tickets_rsvp_after_ticket_row', tribe_events_get_ticket_event( $ticket->id ), $ticket );
 			
-			}elseif ( $ticket->date_is_earlier( time() ) ) { {
-				?>
-				<span class="tickets_nostock"><?php esc_html_e( 'Tickets are not yet on sale.', 'event-tickets' ); ?></span>
-				<?php
-			}else {
-				?>
-				<span class="tickets_nostock"><?php esc_html_e( 'Tickets are no longer on sale.', 'event-tickets' ); ?></span>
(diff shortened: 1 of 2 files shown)


src/views/tickets/rsvp.php
```diff
@@ -143,4 +152,8 @@
 $content = ob_get_clean();
 if ( $is_there_any_product ) {
 	echo $content;
+}else{
```
borkweb: @masteradhoc, I wonder if the logic on lines 89-97 would go nicely down here instead? The `Tickets are no longer on sale` message seems like a nice catch-all that would also satisfy the nostock message you have below! Thoughts?

masteradhoc: @borkweb I dont mind putting it down, so if its better for you i'll take the lines down today evening :)

masteradhoc:
@borkweb finished my work. :)

borkweb:
We're ready to merge this, @masteradhoc! Thank you for the work on this PR :) I'll take it from here and implement similar messaging in our Event Tickets Plus plugin as well!

masteradhoc:
Thanks @borkweb! Are you able to tell me when this will be released? Also i would be interested in translating this one. Any additional coding needed therefore?

borkweb:
We plan to release this with 4.2 in a couple of months!

masteradhoc:
@borkweb thanks for the info. Any info about the translation question?

PR #142, opened 2016-03-21
Only pass a stock quantity to the front end if stock is being managed

borkweb:
If a stock of `0` is provided, then ticket quantities above `0` are not allowed. That's inconvenient for tickets with unlimited stock.

See: https://central.tri.be/issues/44939


barryhughes:
:+1:

PR #145, opened 2016-03-27
Remove redirect from the `init` this will throw 404 on some cases

bordoni:
Whenever we are dealing with Redirects we cannot do stuff on `init`
_Use:_ `template_redirect`

Was running into an issue of `get_permalink( $event_id )` returning he wrong url because it was too early on the execution.


borkweb:
:+1:

PR #150, opened 2016-04-11
Feature/33373 csv import tickets

lucatume:
Ticket https://central.tri.be/issues/33373

This PR builds on https://github.com/moderntribe/the-events-calendar/pull/650 to add the base in the tickets FW to add more ticket imports and the ability to import RSVP tickets


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Whitespace found at end of line | [Tickets.php:80](https://github.com/moderntribe/event-tickets/blob/9f2741b4997dcebe295561ee170187aaeffabd00/src/Tribe/Tickets.php#L80)
:x: | Whitespace found at end of line | [Tickets.php:92](https://github.com/moderntribe/event-tickets/blob/9f2741b4997dcebe295561ee170187aaeffabd00/src/Tribe/Tickets.php#L92)

src/Tribe/CSV_Importer/RSVP_Importer.php
```diff
@@ -0,0 +1,239 @@
...
+
+	/**
+	 * The class constructor proxy method.
+	 *
+	 * @param Tribe__Events__Importer__File_Importer|null $instance The default instance that would be used for the type.
+	 * @param Tribe__Events__Importer__File_Reader        $file_reader
+	 *
+	 * @return Tribe__Tickets__CSV_Importer__RSVP_Importer
+	 */
+	public static function instance( $instance, Tribe__Events__Importer__File_Reader $file_reader ) {
```
barryhughes: Doesn't look like we use/need the `$instance` argument right now - can we remove it, or do you think we'll need it in the future?

lucatume: I'd leave it as this is a filter.
The unusual situation here is the default value is `null` but that makes sense.
I see the case where we or a  third party would like to do the following things on that `$instance`:

* create it to handle the case (what happens now)
* trash it and return a new one suited to a very specific task (think of an instance handling WooCommerce tickets with or without recurrence support)
* augment the instance setting properties on it
* decorate the instance using it as a decorator pattern chain (ok, unlikely but I'd cherish the possibility)

barryhughes:
Alright man, just those two whitespace issues found by the bot and we're good to merge (tagging enhance for those but go ahead and merge once cleared) :+1:

PR #155, opened 2016-04-20
Fix/44221 tickets on cpt

jesseeproductions:
_Ref:_ [C#44221](https://central.tri.be/issues/44221)

After hitting potential road blocks found the issue was resolved, so added back that coding. And ticketed the other issue. 

Also, fixed attendees list link not showing for pages when troubleshooting this.

### DIFF
--- a/src/Tribe/Tickets_Handler.php
+++ b/src/Tribe/Tickets_Handler.php
@@ -57,6 +57,7 @@ class Tribe__Tickets__Tickets_Handler {
 
 		add_action( 'admin_menu', array( $this, 'attendees_page_register' ) );
 		add_filter( 'post_row_actions', array( $this, 'attendees_row_action' ) );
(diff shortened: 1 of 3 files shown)


src/Tribe/Tickets_Handler.php
```diff
@@ -57,6 +57,7 @@ public function __construct() {
 
 		add_action( 'admin_menu', array( $this, 'attendees_page_register' ) );
 		add_filter( 'post_row_actions', array( $this, 'attendees_row_action' ) );
+		add_filter( 'page_row_actions', array( $this, 'attendees_row_action' ) );
```
bordoni: Question, we do need any other Post Types? or only Page and Post?

jesseeproductions: From what I read CPT's will use post_row_actions only pages needed the other action.

bordoni: Sounds good, just read that.

PR #156, opened 2016-04-22
Options to hide ticket form from logged out users (per module basis)

barryhughes:
Makes it easier to stop sales/RSVPs for logged-out users. Setting can be selectively applied to individual ticket modules (ie, so we can disable RSVPing but still allow sales for logged out users, etc).

[#44347](https://central.tri.be/issues/44347)


bordoni:
👍

PR #157, opened 2016-04-22
Record the user ID associated with the creation of a new attendee

barryhughes:
When attendee posts are created, let's associate the current user ID (if any) with them. Coverage for providers besides RSVP will be added in a subsequent PR.

[#44347](https://central.tri.be/issues/44347)


bordoni:
👍

bordoni:
Looks good after the Merge.

PR #159, opened 2016-04-28
Feature/44363 frontend attendee list

barryhughes:
Provides a shortcode allowing us to list those events that the current/a specified user has indicated interest in (insofar as they are linked to an attendance post).

[C#44363](https://central.tri.be/issues/44363)


bordoni:
👍

PR #165, opened 2016-05-04
Login link is not an admin link | #44347

barryhughes:
Change how we form the login link.

https://central.tri.be/issues/44347#note-14


zbtirrell:
💯

PR #168, opened 2016-05-06
Add 'redirect_to' - take user back to event page | #44347

barryhughes:
Don't just send the user to the login page, add a redirect param to bring them back to where they started.

[C#44347](https://central.tri.be/issues/44347)


bordoni:
🎱

PR #171, opened 2016-05-10
New filter for the 'delete ticket' action link

barryhughes:
To satisfy a requirement in Community Tickets we need to make the addition of a delete-ticket link filterable.

(The update to the common commit hash isn't directly related but is correct.)

https://central.tri.be/issues/45683


bordoni:
👍

PR #152, opened 2016-04-13
Create the Tickets JSON-LD

bordoni:
_Ref:_ [C#43595](https://central.tri.be/issues/43595)

PR #172, opened 2016-05-10
Add Offers based on Tickets

bordoni:
_Ref:_ [C#43595](https://central.tri.be/issues/43595)
_Rel:_ #152


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Comment refers to a TODO task | [Order.php:10](https://github.com/moderntribe/event-tickets/blob/732a86b84209f14aba68a768477d5fa7cc47e9f5/src/Tribe/JSON_LD/Order.php#L10)
:warning: | Comment refers to a TODO task | [Order.php:48](https://github.com/moderntribe/event-tickets/blob/732a86b84209f14aba68a768477d5fa7cc47e9f5/src/Tribe/JSON_LD/Order.php#L48)
:warning: | The method parameter $args is never used | [Order.php:72](https://github.com/moderntribe/event-tickets/blob/732a86b84209f14aba68a768477d5fa7cc47e9f5/src/Tribe/JSON_LD/Order.php#L72)
:x: | You must have a space after a closing parenthesis | [Order.php:73](https://github.com/moderntribe/event-tickets/blob/732a86b84209f14aba68a768477d5fa7cc47e9f5/src/Tribe/JSON_LD/Order.php#L73)

src/deprecated/Tribe__Tickets__Google_Event_Data.php
```diff
@@ -0,0 +1,5 @@
+<?php
+_deprecated_file( __FILE__, '4.2', 'Tribe__Events__JSON_LD__Order' );
+
+
+class Tribe__Tickets__Google_Event_Data extends Tribe__Events__JSON_LD__Order {}
```
barryhughes: Shouldn't this extend `Tribe__Tickets__JSON_LD__Order` ('Tickets' not 'Events', or is there a matching PR for core)?

src/Tribe/JSON_LD/Order.php
```diff
@@ -0,0 +1,147 @@
...
+	/**
+	 * Adds the tickets data to the event Object
+	 *
+	 * @param array $data
+	 * @param array $args
+	 * @param WP_Post $post
+	 *
+	 * @return array
+	 */
+	public function add_tickets_data( $data, $args, $post ) {
```
barryhughes: For backwards compatibility, should this method be renamed to `add_ticket_data` (to match the definition in the now deprecated class)?

src/Tribe/JSON_LD/Order.php
```diff
@@ -0,0 +1,147 @@
...
+	 *
+	 * @var int
+	 */
+	protected $low_stock = 5;
+
+	/**
+	 * Setup Google Event Data for tickets.
+	 */
+	public function hook() {
+		add_filter( 'tribe_json_ld_event_object', array( $this, 'add_tickets_data' ) );
```
barryhughes: Looks like the `add_tickets_data()` method expects 3 params, should we set that up here appropriately?

src/Tribe/JSON_LD/Order.php
```diff
@@ -0,0 +1,147 @@
...
+	 * @param WP_Post $event
+	 * @return object
+	 */
+	public function get_offer( $ticket, $event ) {
+		$offer = (object) array(
+			'@type'        => 'Offer',
+			'url'          => $ticket->frontend_link,
+			'price'        => $ticket->price,
+			'category'     => 'primary',
+			'availability' => $this->get_ticket_availability(),
```
barryhughes: The `get_ticket_availability()` method requires a parameter to be passed (`$ticket`).

src/Tribe/JSON_LD/Order.php
bordoni: In this case no, because they are really completely different methods, in it's core.

src/Tribe/JSON_LD/Order.php
```diff
@@ -0,0 +1,147 @@
...
+
+		if ( ! empty( $ticket->end_date ) ) {
+			$offer->validThrough = date( DateTime::ISO8601, strtotime( $ticket->end_date ) );
+		}
+
+		/**
+		 * Allows modifications to be made to the offer object representing a specific
+		 * event ticket.
+		 *
+		 * @param object                  $offer
```
barryhughes: No biggy, might be cleaner to either push the var names out further to actually achieve alignment or else dispense with alignment for this block. Up to you!

src/Tribe/JSON_LD/Order.php
barryhughes: I guess what I'm wondering is, what if someone tries to call `Tribe__Tickets__Google_Event_Data::add_ticket_data()` which until now was a public and valid method?

src/Tribe/JSON_LD/Order.php
```diff
@@ -0,0 +1,147 @@
...
+	 * @param  object  $ticket
+	 * @return string
+	 */
+	public function get_ticket_availability( $ticket ) {
+		$stock = $ticket->stock();
+
+		if ( $stock <= 0 && $stock !== '' ) {
+			return 'SoldOut';
+		}
+		if ( $stock > 1 && $stock <= $this->low_stock ) {
```
barryhughes: Should it be `>= 1` (otherwise 1 will not count as low stock)?

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | The method parameter $args is never used | [Order.php:77](https://github.com/moderntribe/event-tickets/blob/0ade142ea99bc6ba5f8050d4a6c4a12b3facbaa6/src/Tribe/JSON_LD/Order.php#L77)

src/Tribe/Main.php
```diff
@@ -121,6 +121,9 @@ class_exists( 'TribeEvents', false )
 		$this->rsvp();
 
 		$this->user_event_confirmation_list_shortcode();
+
+		// Load the Hooks on JSON_LD
+		Tribe__Tickets__JSON_LD__Order::hook();
```
barryhughes: Isn't `hook()` an instance method? We should either change how we call it or change the method signature.

bordoni: Fixed that.

bordoni:
Fixed all the issues, good for another look.

src/Tribe/JSON_LD/Order.php
```diff
@@ -0,0 +1,149 @@
...
+	 * This can be overridden with the tribe_tickets_json_ld_low_inventory_level filter.
+	 *
+	 * @var int
+	 */
+	protected $low_stock = 5;
+
+	/**
+	 * Setup Google Event Data for tickets.
+	 */
+	public static function hook() {
```
barryhughes: Mostly asking to satisfy my curiosity: but we have `hook()` and `instance()` static methods as well as a public constructor. 

* If it's a singleton, should we make the constructor protected/private? 
* Should the hooks we set up inside the constructor be moved to the `hook()` method?
* Do we _need_ `hook()` *and* `instance()` (or, is there a reason for both to be public)?

src/Tribe/JSON_LD/Order.php
```diff
@@ -0,0 +1,149 @@
...
+	/**
+	 * Returns a string indicating current availability of the ticket.
+	 *
+	 * @param  object  $ticket
+	 * @return string
+	 */
+	public function get_ticket_availability( $ticket ) {
+		$stock = $ticket->stock();
+
+		if ( $stock <= 0 && $stock !== '' ) {
```
barryhughes: Could be something where we just wait-and-see, but I wonder if we should make the return value filterable on the basis negative stock values may be desirable in some situations, such as if back ordering is permitted (and perhaps then we wouldn't want to return `SoldOut` when this condition is met?

src/Tribe/JSON_LD/Order.php
bordoni: I always like to not do anything on `instance()` because it prevents re-attaching the functions to the filters/actions.

src/Tribe/JSON_LD/Order.php
bordoni: The problem is that these value are pre-set from JSON-LD, so making them filterable is not the best idea. But maybe something to consider, if you want to we can do a hangout and talk about it.

src/Tribe/JSON_LD/Order.php
bordoni: There are no hooks on the constructor, just setting values.

src/Tribe/JSON_LD/Order.php
barryhughes: Mmm, I'm not passionate about changing it now, just wanted to moot the idea. If you're happy, I'm happy :smile:

src/Tribe/JSON_LD/Order.php
barryhughes: > There are no hooks on the constructor, just setting values.

Yep, I'm blind.

> I always like to not do anything on instance() because it prevents re-attaching the functions to the filters/actions.

I guess what I mean is, it's conventional to get the instance via the `instance()` method: but here there are three different public methods that can all be used to obtain an `Tribe__Tickets__JSON_LD__Order` object (with different results in each case).

Wouldn't it be cleaner/tighten things up to use an arrangement more like this?

```
pubic static instance();   // Calls the constructor *and* hook()
protected static hook();   // Cannot be called directly
protected __constructor(); // Cannot be called directly
```

bordoni: By having `hook()` been protected and on `instance()` I need to create a `lock` mechanism to prevent hooks from happening twice, Im going with a middle ground. Having `hook()` public but static and the constructor be protected.

bordoni:
Im adding this a discussion point on our next dev chat. Ready for another look.

barryhughes:
Sounds good! Merge monsieur.

PR #179, opened 2016-05-12
add conditional `is_hidden_on` to hide on front end when checked on b…

trishasalas:
https://central.tri.be/issues/45173


bordoni:
👍

PR #184, opened 2016-05-13
Use actions to signal when ET has loaded/failed to load | #61156

barryhughes:
Use actions to signal when the plugin has finished initial loading, or when it has decided to bail out of loading (such as if there are compatibility problems with The Events Calendar). These actions can then be used by dependants like Event Tickets Plus.

https://central.tri.be/issues/61156


bordoni:
💯

PR #185, opened 2016-05-13
Bug/46654 incorrect plugin settings on activation

trishasalas:
https://central.tri.be/issues/46645

See also: https://github.com/moderntribe/the-events-calendar/pull/690

This commit includes updates to `common`. Tower ver generously offered to update the revision pointer for the submodule.  I went ahead and did the commit this way to see what the outcome will be but if it throws things off I'll happily delete this PR and go full CLI ;)

### DIFF
--- a/src/views/tickets/rsvp.php
+++ b/src/views/tickets/rsvp.php
@@ -119,6 +119,14 @@ $messages_class = $messages ? 'tribe-rsvp-message-display' : '';
 								<input type="email" name="attendee[email]" id="tribe-tickets-email">
 							</td>
 						</tr>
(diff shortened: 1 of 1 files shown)


barryhughes:
The change to commons does worry me - it's not something we want to modify unknowingly: 8376026d4 is still the correct hash for use with `release/4.1.4` (also, unless I'm mistaken, the new hash may refer to a local commit that hasn't been pushed?).

trishasalas:
The latter is the case and so Tower fails in not pushing those changes as well as the hash change, ugh....so my experiment fails, back to CLI ;)

PR #187, opened 2016-05-13
When dealing with TB_iframe links we need to add_thickbox

bordoni:
_Ref:_ [C#46198](https://central.tri.be/issues/46198)


borkweb:
:+1:

PR #198, opened 2016-06-03
Make get_attendees_by_transaction() public

elimn:
I see no reason for this to private, but it is a very useful public function.

### DIFF
--- a/src/Tribe/RSVP.php
+++ b/src/Tribe/RSVP.php
@@ -491,7 +491,7 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 		wp_mail( $to, $subject, $content, $headers, $attachments );
 	}
 
-	protected function get_attendees_by_transaction( $order_id ) {
+	public function get_attendees_by_transaction( $order_id ) {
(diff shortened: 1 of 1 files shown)


barryhughes:
It's kind of late to modify the codebase given where we are in the release cycle, but the change seems small enough and safe enough that we could still consider it :+1: 

* We definitely need a docblock for the method if we make it public
* We'd also want a changelog entry
* I'm wondering how a customizer/themer might obtain the `$order_id` - are you expecting they'd always be able to do this by listening out for an action like `event_tickets_rsvp_attendee_created` firing - or should this change be matched with a helper to obtain all RSVP order IDs associated with a given event?

If the answer to that last question means we need more changes, we should probably ticket and slot for NR instead of trying to merge into CR.

elimn:
Thanks for the feedback! This is not crucial so we do not need to slip into the CR in my opinion. I agree we should probably give a good way to obtain the order ID. There are other hooks that pass it but it would be nice to have it obtainable at any point. Per a discussion with Zach I am going to log a ticket for this so we have a change to think it through and discuss it rather than just making the change.

PR #199, opened 2016-06-03
feature/61425-user-facing-tickets-rsvp

trishasalas:
https://central.tri.be/issues/61425

move toggle so that only a single instance shows for multiple meta details, fix a bit of css and html


src/resources/js/rsvp.js
```diff
@@ -14,11 +14,11 @@ var tribe_tickets_rsvp = {
 
 		this.$rsvp.closest( '.cart' )
 			.on( 'submit', this.event.handle_submission );
-
-        $( '.tribe-rsvp-list' ).on( 'click', '.attendee-meta-row .toggle', function() {
```
bordoni: This whole function is with weird indentation.

src/resources/css/rsvp.css
```diff
@@ -187,4 +187,18 @@
...
+.entry-summary a,
+.taxonomy-description a,
+.logged-in-as a,
+.comment-content a,
+.pingback .comment-body > a,
+.textwidget a,
+.entry-footer a:hover,
+.site-info a:hover {
+	box-shadow: none;
+}
```
barryhughes: If this block is only targeting Twenty Sixteen should we make use of `.tribe-theme-twentysixteen` body class in the selectors? (If we are happy to use them with all themes that's fine).

src/resources/js/rsvp.js
```diff
@@ -5,23 +5,20 @@ var tribe_tickets_rsvp = {
 
 (function( $, my ) {
 	'use strict';
-
+
```
barryhughes: Fix whitespace!

src/resources/css/rsvp.css
barryhughes: Doh! This is the generated CSS. But take it as a question about the .pcss source.

src/views/tickets/orders-rsvp.php
```diff
@@ -77,10 +78,9 @@
 							$value = '&nbsp;';
 						}
 					if ( '' != $value ) { ?>
-						<a class="attendee-meta toggle show"><?php esc_html_e( 'Toggle attendee info', 'event-tickets-plus' ); ?></a>
 						<div class="attendee-meta-details">
-							<span class="event-tickets-meta-label <?php echo esc_attr( $field->slug ); ?>"><?php echo esc_html( $field->label ); ?>&nbsp;</span>
-							<span class="event-tickets-meta-data <?php echo esc_attr( $field->slug ); ?>"><?php echo $value ? esc_html( $value ) : '&nbsp;'; ?></span>
+							<span class="event-tickets-meta-label <?php echo esc_attr( $field->slug ); ?>"><?php echo esc_html( $field->label ); ?></span>
```
barryhughes: I suspect this may need to be revised and moved to ET+ (see [this PR](https://github.com/moderntribe/event-tickets/pull/200/)).

src/resources/css/rsvp.css
trishasalas: I'm slightly torn on this one, while it's unlikely that we'll see this anywhere else (for a while at least) I was honestly surprised at it's usage here.  I decided to target all themes on the off chance that there are some 'copycats' out there...this particular style is super ugly on a few of our links.  Input is welcome :)

barryhughes: Hmm, guess it's a bit of a judgement call. Re copycats - though it wouldn't cover _all_ scenarios - we could provide a little more coverage by also using `.tribe-theme-parent-{theme}` (then child themes of Twenty Sixteen would also benefit).

src/views/tickets/orders-rsvp.php
trishasalas: Agreed, I had the same thought while working on it.  I should have spoken up but was deep in the forest ;)

src/resources/css/rsvp.css
trishasalas: That's a balanced approach. (also, I didn't realize that selector was available O_o )

PR #205, opened 2016-06-14
allow parent page to be passed and fix JS error

zbtirrell:
This keeps the Thickbox working on the Community Tickets front end as well as in the Dashboard.

See related pull request: https://github.com/moderntribe/tribe-common/pull/101

See ticket: https://central.tri.be/issues/61823

### DIFF
--- a/src/Tribe/Attendees_Table.php
+++ b/src/Tribe/Attendees_Table.php
@@ -321,6 +321,7 @@ class Tribe__Tickets__Attendees_Table extends WP_List_Table {
 			'TB_iframe' => true,
 			'width' => 410,
 			'height' => 300,
+			'parent' => 'admin.php',
 		) );
 
 		$nav = array(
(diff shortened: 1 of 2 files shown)


src/Tribe/Attendees_Table.php
```diff
@@ -321,6 +321,7 @@ public function extra_tablenav( $which ) {
 			'TB_iframe' => true,
 			'width' => 410,
 			'height' => 300,
+			'parent' => 'admin.php',
```
barryhughes: Basically wondering the same thing I [asked here](https://github.com/moderntribe/tribe-common/pull/101#discussion_r66971372): the submenu page is always registered against `edit.php` rather than `admin.php` - are we confident use of `admin.php` here this won't cause problems?

zbtirrell: This change was to make sure the concerns in the other PR were addressed. I've noted that we should test https://central.tri.be/issues/46198 to make sure there are no regressions.

PR #207, opened 2016-06-14
update path for template part due to directory name change

trishasalas:
https://central.tri.be/issues/62398

_Related:_ https://github.com/moderntribe/event-tickets-plus/pull/186

### DIFF
--- a/src/views/tickets/orders.php
+++ b/src/views/tickets/orders.php
@@ -45,12 +45,12 @@ $is_event_page = class_exists( 'Tribe__Events__Main' ) && Tribe__Events__Main::P
 
 	<?php tribe_tickets_get_template_part( 'tickets/orders-rsvp' ); ?>
 
-		<?php
-		/**
-		 * Adds ticket order form
-		 */
-		do_action( 'event_tickets_plus_orders_tickets' );
-		?>
+	<?php
+	/**
+	 * Fires before the process tickets submission button is rendered
+	 */
+	do_action( 'tribe_tickets_orders_before_submit' );
+	?>
 
 	<?php if ( $view->has_rsvp_attendees( $event_id ) || $view->has_ticket_attendees( $event_id ) ): ?>
 		<div class="tribe-submit-tickets-form">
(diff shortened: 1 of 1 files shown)


src/views/tickets/orders.php
```diff
@@ -44,7 +44,7 @@
 	<form method="post">
 
 	<?php tribe_tickets_get_template_part( 'tickets/orders-rsvp' ); ?>
-	<?php tribe_tickets_get_template_part( 'tickets/orders-tickets' ); ?>
+	<?php tribe_tickets_get_template_part( 'tickets-plus/orders-tickets' ); ?>
```
barryhughes: This definitely fixes things but I don't really feel we should have this call in here at all. 

It's fine for Event Tickets Plus (or any dependent plugin) to call code in Event Tickets (or other base/parent plugin), because that's the way the dependencies flow, but it's not ideal to do that in reverse and have Event Tickets calling Event Tickets Plus.

I realize you didn't introduce this pattern in this PR, but could we use an action/callback approach instead?

trishasalas: Sure, I totally agree! (now that I think about it)

src/views/tickets/orders.php
```diff
@@ -44,7 +44,13 @@
 	<form method="post">
 
 	<?php tribe_tickets_get_template_part( 'tickets/orders-rsvp' ); ?>
-	<?php tribe_tickets_get_template_part( 'tickets/orders-tickets' ); ?>
+
+		<?php
+		/**
+		 * Adds ticket order form
+		 */
+		do_action( 'event_tickets_plus_orders_tickets' );
```
barryhughes: Nice - this seems like the right kind of approach but:

* (Minor) this block is indented too far, we should tab back one
* Naming is slightly off: within Event Tickets, the prefix we generally use for hooks is `tribe_tickets_*` so something like `tribe_tickets_orders_before_submit` might be preferable here (updating the comment to something more generic, such as _"Fires before the process tickets submission button is rendered")_

trishasalas: Thank you! I knew the naming was off but couldn't come up with anything coherent ;p

barryhughes:
:+1:

PR #208, opened 2016-06-15
Bug/61695 export all csv columns

jesseeproductions:
_Ref:_ [C#61695](https://central.tri.be/issues/61695) This fixes the columns for both email and csv and I snuck in the human readable id in the export.

### DIFF
--- a/src/Tribe/Tickets_Handler.php
+++ b/src/Tribe/Tickets_Handler.php
@@ -267,7 +267,7 @@ class Tribe__Tickets__Tickets_Handler {
 		}
 
 		//Add in Columns or get_column_headers() returns nothing
-		$filter_name = 'manage_' . $this->attendees_page . '_columns';
+		$filter_name = "manage_{$this->attendees_page}_columns";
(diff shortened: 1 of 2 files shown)


src/Tribe/Tickets_Handler.php
```diff
@@ -310,6 +314,13 @@ private function generate_filtered_attendees_list( $event_id ) {
 				if ( 'check_in' === $column_id && 1 == $single_item[ $column_id ] ) {
 					$row[ $column_id ] = esc_html__( 'Yes', 'event-tickets' );
 				}
+
+				// Special handling for new human readable id
+				if ( 'attendee_id' === $column_id ) {
+					$ticket_unique_id  = get_post_meta( $single_item[ $column_id ], '_unique_id', true );
```
bordoni: Can we make sure we have the variable to prevent notices? On `$single_item[ $column_id ]`

jesseeproductions: Adding, but that variable should always be present as it is the post id for the attendee so if it is not there we have bigger problems :)

barryhughes:
:+1:

PR #209, opened 2016-06-16
Bug/62398 edit attendee info missing

trishasalas:
https://central.tri.be/issues/62398

### DIFF
--- a/src/views/tickets/orders.php
+++ b/src/views/tickets/orders.php
@@ -43,7 +43,7 @@ $is_event_page = class_exists( 'Tribe__Events__Main' ) && Tribe__Events__Main::P
 
 	<form method="post">
 
-	<?php require( 'tickets/orders-rsvp' ); ?>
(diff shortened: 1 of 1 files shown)


src/views/tickets/orders.php
```diff
@@ -43,7 +43,7 @@
 
 	<form method="post">
 
-	<?php tribe_tickets_get_template_part( 'tickets/orders-rsvp' ); ?>
+	<?php require( 'orders-rsvp.php' ); ?>
```
barryhughes: This seems wrong: themers should be able to override `tickets/orders.php` without also overriding other templates to make it work, but using `require` like this prevents that.

Is there a reason for doing it this way?

trishasalas: Without the require only one instance of the template will load.  So an event with both RSVP and paid tickets will only display meta for the RSVP.

PR #210, opened 2016-06-17
Bug/46279 display attendee list checkbox issues

trishasalas:
https://central.tri.be/issues/46279

_Related:_ https://github.com/moderntribe/event-tickets-plus/pull/188


src/views/tickets/orders-rsvp.php
```diff
@@ -65,13 +67,13 @@
...
-					 * @param WP_Post $post Post object that the tickets are tied to
-					 */
-					do_action( 'event_tickets_orders_attendee_contents', $attendee, $post );
+						/**
+						 * Inject content into an RSVP attendee block on the RVSP orders page
+						 *
+						 * @param array $attendee Attendee array
+						 * @param WP_Post $post Post object that the tickets are tied to
+						 */
+						do_action( 'event_tickets_orders_attendee_contents', $attendee, $post );
```
borkweb: The docblock and `do_action` shouldn't be tabbed in this extra level.

src/views/tickets/orders-rsvp.php
```diff
@@ -46,10 +46,12 @@
 				);
 				?>
 			</p>
+			<?php if ( class_exists( 'Tribe__Tickets_Plus__Attendees_List' ) && ! Tribe__Tickets_Plus__Attendees_List::is_hidden_on( get_the_ID() ) ) : ?>
```
borkweb: Rather than doing a `class_exists()` here, I'd suggest adding a `do_action( 'event_tickets_after_user_details', $attendee_group, $post_id );`

Then move the HTML to `event-tickets-plus` as part of an `add_action()`. This should keep the ET+ specific code in ET+

bordoni: 👍

src/views/tickets/orders-rsvp.php
```diff
@@ -44,12 +44,16 @@
...
 					date_i18n( 'F j, Y', strtotime( esc_attr( $first_attendee['purchase_time'] ) ) )
 				);
+
+				/**
+				* Inject content into the RSVP User Details block on the orders page
+				*
+				* @param array $attendee_group Attendee array
+				* @param WP_Post $post_id Post object that the tickets are tied to
+				*/
+				do_action( 'event_tickets_user_details_rsvp', $attendee_group, $post_id );
```
borkweb: I feel like this action should occur outside of the `<p/>` tag unless you strongly disagree :)

trishasalas: No disagreement from me, makes sense :)

src/views/tickets/orders-rsvp.php
```diff
@@ -46,10 +46,15 @@
...
-			<div class="tribe-tickets attendees-list-optout">
-				<input <?php echo $view->get_restriction_attr( $post_id, esc_attr( $first_attendee['product_id'] ) ); ?> type="checkbox" name="attendee[<?php echo esc_attr( $first_attendee['order_id'] ); ?>][optout]" id="tribe-tickets-attendees-list-optout-<?php echo esc_attr( $first_attendee['order_id'] ); ?>" <?php checked( true, esc_attr( $first_attendee['optout'] ) ) ?>>
-				<label for="tribe-tickets-attendees-list-optout-<?php echo esc_attr( $first_attendee['order_id'] ); ?>"><?php esc_html_e( 'Don\'t list me on the public attendee list', 'event-tickets' ); ?></label>
-			</div>
+			<?php
+				/**
+				* Inject content into the RSVP User Details block on the orders page
+				*
+				* @param array $attendee_group Attendee array
+				* @param WP_Post $post_id Post object that the tickets are tied to
```
barryhughes: Looks like the type should be `int` rather than `WP_Post` here - the note after `$post_id` should also be updated or removed to avoid referring to a post object _or_ you could perhaps pass `$post` as the 2nd param instead.

barryhughes:
:smile_cat:

PR #216, opened 2016-06-28
Send out "you are not attending" emails | 62549

barryhughes:
Instead of sending an email full of tickets to someone who indicated they will not be attending an event, send an acknowledgement (that they RSVP'd) instead. If they later amend the status of one or more attendees to _"Going",_ send the appropriate number of tickets at that point.

https://central.tri.be/issues/62549

### DIFF
--- a/src/Tribe/RSVP.php
+++ b/src/Tribe/RSVP.php
@@ -609,26 +609,6 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 		wp_mail( $to, $subject, $content, $headers, $attachments );
 	}
 
-	/**
-	 * Returns content for emails confirming non-attendance at an event.
-	 *
-	 * @param int $attendees
-	 * @param int $event_id
-	 *
-	 * @return string
-	 */
-	protected function generate_non_attendance_email_content( $attendees, $event_id ) {
-		$file = $this->getTemplateHierarchy( 'tickets/email-non-attendace.php' );
-
-		if ( ! file_exists( $file ) ) {
-			return;
-		}
-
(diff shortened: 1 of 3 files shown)


src/Tribe/RSVP.php
```diff
@@ -468,6 +518,69 @@ public function generate_tickets( ) {
...
+		 *
+		 */
+		$to_send = (array) apply_filters( 'tribe_tickets_rsvp_tickets_to_send', $to_send, $all_attendees, $order_id );
+
+		if ( empty( $to_send ) ) {
+			return;
+		}
+
+		// For now all ticket holders in an order share the same email
+		$to = $all_attendees['0']['holder_email'];
```
borkweb: Why quote the `0`?

borkweb:
This is pretty nice! 🌊

src/Tribe/RSVP.php
barryhughes: Though it's not clear from the diff, this isn't actually new - it's how the code in this method already reads (it's been like this for [close on one year](https://github.com/moderntribe/event-tickets/commit/32012af2e7a662bdd845fa771790f644a75b1bf7)). We can adjust it, though, if you want?

borkweb: Oh weird. Adjust it if you wish. Keep it if you don't mind it/don't want to touch it.

barryhughes: Hmm, seems like a harmless change to make but on balance I think I will leave alone, just in case we cause breakages on the first Tuesday following a new moon when quoted indicies turn out to be needed :cold_sweat:

PR #217, opened 2016-06-28
removing description to eliminate overlap

trishasalas:
I'd like to revisit this as I think the graphical representation that B@rry mentioned in the ticket is a really neat way to solve the problem.  My only concern at this point is having the time to do it well.

https://central.tri.be/issues/61962


barryhughes:
Just needs a changelog entry (unless we have one already) :+1:

PR #220, opened 2016-06-28
Bug/63039 reopened hide attendees list settings issue

trishasalas:
This adds the conditional back that allows the 'attendee optout' checkbox to be hidden when Tickets Plus isn't active or if the 'hide attendee list' checkbox in the admin area is checked

https://central.tri.be/issues/63039


src/views/tickets/orders-rsvp.php
```diff
@@ -47,13 +47,15 @@
 				?>
 			</p>
 			<?php
+			if ( ( class_exists( 'Tribe__Tickets_Plus__Attendees_List' ) ) && ! Tribe__Tickets_Plus__Attendees_List::is_hidden_on( get_the_ID() ) ) {
```
barryhughes: I'd like to understand the reason for this conditional a little more. It seems to me like `event_tickets_user_details_rsvp` should fire regardless; logic like this ought to be inside the relevant callbacks.

barryhughes:
:warning: Changelog entry!

PR #222, opened 2016-06-28
bug/63039-hide-attendees-settings-discrepancy

trishasalas:
https://central.tri.be/issues/63039

Leaving the conditional in `orders-rsvp.php` to continue discussion from previous PR.  

The action `event_tickets_user_details_rsvp` is calling a template part from Tickets Plus that should only be visible if 
1) Tickets plus is active and/or 
2) the attendee list is hidden box is NOT checked in the admin area.

### DIFF
--- a/src/views/tickets/orders-rsvp.php
+++ b/src/views/tickets/orders-rsvp.php
@@ -47,13 +47,15 @@ $attendee_groups = $view->get_event_rsvp_attendees_by_purchaser( $post_id, $user
 				?>
 			</p>
 			<?php
-				/**
-				* Inject content into the RSVP User Details block on the orders page
-				*
-				* @param array $attendee_group Attendee array
-				* @param int $post_id
-				*/
-				do_action( 'event_tickets_user_details_rsvp', $attendee_group, $post_id );
+		if ( ( class_exists( 'Tribe__Tickets_Plus__Attendees_List' ) ) && ! Tribe__Tickets_Plus__Attendees_List::is_hidden_on( get_the_ID() ) ) {
+			/**
+			 * Inject content into the RSVP User Details block on the orders page
+			 *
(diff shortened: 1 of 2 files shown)


src/views/tickets/orders-rsvp.php
```diff
@@ -47,13 +47,15 @@
...
 			</p>
 			<?php
-				/**
-				* Inject content into the RSVP User Details block on the orders page
-				*
-				* @param array $attendee_group Attendee array
-				* @param int $post_id
-				*/
-				do_action( 'event_tickets_user_details_rsvp', $attendee_group, $post_id );
+		if ( ( class_exists( 'Tribe__Tickets_Plus__Attendees_List' ) ) && ! Tribe__Tickets_Plus__Attendees_List::is_hidden_on( get_the_ID() ) ) {
```
barryhughes: > The action event_tickets_user_details_rsvp is calling a template part from Tickets Plus that should only be visible if 
> 1) Tickets plus is active and/or 
> 2) the attendee list is hidden box is NOT checked in the admin area.

My take is this:

* The hook belongs to Event Tickets, not any other plugin
* If Tickets Plus wants to use it, as it does, that's totally fine
* If Tickets Plus only wants to use it under certain conditions, that's fine too - but the conditional should live in Tickets Plus

Removing this if statement and moving the logic elsewhere keeps things clean and means if (for example) a themer wants to use the action or if we want to use it from another of our plugins, we can, whether or not Tickets Plus is active. 

In general, when a parent plugin is aware of and calls code from a dependent, it's a bad sign.

trishasalas: Cool...understood, just needed to change my perspective a bit  👀

trishasalas:
Closing this to open up in Tickets Plus :)

PR #224, opened 2016-06-30
Fix/62692 community events attendees check in

lucatume:
Ticket: https://central.tri.be/issues/62692

This PR adds an `event_ID` attribute on the attendees table when shown on an event context.
That's then used to conditionally reset the event attendees table cache for the event.


src/Tribe/Attendees_Table.php
```diff
@@ -259,9 +259,14 @@ public function column_check_in( $item ) {
 
 			return sprintf( $button_template, $item['order_id_link_src'], __( 'View order', 'event-tickets' ) );
 		}
-
-		$checkin   = sprintf( '<a href="#" data-attendee-id="%d" data-provider="%s" class="button-secondary tickets_checkin">%s</a>', esc_attr( $item['attendee_id'] ), esc_attr( $item['provider'] ), esc_html__( 'Check in', 'event-tickets' ) );
-		$uncheckin = sprintf( '<span class="delete"><a href="#" data-attendee-id="%d" data-provider="%s" class="tickets_uncheckin">%s</a></span>', esc_attr( $item['attendee_id'] ), esc_attr( $item['provider'] ), esc_html__( 'Undo Check in', 'event-tickets' ) );
+		if (empty($this->event)) {
```
barryhughes: Spacing!

`if ( empty( $this->event ) ) {`

src/Tribe/Tickets.php
```diff
@@ -1389,5 +1393,15 @@ public static function get_login_url() {
 			 */
 			return apply_filters( 'tribe_tickets_ticket_login_url', $login_url );
 		}
+
+		/**
+		 * @param $return
```
barryhughes: What's the purpose of `$return` - could we add a short explanation? Not for instance quite understanding why we wouldn't do something like this at the callsite:

```
if ( $this->uncheckin( $order_id ) ) { 
    $this->maybe_update_attendees_cache(); 
}
```

lucatume: if the check-in or uncheck-in failed then there is no use in re-updating the cache, `return` will be `true` or `false` to reflect that

barryhughes: Hmm, ok. Maybe it's just me but `$return` doesn't seem like the best name in that case (it possibly makes more sense at the call site, ie in the `has_permission()` method) but right here it doesn't really reflect what it does.

lucatume: It was the pre-existing var name, but I agree that's not very explicit... something like `did_check_in` e `did_uncheck_in`?

lucatume: How about `did_check_in` and `did_uncheck_in`?

src/Tribe/Tickets.php
```diff
@@ -654,7 +658,7 @@ public static function get_event_attendees( $event_id ) {
 			$attendees = array();
 			if ( ! is_admin() ) {
 				$post_transient = Tribe__Post_Transient::instance();
-
+
```
borkweb: A bunch of stray whitespace was added here.

borkweb:
Get rid of the stray whitespace and this is good to go!

lucatume:
Thanks matt!

PR #225, opened 2016-06-30
Feature/46623 ticket deletion safeguard

trishasalas:
https://central.tri.be/issues/46623


src/admin-views/list.php
```diff
@@ -40,7 +40,7 @@
 		 * @param string $ticket_provider
 		 */
 		if ( apply_filters( 'tribe_tickets_current_user_can_delete_ticket', true, $ticket->ID, $ticket->provider_class ) ) {
-			$controls[] = sprintf( '<span><a href="#" attr-provider="%1$s" attr-ticket-id="%2$s" id="ticket_delete_%2$s" class="ticket_delete">' . esc_html__( 'Delete', 'event-tickets' ) . '</a></span>', $ticket->provider_class, $ticket->ID );
+			$controls[] = sprintf( '<span><a href="#" attr-provider="%1$s" attr-ticket-id="%2$s" onClick="return confirm(\'Are you sure you want to delete This ticket?\')" id="ticket_delete_%2$s" class="ticket_delete">' . esc_html__( 'Delete', 'event-tickets' ) . '</a></span>', $ticket->provider_class, $ticket->ID );
```
borkweb: A good rule of thumb is to avoid the use of event attributes in HTML. Instead, we prefer to keep stuff like this in JS files because it allows for some amount of organization, minification, and customization. This click event should probably live in `src/resources/js/tickets.js`.

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Must use &quot;self::&quot; for local static member reference | [Main.php:607](https://github.com/moderntribe/event-tickets/blob/0135db19b65f911423af30eeb42a01688e976a4f/src/Tribe/Main.php#L607)
:x: | Must use &quot;self::&quot; for local static member reference | [Main.php:608](https://github.com/moderntribe/event-tickets/blob/0135db19b65f911423af30eeb42a01688e976a4f/src/Tribe/Main.php#L608)

trishasalas:
I don't understand what Jenkins is trying to tell me! 🙀 🤖

src/Tribe/Main.php
```diff
@@ -596,4 +597,18 @@ public function embed_head() {
...
 	}
+
+	/**
+	 * Enqueue and localize script for use in ticket delete alert dialogue
+	 *
+	 * @since 4.3
+	 */
+	public function add_ticket_deletion_alert() {
+		$url = Tribe__Tickets__Main::instance()->plugin_url . 'src/resources/js/ticket-delete-alert.js';
+		wp_enqueue_script( 'tribe_tickets_ticket_delete_alert', $url, array( 'jquery' ), Tribe__Tickets__Main::VERSION, true );
```
bordoni: We should be using `tribe_asset` for this.

src/Tribe/Main.php
```diff
@@ -596,4 +597,18 @@ public function embed_head() {
...
 		<?php
 	}
+
+	/**
+	 * Enqueue and localize script for use in ticket delete alert dialogue
+	 *
+	 * @since 4.3
+	 */
+	public function add_ticket_deletion_alert() {
+		$url = Tribe__Tickets__Main::instance()->plugin_url . 'src/resources/js/ticket-delete-alert.js';
```
bordoni: What the 🤖 is trying to tell you here is that since you are inside of the actual class you don't need to fetch an instance of it, in this case `$this->plugin_url` would do, but take a look at my other comment below.

src/Tribe/Main.php
```diff
@@ -596,4 +597,18 @@ public function embed_head() {
...
+		wp_enqueue_script( 'tribe_tickets_ticket_delete_alert', $url, array( 'jquery' ), Tribe__Tickets__Main::VERSION, true );
+		$deletion_data = array(
+			'alert' => __( 'Are you sure you want to delete this ticket?', 'event-tickets' ),
+		);
+		wp_localize_script( 'tribe_tickets_ticket_delete_alert', 'ticket_notices', $deletion_data );
```
bordoni: A better way of improving this script would be to instead of adding one more JS variable via `wp_localize_scripts` would be to use the a `data-attr` on the link to translate, it would give us a little bit more context, like the Ticket name.

src/resources/js/ticket-delete-alert.js
```diff
@@ -0,0 +1,6 @@
+(function( window, $ ) {
+	'use strict';
+	$('.ticket_delete').click( function(){
+		alert( ticket_notices.alert );
+	})
```
bordoni: Missing `;`

src/resources/js/ticket-delete-alert.js
```diff
@@ -0,0 +1,6 @@
+(function( window, $ ) {
+	'use strict';
+	$('.ticket_delete').click( function(){
+		alert( ticket_notices.alert );
```
bordoni: Is an `alert` the best option?

src/Tribe/Main.php
trishasalas: I agree with this as a more lightweight solution. I am also very glad to have another method for this available to me for the right situation.  I was preparing to implement it when I came across this bit of information on MDN 

> Do not store content that should be visible and accessible in data attributes, because assistive technology may not access them. (https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes)

That said, I'm open to discussion and/or other options as long as they are accessible ♿️

src/resources/js/ticket-delete-alert.js
trishasalas: This is my lack of js experience showing.  Alert allows the user to confirm before the action takes place.  I'm not able to make other methods work...I've tried using `prevent.default`.  I'll do a bit more tweaking to see if I can come up with a modal type solution.

src/Tribe/Main.php
trishasalas: Thanks! I should have noticed this...sleeping on it helped!

src/Tribe/Main.php
```diff
@@ -596,4 +597,18 @@ public function embed_head() {
...
 	}
+
+	/**
+	 * Enqueue and localize script for use in ticket delete alert dialogue
+	 *
+	 * @since 4.3
+	 */
+	public function add_ticket_deletion_alert() {
+		$url = $this->plugin_url . 'src/resources/js/ticket-delete-alert.js';
+		wp_enqueue_script( 'tribe_tickets_ticket_delete_alert', $url, array( 'jquery' ), self::VERSION, true );
```
barryhughes: This all looks good, only comment is we should probably use `tribe_asset()` here (I think Gustavo mentioned this also, not sure if you discussed further).

trishasalas: Can you (or @bordoni) elaborate on this, I tried to model this after a usage that I found in the code but it broke my function. I do plan on running through it one more time as I did it pretty quickly...

src/Tribe/Main.php
```diff
@@ -596,4 +597,31 @@ public function embed_head() {
...
+		);
+		tribe_asset(
+			self::instance(),
+			'tribe_tickets_ticket_delete_alert',
+			'ticket-delete-alert.js',
+			array( 'jquery' ),
+			'admin_enqueue_scripts',
+			array(
+				'localize' => array(
+					'name' => 'tribe_tickets_notices',
```
barryhughes: This names the object as `tribe_tickets_notices` (tickets plural) but the script expects `tribe_ticket_notices` (ticket singular), which is why you're getting the undefined object notice you referenced in chat - so we just need to change that here (or in `ticket-delete-alert.js`, whichever you prefer), kill the commented out lines and the transition to `tribe_asset()` will be complete.

barryhughes:
:+1:

PR #229, opened 2016-07-18
Ticket reassignment

barryhughes:
Adds facilities to move ticket types from one event to another, and to move issued tickets from one ticket type to another (whether or not they exist in the same event/post).

Relies on some new additions to tribe-common, see [Tribe Common PR 137](https://github.com/moderntribe/tribe-common/pull/137).

https://central.tri.be/issues/61724


src/Tribe/Admin/Move_Ticket_Types.php
```diff
@@ -0,0 +1,364 @@
...
+			return $data;
+		}
+
+		return array_merge( $data, array(
+			'ticket_type_id' =>
+				absint( $_GET[ 'ticket_type_id' ] ),
+			'src_post_id' =>
+				absint( $_GET[ 'post' ] ),
+			'mode' =>
+				'ticket_type_only'
```
borkweb: This array is formatted in a fairly non-standard way compared to what we do elsewhere. The last element is missing a trailing comma, too.

src/Tribe/Admin/Move_Ticket_Types.php
```diff
@@ -0,0 +1,364 @@
...
+		}
+
+		$args = wp_parse_args( $_POST, array(
+			'post_type'    => '',
+			'search_terms' => '',
+			'ignore'       => '',
+		) );
+
+		wp_send_json_success( array( 'posts' =>  $this->get_possible_matches( $args ) ) );
+		exit();
```
borkweb: The `exit();` isn't needed when doing a `wp_send_json[_success|_error]()` as it will [never be reached](https://github.com/WordPress/WordPress/blob/master/wp-includes/functions.php#L3061).

src/Tribe/Admin/Move_Ticket_Types.php
```diff
@@ -0,0 +1,364 @@
...
+	/**
+	 * Listens out for ajax requests to move a ticket type to a new post.
+	 */
+	public function move_ticket_type_requests() {
+		if ( ! wp_verify_nonce( $_POST['check' ], 'move_tickets' ) ) {
+			wp_send_json_error();
+		}
+
+		$ticket_type_id = absint( $_POST[ 'ticket_type_id' ] );
+		$destination_id = absint( $_POST[ 'target_post_id' ] );
```
borkweb: Should we check if these two $_POST indexes are set to avoid notices?

src/Tribe/Admin/Move_Tickets.php
```diff
@@ -0,0 +1,640 @@
...
+	}
+
+	/**
+	 * Enqueues all assets and data required by the move tickets dialog.
+	 */
+	protected function dialog_assets() {
+		// Ensure common admin CSS is enqueued within this screen
+		add_filter( 'tribe_asset_enqueue_tribe-common-admin', '__return_true', 20 );
+
+		// @todo consider switching to tribe_asset() following resolution of https://github.com/moderntribe/tribe-common/pull/111#discussion_r68219366
```
borkweb: Good thought on switching to `tribe_asset()`

src/Tribe/Admin/Move_Tickets.php
```diff
@@ -0,0 +1,640 @@
...
+	/**
+	 * Responds to ajax requests for a list of supported post types.
+	 */
+	public function get_post_types() {
+		if ( ! wp_verify_nonce( $_POST['check' ], 'move_tickets' ) ) {
+			wp_send_json_error();
+		}
+
+		wp_send_json_success( array( 'posts' => $this->get_post_types_list() ) );
+		exit();
```
borkweb: Unneeded `exit();`

src/Tribe/Admin/Move_Tickets.php
```diff
@@ -0,0 +1,640 @@
...
+		}
+
+		$args = wp_parse_args( $_POST, array(
+			'post_type'    => '',
+			'search_terms' => '',
+			'ignore'       => '',
+		) );
+
+		wp_send_json_success( array( 'posts' =>  $this->get_possible_matches( $args ) ) );
+		exit();
```
borkweb: Unneeded `exit();`

src/Tribe/Admin/Move_Tickets.php
```diff
@@ -0,0 +1,640 @@
...
+	 * looks in $_POST for the corresponding values):
+	 *
+	 * - 'post_type': string or array of post types
+	 * - 'search_term': string used for searching posts to narrow the field
+	 *
+	 * @param array|null $request post parameters (or looks at $_POST if not set)
+	 *
+	 * @return array
+	 */
+	protected function get_possible_matches( array $request = null ) {
```
borkweb: This method is identical to the one declared [here](https://github.com/moderntribe/event-tickets/pull/229/files#diff-41c69c826caa255771804f564897194fR91).

src/Tribe/Admin/Move_Tickets.php
```diff
@@ -0,0 +1,640 @@
...
+
+	/**
+	 * Given an array of WP_Post objects, returns an array containing the post title
+	 * of each (with the post ID as the index).
+	 *
+	 * @param array $query_results
+	 *
+	 * @return array
+	 */
+	protected function format_post_list( array $query_results ) {
```
borkweb: This appears identical to [this method](https://github.com/moderntribe/event-tickets/pull/229/files#diff-41c69c826caa255771804f564897194fR135).

src/Tribe/Admin/Move_Tickets.php
```diff
@@ -0,0 +1,640 @@
...
+			wp_send_json_error();
+		}
+
+		$args = wp_parse_args( $_POST, array(
+			'post_id'  => '',
+			'provider' => '',
+		) );
+
+		wp_send_json_success( array( 'posts' =>  $this->get_ticket_type_matches( $args[ 'post_id' ], $args[ 'provider' ] ) ) );
+		exit();
```
borkweb: Unneeded `exit()`

src/Tribe/Admin/Move_Tickets.php
```diff
@@ -0,0 +1,640 @@
...
+		wp_send_json_success( array(
+			'message' => sprintf(
+				__( 'The tickets were successfully moved and can be found within %1$sthis post%2$s. Please wait a moment while we refresh the editor screen.', 'event-tickets' ),
+				'<a href="' . esc_url( get_admin_url( null, '/post.php?post=' . $args[ 'target_post_id' ] . '&action=edit' ) ) . '" target="_blank">',
+				'</a>'
+			),
+			'redirect_top' => $redirect_url,
+		) );
+
+		exit();
```
borkweb: Unneeded `exit()`

src/Tribe/Admin/Move_Tickets.php
```diff
@@ -0,0 +1,640 @@
...
+
+			if ( ! isset( $to_notify[ $attendee[ 'purchaser_email' ] ] ) ) {
+				$to_notify[ $attendee[ 'purchaser_email' ] ] = array( $attendee );
+			} else {
+				$to_notify[ $attendee[ 'purchaser_email' ] ][] = $attendee;
+			}
+		}
+
+		foreach ( $to_notify as $email_addr => $affected_tickets) {
+			$to = apply_filters( 'tribe_tickets_ticket_type_moved_email_recipient', $email_addr );
```
borkweb: Should this filter be changed to `tribe_tickets_ticket_moved_email_recipient`?

src/Tribe/Admin/Move_Tickets.php
barryhughes: You're right, I can remove the subclass's copy.

src/Tribe/Admin/Move_Tickets.php
borkweb: Side note: it'd be snazzy if these filters had docblocks.

src/Tribe/Tickets.php
```diff
@@ -207,6 +233,20 @@ public function get_event_for_ticket( $possible_ticket ) {
...
 		/**
+		 * Returns the email address associated with an attendee record (if any).
+		 *
+		 * Individual providers ought to override this method.
+		 *
+		 * @param int $attendee_id
+		 *
+		 * @return string|null
+		 */
+		public function get_attendee_email_address( $attendee_id ) {
```
borkweb: Is this method still needed? If so, should it be fleshed out more?

src/resources/js/move-tickets-dialog.js
```diff
@@ -0,0 +1,496 @@
+/**
+ * Handles the move tickets/move ticket type dialogs.
+ *
+ * @var object tribe_move_tickets
+ * @var object tribe_move_tickets_data
+ */
+
+var tribe_move_tickets = ( 'object' === typeof tribe_move_tickets ) ? tribe_move_tickets : {};
```
borkweb: You could simplify this by doing:

```js
var tribe_move_tickets = tribe_move_tickets || {};
```

src/resources/js/move-tickets-dialog.js
```diff
@@ -0,0 +1,496 @@
...
+		$stages,
+		$back,
+		$next,
+		last_direction,
+		skip_choose_event,
+		target_post_id,
+		target_ticket_type_id;
+
+
+	function init() {
```
borkweb: The way all of these functions are declared makes it pretty hard to debug/extend/modify/test (I'm just pretending that we have tests for JS as I type this). These functions are essentially all private and untouchable outside the confines of this anonymous block.

Using this approach resolves those issues:

```js
obj.init = function() {
};
```

barryhughes: Hmm, well I dig the approach in general (and of course `obj` is in scope for that very reason - to let us make methods public, even if right now none are), but our coding standards state that methods should be private by default. 

This is our first crack at a wizard-style interface and I'm not sure we want to make every method public by default here as that means a lot of work in the future if we want to deprecate stuff and move things around.

So I guess my questions are: 

* Are we deviating from this rule (members should by private by default) in the case of our JS code? (Why?)
* Or are there specific items you'd like to be public (rather than making them all public)?
* I already trigger numerous events and external code is free to utilize those - is _that_ an approach we could extend in preference to making more things public?

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Expected 1 space after &quot;=&gt;&quot;; 2 found | [Move_Ticket_Types.php:32](https://github.com/moderntribe/event-tickets/blob/255f2ef5e49b4823b07da5231f816c521789af74/src/Tribe/Admin/Move_Ticket_Types.php#L32)

src/Tribe/Admin/Move_Tickets.php
barryhughes: (Removed in the subclass.)

src/resources/js/move-tickets-dialog.js
borkweb: Our _PHP_ coding standards state that methods should be private by default and I suppose there is merit to our JS following suit, though that line hasn't been explicitly drawn in the sand to this point.

Perhaps I have a jaded view of plugins/services that provide JS that isn't easily modifiable/easy to override. In my previous positions privately scoped third-party JS was the bane of my existence when tasked with mods/complex behaviors for products I was working on. There's probably a pleasant middle ground (and your third bullet is most likely it).

I guess my thoughts on your questions are:

* Perhaps we don't have to. We can aim to expose stuff that _should_ be exposed and keep private the stuff that needs to be private.
* If our goal is to avoid customizations of the wizard, then I suppose not. As it stands, if I wanted to inject an additional stage into the wizard and utilize some of the same button behaviors that are defined, that would prove troublesome (unless I'm missing something). That's probably ok for now and we can just wait for a feature request which may never come!
* Events are great.

barryhughes: OK, let me ponder what changes would make sense here then.

> As it stands, if I wanted to inject an additional stage into the wizard and utilize some of the same button behaviors that are defined, that would prove troublesome (unless I'm missing something).

borkweb:
:noice:

PR #230, opened 2016-07-25
Bugfixes for moving tickets/ticket types

barryhughes:
Separates out the process of adding actions/filters from object construction, so we can still inherit methods but don't double up on callbacks unnecessarily.

[C#61724](https://central.tri.be/issues/61724)


borkweb:
:+1:

PR #235, opened 2016-08-11
fix/64778-rsvp-mail-apostrophe-support

lucatume:
Ticket: https://central.tri.be/issues/64778

Parse the site title before adding it as a subject to the RSVP email confirmation.


borkweb:
🍦

PR #238, opened 2016-08-18
Fix targeting of ticket deletion confirmation

borkweb:
Additionally, I've moved the contents of the ticket alert JS file into the larger tickets.js file to:

1. Avoid extra JS files
2. Avoid having the file (and event) executed on every admin page - we only need it for the meta box

See: https://central.tri.be/issues/46623


zbtirrell:
👍

PR #241, opened 2016-08-26
Fix/61992 enhancements

barryhughes:
Address QA feedback re styling and appearance of the attendee screen's summary box.

[#61992](https://central.tri.be/issues/61992)


jesseeproductions:
:100:

PR #242, opened 2016-08-26
Maintain expected set of attendee export columns

barryhughes:
Following some reworking of the attendee screen, we still need to maintain the expected set of export columns.

[#61993](https://central.tri.be/issues/61993)


jesseeproductions:
:+1:

PR #243, opened 2016-08-29
tribe_is_past_event conditional and ticket messages

jesseeproductions:
_Ref:_ [C#62303 ](https://central.tri.be/issues/62303 ) Utilize new tribe_is_event_past() conditional to change tickets not available message.

### DIFF
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -1278,10 +1278,15 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 		 * @return string
 		 */
 		public function get_tickets_unavailable_message( $tickets ) {
-			$availability_slug = $this->get_availability_slug_by_collection( $tickets );
-			$message = null;
 
-			if ( 'availability-future' === $availability_slug ) {
+			$availability_slug = $this->get_availability_slug_by_collection( $tickets );
+			$message           = null;
+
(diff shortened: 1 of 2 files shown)


src/Tribe/Tickets.php
```diff
@@ -1278,10 +1278,14 @@ public function get_availability_slug_by_collection( $tickets, $datetime = null
 		 * @return string
 		 */
 		public function get_tickets_unavailable_message( $tickets ) {
+
 			$availability_slug = $this->get_availability_slug_by_collection( $tickets );
-			$message = null;
+			$message           = null;
 
-			if ( 'availability-future' === $availability_slug ) {
+			if ( function_exists( 'tribe_is_past_event' ) && tribe_is_past_event() ) {
```
barryhughes: Thinking we might also need a check against the current post's post type here. For instance, suppose the tickets here belong to a page rather than an event and that `tribe_is_past_event()` is indeed defined - I _think_ this will evaluate to true every time in that situation.

* Over in `tribe_is_past_event()` we look at the current post
* We call `tribe_get_end_date()` which returns null (for a non-event)
* Null evaluates to less than `time()` so `tribe_is_past_event()` returns true

jesseeproductions: added a check for the tribe event post type so it prevents this.

barryhughes:
:pear:

PR #244, opened 2016-08-31
Action for 3rd Party plugins

sabitertan:
I added it for 3rd party plugins can hook to this action.
I customized error_handler to achieve what i need, but should consider to create a global error handler on Main class, so internally or 3rd party plugins can throw that specific error. 
So IMO it should be like this

do_action( 'tribe_tickets_rsvp_before_tickets_generated', $event_id, $_POST);
/*check tribe_tickets_global_error after this line and redirect to page if there is any error*/

### DIFF
--- a/src/Tribe/RSVP.php
+++ b/src/Tribe/RSVP.php
@@ -408,7 +408,14 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 		}
 
 		$has_tickets = $event_id = false;
-
+		/**
+		* RSVP specific action fired just before a RSVP-driven attendee ticket for an event is generated
+		* 
+		* @param $event_id ID of event
+		* @param $_POST post paremeters comes from RSVP Form
+		*/
+		do_action( 'tribe_tickets_rsvp_before_tickets_generated', $event_id, $_POST);
+		
 		$order_id = md5( time() . rand() );
 
 		$attendee_email = empty( $_POST['attendee']['email'] ) ? null : sanitize_email( $_POST['attendee']['email'] );


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Whitespace found at end of line | [RSVP.php:413](https://github.com/moderntribe/event-tickets/blob/6da24e49125060699c412df18a515104de01e3f4/src/Tribe/RSVP.php#L413)
:x: | There must be a single space before the last parenthesis | [RSVP.php:417](https://github.com/moderntribe/event-tickets/blob/6da24e49125060699c412df18a515104de01e3f4/src/Tribe/RSVP.php#L417)
:x: | Whitespace found at end of line | [RSVP.php:418](https://github.com/moderntribe/event-tickets/blob/6da24e49125060699c412df18a515104de01e3f4/src/Tribe/RSVP.php#L418)

borkweb:
This is great, @sabitertan! Thank you for this contribution. We'll be adding this into the next major release!

borkweb:
@sabitertan: Just as a heads up, I've made a few changes in a subsequent [Pull Request](https://github.com/moderntribe/event-tickets/pull/245/files#diff-cf9dea489ace789fa7f408943aa26cd3R468).

1. The original action was in a place that did not have the `$event_id` available (just declared as `false`). I've replaced that action with the following: [`tribe_tickets_rsvp_before_order_processing`](https://github.com/moderntribe/event-tickets/pull/245/files#diff-cf9dea489ace789fa7f408943aa26cd3R417). Its only parameter is the `$_POST` array.

2. I moved (and slightly renamed) the action to a location in `generate_tickets()` where the `$event_id` is present, though it _may_ be possible that it can fire multiple times if a user purchases multiple RSVP ticket types at the same time. This action has been named: [`tribe_tickets_rsvp_before_attendee_ticket_creation`](https://github.com/moderntribe/event-tickets/pull/245/files#diff-cf9dea489ace789fa7f408943aa26cd3R475) and has a new parameter of Ticket Type (`$ticket_type`).

PR #245, opened 2016-08-31
Cleaning up some tr1b0t sniffs and adding changelog

borkweb:
Follow-up to: https://github.com/moderntribe/event-tickets/pull/244

See: https://central.tri.be/issues/65836

### DIFF
--- a/src/Tribe/RSVP.php
+++ b/src/Tribe/RSVP.php
@@ -447,7 +447,7 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 				continue;
 			}
 
-			$ticket = $this->get_ticket( $event_id, $product_id );
+			$ticket_type = $this->get_ticket( $event_id, $product_id );
 
 			// if there were no RSVP tickets for the product added to the cart, continue
 			if ( empty( $_POST[ "quantity_{$product_id}" ] ) ) {
@@ -457,7 +457,7 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 			$qty = max( intval( $_POST[ "quantity_{$product_id}" ] ), 0 );
 
 			// Throw an error if Qty is bigger then Remaining
-			if ( $ticket->managing_stock() && $qty > $ticket->remaining() ) {
+			if ( $ticket_type->managing_stock() && $qty > $ticket_type->remaining() ) {
 				$url = add_query_arg( 'rsvp_error', 2, get_permalink( $event_id ) );
 				wp_redirect( esc_url_raw( $url ) );
 				die;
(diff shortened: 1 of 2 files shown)


src/Tribe/RSVP.php
```diff
@@ -408,14 +408,15 @@ public function generate_tickets( ) {
...
 		/**
 		* RSVP specific action fired just before a RSVP-driven attendee ticket for an event is generated
-		* 
+		*
 		* @param $event_id ID of event
 		* @param $_POST post paremeters comes from RSVP Form
 		*/
-		do_action( 'tribe_tickets_rsvp_before_tickets_generated', $event_id, $_POST);
-		
+		do_action( 'tribe_tickets_rsvp_before_tickets_generated', $event_id, $_POST );
```
barryhughes: Seems a little odd to pass a superglobal that the callback can access anyway (though I guess we might switch it up in future and pass a modified/untainted array). More importantly though, isn't `$event_id` always going to be false at this point?

borkweb: Yeah. I guess I looked over this too loosely. I've moved the action down to the product loop.

src/Tribe/RSVP.php
```diff
@@ -466,6 +458,15 @@ public function generate_tickets( ) {
...
 			$has_tickets = true;
 
+			/**
+			* RSVP specific action fired just before a RSVP-driven attendee ticket for an event is generated
+			*
+			* @param $event_id ID of event
+			* @param $ticket Ticket of product
+			* @param $_POST post paremeters comes from RSVP Form
+			*/
+			do_action( 'tribe_tickets_rsvp_before_tickets_generated', $event_id, $ticket, $_POST );
```
barryhughes: Maybe I'm grinding a personal axe here, but lots of our ticket code is confusing because of our loose use of terms like tickets. For instance, here we're passing `$ticket` to a callback that fires before tickets are generated. Might `$ticket_type` help to differentiate between the two types of 'ticket'?

Optional, might just be the way I read things.

src/Tribe/RSVP.php
```diff
@@ -465,6 +458,15 @@ public function generate_tickets( ) {
 
 			$has_tickets = true;
 
+			/**
+			* RSVP specific action fired just before a RSVP-driven attendee ticket for an event is generated
+			*
+			* @param $event_id ID of event
+			* @param $ticket Ticket of product
+			* @param $_POST post paremeters comes from RSVP Form
```
barryhughes: Should have noted this before also. True story: I have seen in our own code a case where we did this (or it could have been a different superglobal like `$_GET`) and it was written that way in the callback function, ie:

``` php
function fire_sometime( $_POST, ... ) {}
```

Simply to discourage others from that (it will trigger errors on newer php runtimes) how about naming the variable differently here in the docblock?

zbtirrell:
I'm not sure that location is where the customer wanted the action though...

barryhughes:
:egg:

PR #247, opened 2016-09-06
Don't allow recurring events safeguard to impact non-events | #45008

barryhughes:
In the case of recurring events (child posts) we deliberately load the parent event's tickets - however this same logic should not apply to other post types such as pages.

[#45008](https://central.tri.be/issues/45008)


borkweb:
:+1:

PR #248, opened 2016-09-06
Revise "tickets unavailable" messaging, prevent duplicates when multiple ticket providers are active

barryhughes:
Introduce a system for displaying "tickets unavailable" messages that can be shared by multiple active ticket providers. We should only ever see one "tickets are unavailable" message per event.

[#65687](https://central.tri.be/issues/65687)


zbtirrell:
👍

PR #253, opened 2016-09-08
conditionally update the ticket cache

lucatume:
Ticket: https://central.tri.be/issues/63959

This PR adds a check to make sure TEC and PRO related actions are not attempted when the plugins are not active.

### DIFF
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -497,7 +497,9 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 			// Pass the control to the child object
 			$did_checkin = $this->checkin( $order_id );
 
-			$this->maybe_update_attendees_cache( $did_checkin );
+			if ( class_exists( 'Tribe__Events__Main' ) ) {
+				$this->maybe_update_attendees_cache( $did_checkin );
+			}
 
 			$this->ajax_ok( $did_checkin );
 		}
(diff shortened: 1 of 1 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | There must be a single space after the first parenthesis | [Tickets.php:532](https://github.com/moderntribe/event-tickets/blob/71761e1a68379fd00510fbe4e875bcf62a7b3ab0/src/Tribe/Tickets.php#L532)
:x: | There must be a single space before the last parenthesis | [Tickets.php:532](https://github.com/moderntribe/event-tickets/blob/71761e1a68379fd00510fbe4e875bcf62a7b3ab0/src/Tribe/Tickets.php#L532)

barryhughes:
Guessing this is because `Tribe__Tickets__Tickets::maybe_update_attendees_cache()` calls `tribe_is_event()`.

When I look at `Tribe__Tickets__Tickets::get_event_attendees()` though (which gets/sets the cached value) it doesn't seem to care whether the 'event' is an actual `'tribe_event'` post or not - so what I'm wondering is: do we need these checks, or should we adjust `Tribe__Tickets__Tickets::maybe_update_attendees_cache()` and remove the `tribe_is_event()` call?

lucatume:
I think the call should be removed entirely and re-instate the cache update.
Tickets could be associated to any post type, not just events, and we want to update cache for any post type.

barryhughes:
:8ball:

PR #254, opened 2016-09-08
added tjhe `bumpdown-class` data attribute

lucatume:
Ticket https://central.tri.be/issues/65617

Builds on https://github.com/moderntribe/tribe-common/pull/191

Pass the `bumpdown-class` over to the bumpdown code

### DIFF
--- a/src/admin-views/rsvp-metabox-advanced.php
+++ b/src/admin-views/rsvp-metabox-advanced.php
@@ -9,6 +9,7 @@
 
 <?php
 if ( class_exists( 'Tribe__Events__Pro__Main' ) ) {
+	$done = true;
 	if ( is_admin() ) {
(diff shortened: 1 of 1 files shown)


src/admin-views/rsvp-metabox-advanced.php
```diff
@@ -9,6 +9,7 @@
 
 <?php
 if ( class_exists( 'Tribe__Events__Pro__Main' ) ) {
+	$done = true;
```
barryhughes: Can this be removed? Don't _think_ it's actually used any place.

barryhughes: ... Otherwise, this looks great - merge away when resolved.

PR #255, opened 2016-09-08
Set current screen to avoid notices/warnings | spotfix

barryhughes:
The move tickets/ticket type dialog is generated quite early on in the request and the `$current_screen` global is not yet set. This creates some problems which we can resolve by setting it ourselves.


zbtirrell:
👍

PR #261, opened 2016-09-28
Fix/61992 bulk row actions

barryhughes:
Removes attempt to redirect after processing bulk/row actions within the attendee table (it's too late in the request to do this).

Additionally, adds a safeguard in the RSVP provider to avoid notices if an invalid or deleted ticket ID is passed to the delete handler.


bordoni:
👍

PR #262, opened 2016-09-28
Optionally adjust total_sales following deletion of an RSVP ticket/attendee

barryhughes:
Previous behaviour: `total_sales` was being reduced by one every time an attendee was deleted (this was done to re-stock RSVP tickets). New behaviour: it is only reduced if configured to do so - by default this means if the deleted attendee was 'Going' we re-stock, but not otherwise.

https://central.tri.be/issues/64669

### DIFF
--- a/src/Tribe/RSVP.php
+++ b/src/Tribe/RSVP.php
@@ -858,9 +858,16 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 		}
 		$product_id = get_post_meta( $ticket_id, self::ATTENDEE_PRODUCT_KEY, true );
 
-		// Decrement the sales figure
-		$sales = (int) get_post_meta( $product_id, 'total_sales', true );
-		update_post_meta( $product_id, 'total_sales', -- $sales );
+		// For attendees whose status ('going' or 'not going') for whom a stock adjustment is required?
+		$rsvp_options    = $this->tickets_view->get_rsvp_options( null, false );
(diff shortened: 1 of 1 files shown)


src/Tribe/RSVP.php
```diff
@@ -858,9 +858,16 @@ public function delete_ticket( $event_id, $ticket_id ) {
 		}
 		$product_id = get_post_meta( $ticket_id, self::ATTENDEE_PRODUCT_KEY, true );
 
-		// Decrement the sales figure
-		$sales = (int) get_post_meta( $product_id, 'total_sales', true );
-		update_post_meta( $product_id, 'total_sales', -- $sales );
+		// For attendees whose status ('going' or 'not going') for whom a stock adjustment is required?
+		$rsvp_options    = $this->tickets_view->get_rsvp_options( null, false );
+		$attendee_status = get_post_meta( $ticket_id, self::ATTENDEE_RSVP_KEY, true );
+		$adjustment      = absint( $rsvp_options[ $attendee_status ]['decrease_stock_by'] );
```
bordoni: Is it safe to always fetch this Array like this?

barryhughes: Good point, we don't know for sure that `$attendee_status` will be set because `$rsvp_options` is filtered; will add a safety check :+1:

barryhughes: Updated!

borkweb:
:boom:

PR #263, opened 2016-09-28
Do not calculate attendance totals until after the table is prepared …

barryhughes:
Moves calculation of event attendance to a later point in time.

This is required because code handling things such as check-in and deletion lives in the attendee list table class, which doesn't run until right before the table is rendered.


bordoni:
💯

PR #259, opened 2016-09-22
Implement support for RSVP and sales support

lucatume:
Ticket: https://central.tri.be/issues/64669

This PR updates the RSVP tickets code to:

* not increase sales when an RSVP is Not Going
* increase sales when Going
* allow a robust filtering of the RSVP states with variable stock sizes
* tests for all the code (integration as it's as small as I could go)
* add some filters to prevent email and anywhere I felt we were lacking

PR #264, opened 2016-09-28
Remove unused var | spotfix

barryhughes:
Removes unused var. 

Discussion [here](https://github.com/moderntribe/event-tickets/pull/259#discussion_r80913711) and [here](https://tribe.slack.com/archives/products-dev/p1475069038003538) (basically - safe to remove).


lucatume:
Go for it, I could not possibly imagine how bad a developer someone could be to do that.

PR #266, opened 2016-09-28
Confirm for ticket deletion in the main delete handling function

borkweb:
See: https://central.tri.be/issues/46623


trishasalas:
🐳

PR #277, opened 2016-10-12
Allow localization of the decimal character to happen outside of the admin environment

barryhughes:
Move the decimal localization code into its own method, so we can reuse it from within Community Tickets without having to duplicate it.

https://central.tri.be/issues/67913


barryhughes:
Wrong branch ... re-targeting.

PR #280, opened 2016-10-12
Allow localization of the decimal character to happen outside of the admin env | #67913

barryhughes:
Sorry folks. One last time.


borkweb:
Hahahahha :+1:

PR #285, opened 2016-10-18
add response param to 2nd edit-ticket.tribe trigger 

helgatheviking:
in edit ticket `$post` success callback.

For some reason `edit-ticket.tribe` is triggered twice, once at the very [beginning](https://github.com/moderntribe/event-tickets/blob/develop/src/resources/js/tickets.js#L473) and again at the [very end](https://github.com/moderntribe/event-tickets/blob/develop/src/resources/js/tickets.js#L615) of the success callback. But the second trigger doesn't have the `response`  variable passed through so any listener trying to set the values for some custom inputs will fail (ie. *not* properly set the inputs in the metabox with the response values) when running on the 2nd trigger because the `response` variable is no longer there. 

example of some custom inputs:
image

### DIFF
--- a/src/resources/js/tickets.js
+++ b/src/resources/js/tickets.js
@@ -612,7 +612,7 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
 
 						$tribe_tickets
 							.trigger( 'set-advanced-fields.tribe' )
-							.trigger( 'edit-ticket.tribe' );
+							.trigger( 'edit-ticket.tribe', response );
 
 					},
 					'json'


helgatheviking:
Looks like the same issue exists in Events Calendar, though I don't know if that's having any impact:
https://github.com/moderntribe/event-tickets/blob/0d3dc0c7e46afb5ba0a3daab4f70125254568243/src/resources/js/tickets.js#L615

barryhughes:
[Internal ticket added: #68577](https://central.tri.be/issues/68557)

barryhughes:
Hi @helgatheviking - thanks for submitting this fix!

> Looks like the same issue exists in Events Calendar

For a number of reasons, mostly connected with legacy usage of our ticketing plugins, we bundle `event-tickets` as a vendor lib of The Events Calendar. I'm guessing that's where you're seeing the code you referenced and, if so, merging your change will effectively fix the same thing there, too.

helgatheviking:
@barryhughes for purely selfish reasons, but you're welcome! I think I eventually realized that the events calendar was sharing this code, when I was digging through Events Calendar to try to prepare a PR for that.

bordoni:
Resolved on #291

PR #289, opened 2016-10-26
Attendees report Screen Options

lucatume:
Ticket: https://central.tri.be/issues/66016

This PR registers the Attendees table columns in the screen options to allow those to be hidden on a per-user base.

### DIFF
--- a/src/Tribe/Admin/Screen_Options/Attendees.php
+++ b/src/Tribe/Admin/Screen_Options/Attendees.php
@@ -1,0 +1,63 @@
+<?php
+
+
+/**
+ * Class Tribe__Tickets__Admin__Screen_Options__Attendees
+ *
+ * Handles the Attendees report table screen options.
+ */
+class Tribe__Tickets__Admin__Screen_Options__Attendees {
+
+	/**
+	 * @var string The screen id these screen options should render on.
+	 */
+	protected $screen_id;
+
+	/**
+	 * @var WP_Screen Either the globally defined WP_Screen instance or an injected dependency.
+	 */
+	protected $screen;
+
+	/**
+	 * Tribe__Tickets__Admin__Screen_Options__Attendees constructor.
+	 *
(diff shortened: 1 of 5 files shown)


src/Tribe/Admin/Screen_Options/Attendees.php
```diff
@@ -0,0 +1,63 @@
...
+	 * @return bool Whether the screen options were added or not.
+	 */
+	public function add_options() {
+		$this->screen = $this->screen ? $this->screen : get_current_screen();
+
+		if ( ! is_object( $this->screen ) || $this->screen->id !== $this->screen_id ) {
+			return false;
+		}
+
+		$this->add_column_headers_options();
```
barryhughes: Might it be worth inlining the add_filter() call from this method?

src/Tribe/Admin/Screen_Options/Attendees.php
```diff
@@ -0,0 +1,63 @@
...
+
+		return true;
+	}
+
+	protected function add_column_headers_options() {
+		add_filter( "manage_{$this->screen->id}_columns", array( $this, 'filter_manage_columns' ) );
```
barryhughes: Might it be worth supplying `Tribe__Tickets__Attendees_Table::get_table_columns()` as the callback without an intermediate step?

src/Tribe/Admin/Screen_Options/Attendees.php
lucatume: the `add_filter` calls is clear about how the thing is done, not the "what"; the "what" here is that I'm adding the columns screen options by listing them on the table. I think the one line method clear about the "what" is happening (adding screen options for columns) beats inlining the `add_filter` call and then adding a comment; personal taste really. Furthermore methods can be overridden in child classes.

barryhughes:
:nerd_face:

PR #291, opened 2016-10-29
Ensure response data is always passed when edit-ticket.tribe is triggered 

bordoni:
_Ref:_ [C#68557](https://central.tri.be/issues/68557)
_Rel:_ #285


bordoni:
Since I'm just including the code created by @helgatheviking no need for a Code Review here.

PR #294, opened 2016-11-04
adding a filter and fixing a filter

zbtirrell:
Adding `tribe_tickets_get_modules` filter that allows a customization like removng a specific ticketing module for Community. Example: https://gist.github.com/969a61a8217cb705b7864ae2b5b3fec6

Also, fixed the existing `tribe_events_tickets_module_name` filter so it works with community events as originally intended. https://central.tri.be/issues/45042

And... a couple code sniffer fixes

### DIFF
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -992,7 +992,7 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 			/**
 			 * Filters the available tickets modules
 			 *
-			 * @var array ticket modules
+			 * @var string[] ticket modules
 			 */
 			return apply_filters( 'tribe_tickets_get_modules', self::$active_modules );
 		}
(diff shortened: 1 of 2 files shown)


src/Tribe/Tickets.php
```diff
@@ -989,7 +989,12 @@ public static function enqueue_frontend_stock_data() {
 		 * @return array
 		 */
 		public static function modules() {
-			return self::$active_modules;
+			/**
+			 * Filters the available tickets modules
+			 *
+			 * @var array ticket modules
```
barryhughes: Optional: might `string[]` be a more descriptive and accurate type description?

barryhughes:
This is a neat idea; but perhaps if we're filtering it we also should update all internal usage of the static `$active_modules` property to go through the `modules()` method instead: otherwise, there will be potential for all kinds of mismatches.

Example: `get_event_attendees()` will (currently) be oblivious of the fact that a module has been removed - or added - via the new filter hook.

barryhughes:
:boat:

PR #296, opened 2016-11-09
trigger pre and post print events

lucatume:
Ticket: https://central.tri.be/issues/66007

This PR:

* triggers an event before and after the attendess report is printed
* applies a css on the fly to make available columsn fill the width

On this second point I'd like to have an opinion and possibly a better approach than hardcoding the number of columns that will be hidden.  
My idea was to append a `data` attribute to the table (?) with the number of columns that will be printed; that number would come from a PHP filter that would default to "all the manage columns unless excluded"; in the print CSS one could then style according to that data attribute. e.g.:

    table.wp-list-table.attendees[data-visible-columns="4"] th,td{
        width: 25%;
    }

    table.wp-list-table.attendees[data-visible-columns="5"] th,td{
        width: 20%;
    }

And so on, but really just  a mock idea.


src/resources/js/tickets-attendees.js
```diff
@@ -32,8 +32,23 @@ var tribe_event_tickets_attendees = tribe_event_tickets_attendees || {};
 			var $pointer = $( AttendeesPointer.target ).pointer( options ).pointer( 'open' ).pointer( 'widget' );
 		}
 
-		$( 'input.print' ).on( 'click', function( e ) {
+		$( 'input.print' ).on( 'click', function() {
+			$( window ).trigger( 'tribe-tickets.attendees-report.before-print' );
```
barryhughes: Hmm; should we stick to the `event.namespace` notation for events (and use something tribe/event related as the namespace)?

lucatume: this is but a ticket related action that will apply to any ticket-able post type, does it make sense to use `events.`?

lucatume:
Ok, I've renamed the events

barryhughes:
:goat:

PR #300, opened 2016-11-18
Fix lagging ticket count and initial state

lucatume:
Ticket: https://central.tri.be/issues/67175

This PR:
* fixes the count of ticketed/unticketed to not lag, when on the admin screen, due to cache
* hence adds some control on the cache
* fixes an initial state issue where the events post type was misspelled


barryhughes:
:ear_of_rice:

PR #305, opened 2016-12-07
Datepicker formatting for the tickets UI

barryhughes:
Let Event Tickets datepickers inherit the same properties as their TEC cousins, if TEC is active.

It would probably be smarter to relocate the `TEC` object so it lives in tribe-common and rename it so it is clearly concerned with datepicker formatting (right now it is used for nothing else that I can see) - but I figured I'd keep the size of the change small given where we are in the release cycle.

https://central.tri.be/issues/69246


bordoni:
👍

PR #309, opened 2016-12-09
Whitelist various Select2 assets

barryhughes:
These are whitelisted in TEC, but if ET's common loads first / if only ET is active we're missing various assets needed by the dropdowns.


barryhughes:
Updated: adds `jquery-timepicker` to the whitelist.

bordoni:
👍

PR #312, opened 2016-12-14
Fix purchaser name search (attendee screen) | #70682

barryhughes:
Attendee search for purchaser names had ceased working as `td.purchaser` no longer exists (`.purchaser_name` is now contained inside of `td.purchaser_info`). This change should resolve that issue.

https://central.tri.be/issues/70682


bordoni:
👍

PR #313, opened 2016-12-14
'Subsubsub' admin view links should always take users to 'page 1'

barryhughes:
If a user is browsing the admin events list and is currently viewing page 10, then the `ticketed` and `unticketed` view links will try to take them to page 10 of those views.

We don't actually want to do this, though, because it's inconsistent with regular views like `published`, `draft` etc and also because there may not be 10 pages of ticketed or unticketed events.

https://central.tri.be/issues/70694


bordoni:
👍

PR #314, opened 2016-12-14
Get more specific with formatting rules for our attendee action buttons

barryhughes:
Increases specificity of our styles targetting the attendee screen action buttons, preventing default WP styles from knocking them for six.

https://central.tri.be/issues/70431


bordoni:
👍

PR #320, opened 2016-12-21
Release/4.3.5

bordoni:
_Ref:_ [C#70645](https://central.tri.be/issues/70645)


src/views/tickets/email-tickets-moved.php
```diff
@@ -12,7 +16,7 @@
  * @var string $ticket_type_name
  * @var array  $affected_tickets
  *
- * @version 4.3
+ * @version 4.3.5
```
barryhughes: Here and elsewhere, were the `@version` tag updates made to satisfy the packaging script? Not a big deal I guess, but then again it doesn't seem like we should alert users that they need to update template overrides just because extra instructions were added to the docblock.

PR #322, opened 2017-01-02
Remove meta details column during attendee export

barryhughes:
The meta details column (which contains the _view details_ link but not the submitted details themselves) is superfluous during CSV/email export of attendee data.

https://central.tri.be/issues/71251


bordoni:
👍

PR #324, opened 2017-01-09
Spotfix/safety check common

barryhughes:
Applies the same safety check (min version of Tribe Common) applied to TEC.


borkweb:
This one looks good too :+1:

PR #330, opened 2017-01-25
Email: add State/Province & Zip to Venue's address

cliffordp:
https://central.tri.be/issues/72475

**Before**
aeacd908e1787d2ede9416d3924faef0 _image 202017-01-25 20at 204 53 31 20pm

**After**
f6020f7787e27823711e68c998816109 _image 202017-01-25 20at 204 55 10 20pm


tr1b0t:
:bangbang: This Pull Request is against a Central ticket for "To Be Determined". It shouldn't be merged until the Central ticket has been changed to "Current Release" or "CR Maintenance Release".

src/views/tickets/email.php
```diff
@@ -369,7 +374,7 @@
 																		<span style="color:#0a0a0e !important; font-family: 'Helvetica Neue', Helvetica, sans-serif; font-size:13px; display:block; margin-bottom:5px;"><?php echo $venue_name; ?></span>
 																		<a style="color:#006caa !important; display:block; margin:0; font-family: 'Helvetica Neue', Helvetica, sans-serif; font-size:13px; text-decoration:underline;">
 																			<?php echo $venue_address; ?><br />
-																			<?php echo $venue_city; ?>
+																			<?php printf( '%s, %s %s', $venue_city, $venue_state, $venue_zip ); ?>
```
borkweb: If both the state and the zip are missing, the following would be printed:

```
Phoenix,
```

We might need to do something like:

```php
<?php
if ( $venue_city && ( $venue_state || $venue_zip ) ) :
  printf( '%s, %s %s', $venue_city, $venue_state, $venue_zip );
else:
  echo $venue_city;
endif;
```

elimn:
What about replacing the entire address output with `tribe_get_full_address()` ? It would take care of edge cases like @borkweb pointed out, plus it would run it through all the usual filters and things. Do you think that would work here?

cliffordp:
I dug into that a bit, @elimn.

Doing this:
`$venue_address = Tribe__Events__Main::instance()->fullAddressString( $event );`

Results in the address not having any `<br>` and also includes the Country. Screenshot:
image

Reading the docblocks for `fullAddress()` (instead of `fullAddressString()`), I don't think we want or need the extra Google microformats and whatever other formatting.

We could replace the first comma with `<br>`, like this:
image
but that doesn't feel any less-hacky than what's already in place here so I'm just going with what we have already (plus, then we don't have to deal with the country at the end).

However, I did notice the address was wrapped in `<a>` without `href` so I added `href` now too.

I also linked the Event Title to the Event Single page.

barryhughes:
Some notes!

* This will help some users, others may not care and others still may prefer how it was previously - so it may be better yet to introduce some sort of filterable approach
* If we change a template as we're doing here we should update the `@version` tag (or add it if it's missing) to reflect when the change was made
* Probably also warrants a changelog entry

elimn:
@cliffordp That makes sense. Ideally, we would probably offer an alternative mode for for `tribe_get_full_address()` that outputs what we need. But that's getting outside the scope of this little ticket.

It's generally bad form to call class methods directly inside of templates. WP itself avoids this as much as possible, and uses functions when it can. But on the other hand, we are further complicating this once simple ticket if we start worrying about this. Thoughts?

Personally, I think we should honor the "show map link" setting here as well. Typically when people disable that it'd be because the map points to an inaccurate location, or at least so I'd guess. So we probably wouldn't want to include the link here either, right?

Regardless of what happens here, I think we need a ticket to involve completely updating this view. I see a lot of variables we could remove, and replace them with function calls.

borkweb:
Nice work here, @cliffordp! I've redirected the Pull Request from `develop` to `release/4.4.2` and in doing so it appears that there's a conflict with `common` and the `readme.txt`. Merging `release/4.4.2` into your branch should help, though you'll have to fix the merge conflict in `readme.txt` manually. To fix the `common/` merge conflict, just pull the latest `release/4.4.2` branch of common and commit that.

PR #335, opened 2017-02-03
qa fixes

trishasalas:
https://central.tri.be/issues/72772


src/resources/postcss/tickets-attendees-print.pcss
```diff
@@ -51,11 +49,35 @@
...
 
 	#tribe-attendees-summary .welcome-panel-column {
-		width: 100%;
 		padding: 0;
 		float: none;
 	}
 
+	#tribe-attendees-summary {
+		margin-bottom: 30px;
+		.welcome-panel-column {
```
barryhughes: Add a line break between the properties of the element itself and the start of any nested selectors?

trishasalas: yes, yes...I know and preach this very thing! A sure sign I need to take a nap! ;)

PR #336, opened 2017-02-03
Also check for _VenueStateProvince meta field

cliffordp:
To fix for https://central.tri.be/issues/72475#note-6


barryhughes:
Bueno

PR #340, opened 2017-02-14
Only execute attendee list logic if not found in the transient

borkweb:
Related: https://github.com/moderntribe/tribe-common/pull/332

See: https://central.tri.be/issues/70485


bordoni:
👍

PR #341, opened 2017-02-15
Fix/73539 confirm rsvp visibility

barryhughes:
We were iterating across the available (RSVP) tickets within the ticket form and setting a truthy flag each iteration, indicating if stock was available or not. 

This same flag was then used to decide if the "Confirm RSVP" button should show, meaning if the _last_ ticket had no availability the button wouldn't show even if other tickets _did_ have availability. This change should resolve that.

https://central.tri.be/issues/73539

### DIFF
--- a/src/views/tickets/rsvp.php
+++ b/src/views/tickets/rsvp.php
@@ -92,7 +92,7 @@ $now = current_time( 'timestamp' );
 		}
 		?>
 
-		<?php if ( $are_any_products_on_sale ): ?>
+		<?php if ( $are_any_products_on_sale ) : ?>
 			<tr class="tribe-tickets-meta-row">
 				<td colspan="4" class="tribe-tickets-attendees">
 					<header><?php esc_html_e( 'Send RSVP confirmation to:', 'event-tickets' ); ?></header>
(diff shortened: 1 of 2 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | You must have a space after a closing parenthesis | [rsvp.php:95](https://github.com/moderntribe/event-tickets/blob/24a29c556ff8c3b7cd74c8ec16b51c24695e8c90/src/views/tickets/rsvp.php#L95)

src/views/tickets/rsvp.php
```diff
@@ -51,6 +52,10 @@
 
 			$is_there_any_product = true;
 			$is_there_any_product_to_sell = $ticket->is_in_stock();
+
+			if ( $is_there_any_product_to_sell ) {
+				$are_any_products_on_sale = true;
```
borkweb: `_on_sale` makes me think we are talking about discounts. But I may be alone on that front. Just a thought `¯\_(ツ)_/¯`. Perhaps `$are_products_available`? Or leave it as-is, too.

barryhughes: I didn't think of that, but you're right. Updating.

borkweb:
🚶 :dash:

PR #345, opened 2017-02-22
#71094 - Add Support For Cost Field Without Tickets

jesseeproductions:
_Ref:_ [C#71094](https://central.tri.be/issues/71094) - Do not show Buy Now or Ticket Counts if No Tickets, but there is a cost field.

### DIFF
--- a/src/template-tags/tickets.php
+++ b/src/template-tags/tickets.php
@@ -123,9 +123,9 @@ if ( ! function_exists( 'tribe_tickets_display_count_and_ticket_button' ) ) {
 		// get an array for ticket and rsvp counts
 		$types = Tribe__Tickets__Tickets::get_ticket_counts( $event_id );
 
-		// if no rsvp or tickets do not show
+		// if no rsvp or tickets return
 		if ( ! $types ) {
-			return;
+			return null;
 		}
 
 		// If we have tickets or RSVP, but everything is Sold Out then display the Sold Out messages
--- a/src/template-tags/tickets.php
+++ b/src/template-tags/tickets.php
@@ -123,6 +123,11 @@ if ( ! function_exists( 'tribe_tickets_display_count_and_ticket_button' ) ) {
 		// get an array for ticket and rsvp counts
 		$types = Tribe__Tickets__Tickets::get_ticket_counts( $event_id );
 
+		// if no rsvp or tickets do not show
+		if ( ! $types ) {
+			return;
+		}
+
 		// If we have tickets or RSVP, but everything is Sold Out then display the Sold Out messages
 		if ( ( $types['tickets']['count'] || $types['rsvp']['count'] ) && ( ! $types['tickets']['available'] && ! $types['rsvp']['available'] ) ) {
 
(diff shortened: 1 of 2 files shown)


masteradhoc:
Hi @jesseeproductions - if you want to make some people happy please update the status here:
https://tribe.uservoice.com/forums/195723-feature-ideas/suggestions/5605710--sold-out-only-x-positions-left-etc-type-label
and https://tribe.uservoice.com/forums/195723-feature-ideas/suggestions/6482848-add-buy-ticket-on-event-list-page

Thanks ;)

src/template-tags/tickets.php
```diff
@@ -123,6 +123,11 @@ function tribe_tickets_display_count_and_ticket_button() {
 		// get an array for ticket and rsvp counts
 		$types = Tribe__Tickets__Tickets::get_ticket_counts( $event_id );
 
+		// if no rsvp or tickets return
+		if ( ! $types ) {
+			return null;
```
barryhughes: 💡 Not introduced as part of this change, so this isn't essential, but I notice the function either echoes a bit of text or else bails out - it doesn't look like it ever returns a string, though (so might be nice to revise or remove the `@return` tag in the docblock).

src/template-tags/tickets.php
```diff
@@ -123,6 +123,11 @@ function tribe_tickets_display_count_and_ticket_button() {
 		// get an array for ticket and rsvp counts
 		$types = Tribe__Tickets__Tickets::get_ticket_counts( $event_id );
 
+		// if no rsvp or tickets return
+		if ( ! $types ) {
```
barryhughes: Is `0 === $types['ticket'] && 0 === $types['rsvp']` a possibility we need to test for?

src/template-tags/tickets.php
jesseeproductions: I do not get what you are saying here. 

The return for tribe_tickets_display_count_and_ticket_button is either a string or null

src/template-tags/tickets.php
jesseeproductions: If there are no tickets it returns an empty array.  both of those indexes will be arrays as well and not equal 0 them selves. 

$types['rsvp']    = array(
				'count'     => 0,
				'stock'     => 0,
				'unlimited' => 0,
				'available' => 0,
			);

jesseeproductions:
Updated the Uservoice entries @masteradhoc

src/template-tags/tickets.php
barryhughes: It can return null/void, but there's no point I see where it returns a string (it can echo a string, but that's not the same thing ... `@return string` is telling devs they can get back a string, which they can't)

jesseeproductions: Oh I get it now.... I guess I was thinking echo is returning, but it is not.

src/template-tags/tickets.php
barryhughes: Got you now :+1:

PR #346, opened 2017-03-02
Rename tribe_rsvp_email_subject filters | spotfix

cliffordp:
`tribe_rsvp_email_subject` filter was used to filter different strings in different contexts

Might this change warrant a changelog entry too?


cliffordp:
Closing in favor of https://central.tri.be/issues/74412, which addresses issues with this and other filters with the same issue.

PR #347, opened 2017-03-02
improve the row action check in link

zbtirrell:
This makes the check-in and Undo check in links that are row actions work identically to the buttons

See: https://central.tri.be/issues/68414


borkweb:
It took me a few moments to see what was going on here. I got it now. Looks good :+1:

PR #360, opened 2017-05-11
Switch to tribe_tickets_get_template_part() for increased flexibility

barryhughes:
Switch to using `tribe_tickets_get_template_part()` for cleaner code and greater flexibility (themers/3rd party devs can then take advantage of sundry hooks and other goodies).

:ticket: [#69660](https://central.tri.be/issues/69660)


borkweb:
:+1:

PR #361, opened 2017-05-11
Remove unneeded and undefined var

barryhughes:
The use of `$break` in the `tickets/email-non-attendance.php` template appears to be a result of it being based upon `tickets/email.php` - however in the case of the former template it is not required and serves no purpose.

:ticket: [#77421](https://central.tri.be/issues/77421)


borkweb:
:+1:

PR #362, opened 2017-05-17
Fix the way "moved to/assigned to" strings are composed | #68432

barryhughes:
Strips duplicate 'moved to' text (the `%3$s` placeholder/`$moved_to`) already contains either 'moved to' or 'assigned to' as appropriate.

:ticket: [#68432](https://central.tri.be/issues/68432)


zbtirrell:
I bet this is really hard to translate in a way that makes sense.

zbtirrell:
That being said, it's better than it was.

PR #363, opened 2017-05-19
:fast_forward: common to 3ea523f60a998e9cde3cccbbc641f93915cb6855

elimn:
Master currently errors out, Common is out of sync with its master. We can either fast forward common here or revert commits. Wasn't sure what was best so I figured I'd at least start us in one direction.


elimn:
Barry is handing this as part of the release process.

PR #365, opened 2017-05-24
#70578 - add documentation and cast event ids as int

jesseeproductions:
_Ref:_ [C#70578](https://central.tri.be/issues/70578)


bordoni:
👍

PR #377, opened 2017-06-09
Fix/77608 csv no time

elimn:
Ticket: https://central.tri.be/issues/77608

### DIFF
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -471,34 +471,24 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 			}
 
 			if ( ! empty( $data['ticket_start_date'] ) ) {
-				$start_hour     = ! empty( $data['ticket_start_hour'] ) ? $data['ticket_start_hour'] : '00';
-				$start_minute   = ! empty( $data['ticket_start_minute'] ) ? $data['ticket_start_minute'] : '00';
-				$start_meridian = ! empty( $data['ticket_start_meridian'] ) ? $data['ticket_start_meridian'] : '';
-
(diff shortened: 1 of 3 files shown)


src/Tribe/Tickets.php
```diff
@@ -471,13 +471,35 @@ public function has_permission( $post, $data, $nonce_action ) {
 			}
 
 			if ( ! empty( $data['ticket_start_date'] ) ) {
-				$meridian           = ! empty( $data['ticket_start_meridian'] ) ? ' ' . $data['ticket_start_meridian'] : '';
-				$ticket->start_date = date( Tribe__Date_Utils::DBDATETIMEFORMAT, strtotime( $data['ticket_start_date'] . ' ' . $data['ticket_start_hour'] . ':' . $data['ticket_start_minute'] . ':00' . $meridian ) );
+				$start_hour     = ! empty( $data['ticket_start_hour'] ) ? $data['ticket_start_hour'] : '00';
```
bordoni: Can I Suggest we use the `Tribe__Utils__Array::get()` for these?

elimn: @bordoni I used exactly that, but my last commit undid it because our new code review bot complains about the dependency. 

I'm green with dependency injection, so maybe I'm wrong about the following. My thinking was, since this is an abstract class I would need to put each of the children in a container. Which is a bit beyond the scope of this little ticket IMO. Unless, you can think of an easier way to account for this?

bordoni: What dependency that it was complaining? I've never saw that one. Merge away I will check.

barryhughes:
Needs enhancement [per team chat](https://tribe.slack.com/messages/C3XGB48QK/).

PR #388, opened 2017-07-06
Fix/77050 limit email list

elimn:
Ticket: https://central.tri.be/issues/77050

### DIFF
--- a/src/admin-views/attendees-email.php
+++ b/src/admin-views/attendees-email.php
@@ -1,4 +1,26 @@
-<div id="tribe-loading"><span></span></div>
+<?php
+
+$users_args = array(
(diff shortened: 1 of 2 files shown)


src/admin-views/attendees-email.php
```diff
@@ -1,4 +1,26 @@
...
+);
+
+if ( ! current_user_can( 'list_users' ) ) {
+	$users_args['include'] = get_current_user_id();
+}
+
+/**
+ * Filters the args for the Email Users Dropdown menu
+ *
+ * @see wp_dropdown_users()
```
barryhughes: Add `@since`

PR #389, opened 2017-07-06
Feature/66141 User Event Confirmations shortcode add shortcode name to shortcode_atts call

niconerd:
https://central.tri.be/issues/66141

### DIFF
--- a/src/Tribe/Shortcodes/User_Event_Confirmation_List.php
+++ b/src/Tribe/Shortcodes/User_Event_Confirmation_List.php
@@ -16,24 +16,22 @@
  *     [tribe-user-event-confirmations limit="16"]
  */
 class Tribe__Tickets__Shortcodes__User_Event_Confirmation_List {
+	protected $shortcode_name = 'tribe-user-event-confirmations';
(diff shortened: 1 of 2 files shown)


src/Tribe/Shortcodes/User_Event_Confirmation_List.php
```diff
@@ -16,24 +16,22 @@
...
  */
 class Tribe__Tickets__Shortcodes__User_Event_Confirmation_List {
+	protected $shortcode_name = 'tribe-user-event-confirmations';
 	protected $params = array();
 
 	/**
-	 * Registers a user event confirmation list shortcode using the specified
-	 * name (or else defaults to "tribe-user-event-confirmations").
-	 *
-	 * @param string $shortcode_name
```
bordoni: Add a `@since` tag describing why this param was removed on this version.

niconerd: thanks @bordoni I was not aware of the use of @since will keep it in mind for the future. Can you re-check if I understood this correctly?

PR #397, opened 2017-07-14
Merge master to bucket

Camwyn:
I cannot push to the bucket branch so here's a PR 🐒 

This adds latest master and makes sure this bucket uses the same bucket for common to avoid confusion.


borkweb:
We always PR the merges anyway :)

PR #400, opened 2017-07-20
Update Tickets.php for Bug #82684

jentheo:
Stock for tickets with global stock enabled were being counted each time through the loop, resulting in a multiplied final stock count being displayed next to the "buy now" button in list view. Added in a condition to skip tickets with global stock during the loop, then added in the global stock count at the end.

### DIFF
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -1051,73 +1051,87 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 		 */
 		public static function get_ticket_counts( $event_id ) {
 
-			$tickets = self::get_all_event_tickets( $event_id );
+		    $tickets = self::get_all_event_tickets( $event_id );
 
-			// if no tickets or rsvp return empty array
-			if ( ! $tickets ) {
-				return array();
+		    // if no tickets or rsvp return empty array
+		    if ( ! $tickets ) {
+			return array();
+		    }
+
+		    /**
+		     * This order is important so we that tickets overwrite RSVP on
+		     * the Buy Now Button on the front-end
+		     */
+		    $types['rsvp']    = array(
+			'count'     => 0,
+			'stock'     => 0,
+			'unlimited' => 0,
+			'available' => 0,
+		    );
+		    $types['tickets'] = array(
+			'count'     => 0, // count of tickets currently for sale
+			'stock'     => 0, // current stock of tickets available for sale
+			'global'    => 0, // global stock ticket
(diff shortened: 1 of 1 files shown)


src/Tribe/Tickets.php
```diff
@@ -1051,73 +1051,87 @@ public static function add_frontend_stock_data( array $tickets ) {
...
-						$types['tickets']['available'] ++;
-					}
-
-					if ( ! $ticket->manage_stock() ) {
-						$types['tickets']['unlimited'] ++;
-						$types['tickets']['available'] ++;
-					}
-				} else {
-					$types['rsvp']['count'] ++;
+		    $tickets = self::get_all_event_tickets( $event_id );
```
barryhughes: Something weird has happened with the initial indent (it's changed from 3 tabs to 2 tabs and a bunch of spaces) - be great to fix this!

src/Tribe/Tickets.php
```diff
@@ -1051,73 +1051,87 @@ public static function add_frontend_stock_data( array $tickets ) {
...
+			    continue;
+			}
+
+			// if ticket and not rsvp add to ticket array
+			if ( 'Tribe__Tickets__RSVP' !== $ticket->provider_class ) {
+			    $types['tickets']['count'] ++;
+
+			    $global_stock_mode = $ticket->global_stock_mode();
+
+			    if ( Tribe__Tickets__Global_Stock::GLOBAL_STOCK_MODE === $global_stock_mode && 0 === $types['tickets']['global'] ) {
```
barryhughes: This is a nice change and avoids unnecessary function calls - but we should 'yoda-ify' it, ie for the first part we should switch to:

`$global_stock_mode === Tribe__Tickets__Global_Stock::GLOBAL_STOCK_MODE`

src/Tribe/Tickets.php
```diff
@@ -1051,73 +1051,85 @@ public static function add_frontend_stock_data( array $tickets ) {
 		 */
 		public static function get_ticket_counts( $event_id ) {
 
-			$tickets = self::get_all_event_tickets( $event_id );
+		    	$tickets = self::get_all_event_tickets( $event_id );
```
barryhughes: I hate to say it ... but throughout the change the initial indent is still out of whack. Right now we have `TAB TAB 4xSPACES TAB` but what we _want_ is `TAB TAB TAB` (`"draw_white_space": "all"` can help if you're using Sublime for example)

barryhughes:
@jentheo - some further whitespace/formatting fixes and I think we'll be good to go here :+1:

PR #415, opened 2017-07-29
Moving all asset register/enqueue to Assets.php via Service_Provider

Camwyn:
This utilizes the service provider in common, and moves all asset handling to `src/Tribe/Assets.php`

Ticket: https://central.tri.be/issues/83302


src/Tribe/Assets.php
```diff
@@ -0,0 +1,107 @@
...
+
+	/**
+	 * Enqueue scripts for admin views
+	 *
+	 * @since TBD
+	 */
+	public function admin_enqueue_scripts() {
+		global $post;
+
+		$modules = apply_filters( 'tribe_events_tickets_modules', null );
```
bordoni: Add Docblock.

src/Tribe/Assets.php
```diff
@@ -0,0 +1,114 @@
...
+	 * @since TBD
+	 */
+	public function admin_enqueue_scripts() {
+		global $post;
+
+		/**
+		 * Filter the array of module names.
+		 *
+		 * @since TBD
+		 *
```
bordoni: Add `@param`.

src/Tribe/Assets.php
```diff
@@ -0,0 +1,114 @@
...
+							'data' => array(
+								'decimal' => $decimal,
+								'decimal_error' => __( 'Please enter in without thousand separators and currency symbols.', 'event-tickets' ),
+							),
+						),
+					),
+				)
+			);
+
+			wp_enqueue_script( 'tribe-bumpdown' );
```
bordoni: Why not add this as a dependency of something?

src/Tribe/Assets.php
```diff
@@ -0,0 +1,116 @@
...
+		global $post;
+
+		/**
+		 * Filter the array of module names.
+		 *
+		 * @since TBD
+		 *
+		 * @param array the array of modules
+		 *
+		 * See event-tickets/src/Tribe/Tickets.php->modules()
```
barryhughes: Optional - but we could make this into a `@see` tag

src/Tribe/Assets.php
```diff
@@ -0,0 +1,116 @@
...
+		 */
+		$modules = apply_filters( 'tribe_events_tickets_modules', null );
+
+		// For the metabox
+		if ( ! empty( $post ) &&
```
barryhughes: Let's bump the first part of the conditional to the next line so we're following [this standard](http://moderntribe.github.io/products-engineering/php/#php-control-multi) - we should also move hte `&&` operator to the start of the line that follows - ie:

```php
if (
	$first
	&& $second
	&& $third
) {
	// Body
}
```

src/Tribe/Assets.php
```diff
@@ -0,0 +1,116 @@
...
+			 ! empty( $modules ) &&
```
barryhughes: Ditto, move `&&` to the start of the next line

src/Tribe/Assets.php
```diff
@@ -0,0 +1,116 @@
...
+			$locale  = localeconv();
+			$decimal = isset( $locale['decimal_point'] ) ? $locale['decimal_point'] : '.';
+
+			/**
+			 * Filter the decimal point character used in the price
+			 * @param string $decimal the decimal character to filter
+			 *
+			 * @since TBD
+			 *
+			 * @param array the array of modules
```
barryhughes: C+P error? Looks like this should probably be `@param string $decimal`

Camwyn: Yes, y'know I can't find anywhere we _add_ this filter, so 🤞 it's a string 😉

PR #420, opened 2017-08-07
Feature/81726 conditional warnings

jesseeproductions:
_Ref:_ [C#81726](https://central.tri.be/issues/81726)


jesseeproductions:
I believe updating common here is ok, as the bucket refresh is behind a bit on the common version and this gets it up to date.

PR #433, opened 2017-08-21
Ticket Editor QA

Camwyn:
Tickets: https://central.tri.be/issues/80625 and https://central.tri.be/issues/85567

Required by https://github.com/moderntribe/event-tickets-plus/pull/365


src/Tribe/Ticket_Object.php
```diff
@@ -660,5 +698,16 @@ public function get_event() {
 
 			return null;
 		}
+
+		/**
+		 *
+		 */
+		public function show_description() {
+				$show = ( metadata_exists( 'post', $this->ID, $this->show_description_key ) ) ?
```
borkweb: If the ternary is too long to be readable on one line, might as well make it readable by switching to an `if` statement.

```php
$show = true;

if ( metadata_exists( 'post', $this->ID, $this->show_description_key ) ) {
  $show = get_post_meta( $this->ID, $this->show_description_key, true );
}
```

If you really like the ternary, then formatting like this would be preferred:

```php
$show = metadata_exists( 'post', $this->ID, $this->show_description_key )
  ? get_post_meta( $this->ID, $this->show_description_key, true )
  : true;
```

src/Tribe/Ticket_Object.php
```diff
@@ -660,5 +698,16 @@ public function get_event() {
 
 			return null;
 		}
+
+		/**
+		 *
```
borkweb: Add text to the docblock?

src/Tribe/Ticket_Object.php
```diff
@@ -660,5 +698,16 @@ public function get_event() {
...
+		 */
+		public function show_description() {
+				$show = ( metadata_exists( 'post', $this->ID, $this->show_description_key ) ) ?
+				get_post_meta( $this->ID, $this->show_description_key, true ) :
+				true;
+
+				return apply_filters( 'tribe_show_ticket_description', $show );
```
borkweb: Add a docblock for this?

borkweb: Also: event-tickets actions/filters have a lot of legacy names. However, we do have a [naming standard](http://moderntribe.github.io/products-engineering/plugins/) that says we should prefix action/filter names with `tribe_tickets_`.

src/Tribe/Ticket_Object.php
Camwyn: Trying to be mysterious... :no_mouth:

PR #450, opened 2017-09-01
Fix/86616 fix rsvp products available

vicskf:
https://central.tri.be/issues/86616

### DIFF
--- a/src/views/tickets/rsvp.php
+++ b/src/views/tickets/rsvp.php
@@ -65,6 +65,11 @@ $now = current_time( 'timestamp' );
 
 			$is_there_any_product = true;
 			$is_there_any_product_to_sell = $ticket->is_in_stock();
+			
+			if ( $is_there_any_product_to_sell ) {
+				$are_products_available = true;
+			}
+			
 			?>
 			<tr>
 				<td class="tribe-ticket quantity" data-product-id="<?php echo esc_attr( $ticket->ID ); ?>">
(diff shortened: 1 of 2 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Whitespace found at end of line | [rsvp.php:68](https://github.com/moderntribe/event-tickets/blob/417a2924d97422eb8d5d2512db6af228392e09d6/src/views/tickets/rsvp.php#L68)
:x: | Whitespace found at end of line | [rsvp.php:72](https://github.com/moderntribe/event-tickets/blob/417a2924d97422eb8d5d2512db6af228392e09d6/src/views/tickets/rsvp.php#L72)

src/views/tickets/rsvp.php
```diff
@@ -108,7 +113,7 @@ class="tribe-ticket-quantity"
 		}
 		?>
 
-		<?php if ( $is_there_any_product_to_sell ) : ?>
+		<?php if ( $are_products_available ) : ?>
```
Camwyn: Why are we testing a boolean to create a boolean with the same value? This is the only place we use this, but `is_in_stock()` (and thus `$is_there_any_product_to_sell`)  should return a boolean. Feels like this is a band-aid for another issue?

Camwyn: There _will_ be an easier way to do this check later (post TER) but this is 👍  for now.

PR #454, opened 2017-09-09
#62976 Unhook Email Content-Type Filter After Email is Sent

ggwicz:
_Ref:_ [C#62976](http://central.tri.be/issues/62976)


ggwicz:
Resolved conflicts, needs another review/approval unfortunately.

readme.txt
```diff
@@ -235,6 +235,7 @@ The plugin is produced by [Modern Tribe Inc](http://m.tri.be/18uc).
 
 = [4.5.6] TBD =
 
+* Fix - Prevent occasional issue with email content-type not being reset after Tickets emails were sent. Props to @jappleton in the forums for reporting this! [62976]
```
barryhughes: I think I get why you worded it this way, but I wonder if it would be smoother to write _"after ticket emails are sent"_ (vs _"after Tickets emails were sent")_

barryhughes: need not block merge though

PR #457, opened 2017-09-11
Added is_edit_page() for view tickets, and support for CPTs

elimn:
Ticket: https://central.tri.be/issues/67570

The meat of this is setting the post type of our query, since it defaults to blog posts. I also needed to hide the purchase tickets form on the CPT tickets edit page. Everything else is incidental code cleanup that happened as I was playing around trying to find the problem.


elimn:
This has already been code review, I just added a readme entry.

PR #458, opened 2017-09-11
#70750 Prevent Weird Chrome Error

ggwicz:
_Ref:_ [C#70750](http://central.tri.be/issues/70750)


barryhughes:
Looks good - should we have a changelog entry though?

ggwicz:
yeah, I had that change in place but forgot to commit it!

ggwicz:
Note: There are no "props" for this change; this one comes from Sydney, not the community.

PR #461, opened 2017-09-12
#80629 - add filter for default capacity type

jesseeproductions:
_Ref:_ [C#80629](https://central.tri.be/issues/80629)


jesseeproductions:
Another review please as I had to fix a conflict with minfied js

src/Tribe/Assets.php
```diff
@@ -69,6 +69,18 @@ public function admin_enqueue_scripts() {
...
 
+			/**
+			 *  Filter Default Ticket Capacity Type
+			 *
+			 * @since TBD
+			 *
+			 * @param string 'global'
+			 *
+			 * @return string (global,own,unlimited)
+			 *
```
ggwicz: not a blocker, but you can remove `@return` on hook docblocks

also I think `@param` should be more like `@param string $mode`

but again these aren't blockers so "fixing" is optional — tagging this both `merge` and `enhance` to reflect that

Camwyn: I'll second the docblock comments and second that they're non-blockers 😉

PR #470, opened 2017-09-14
Ensure we hide the global cap field once we've got a value.

Camwyn:
Ticket: https://central.tri.be/issues/80625

Also cleaned up some unused/non-working (don't want them cached) vars.

Requires: https://github.com/moderntribe/event-tickets-plus/pull/392


src/resources/js/tickets.js
```diff
@@ -813,7 +816,7 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
 		 * cases).
  		*/
 		$( 'input[type="submit"]' ).click( function() {
```
bordoni: Should we remove this event?

PR #477, opened 2017-09-15
Set the admin-menu parent-item to the parent post type of a ticket on…

elimn:
Ticket: https://central.tri.be/issues/81284

For some reason this variable gets set to blank sometime between where it gets set in WP ( /wp-admin/admin.php:124 ) and here. So I just copied the code from WP which sets this:

`$typenow = $_REQUEST['post_type'];`

Since that what WP uses, there is no security vulnerabilities. $typenow is whitelisted, so if it doesn't match a known post_type WP errors out.

### DIFF
--- a/src/Tribe/Tickets_Handler.php
+++ b/src/Tribe/Tickets_Handler.php
@@ -178,6 +178,9 @@ class Tribe__Tickets__Tickets_Handler {
 		add_action( 'admin_enqueue_scripts', array( $this, 'attendees_page_load_pointers' ) );
 		add_action( 'load-' . $this->attendees_page, array( $this, 'attendees_page_screen_setup' ) );
 
+		// Set the post type to get highlighted in the admin menu.
+		add_filter( 'admin_head', array( $this, 'set_typenow' ) );
+
 		/**
 		 * This is a workaround to fix the problem
 		 *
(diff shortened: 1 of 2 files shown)


src/Tribe/Tickets_Handler.php
```diff
@@ -789,4 +792,15 @@ public function get_attendee_report_link( $post ) {
...
+	/**
+	 * Sets the `$typenow` var to the post_type
+	 *
+	 * Causes parent post type to be highlighted in admin menus.
+	 *
+	 * @see `admin_head` action
+	 */
+	public function set_typenow() {
+		global $typenow;
+		$typenow = $_REQUEST['post_type'];
```
borkweb: Can you bail out of this function if `empty( $_REQUEST['post_type'] )`?

elimn: Sure. That will only happen if it's not attached to `admin_head`, but just in case I can add that check.

src/Tribe/Tickets_Handler.php
```diff
@@ -789,4 +792,19 @@ public function get_attendee_report_link( $post ) {
...
+	 * Sets the `$typenow` var to the post_type
+	 *
+	 * Causes parent post type to be highlighted in admin menus.
+	 *
+	 * @see `admin_head` action
+	 */
+	public function set_typenow() {
+		global $typenow;
+
+		if ( ! empty( $_REQUEST['post_type'] ) ) {
```
bordoni: Why not use `tribe_get_request_var()`? Avoid usage of `_COOKIE`, will only check `_GET` and `_POST`

elimn: WordPress itself sets `$typenow` based on `$_COOKIE` as well. As mentioned in the description I actually copied that line right from WP core (  /wp-admin/admin.php:124 ) Either way I will switch to tribe_get_request_var(), even at the risk of introducing a bug. I doubt this gets set via cookies very often.

PR #485, opened 2017-09-22
Resolve some Validation for Datepickers

bordoni:
_Ref:_ [C#87205](http://central.tri.be/issues/87205)
---


src/admin-views/edit_admin_panel.php
```diff
@@ -113,9 +122,9 @@ class="ticket_field ticket_provider"
 				do_action( 'tribe_events_tickets_metabox_edit_main', $post_id, null ); ?>
 			</section>
 			<div class="accordion">
-				<button class="accordion-header tribe_advanced_meta" type="button">
+				<div class="accordion-header tribe_advanced_meta">
```
Camwyn: The accordion headers are buttons on purpose - accessibility. 
If you click it and go somewhere else it's a link. 
If you click it and something happens on the page, it's a button.
Nothing else should be clickable (unless it's a purely visual effect)

bordoni: I needed it to change due to how Validation.js treats buttons inside of a form, it transforms it into validation points. But I can work around that. Didn't know it was specifically for that

bordoni:
Ready for review again

PR #486, opened 2017-09-25
Use default module information to check correct option

lucatume:
Ticket: https://central.tri.be/issues/81217

in the `bucket/ticket-editor-refresh` the `Tribe__Tickets__Tickets::get_default_module` method is added and in this code is used.

### DIFF
--- a/src/admin-views/meta-box.php
+++ b/src/admin-views/meta-box.php
@@ -128,8 +128,15 @@ $modules = Tribe__Tickets__Tickets::modules();
 					<td width="20%"><label for="ticket_provider"><?php esc_html_e( 'Sell using:', 'event-tickets' ); ?></label></td>
 					<td>
 						<?php
-						$checked = true;
+						$checked        = true;
+						$default_module = null;
+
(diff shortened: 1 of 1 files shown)


lucatume:
Putting this on hold while waiting for reply from @borkweb to https://central.tri.be/issues/81217#note-15

src/admin-views/meta-box.php
```diff
@@ -128,8 +128,15 @@
 					<td width="20%"><label for="ticket_provider"><?php esc_html_e( 'Sell using:', 'event-tickets' ); ?></label></td>
 					<td>
 						<?php
-						$checked = true;
+						$checked        = true;
+						$default_module = null;
+
+						if ( method_exists( 'Tribe__Tickets__Tickets', 'get_default_module' ) ) {
```
borkweb: Is this `if` statement needed? We are currently in the event-tickets so this method should be known to exist.

lucatume: The method does not exist in the class currently; it's a TER method

PR #488, opened 2017-09-26
Ticket Editor: Spinner and Leave Warning

bordoni:
_Ref:_ [C#80641](http://central.tri.be/issues/80641)
---


src/resources/postcss/tickets-refresh.pcss
```diff
@@ -45,6 +45,29 @@
 	margin-top: 12px;
 }
 
+
+.tribe-tickets-editor-block {
+	display: none;
```
Camwyn: Alphabetize.

src/resources/postcss/tickets-refresh.pcss
```diff
@@ -45,6 +45,29 @@
...
+	position: absolute;
+	background: rgba(255, 255, 255, .7);
+	top: 0;
+	right: 0;
+	bottom: 0;
+	left: 0;
+	z-index: 100000;
+
+	.spinner {
+		position: absolute;
```
Camwyn: Alphabetize

src/resources/js/tickets.js
```diff
@@ -329,6 +325,12 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
 
 			// then show the one we want
 			$panel.attr( 'aria-hidden', false );
+
+			if ( ! $panel.is( $base_panel ) ) {
+				$( window ).on( 'beforeunload', beforeUnload );
+			} else {
+				$( window ).off( 'beforeunload' );
```
Camwyn: You should add the handler here so you're not removing _allthethings_?(there may be third-party handlers)

src/resources/js/tickets.js
```diff
@@ -424,6 +426,26 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
 
+		/**
+		 * If the user attempts to nav away without saving global stock setting
+		 * changes then try to bring this to their attention!
+		 */
+		var beforeUnload = function( event ) {
+			var returnValue = false;
+
+			// If we are not on the base panel we alert the user about leaving
+			if ( $metabox.find( '.ticket_panel' ).filter( '[aria-hidden="false"]' ).is( $base_panel ) ) {
```
Camwyn: Would this be more performant (no DOM hunting) if we just did (rough code)
```
if ( false === $base_panel.attr( 'aria-hidden' ) ) {
```

src/admin-views/meta-box.php
```diff
@@ -16,6 +16,10 @@
 $attendees_url = Tribe__Tickets__Tickets_Handler::instance()->get_attendee_report_link( get_post( $post_id ) );
 ?>
 
+<div class="tribe-tickets-editor-block">
```
Camwyn: Would it be better if this was `tribe-tickets-editor-blocker` as "block" is pretty generic and sounds like a container/wrapper?

src/resources/js/tickets.js
```diff
@@ -8,6 +8,11 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
 	var $tribe_tickets                   = $( document.getElementById( 'tribetickets' ) );
 	var $tickets_container               = $( document.getElementById( 'event_tickets' ) );
 	var $post_id                         = $( document.getElementById( 'post_ID' ) );
+
+	var $metabox                         = $tribe_tickets.find( '.inside' );
```
Camwyn: This variable seems...extraneous? While it narrows down the later searches some, they could all be run from `$tribe_tickets` - no?

You could even do `$tribe_tickets.find( '.inside > .tribe-tickets-editor-block' );`

Up for discussion! 😄

bordoni: Yeah we could mostly run it from `$tribe_tickets` Since it's done only once in a very narrow scope. It allows for better control later on if we need.

src/resources/js/tickets.js
bordoni: So the idea here is that if in the future the `$base_panel` is replaced it would still trigger the message.

bordoni:
Ready for another pair of 👁 👁

PR #490, opened 2017-09-27
Generate a cookie-based invoice number (and purge it after complete)

borkweb:
This does not, however, cause the sandbox to _observe_ the [invoice number](https://developer.paypal.com/docs/classic/paypal-payments-standard/integration-guide/Appx_websitestandard_htmlvariables/) despite it being passed/regenerated after successful completion. The issue in the PayPal sandbox that is causing subsequent attempts at checking out to fail appears to be a bug. A quick clearing of Domain Cookies on the PayPal side resolves the issue.

See: https://central.tri.be/issues/81566


lucatume:
Go for it!

PR #502, opened 2017-10-04
Only disable Tribe Commerce PayPal rows - leave other ticket forms alone

borkweb:
See: https://central.tri.be/issues/90209


lucatume:
Approving and merging this into the bucket branch.

PR #504, opened 2017-10-05
Re-organize Tickets.js to use $document as base

bordoni:
_Ref:_ [C#80632](http://central.tri.be/issues/80632)
---


src/resources/js/tickets.js
```diff
@@ -193,879 +132,962 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
+
+		if ( ! force_rsvp && $checked_provider.length > 0 ) {
+			provider_id = $checked_provider.val() + '_radio';
+		 }
+
+		$( document.getElementById( provider_id ) ).prop( 'checked', true ).trigger('change');
+	}
+
+	/**
+	 * Returns the current global capacity (via the settings panel.
```
barryhughes: Missing `)`

src/resources/js/tickets.js
```diff
@@ -193,879 +132,962 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	function get_global_cap() {
+		var $global_capacity_edit = $( document.getElementById( 'settings_global_capacity_edit' ) )
```
barryhughes: No kittens will die because of this, but it would be good style to terminate with a `;`

src/resources/js/tickets.js
```diff
@@ -193,879 +132,962 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
+
+		$( document.getElementById( provider_id ) ).prop( 'checked', true ).trigger('change');
+	}
+
+	/**
+	 * Returns the current global capacity (via the settings panel.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
```
barryhughes: Hate to be a pedant - but this could also return type `number` (looking at the last line of the function, we can return `0` depending on how the ternary is evaluated)

src/resources/js/tickets.js
```diff
@@ -193,879 +132,962 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
-			},
-			onSelect       : function( dateText, inst ) {
-				var the_date = $.datepicker.parseDate( 'yy-mm-dd', dateText );
-				if ( inst.id === 'ticket_start_date' ) {
-					$ticket_end_date.datepicker( 'option', 'minDate', the_date );
-				} else {
-					$ticket_start_date.datepicker( 'option', 'maxDate', the_date );
-				}
+		var $toggle_link      = $history.find( 'a.toggle-history' );
+		var $toggle_link_text = $toggle_link.find( 'span' );
```
barryhughes: Don't think we use `$toggle_link_text` - can we get rid of it?

src/resources/js/tickets.js
```diff
@@ -193,879 +132,962 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
+		var $history_list     = $history.find( 'ul' );
```
barryhughes: Ditto here - seems like we define `$history_list` but don't use it after this point

src/resources/js/tickets.js
```diff
@@ -193,879 +132,962 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
+	 * @param  object ($base_panel) $panel jQuery object containing the panel we want to switch to
+	 * @return void
+	 */
+	function show_panel( e, $panel ) {
+		if ( e ) {
+			e.preventDefault();
+		}
+
+		// this way if we don't pass a panel, it works like a 'reset'
+		if ( undefined == $panel ) {
```
barryhughes: Should this be `'undefined' === typeof $panel`?

src/resources/js/tickets.js
```diff
@@ -193,879 +132,962 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
+			if ( $history.hasClass( '_show' ) ) {
+				$history.removeClass( '_show' );
+			} else {
+				$history.addClass( '_show' );
 			}
+		} );
+	}
+
+	/**
+	 * Switch from one panel to another
```
barryhughes: If this function is also new, should we add a `@since` tag?

src/resources/js/tickets.js
```diff
@@ -193,879 +132,962 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string optional notice to prepend to the ticket table
+	 * @param bool (true) flag for panel swap
+	 * @return void
+	 */
+	function refresh_panels( notice, swap ) {
+		// make sure we have this for later (default to true)
+		swap = undefined === swap ? true : false;
```
barryhughes: * Should we use `typeof` here to test if it is undefined?
* The ternary will result in true or false ... can't we simplify to just capture the result of the test? (ie, something like `swap = ( 'undefined' === typeof swap );`

barryhughes:
Not sure if we have an official position on this, but throughout this change shouldn't we use JSDoc style annotations vs PHPDoc style, ie `@param {number} foo` vs `@param int foo` and so on?

bordoni:
ready for another look.

src/resources/js/tickets.js
```diff
@@ -178,9 +106,7 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
 			return;
 		}
 
-		var $toggle_link      = $history.find( 'a.toggle-history' );
-		var $toggle_link_text = $toggle_link.find( 'span' );
-		var $history_list     = $history.find( 'ul' );
+		var $toggle_link = $history.find( 'a.toggle-history' );
```
barryhughes: We aren't using `$toggle_link` after this - could remove?

PR #507, opened 2017-10-06
Resolve bugs related to Table and Panel Swap

bordoni:
_Ref:_ [C#90281](http://central.tri.be/issues/90281)
---


src/resources/js/tickets-tables.js
```diff
@@ -50,19 +50,18 @@
 
 	function tribe_toggle_sortable() {
 		if ( window.matchMedia( '( min-width: 786px )' ).matches ) {
-			if ( ! $( $table ).hasClass( 'ui-sortable' ) ) {
```
Camwyn: `$table` should already be a jQuery object, not sure why we need to wrap it in `$()`?

bordoni: We don't, that is why I removed it.

Camwyn: Ah, ignore me - I saw that as adding it... 🤦‍♂️

PR #509, opened 2017-10-07
Review of the Capacity and Stock

bordoni:
_Ref:_ [C#90269](http://central.tri.be/issues/90269)
---

PR #510, opened 2017-10-12
Prevent users from clicking on Update button when Not on Base Panel

bordoni:
_Ref:_ [C#80624](http://central.tri.be/issues/80624)
---
_Depends on #509_


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$html' | [tickets.php:780](https://github.com/moderntribe/event-tickets/blob/10a9c11a687f7665c3ecb97c6a0d3f6afd111e4a/src/template-tags/tickets.php#L780)
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$ticket' | [rsvp.php:89](https://github.com/moderntribe/event-tickets/blob/10a9c11a687f7665c3ecb97c6a0d3f6afd111e4a/src/views/tickets/rsvp.php#L89)
:warning: | Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$ticket' | [rsvp.php:89](https://github.com/moderntribe/event-tickets/blob/10a9c11a687f7665c3ecb97c6a0d3f6afd111e4a/src/views/tickets/rsvp.php#L89)

src/Tribe/Global_Stock.php
```diff
@@ -62,10 +99,15 @@ public function __construct( $post_id ) {
 	 * As a convenience, false can be passed to this method to disable rather
 	 * than enable global stock.
 	 *
-	 * @param bool $yes
+	 * @since  4.1
+	 * @since  TBD Added a return so we can check if it was enabled correctly
+	 *
+	 * @param  bool $yes
+	 *
+	 * @return bool
```
barryhughes: Should be `int|bool`? (If the `self::GLOBAL_STOCK_ENABLED` key didn't previously exist, `update_post_meta()` will return the ID)

src/Tribe/Global_Stock.php
```diff
@@ -74,41 +116,69 @@ public function enable( $yes = true ) {
 	 * As a convenience, false can be passed to this method to enable rather
 	 * than disable global stock.
 	 *
-	 * @var bool $yes
+	 * @since  4.1
+	 * @since  TBD Added a return so we can check if it was enabled correctly
+	 *
+	 * @param  bool $yes
+	 *
+	 * @return bool
```
barryhughes: Ditto re `int|bool`

src/Tribe/Global_Stock.php
```diff
@@ -74,41 +116,69 @@ public function enable( $yes = true ) {
...
 	/**
 	 * Sets the global stock level for the current post.
 	 *
-	 * @param int $quantity
+	 * @since  4.1
+	 * @since  TBD  Added a Return
+	 *
+	 * @param  int $quantity
+	 *
+	 * @return bool
```
barryhughes: `int|bool` since we return the result of `update_post_meta()`

src/Tribe/RSVP.php
```diff
@@ -927,10 +927,39 @@ public function save_ticket( $event_id, $ticket, $raw_data = array() ) {
...
+			$data['capacity'] = -1;
+		}
+
+		// The only available value lower than zero is -1 which is unlimited
+		if ( 0 > $data['capacity'] ) {
+			$data['capacity'] = -1;
+		}
+
+		// Fetch the stock if defined, otherwise use Capacity field
+		$data['stock'] = Tribe__Utils__Array::get( $data, 'stock', $data['capacity'] );
```
barryhughes: Not sure it's needed, but is there value in either performing a `trim()` (like we do in similar code above) or else softening the next condition to test using `empty()` vs strictly testing for an empty string?

src/Tribe/Ticket_Object.php
```diff
@@ -401,91 +421,49 @@ public function is_in_stock() {
...
-
-		/**
-		 * Method to get the total `stock` property of the Object.
+		 *	@since  TBD
 		 *
-		 * Returns the current ticket total capacity: either an integer or a
-		 * string if stock is unlimited.
-		 *
-		 * @return int|string
+		 * @return  string|int
```
barryhughes: Just `int` here (seems like we cast to int any time we return from this func)?

src/Tribe/Tickets_Handler.php
```diff
@@ -531,22 +365,25 @@ public function get_event_shared_tickets( $post = null ) {
...
+			return $ticket_list;
+		}
 
-				// Failsafe - should not include unlimited tickets
-				if ( $this->is_unlimited_ticket( $ticket ) ) {
-					continue;
-				}
+		foreach ( $tickets as $ticket ) {
+			$stock_mode = $ticket->global_stock_mode();
+			if ( empty( $stock_mode ) || Tribe__Tickets__Global_Stock::OWN_STOCK_MODE == $stock_mode ) {
```
barryhughes: Yodify second part of condition

src/resources/js/tickets.js
```diff
@@ -31,8 +37,24 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
+		'dd/mm/yy',
+		'm-d-yy',
+		'mm-dd-yy',
+		'd-m-yy',
+		'dd-mm-yy',
+		'yy.mm.dd',
+		'mm.dd.yy',
+		'dd.mm.yy'
+	];
+	var dateFormat = datepicker_formats[ 0 ];
```
barryhughes: Don't need to change it necessarily, but it's weird that we're using camelCase and snake_case for different vars.

src/template-tags/tickets.php
```diff
@@ -707,3 +707,78 @@ function tribe_tickets_has_meta_fields( $id, $context = null ) {
...
+		$html[] = esc_html( tribe( 'tickets.handler' )->unlimited_term );
+	} else {
+		$html[] = esc_html( $number );
+	}
+
+	if ( $show_parens ) {
+		$html[] = ')';
+	}
+
+	$html = implode( '', $html );
```
barryhughes: Hmm ... is building the string this way good i18n practice? I realize the `unlimited_term` is already going to be translated, but word order may not be the same when it's used here. Could a translatable, sprintf()'d string be a better way to do this than joining array elements?

src/Tribe/Tickets_Handler.php
bordoni: That is the YODA version. You cannot assign values to Constants.

src/template-tags/tickets.php
bordoni: In this case is better to avoid strings there because we are using the `()` as a part of the UI to display Shared capacity.

bordoni:
@barryhughes fixed your comments on #509

src/template-tags/tickets.php
barryhughes: Not sure I understand that argument but I also don't have as good a view as you of where this will be used :smile:

PR #515, opened 2017-10-13
TImepicker Moved into Common

bordoni:
_Ref:_ [C#84936](http://central.tri.be/issues/84936)
---
Depends on moderntribe/tribe-common#550


src/admin-views/tickets-advanced.php
```diff
@@ -30,7 +29,6 @@ class="tribe-datepicker tribe-field-start_date ticket_field"
 			<span class="datetime_seperator"> <?php esc_html_e( 'at', 'event-tickets' ); ?> </span>
 			<input
 				autocomplete="off"
-				tabindex="<?php tribe_events_tab_index(); ?>"
```
ggwicz: is removing tabindex necessary and intentional? i'm assuming it is, so marking this as `merge` and this is not a blocker — but am curious about this and just wanted to flag it in case it's useful...

bordoni: Because this function is a Events one, which will throw fatals when we don't have TEC.

PR #525, opened 2017-10-16
Styles for Community Tickets fixed

bordoni:
_Ref:_ [C#81724](http://central.tri.be/issues/81724)
---


src/resources/postcss/accordion.pcss
```diff
@@ -6,9 +6,10 @@
 .accordion-header {
 	background: none;
 	border: 0;
+    color: inherit;
```
borkweb: Tabbing

PR #532, opened 2017-10-19
Inventory is now Event Aware

bordoni:
_Ref:_ [C#91132](http://central.tri.be/issues/91132)
---

### DIFF
--- a/src/Tribe/Ticket_Object.php
+++ b/src/Tribe/Ticket_Object.php
@@ -38,15 +38,6 @@ if ( ! class_exists( 'Tribe__Tickets__Ticket_Object' ) ) {
 		public $show_description = true;
 
 		/**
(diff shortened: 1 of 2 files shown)


src/Tribe/Ticket_Object.php
```diff
@@ -416,6 +385,37 @@ public function is_in_stock() {
...
+		 *
+		 * @param boolean $manages_stock Boolean to set stock management state
+		 * @return boolean
+		 */
+		public function manage_stock( $manages_stock = null ) {
+
+			if ( null !== $manages_stock ) {
+
+				// let's catch a truthy string and consider it false
+				if ( 'no' === $manages_stock ) {
```
ggwicz: if looking for a truthy string any reason why not using `tribe_is_truthy`?

PR #533, opened 2017-10-20
Resolving the Show Description Ticket

bordoni:
_Ref:_ [C#89247](http://central.tri.be/issues/89247)
---

### DIFF
--- a/src/Tribe/RSVP.php
+++ b/src/Tribe/RSVP.php
@@ -926,6 +926,10 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 			return false;
 		}
 
+		// Updates if we should show Description
+		update_post_meta( $ticket->ID, tribe( 'tickets.handler' )->key_show_description, $ticket->show_description );
+
+		// Adds RSVP price
 		update_post_meta( $ticket->ID, '_price', $ticket->price );
 
 		$data = Tribe__Utils__Array::get( $raw_data, 'tribe-ticket', array() );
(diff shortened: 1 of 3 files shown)


src/Tribe/Tickets.php
```diff
@@ -882,9 +882,8 @@ public function ajax_refresh_settings() {
 			$ticket->description      = isset( $data['ticket_description'] ) ? esc_html( $data['ticket_description'] ) : null;
 			$ticket->price            = ! empty( $data['ticket_price'] ) ? filter_var( trim( $data['ticket_price'] ), FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION | FILTER_FLAG_ALLOW_THOUSAND ) : 0;
 			$ticket->purchase_limit   = isset( $data['ticket_purchase_limit'] ) ? absint( $data['ticket_purchase_limit'] ) : apply_filters( 'tribe_tickets_default_purchase_limit', 0, $ticket->ID );
-
-			$show_description = isset( $data['ticket_show_description'] ) ? absint( $data['ticket_show_description'] ) : 0;
-			$ticket->show_description = $ticket->show_description();
+			$ticket->show_description = isset( $data['ticket_show_description'] ) ? 'yes' : 'no';
```
zbtirrell: everywhere else this seems to use `1` or `0` (maybe `true`/`false` - why introduce `yes` and `no` here?

zbtirrell: why not call `$ticket->show_description();` here like in Tickets Plus?

src/Tribe/RSVP.php
```diff
@@ -926,6 +926,10 @@ public function save_ticket( $event_id, $ticket, $raw_data = array() ) {
 			return false;
 		}
 
+		// Updates if we should show Description
+		update_post_meta( $ticket->ID, tribe( 'tickets.handler' )->key_show_description, $ticket->show_description );
```
zbtirrell: same concern here as in https://github.com/moderntribe/event-tickets-plus/pull/436

src/Tribe/Tickets.php
bordoni: Because on this case it is where we initially start to save the data that is sent to ET+

bordoni: For consistency.  I used `yes`/`no` like WooCommerce uses.

PR #537, opened 2017-10-20
Update Stock Levels for Event Shared Capacity

bordoni:
_Ref:_ [C#91132](http://central.tri.be/issues/91132) [C#91161](http://central.tri.be/issues/91161)
---

### DIFF
--- a/src/Tribe/Tickets_Handler.php
+++ b/src/Tribe/Tickets_Handler.php
@@ -201,10 +201,16 @@ class Tribe__Tickets__Tickets_Handler {
 			}
 
 			$totals = $this->get_ticket_totals( $ticket );
+			$completes[] = $complete = $totals['pending'] + $totals['sold'];
 
-			$stock = $event_capacity - $totals['pending'] - $totals['sold'];
+			$stock = $event_capacity - $complete;
 			update_post_meta( $ticket, '_stock', $stock );
 		}
+
+		// Make sure we are updating the Global Stock when we update it's capacity
+		$shared_stock = new Tribe__Tickets__Global_Stock( $object_id );
+		$shared_stock_level = $event_capacity - array_sum( $completes );
+		$shared_stock->set_stock_level( $shared_stock_level );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/Tribe/Tickets_Handler.php
```diff
@@ -201,10 +201,16 @@ public function update_shared_tickets_capacity( $meta_id, $object_id, $meta_key,
...
+			$completes[] = $complete = $totals['pending'] + $totals['sold'];
 
-			$stock = $event_capacity - $totals['pending'] - $totals['sold'];
+			$stock = $event_capacity - $complete;
 			update_post_meta( $ticket, '_stock', $stock );
 		}
+
+		// Make sure we are updating the Global Stock when we update it's capacity
+		$shared_stock = new Tribe__Tickets__Global_Stock( $object_id );
+		$shared_stock_level = $event_capacity - array_sum( $completes );
```
barryhughes: Not sure if this would happen in practice, but it _looks_ like there is potential for the foreach loop not to run (ie, if `$tickets` is empty) ... if that is indeed possible we might create some noise and unexpected results here, as `$completes` won't be defined

src/Tribe/Tickets_Handler.php
```diff
@@ -1205,7 +1211,12 @@ public function render_ticket_row( $ticket ) {
 			&& -1 !== $capacity
 		) {
 			$product = wc_get_product( $ticket->ID );
-			$needs_warning = (int) $inventory !== (int) $stock;
+			$shared_stock = new Tribe__Tickets__Global_Stock( $ticket->get_event()->ID );
```
barryhughes: It's _possible_ `$ticket->get_event()` will return null, so a nice bit of defensive coding could be to protect against that outcome (and remove the potential for an accessing-property-on-non-object error)

src/Tribe/Tickets_Handler.php
bordoni: Ohh good catch. Will add a base definition.
But yes it wouldn't happen.

src/Tribe/Tickets_Handler.php
bordoni: Yes, it could happen. But only if you edit the DB directly

barryhughes: Or if a third party extension / user customization runs amok ;-)

PR #567, opened 2017-11-09
Escape underscores in LIKE query parameter so the column's index can be used properly

misenhower:
MySQL treats underscores as a single-character wildcard in LIKE comparisons. Starting a search with a wildcard prevents MySQL from using any indexes on the column being compared.

This change enables MySQL to use the index on meta_key for this query and better reflects the intended behavior of the query.

In our case, we have a very large wp_postmeta table so this query was taking a long time to complete since it had to scan every row.

References:
https://dev.mysql.com/doc/refman/5.7/en/string-comparison-functions.html#operator_like
https://dev.mysql.com/doc/refman/5.7/en/index-btree-hash.html

### DIFF
--- a/src/Tribe/Cache/Abstract_Cache.php
+++ b/src/Tribe/Cache/Abstract_Cache.php
@@ -61,7 +61,7 @@ abstract class Tribe__Tickets__Cache__Abstract_Cache implements Tribe__Tickets__
 				LEFT JOIN {$wpdb->posts} p
 				ON pm.meta_value = p.ID
 				WHERE p.post_type IN {$post_types}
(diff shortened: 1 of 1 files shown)


jmalko:
+1

misenhower:
Hey, just wanted to check on this since it looks like it was approved but didn't actually get merged in. Was this merged into a hidden branch somewhere?

bordoni:
Sorry for closing this one @misenhower it was due to the `release/M17.23` branch deletion, also that release had a very small scope internally. But on the next release we will merge and test this one.

misenhower:
Cool, thanks!

PR #572, opened 2017-11-15
#92550 - add spacing to logged in user view links

jesseeproductions:
_Ref:_ [C#92550](https://central.tri.be/issues/92550)


barryhughes:
Looks like we need the same change made in the `.pcss` file to be compiled into the corresponding `.css` and `.min.css` files?

PR #580, opened 2017-11-17
Add connection method to Tickets Handler and Fix (Un)Checkin

bordoni:
_Ref:_ [C#93720](http://central.tri.be/issues/93720)
---


src/Tribe/Metabox.php
```diff
@@ -473,6 +471,36 @@ public function has_permission( $post, $data, $nonce_action ) {
 		return ! empty( $data['nonce'] ) && wp_verify_nonce( $data['nonce'], $nonce_action ) && current_user_can( get_post_type_object( $post->post_type )->cap->edit_posts );
 	}
 
+	/**
+	 * Tests if the user has the specified capability in relation to whatever post type
+	 * the attendee object relates to.
+	 *
+	 * For example, if the attendee was generated for a ticket set up in relation to a
+	 * post of the banana type, the generic capability "edit_posts" will be mapped to
```
lucatume: What about quoting `banana` with backtick here to distinguish it from normal words?

src/Tribe/Tickets_Handler.php
```diff
@@ -271,6 +271,121 @@ public function update_start_date( $post_id, $post, $update ) {
...
+	 *             'provider' => (mixed|null)
+	 *             'event' => (int|null)
+	 *             'product' => (int|null)
+	 *             'order' => (int|string|null)
+	 *             'order_item' => (int|null)
+	 *         }
+	 */
+	public function get_object_connections( $object ) {
+		// If you add any new Items here, update the Docblock
+		$connections = (object) array(
```
lucatume: Nitpicking here but since in this line you are essentially initializing an empty connections template why not refactor that in a `public` method? E.g. `$connections = $this->get_connections_template();`.
That method would then be available to any client code that needs/wants to know how the empty template looks like.

bordoni: Fair, will do that.

PR #582, opened 2017-11-17
Resolve some New Panels problems

bordoni:
_Ref:_ [C#93625](http://central.tri.be/issues/93625) [C#93633](http://central.tri.be/issues/93633)
---


src/admin-views/editor/fieldset/advanced.php
```diff
@@ -135,7 +135,7 @@ class="dashicons dashicons-editor-help"
 		 * @param int Post ID
 		 * @param null Ticket ID
```
borkweb: Should this parameter comment be updated?

PR #583, opened 2017-11-20
Attendees Column now with the correct calculation

bordoni:
_Ref:_ [C#88047](http://central.tri.be/issues/88047)
---

### DIFF
--- a/src/Tribe/Admin/Columns/Tickets.php
+++ b/src/Tribe/Admin/Columns/Tickets.php
@@ -138,7 +138,7 @@ class Tribe__Tickets__Admin__Columns__Tickets {
 		}
 		// In all other cases, calculate the actual percentage
 		else {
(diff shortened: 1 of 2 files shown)


src/Tribe/Admin/Columns/Tickets.php
```diff
@@ -138,7 +138,7 @@ protected function get_percentage_string( $post_id, $deprecated = null ) {
 		}
 		// In all other cases, calculate the actual percentage
 		else {
-			$percentage = round( $totals['sold'] * 100 / $stock, 0 );
+			$percentage = round( ( 100 / $totals['capacity'] ) * $totals['sold'], 0 );
```
lucatume: What if capacity is 0? Is that case handled before?

bordoni: This will be handled by the other conditionals above.

PR #589, opened 2017-11-24
Makes sure Attendees meta are initialized when configured

bordoni:
_Ref:_ [C#94142](http://central.tri.be/issues/94142)
---


src/resources/js/tickets.js
```diff
@@ -345,6 +345,11 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
 			}
 		}
 
+		// When we have Meta fields for Attendees
+		if ( 'undefined' !== typeof tribe_event_tickets_plus && 'undefined' !== typeof tribe_event_tickets_plus.meta ) {
```
lucatume: Does such a strict check on `undefined` make sense here?
What if the two variables, `tribe_event_tickets_plus` and `tribe_event_ticket_plus.meta` are defined but not in the correct format? E.g. `[ meta: 'foo' ]

bordoni: Specifically for these I think it might, due to only been set on by our own code. But what would suggest instead?

lucatume: A more informed check; if we expect meta to come as an array check if it's an array and so on.
Yet it might not make sense here so that's your call.

bordoni: Fixed with a better check

PR #594, opened 2017-11-28
reload after ticket move

lucatume:
Ticket: https://central.tri.be/issues/91628

This PR will make sure the ticket list is reloaded after a successful, failed or aborted ticket move.
Mind that I'm usign a `setTimeout` to wait for the iFrame to appear as more modern approaches might not be supported by older/weird browsers; time is set to a reasonable value.


ggwicz:
sorry, just noticed that @bordoni needs to review this per @lucatume -- i think it's good to merge but will remove that tag so @bordoni sees it

PR #597, opened 2017-11-29
Resolve the CSS for the Purchase limit field

bordoni:
_Ref:_ [C#92914](http://central.tri.be/issues/92914)
---


src/resources/postcss/tickets-refresh.pcss
```diff
@@ -76,7 +76,7 @@
 	fieldset {
 		border: 0;
 		padding: 0;
-		margin: 0;
+		margin: 0 0 1em 0;
```
borkweb: Technically you don't need that final `0`. But you don't have to remove it in this PR.

PR #603, opened 2017-12-06
#81629 Fix Front-End Check-in

ggwicz:
**Ticket:** [**#81629**](http://central.tri.be/issues/81629)


borkweb:
This PR is named horribly. Try to use a descriptive name rather than just a ticket number.

PR #614, opened 2018-01-09
changed to use the correct filter

hadamlenz:
Unless there is a filter somewhere in your own code base called manage_{$post_type}_columns that I cannot find with a search.  the correct filter should be manage_{$post_type}_post_columns


hadamlenz:
sorry, I'm dumb, ignore me

PR #634, opened 2018-02-09
Rework Oversell notice look and behaviour, fix bug

lucatume:
Ticket: https://central.tri.be/issues/98563#note-8

This PR:
* remakes the oversell notice look and behaviour to make it (kinda) stick to the mock (https://cl.ly/14383c1B2x34)
* fixes a bug that would prevent oversell

Note: the notice contains HTML and CSS code; the former generated by a number of `sprintf` and the latter as there is no good place to style it.
As for the inline CSS:
1. it would be custom to the notice
2.  the notice might/should display *anywhere* in the admin area and I did not want to burde the common-admin file with its style; think React

### DIFF
--- a/src/Tribe/Commerce/PayPal/Links.php
+++ b/src/Tribe/Commerce/PayPal/Links.php
@@ -9,6 +9,15 @@
  */
 class Tribe__Tickets__Commerce__PayPal__Links {
 
+	/**
+	 * Returns the link to the IPN notification history page on PayPal.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $what Either `link` to return the URL or `tag` to return an `a` tag.
+	 *
+	 * @return string
+	 */
 	public function ipn_notification_history( $what = 'tag' ) {
 		$link = add_query_arg(
 			array( 'cmd' => '_display-ipns-history' ),
@@ -27,6 +36,15 @@ class Tribe__Tickets__Commerce__PayPal__Links {
 		return Tribe__Utils__Array::get( $map, $what, '' );
 	}
 
+	/**
+	 * Returns the link to the IPN notification settings page on PayPal.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $what Either `link` to return the URL or `tag` to return an `a` tag.
+	 *
+	 * @return string
+	 */
 	public function ipn_notification_settings( $what = 'tag' ) {
 		$link = add_query_arg(
 			array( 'cmd' => '_profile-ipn-notify' ),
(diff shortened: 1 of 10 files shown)


src/Tribe/Commerce/PayPal/Links.php
```diff
@@ -45,4 +63,33 @@ public function ipn_notification_settings( $what = 'tag' ) {
...
+	 *                  the Order PayPal ID (hash).
+	 *
+	 * @return string
+	 */
+	public function order_link( $what, $order_id, $text = null ) {
+		$text = null !== $text ? $text : $order_id;
+
+		$link = Tribe__Tickets__Commerce__PayPal__Order::get_order_link( $order_id );
+		$tag  = '<a href="'
+		        . esc_attr( $link )
```
bordoni: `esc_url`

src/Tribe/Commerce/PayPal/Oversell/Admin_Notice_Decorator.php
```diff
@@ -50,62 +50,112 @@ public function allows_overselling() {
...
+	 * Returns the embedded styles for the notice.
 	 *
 	 * @since TBD
 	 *
 	 * @return string
 	 */
-	public function get_order_id() {
-		return $this->policy->get_order_id();
+	protected function style() {
+		return '<style>
```
bordoni: Can we move this into a file?

src/Tribe/Commerce/PayPal/Oversell/Admin_Notice_Decorator.php
```diff
@@ -50,62 +50,112 @@ public function allows_overselling() {
...
+		if ( ! empty( $post ) ) {
+			$edit_link  = $this->get_user_insensible_edit_link( $post_type, $post_id );
+			$post_title = sprintf(
+				__(
+					'%s "%s" (ID %d)',
+					'event-tickets' ),
+				ucwords( $post_type_label ),
+				sprintf(
+					'<a href="%s">%s</a>',
+					esc_attr( $edit_link ),
```
bordoni: `esc_url`

src/Tribe/Commerce/PayPal/Oversell/Admin_Notice_Decorator.php
```diff
@@ -120,14 +170,31 @@ public function get_post_id() {
...
+	 *
+	 * @param object $post_type
+	 * @param int    $post_id
+	 *
+	 * @return string
+	 */
+	protected function get_user_insensible_edit_link( $post_type, $post_id ) {
+		$action = '&amp;action=edit';
+
+		return admin_url( sprintf( $post_type->_edit_link . $action, $post_id ) );
```
bordoni: use `add_query_arg` to add the action, so we make sure it replaces if it already has it.

src/Tribe/Commerce/PayPal/Oversell/Admin_Notice_Decorator.php
```diff
@@ -142,26 +209,54 @@ public function get_ticket_id() {
...
+			'sell-available' => __( 'Confirm attendee records and send emails for some tickets in this order without overselling the event', 'event-tickets' ),
+			'no-oversell'    => __( 'Delete all attendees for this order and do not send any email', 'event-tickets' ),
+		);
+
+		foreach ( $options as $policy => $label ) {
+			$form_inside .= sprintf(
+				'<div class="tribe-tickets-paypal-oversell-radio"><input type="radio" radiogroup="order-%1$s-actions" value="%2$s" name="tpp_policy" '
+				. checked( $default, $policy, false )
+				. '><label >%3$s</label></div>',
+
```
bordoni: remove blank line

src/Tribe/Commerce/PayPal/Oversell/Admin_Notice_Decorator.php
```diff
@@ -50,62 +50,112 @@ public function allows_overselling() {
...
-					'event-tickets'
-				),
-				$this->get_order_id(),
-				$inventory,
-				$qty,
-				strtolower( $this->policy->get_name() )
-			)
-		);
+		$output = $this->style();
+		$output .= $this->header_html( $qty, $inventory );
```
bordoni: We could make this class extend `Tribe__Template`, so we could move most of the HTML we have here into a `src/admin-view/` file.

Take a look at: https://github.com/moderntribe/event-tickets/blob/bucket/tribe-commerce/src/Tribe/Admin/Views.php

To see how to setup.

lucatume: I've taken that approach into account but let go for two reasons:
1. the template would be just a mangled sequence of `if` that would defeat the purpose of a "logic-less" template
2. since this notice could/should show up anywhere in the admin area (whether a ticket-able post is being edited or not) the only place to put this notice very specialized CSS style would be either a separated CSS file loaded in all the admin screens or (pretty much the same) the common (as "in the `common` package) admin file... but that seems to lead on the same path of putting everything into a `Main` file (nope, if possible); adding inline CSS on top of a ton of logic seems to defeat the purpose of templates again.

Open to suggestions about this, though.

lucatume:
Good for another look.

PR #637, opened 2018-02-12
add some checks and casts to avoid notices

lucatume:
Ticket: https://central.tri.be/issues/98251

Added a little more defense from notices in the code

### DIFF
--- a/src/Tribe/Commerce/PayPal/Main.php
+++ b/src/Tribe/Commerce/PayPal/Main.php
@@ -1141,7 +1141,7 @@ class Tribe__Tickets__Commerce__PayPal__Main extends Tribe__Tickets__Tickets {
 		// Makes sure it's an Int after this point
 		$data['stock'] = (int) $data['stock'];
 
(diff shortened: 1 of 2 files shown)


src/Tribe/Commerce/PayPal/Main.php
```diff
@@ -1141,7 +1141,7 @@ public function save_ticket( $post_id, $ticket, $raw_data = array() ) {
 		// Makes sure it's an Int after this point
 		$data['stock'] = (int) $data['stock'];
 
-		if ( '' !== $data['mode'] ) {
+		if ( isset( $data['mode'] ) && '' !== $data['mode'] ) {
```
Camwyn: Suggestion: If `$data['mode']` can't be 0 or `false`, you could just use `empty()` here?

If it can ignore the suggestion. This looks 👍

lucatume: I'm not really sure that `0` or `false` would not be legit values... I had the same thought though.

PR #649, opened 2018-02-19
Adjust the PayPal bn value

lucatume:
Ticket: https://central.tri.be/issues/99690

This PR updates PayPal `bn` value to one conforming with [PayPal documentation](https://developer.paypal.com/docs/classic/paypal-payments-standard/integration-guide/Appx_websitestandard_htmlvariables/)


lucatume:
I've removed the display fix to move it in a stand-alone PR

zbtirrell:
According to the latest email from PayPal, we were doing it right before…

PR #651, opened 2018-02-19
pass the post ID when formatting the line total entry

lucatume:
Ticket:  n/a

Make sure Tribe Commerce currency code is used to format line totals in Tribe Commerce reports.

### DIFF
--- a/src/Tribe/Commerce/PayPal/Orders/Table.php
+++ b/src/Tribe/Commerce/PayPal/Orders/Table.php
@@ -179,7 +179,9 @@ class Tribe__Tickets__Commerce__PayPal__Orders__Table extends WP_List_Table {
 	 * @return string
 	 */
 	public function column_total( $item ) {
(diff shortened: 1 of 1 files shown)


src/Tribe/Commerce/PayPal/Orders/Table.php
```diff
@@ -179,7 +179,9 @@ public function column_order( $item ) {
 	 * @return string
 	 */
 	public function column_total( $item ) {
-		return tribe_format_currency( number_format( $item['line_total'], 2 ) );
+		$post_id = Tribe__Utils__Array::get( $_GET, 'post_id', null );
```
bordoni: why not `tribe_get_request_var()`?

lucatume: to skip a check on the `$_POST` array; it seems unlikely there would ever be the need to look up `$_POST` when rendering the table

PR #655, opened 2018-02-21
Spotfix/update versions

mitogh:


mitogh:
Mergin master into latest version of tribe commerce.

PR #656, opened 2018-02-21
Use providers if available to format cost

mitogh:
**Ticket** https://central.tri.be/issues/70361

Hook into the filter for the cost if available to use providers


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | There must be a single space after the first parenthesis | [Currency.php:401](https://github.com/moderntribe/event-tickets/blob/4fe4627afa123d58dea6975f44ece2ff90ab55fb/src/Tribe/Commerce/Currency.php#L401)

PR #657, opened 2018-02-21
Use providers if available to format cost

mitogh:
**Ticket** https://central.tri.be/issues/70361

Hook into the filter for the cost if available to use providers

### DIFF
--- a/src/Tribe/Commerce/Currency.php
+++ b/src/Tribe/Commerce/Currency.php
@@ -26,6 +26,7 @@ class Tribe__Tickets__Commerce__Currency {
 	 */
 	public function hook(  ) {
 		add_filter( 'tribe_currency_symbol', array( $this, 'filter_currency_symbol' ), 10, 2 );
+		add_filter( 'tribe_currency_cost', array( $this, 'filter_currency_cost' ), 10, 2 );
 		add_filter( 'tribe_reverse_currency_position', array( $this, 'reverse_currency_symbol_position' ), 10, 2 );
 		add_filter( 'get_post_metadata', array( $this, 'filter_cost_meta' ), 10, 4 );
 	}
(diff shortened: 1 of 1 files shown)


src/Tribe/Commerce/Currency.php
```diff
@@ -367,6 +383,38 @@ protected function get_provider_symbol( $provider, $object_id ) {
 		return $this->get_currency_symbol( $object_id );
 	}
 
+	/**
+	 * Returns he cost applying separators from Woo or EDD depends on the provider.
```
ggwicz: typo - `he` should be `the`

src/Tribe/Commerce/Currency.php
```diff
@@ -367,6 +383,38 @@ protected function get_provider_symbol( $provider, $object_id ) {
...
+	 * @since TBD
+	 *
+	 * @param string $provider
+	 * @param string $cost
+	 *
+	 * @return string
+	 */
+	protected function get_provider_cost( $provider = '', $cost = '' ) {
+		if ( ! class_exists( $provider ) ) {
+			return $cost;
```
ggwicz: curious about just returning the cost -- we have some `Cost_Utils` in tribe common, with some cost-formatting methods of our own. should we try to format the cost using those methods if no provider methods are available? 

sorry if this was addressed already or not relevant

mitogh: Thanks for asking at this stage the cost has been already formatted by `number_format` so that's why is a `string` and not a `number` so we are just falling back to the initial formatting of cost. Not sure if is worth it.

mitogh:
Hey folks should we update `common` module hash here as well to point to the new branch `bucket/tribecommerce`?

Thanks.

cc @ggwicz @barryhughes @bordoni

barryhughes:
Hmm, yeah we need that to happen it _should_ be taken care of automatically (ie, the bot should take care of syncing common) ... but perhaps we'll need to do it manually to get things started. I'd say, go ahead and update it.

PR #665, opened 2018-02-28
Fix the login requirement issue with EDD and Woo tickets

lucatume:
Ticket: https://central.tri.be/issues/100398

This PR:
* fixes an issue in the function that would fetch the default ticket provider that would make it return an array when a string is expected
* fixes the check for the login requirements to make sure that the correct option slug is used


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Must use &quot;self::&quot; for local static member reference | [Tickets.php:1266](https://github.com/moderntribe/event-tickets/blob/8512ea860d7b146941210a13889464624437c3f6/src/Tribe/Tickets.php#L1266)

PR #669, opened 2018-03-02
Cache calls to `get_all_event_tickets` to avoid duplicates

mitogh:
**Ticket** https://central.tri.be/issues/99267

In order to avoid multiple calls to the same function we are using a transient that is clear every time a post is updated or a ticket is added or removed.


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Missing space after array opener. | [Tickets.php:695](https://github.com/moderntribe/event-tickets/blob/d5e976555fcad2965cec7114748efc49733b5152/src/Tribe/Tickets.php#L695)
:x: | There must be a single space after the first parenthesis | [Tickets.php:695](https://github.com/moderntribe/event-tickets/blob/d5e976555fcad2965cec7114748efc49733b5152/src/Tribe/Tickets.php#L695)
:x: | There must be a single space after the first parenthesis | [Tickets.php:695](https://github.com/moderntribe/event-tickets/blob/d5e976555fcad2965cec7114748efc49733b5152/src/Tribe/Tickets.php#L695)

mitogh:
**Note** this still needs a changelog entry as we need to setup `MR18.05` with the appropriate version number changes and changes on `readme`

src/Tribe/Tickets.php
```diff
@@ -33,6 +33,16 @@ class Tribe__Tickets__Tickets {
...
 
+		/**
+		 * Function that is used to store the cache of a specific post associated with a set of tickets, where %d is the
+		 * ID of the post being affected.
+		 *
+		 * @since TBD
+		 *
+		 * @var string
+		 */
+		private static $cache_key_all_tickets = 'tribe_event_tickets_from_%d';
```
zbtirrell: if this was `$cache_key_prefix` you could avoid all the `sprintf` calls later and just concatenate

zbtirrell: also, making this variable `static` is a surprising choice. Since it is private, you can't access it outside the object anyway. There are other variables surrounding it which _are_ static, so, perhaps this was just pattern matching with the rest?

src/Tribe/Tickets.php
```diff
@@ -683,6 +693,10 @@ public function __construct() {
 
 			// Ensure ticket prices and event costs are linked
 			add_filter( 'tribe_events_event_costs', array( $this, 'get_ticket_prices' ), 10, 2 );
+			// Flush caches if any of the following actions has occurred
+			add_action( 'save_post', array( $this, 'clear_cache' ) );
```
zbtirrell: I'm _slightly_ concerned that this action will run on save post for all posts no matter what. I'd be tempted to check that the post we are acting on is of a post type that could have tickets. It's probably OK though. If the `delete_transient` function was less complicated I wouldn't have even raised the concern

src/Tribe/Tickets.php
mitogh: 1. I think having the variable as a prefix makes more sense thanks.
2. `private` to prevent inherit it down any child class and `static` just to follow the same pattern on the class.  I think just `private` should be enough.  As I don't think we need to create a new instance of the class in order to access the variable to be referenced inside of the class.

src/Tribe/Tickets.php
mitogh: As any post type can have tickets you're right we need to check before attaching the hooks to know on which post types we only need to clear caches, however as every post has a unique ID and this ID is associate independently from each other by clearing or removing a cache on save post even if the post does not have the cache removed is only the one associated with the post being modified. 

Let me know what do you think and if you think is best to do it dynamically based on the options from the admin on the post types that has tickets associated with it.

readme.txt
```diff
@@ -235,6 +235,8 @@ The plugin is produced by [Modern Tribe Inc](http://m.tri.be/18uc).
 
 = [M18.05] TBD =
 
+* Fix - Prevent duplicates queries on the admin, by caching them [99267]
```
ggwicz: Can you specify _which_ queries here? Also, use `duplicate`, not `duplicates` 🍻

src/Tribe/Tickets.php
```diff
@@ -784,6 +806,8 @@ public static function get_all_event_tickets( $post_id ) {
 				$tickets = array_merge( $tickets, $obj->get_tickets( $post_id ) );
 			}
 
+			set_transient( $cache_key, $tickets, HOUR_IN_SECONDS );
```
ggwicz: curious about the `HOUR_IN_SECONDS` choice here — if we're going to be clearing cache anytime the post is updated, maybe we can make this a longer time period? like six hours or something?

readme.txt
mitogh: Sure thanks

mitogh:
Removing _enhance_ to ask for another set of eyes.

ggwicz:
Hey @mitogh, it seems like there are a number of changes you can make here:

1. In regards to [this comment](https://github.com/moderntribe/event-tickets/pull/669#discussion_r172214569), sounds like you agree with Zach about using `$cache_key_prefix`, so just implement that and leave as `private static` for now (your reasons for using `private static` make sense).

2. In regards to [this comment](https://github.com/moderntribe/event-tickets/pull/669#discussion_r172216573), it sounds like you agree that this should be limited to the `save_post` action _only_ on post types that have tickets. So proceed with implementing that for now, and yes you can just use the option from the wp-admin about tickets-enabled post types.

3. Next, implement the [readme changes](https://github.com/moderntribe/event-tickets/pull/669#discussion_r174475053) that were flagged.

4. If you agree with [my comment](https://github.com/moderntribe/event-tickets/pull/669#discussion_r172244861) about `HOURS_IN_SECONDS`, make that change, but if not, just note your thoughts and keep it in place.

☝️ Does this give you what you need to move this PR forward a bit?

mitogh:
Thank you for the action items here @ggwicz 🎉

src/Tribe/Tickets.php
```diff
@@ -330,7 +340,16 @@ public function get_tickets_ids( $post = null ) {
...
+			$cache = new Tribe__Cache();
+			$cache_key = $cache->make_key( $args );
+			$query = $cache->get( $cache_key );
+
+			if ( $query instanceof WP_Query ) {
+				return $query->posts;
+			}
+
 			$query = new WP_Query( $args );
+			$cache->set( $cache_key, $query, $cache::NO_EXPIRATION );
```
barryhughes: 99% sure referencing a class constant via the object reference works only with PHP 5.3+. For compatibility with 5.2, we probably need `Tribe__Cache::NO_EXPIRATION`.

src/Tribe/Tickets.php
```diff
@@ -924,7 +952,10 @@ public static function get_all_event_tickets( $post_id ) {
 				}
 			}
 
-			return ! empty( $tickets ) ? call_user_func_array( 'array_merge', $tickets ) : array();
+			$tickets = empty( $tickets ) ? array() : call_user_func_array( 'array_merge', $tickets );
+			$cache->set( $cache_key, $tickets, $cache::NO_EXPIRATION );
```
barryhughes: Ditto my note re PHP 5.2 compatibility

src/Tribe/Tickets.php
bordoni: Yeah @barryhughes, `constant( $cache . '::NO_EXPIRATION' )` is what we want.

src/Tribe/Tickets.php
bordoni: Yeah @barryhughes, `constant( $cache . '::NO_EXPIRATION' )` is what we want.

readme.txt
```diff
@@ -241,6 +241,7 @@ The plugin is produced by [Modern Tribe Inc](http://m.tri.be/18uc).
 
 * Feature - Add updater class to enable changes on future updates [84675]
 * Feature - Add JSON-LD for custom post types with support for tickets (Thanks to Albert for flag this in our forums) [95034]
+* Fix - Prevent duplicate calls to `get_all_event_tickets` on the admin, by caching them [99267]
```
barryhughes: Could be worth sending props to Gabriel

src/Tribe/Tickets.php
mitogh: 🥇

mitogh:
We are ready over here again 🚀

PR #673, opened 2018-03-05
Count ticketd / unticketd events that are in trash

mitogh:
**Ticket** https://central.tri.be/issues/92236

Make sure the count on the admin does not consider events under draft or trash status.


readme.txt
```diff
@@ -235,7 +235,7 @@ The plugin is produced by [Modern Tribe Inc](http://m.tri.be/18uc).
 
 = [M18.05] TBD =
 
-
+* Fix - Make sure the ticket / unticked events does not count events under `trash` [92236]
```
ggwicz: I'd recommend adjusting to something like this:
```
* Fix - Prevent trashed events from affecting ticketed/unticketed counts in the wp-admin [92236]
```

Also, the Central ticket [cites this WP.org thread](https://wordpress.org/support/topic/bug-unticketed-events-count-in-admin-counts-trashed-events/) as the source of the bug, so it looks like there's a user to call out with "props" on this readme entry 👍

mitogh:
Changes has been applied, thank you @ggwicz

src/Tribe/Cache/Abstract_Cache.php
```diff
@@ -74,8 +74,29 @@ protected function fetch_posts_with_ticket_types( array $post_types = null ) {
...
+		}
+
+		/**
+		 * The above will fetch posts based on the meta data regardless of the status of the post, however post
+		 * under the status of `trash` or `auto-draft` shouldn't be in the list.
+		 */
+		$query = $wpdb->prepare(
+			"SELECT DISTINCT(ID) 
+				FROM {$wpdb->posts}
+				WHERE post_status != 'auto-draft'
```
bordoni: We should allow filtering of these post status and use `NOT IN` to allow easier filtering.

src/Tribe/Cache/Abstract_Cache.php
```diff
@@ -104,7 +125,8 @@ protected function fetch_posts_without_ticket_types( array $post_types = null )
 
 		$query = "SELECT DISTINCT(ID) FROM {$wpdb->posts}
 				WHERE post_type IN {$post_types}
-				AND post_status != 'auto-draft'";
+				AND post_status != 'auto-draft'
+				AND post_status != 'trash'";
```
bordoni: We should allow filtering of these post status and use `NOT IN` to allow easier filtering.

jesseeproductions:
👍

PR #674, opened 2018-03-06
Restore RSVP email functionality

lucatume:
Ticket: https://central.tri.be/issues/100856

An RSVP confirmation has tickets when *at least* one ticket was generated (logic OR) not when *all* tickets were generated (logic AND).
The latter would only happen when running out of stock while generating the tickets or via filtering in code.


jesseeproductions:
💯

PR #675, opened 2018-03-06
Fix back-compat with out of date ET+ plugin

lucatume:
Ticket: https://central.tri.be/issues/100879

This PR modifies the code of the `Tickets` class to make sure extending classes defined  in ET+ (Woo, EDD) that might still be one version behind will not generate fatals or warnings. 
This **does** create duplicate code (the `generate_security_code` method is now defined in all classes) and makes the inheritance tree a little less "stable" but avoids issues during upgrades.


jesseeproductions:
🚀

PR #683, opened 2018-03-08
Fix/91471 stock manage warning

jesseeproductions:
_Ref:_ [C#91471](https://central.tri.be/issues/91471)


readme.txt
```diff
@@ -236,6 +236,7 @@ The plugin is produced by [Modern Tribe Inc](http://m.tri.be/18uc).
 = [M18.05] TBD =
 
 * Fix - Allow access when visiting the attendees list on venue or organizer post type (props to Antonio Jose in our forums for flagging this problem) [90062]
+* Fix - Add warning if a ticket has stock management turned off in the related Woocommerce product, but has capacity enabled for the ticket [91471]
```
barryhughes: * Can we add something like `(our thanks to Isaiah Baker and others for highlighting this)` before the ticket ref?
* Nitpick ... should be `WooCommerce`

PR #690, opened 2018-03-12
95257 Prevent Overriding of Start|End Sale Dates on RSVPs and Tickets

ggwicz:
**Ticket:** [**#95257**](http://central.tri.be/issues/95257)

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -235,13 +235,14 @@ The plugin is produced by [Modern Tribe Inc](http://m.tri.be/18uc).
 
 = [M18.05] TBD =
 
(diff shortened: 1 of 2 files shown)


bordoni:
On this one @ggwicz we need to make sure we are validating if there were any manual changes to the sales DateTime on the ticket or RSVP, because otherwise it's the intended behavior to prevent fails of tickets with no Start and End.

mitogh:
I think this is the same as #699

ggwicz:
It looks so!

mitogh:
💃

PR #695, opened 2018-03-14
Add query params as input values to the form

mitogh:
**Ticket**: https://central.tri.be/issues/96640

In order to send the params as the type of Request is $_GET and all the
values that are on the permalink as $query params are not consider when
doing the request.

### DIFF
--- a/src/template-tags/tickets.php
+++ b/src/template-tags/tickets.php
@@ -238,8 +238,24 @@ if ( ! function_exists( 'tribe_tickets_buy_button' ) ) {
 					$button_anchor = '#buy-tickets';
 				}
 
-				$button = '<form method="get" action="' . esc_url( get_the_permalink( $event_id ) . $button_anchor ) . '">'
-					. '<button type="submit" name="tickets_process" class="tribe-button">' . $button_label . '</button>'
+				$permalink = get_the_permalink( $event_id );
+				$query_string = parse_url( $permalink, PHP_URL_QUERY );
+				$query_params = empty( $query_string ) ? array() : (array) explode( '&', $query_string );
+
+				$button = '<form method="get" action="' . esc_url( $permalink . $button_anchor ) . '">';
+
+				// Add any query attribute as a hidden input as the action of the form is GET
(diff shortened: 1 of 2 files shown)


bordoni:
@mitogh check the tests here.

src/template-tags/tickets.php
```diff
@@ -238,8 +238,24 @@ function tribe_tickets_buy_button( $echo = true ) {
...
+				$query_params = empty( $query_string ) ? array() : (array) explode( '&', $query_string );
+
+				$button = '<form method="get" action="' . esc_url( $permalink . $button_anchor ) . '">';
+
+				// Add any query attribute as a hidden input as the action of the form is GET
+				foreach ( $query_params as $param ) {
+					$parts = explode( '=', $param );
+					// a query string must be 2 parts only a name and a value
+					$number_of_parts = 2;
+					if ( is_array( $parts ) && $number_of_parts === count( $parts ) ) {
```
barryhughes: `$number_of_parts` will never vary, so it may be better to just use a literal (ie `2 === count( $parts )`) ... though it's subjective this for me would improve readability, besides shortening the amount of code we have here.

mitogh: Yeah I was going to use the 2 directly but wanted to be explicit about what 2 in this case. But let me know what do you think in favor over readability here.

mitogh:
_Working on Fixing the tests_

mitogh:
_Tests are working correctly_

src/template-tags/tickets.php
mitogh: @bordoni @barryhughes ⬆️

barryhughes: I would veer toward using a literal vs a variable and feel doing so improves readability (in my eyes, this is _too_ verbose) but in the grand scheme of things it isn't so important it need hold up this work.

mitogh: 👍 thank you ;)

mitogh:
_Seems like last commit on this PR requires another review_

PR #685, opened 2018-03-08
Fix/100165 rsvp change status capacity

jesseeproductions:
_Ref:_ [C#100165](https://central.tri.be/issues/100165)

PR #696, opened 2018-03-15
Allow to set the number of stock for an RSVP on tests

mitogh:


mitogh:
This fix the tests failure is introduced on #685

PR #705, opened 2018-03-19
Fix/1007336 unify upsell

niconerd:
https://central.tri.be/issues/100736

### DIFF
--- a/src/admin-views/tribe-options-tickets.php
+++ b/src/admin-views/tribe-options-tickets.php
@@ -131,12 +131,12 @@ $is_tickets_plus_available = array_key_exists( $tickets_plus_plugin, $available_
 
 if ( ! $is_tickets_plus_available ) {
 	$plus_link = sprintf(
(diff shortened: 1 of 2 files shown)


src/admin-views/tribe-options-tickets.php
```diff
@@ -135,9 +135,15 @@
 		__( 'Events Tickets Plus', 'tribe-common' )
 	);
 
+	$plus_link_2 = sprintf(
+		'<a href="https://theeventscalendar.com/product/wordpress-event-tickets-plus/?utm_campaign=in-app&utm_medium=plugin-tickets&utm_source=post-editor" target="_blank">%s</a>',
+		__( 'Check it out!', 'tribe-common' )
+	);
```
jesseeproductions: These links we already here, but please use the m.tri.be shortlinks
https://m.tri.be/new/

That way we can update there and not in the plugins. If there are related links this could be a good time to change them to m.tri.be

PR #708, opened 2018-03-20
Use timezones when comparing dates on a ticket

mitogh:
**Ticket**: https://central.tri.be/issues/99721

Use the timezone from the ticket if any if not uses the timezone from
the global settings.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -251,7 +251,7 @@ The plugin is produced by [Modern Tribe Inc](http://m.tri.be/18uc).
 * Fix - Fixes the missing notification on the email and removes the notification when there's none [99979]
 * Fix - Prevent overwriting the start date of a ticket if it was already set [99601]
 * Fix - The redirection to the correct post type URL when site has a plain permalink structure on the buy ticket form [96640]
-* Fix - Make sure the dates to compare if a ticket is in range uses the event time zone as the reference date [99721]
+* Fix - Make sure the dates to compare if a ticket is in range uses the event time zone as the reference date (props to @evolutionstartup for report this in our help desk) [99721]
 * Fix - Fixes a glitch, where adding an RSVP results in "NaN" in the counter when using Event Tickets, Enfold and WooCommerce. (Thanks to @tbo24 for the contribution.) [93027]
 * Tweak - Change Event tickets slug from 3 different types into 2 variants for post types and events types [88569]
 * Tweak - Made it easier to set Tribe Commerce as the default ticket module (when multiple ticketing modules are active) [96538]
--- a/readme.txt
+++ b/readme.txt
@@ -251,6 +251,7 @@ The plugin is produced by [Modern Tribe Inc](http://m.tri.be/18uc).
 * Fix - Fixes the missing notification on the email and removes the notification when there's none [99979]
 * Fix - Prevent overwriting the start date of a ticket if it was already set [99601]
 * Fix - The redirection to the correct post type URL when site has a plain permalink structure on the buy ticket form [96640]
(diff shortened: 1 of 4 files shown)


src/Tribe/Ticket_Object.php
```diff
@@ -211,14 +211,34 @@ class Tribe__Tickets__Ticket_Object {
 		 */
 		public $purchase_limit;
 
+		/**
+		 * Variable used to save the DateTimeZone object of the parent event
+		 *
+		 * @since TBD
+		 *
+		 * @var null
```
bordoni: It's not really only a null

src/Tribe/Ticket_Object.php
```diff
@@ -211,14 +211,34 @@ class Tribe__Tickets__Ticket_Object {
...
+		 */
+		private $event_timezone = null;
+
+		/**
+		 * ID of the parent event of the current Ticket
+		 *
+		 * @since TBD
+		 *
+		 * @var null
```
bordoni: It's not really only a `null`.

src/Tribe/Ticket_Object.php
```diff
@@ -211,14 +211,34 @@ class Tribe__Tickets__Ticket_Object {
...
+		 */
+		private $event_id = null;
```
bordoni: If this is private, It wold be good to have a `get` method.

mitogh: This one is available already on a public method in this class as `$this->get_event()` the `private` variable was created only to store the value on memory when used in calls between the same methods of the class.

src/Tribe/Ticket_Object.php
```diff
@@ -263,17 +284,83 @@ public function end_date() {
...
+		 * Return a DateTimeZone associated with the parent Event of the current ticket
+		 *
+		 * @since TBD
+		 *
+		 * @return DateTimeZone|null
+		 */
+		public function get_event_timezone() {
+
+			if ( is_null( $this->event_id ) ) {
+				$this->event_id = $this->get_event();
```
bordoni: `get_event()` will fetch the whole object, we should either save the whole object in memory, but I would rather see it been fetched when you hit the `get_event()` method instead.

mitogh: `getter` has been added 👍

readme.txt
```diff
@@ -251,6 +251,7 @@ The plugin is produced by [Modern Tribe Inc](http://m.tri.be/18uc).
 * Fix - Fixes the missing notification on the email and removes the notification when there's none [99979]
 * Fix - Prevent overwriting the start date of a ticket if it was already set [99601]
 * Fix - The redirection to the correct post type URL when site has a plain permalink structure on the buy ticket form [96640]
+* Fix - Make sure the dates to compare if a ticket is in range uses the event time zone as the reference date [99721]
```
barryhughes: We could credit evolutionstartup for flagging the situation here (but don't let that hold up merging this work)

src/Tribe/Ticket_Object.php
```diff
@@ -834,6 +917,20 @@ public function show_description() {
...
+		/**
+		 * Access the ID of the Event parent of the current Ticket.
+		 *
+		 * @since TBD
+		 *
+		 * @return int|null
+		 */
+		public function get_event_id() {
+			if ( is_null( $this->event_id ) ) {
+				$this->event_id = $this->get_event();
```
bordoni: we should be doing `$this->get_event()->ID`  because it returns a `WP_Post` object.

mitogh:
Ready for another set of eyes 👯‍♂️

PR #712, opened 2018-03-22
Use a generic type for all other types for JSON-LD

mitogh:
**Ticket**: https://central.tri.be/issues/95034

Does not require a changelog entry as this one was already added.

The following PR introduces / changes the following items for the ticket
described above.

- Adds priceCurrency
- Adds Product as generic type
- Fixes the problem when the ticket is unlimited to be in stack all the
time

### DIFF
--- a/src/Tribe/JSON_LD/Order.php
+++ b/src/Tribe/JSON_LD/Order.php
@@ -56,6 +56,19 @@ class Tribe__Tickets__JSON_LD__Order {
 			if ( $event_type === $post_type ) {
 				continue;
 			}
+
+			/**
+			 * This will allow you to change the type for the Rich Snippet, by default it will use the type Product for
+			 * any Post type or Page. If this is runs in a book post type the filter becomes something like.
+			 *
+			 * @example tribe_events_json_ld_book_type
+			 *
+			 * @see http://schema.org/Product
+			 *
+			 * @since TBD
+			 *
+			 * @return string
+			 */
 			$filters[] = strtolower( (string) apply_filters( "tribe_tickets_json_ld_{$post_type}_type", 'Product' ) );
 		}
 
--- a/src/Tribe/JSON_LD/Order.php
+++ b/src/Tribe/JSON_LD/Order.php
@@ -185,19 +185,17 @@ class Tribe__Tickets__JSON_LD__Order {
 	/**
 	 * Return the price currency used on the Ticket
 	 *
+	 * @since TBD
+	 *
 	 * @param $ticket
 	 *
 	 * @return mixed
 	 */
 	public function get_price_currency( $ticket ) {
+
 		$currency = tribe_get_option( 'ticket-commerce-currency-code', 'USD' );
 
-		if ( class_exists( $ticket->provider_class )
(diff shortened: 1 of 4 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Usage of ELSE IF is discouraged; use ELSEIF instead | [Order.php:165](https://github.com/moderntribe/event-tickets/blob/8404f20d076c21f6716cb59d64e28f40227f3950/src/Tribe/JSON_LD/Order.php#L165)

src/Tribe/JSON_LD/Order.php
```diff
@@ -51,12 +51,20 @@ public static function hook() {
 		/**
 		 * Other types can have tickets as well we might need to hook into each type to add tickets if any has tickets
 		 */
+		$filters = array();
 		foreach ( $post_types as $post_type ) {
 			if ( $event_type === $post_type ) {
 				continue;
 			}
+			$filters[] = strtolower( (string) apply_filters( "tribe_tickets_json_ld_{$post_type}_type", 'Product' ) );
```
bordoni: We should add a docblock for this filter.

src/Tribe/JSON_LD/Order.php
```diff
@@ -150,13 +160,35 @@ public function get_offer( $ticket, $event ) {
 	public function get_ticket_availability( $ticket ) {
 		$stock = $ticket->stock();
 
-		if ( $stock <= 0 && $stock !== '' ) {
+		if ( $stock === 0 ) {
```
bordoni: Yoda it up.

src/Tribe/JSON_LD/Order.php
```diff
@@ -150,13 +160,35 @@ public function get_offer( $ticket, $event ) {
...
+	 * Return the price currency used on the Ticket
+	 *
+	 * @param $ticket
+	 *
+	 * @return mixed
+	 */
+	public function get_price_currency( $ticket ) {
+		$currency = tribe_get_option( 'ticket-commerce-currency-code', 'USD' );
+
+		if ( class_exists( $ticket->provider_class )
```
bordoni: This seems like an overkill.
We should be able to check for `$ticket->provider_class instanceof Tribe__Tickets__Tickets`, and with the change I suggested on the other we will be always safe to call `get_instance()` and `get_currency()`

mitogh:
Ready for another look over here

src/Tribe/Tickets.php
```diff
@@ -2189,5 +2190,17 @@ public static function get_event_ticket_provider( $event_id = null ) {
...
+
+		/**
+		 * Parent method to be pass to any child of this class.
+		 *
+		 * @since TBD
+		 *
+		 * @return string
+		 */
+		public function get_currency() {
+			return (string) apply_filters( 'tribe_tickets_default_currency', 'USD' );
```
ggwicz: sorry to flag another one @mitogh, but add a docblock for this filter

mitogh:
Docblock changes updated thanks @ggwicz

PR #717, opened 2018-03-26
Make sure `tribe_events_get_the_excerpt` is available

mitogh:
Before usage of the function make sure is available as it's called by
the parent class.

### DIFF
--- a/src/Tribe/JSON_LD/Type.php
+++ b/src/Tribe/JSON_LD/Type.php
@@ -113,7 +113,10 @@ class Tribe__Tickets__JSON_LD__Type extends Tribe__JSON_LD__Abstract {
 				continue;
 			}
 
-			$data = parent::get_data( $post, $args );
+			$data = array();
+			if ( function_exists( 'tribe_events_get_the_excerpt' ) ) {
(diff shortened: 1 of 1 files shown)


mitogh:
Unfortunately the function is part of TEC only and not from `common`

src/Tribe/JSON_LD/Type.php
```diff
@@ -113,7 +113,10 @@ public function get_data( $posts = null, $args = array() ) {
 				continue;
 			}
 
-			$data = parent::get_data( $post, $args );
+			$data = array();
+			if ( function_exists( 'tribe_events_get_the_excerpt' ) ) {
```
bordoni: I mean we should add the fix on Common to avoid calling `tribe_events_get_the_excerpt` if not available, fall back on the WordPress `$post->post_excerpt`.

mitogh: Closing this PR here then an opening one on TEC as is where the parent class is located.

bordoni:
@mitogh delete the branch?

mitogh:
💃

PR #739, opened 2018-04-18
fixed de translation

logTom:
"Ihren Karten von" is not grammatically correct german.


bordoni:
Hi @logTom,
We are no longer using the Repository to manage the Translation files, we use the GlotPress from WordPress.org for the Free Plugins and for our premium plugins you can help with translations here:

http://translations.theeventscalendar.com/projects

PR #745, opened 2018-05-08
Display the correct number of available tickets on list and day view

juanfra:
https://central.tri.be/issues/100340

Couldn't find a more elegant way to fix this. Basically, the shared capacity was adding to the stock when an event had all tickets with their own stock. Added a helper function and modified the addition for that special case.


jesseeproductions:
👍

PR #751, opened 2018-05-18
Feature/79549 rsvp acceptance

niconerd:
MR ticket: https://central.tri.be/issues/106603
Original research ticket: https://central.tri.be/issues/79549


tests/_support/AcceptanceTester.php
```diff
@@ -23,4 +23,39 @@ class AcceptanceTester extends \Codeception\Actor
...
+		$I->haveOptionInDatabase( 'tribe_events_calendar_options', $tribe_options );
+
+		/*$code = <<< PHP
+add_filter( 'tribe_tickets_post_types', 'test_ticketable_post_types' );
+function test_ticketable_post_types(){
+	return array( 'post' );
+}
+PHP;
+
+		$I->haveMuPlugin('ticketable-post-types.php',$code);*/
```
ggwicz: any reason to keep these comments in place? if not, remove

tests/_support/AcceptanceTester.php
```diff
@@ -23,4 +23,39 @@ class AcceptanceTester extends \Codeception\Actor
...
     */
+
+   	/**
+	 * Sets tickets enabled post types
+	 *
+	 * @param  array $post_types The post types to be enabled for tickets. Can be empty.
+	 */
+	public function haveTicketablePostTypes ( array $post_types = array( ) ) {
+
+		$I = $this;
```
ggwicz: you don't _need_ to change this, but I'd recommend just using `$this` all of the time instead of aliasing it to `$I`

although if this is a common (or even recommended) pattern with acceptance tests, which it seems to be, then please ignore this comment 😎

tests/_support/AcceptanceTester.php
```diff
@@ -23,4 +23,39 @@ class AcceptanceTester extends \Codeception\Actor
    /**
     * Define custom actions here
     */
+
+   	/**
+	 * Sets tickets enabled post types
+	 *
+	 * @param  array $post_types The post types to be enabled for tickets. Can be empty.
+	 */
+	public function haveTicketablePostTypes ( array $post_types = array( ) ) {
```
ggwicz: our method/function/var names should always be `snake_case` in PHP

tests/_support/AcceptanceTester.php
```diff
@@ -23,4 +23,39 @@ class AcceptanceTester extends \Codeception\Actor
...
+}
+PHP;
+
+		$I->haveMuPlugin('ticketable-post-types.php',$code);*/
+	}
+
+	/**
+	 * Activates The Events Calendar
+	 */
+	public function haveTheEventsCalendarActive ( ) {
```
ggwicz: `snake_case`

tests/acceptance/Commerce/RSVP/TicketCreationCest.php
```diff
@@ -0,0 +1,83 @@
...
+		$I->dontSeeElement('#rsvp_form_toggle');
+	}
+
+	/**
+	 *
+	 * Provides post types as examples for tests
+	 *
+     * @return array
+     */
+	protected function post_types_provider () {
```
ggwicz: fix  php docblock formatting here

tests/_support/AcceptanceTester.php
niconerd: @ggwicz this seems to be the good practice, found it this way in the official codeception docs.

tests/_support/AcceptanceTester.php
niconerd: @lucatume I found the DB method slighly faster while running the tests but I'd defer to your opinion here.

tests/_support/AcceptanceTester.php
niconerd: @ggwicz in this case I'm following the casing of the 'standard' of the existing Actor methods and the ones added by WP Browser. Not sure if my approach is the correct one.

niconerd:
@ggwicz thanks a lot for the review. I added some comments on specific things and fixed the other stuff (commit incoming). Regarding the comments these tests will serve as example for other QA folks digging into this, that's why there are almost obvious comments. But if you think this should be documented in other fashion anyway, please let me know.

tests/_support/AcceptanceTester.php
lucatume: If they proved to be faster then use the method you found best and remove this code, thanks!

tests/_support/AcceptanceTester.php
lucatume: Both ways are really fine. My personal way is: anything I add in the context of WordPress related work I `snake_case`.

tests/_support/AcceptanceTester.php
niconerd: Actually re-tested and it's not. I'll change back to the muplugin method.

niconerd:
@ggwicz can you take a new look? I decided to keept the `$I` and the camel casing for this cases.

@lucatume I see an integration test failing in the CI. Can you help me fix that one? Also seems like the job time is exceeded. I took a look at the code but it far above my technical chops. Thanks.

bordoni:
We need to resolve these conflicts, after that I feel like this is good for a merge.

niconerd:
@bordoni I think there's more work to be done in the setup. Right now running the tests as they are (with current acceptance setup) would produce failures.

Camwyn:
A lot has changed since the project was started, and things interfered with this being finished prior - now the environment is vastly different (blocks!) so this PR (and the associated ticket) is on hold until we determine the direction we're going with it.

sc0ttkclark:
Going to close this PR because we've got a lot to revisit on this. We can open up a new PR whenever we have time to resolve the problems.

sc0ttkclark:
Related: https://central.tri.be/issues/106603

PR #755, opened 2018-05-25
⌗41649 Ensure QR Code Option Only Shows when Event Tickets Plus is Active

ggwicz:
**Ticket:** [**⌗41649**](http://central.tri.be/issues/41649)

**Code Review Note:** No changelog adjustments are needed for this.

**Companion _Event Tickets Plus_ PR:** https://github.com/moderntribe/event-tickets-plus/pull/569

### DIFF
--- a/src/admin-views/tribe-options-tickets.php
+++ b/src/admin-views/tribe-options-tickets.php
@@ -96,22 +96,6 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 	) );
 }
 
-if ( class_exists( 'Tribe__Tickets_Plus__Main' ) ) {
-
-	$tickets_fields = array_merge( $tickets_fields, array(
-		'ticket-enable-qr-codes-heading' => array(
-			'type' => 'html',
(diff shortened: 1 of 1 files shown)


src/admin-views/tribe-options-tickets.php
```diff
@@ -93,12 +93,21 @@
 			'parent_option'   => Tribe__Events__Main::OPTIONNAME,
 			'default'         => reset( $ticket_form_location_options ),
 		),
+	) );
+}
+
+if ( class_exists( 'Tribe__Tickets_Plus__Main' ) ) {
```
mitogh: I think using the filter `tribe_tickets_settings_tab_fields` instead on Tickets+ would remove the need of the class `class_exists` as will be created directly on event Tickets+, and also favors extensibility.

By making sure this item is modified only where is required by the item that requires this change.

Let me know

ggwicz:
That's the right call, thanks @mitogh. This PR will now just remove that code from ET. The code to add it is now over in this new ET+ PR → https://github.com/moderntribe/event-tickets-plus/pull/569

ggwicz:
Both are now ready for a review! 🔍

PR #771, opened 2018-06-10
Add notice if user doesn't have tickets

nefeline:
Ref. https://central.tri.be/issues/89201

Display a notice if the user accesses the tickets page and doesn't have tickets.

### DIFF
--- a/src/views/tickets/orders.php
+++ b/src/views/tickets/orders.php
@@ -16,6 +16,14 @@ $view = Tribe__Tickets__Tickets_View::instance();
 $event_id = get_the_ID();
 $event = get_post( $event_id );
 $post_type = get_post_type_object( $event->post_type );
+$user_id = get_current_user_id();
+
+/**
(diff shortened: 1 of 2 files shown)


src/views/tickets/orders.php
```diff
@@ -16,6 +16,14 @@
...
 $event = get_post( $event_id );
 $post_type = get_post_type_object( $event->post_type );
+$user_id = get_current_user_id();
+
+/**
+ * Display a notice if the user doesn't have tickets
+ */
+if ( ! $view->has_ticket_attendees( $event_id, $user_id ) && ! $view->has_rsvp_attendees( $event_id, $user_id ) ) {
+	Tribe__Notices::set_notice( 'ticket-no-results', esc_html__( "You don't have tickets for this event", 'event-tickets' ) );
+};
```
bordoni: remove that `;`

PR #780, opened 2018-06-14
Spotfix/108436 paypal purchase record

juanfra:
🎫 https://central.tri.be/issues/108436

Some PayPal orders were not being recorded because we were receiving the custom parameter in the following format instead of a JSON: `user_id:N,tribe_handler:tpp,pid:N`

### DIFF
--- a/src/Tribe/Commerce/PayPal/Custom_Argument.php
+++ b/src/Tribe/Commerce/PayPal/Custom_Argument.php
@@ -58,7 +58,7 @@ class Tribe__Tickets__Commerce__PayPal__Custom_Argument {
 		}
 
 		// in case we receive the param in non json 'user_id:0,tribe_handler:tpp,pid:4' format
-		if ( 'false' === strpos( $encoded, '{' ) && 'false' === strpos( $encoded, '}' ) ) {
+		if ( false === strpos( $encoded, '{' ) && false === strpos( $encoded, '}' ) ) {
 			// we create the json from that string
 			$encoded = explode( ',', $encoded );
 
--- a/src/Tribe/Commerce/PayPal/Custom_Argument.php
+++ b/src/Tribe/Commerce/PayPal/Custom_Argument.php
@@ -79,7 +79,7 @@ class Tribe__Tickets__Commerce__PayPal__Custom_Argument {
 
 
 		// in case we receive the param in non json 'user_id:0,tribe_handler:tpp,pid:4' format
-		if ( false === json_decode( $encoded ) ) {
+		if ( false === $decoded = json_decode( urldecode_deep( $encoded ), $assoc_array ) ) {
 			// we create an array that string
 			$encoded = explode( ',', $encoded );
 
@@ -89,8 +89,6 @@ class Tribe__Tickets__Commerce__PayPal__Custom_Argument {
 			// we convert it into a json
 			$encoded = json_encode( $encoded );
 		}
-
-		$decoded = json_decode( urldecode_deep( $encoded ), $assoc_array );
 
 		if ( null === $decoded ) {
 			return $assoc_array ? array() : new stdClass();
--- a/src/Tribe/Commerce/PayPal/Custom_Argument.php
+++ b/src/Tribe/Commerce/PayPal/Custom_Argument.php
@@ -77,13 +77,15 @@ class Tribe__Tickets__Commerce__PayPal__Custom_Argument {
 			$encoded = str_replace( '\"', '"', $encoded );
 		}
 
+
 		// in case we receive the param in non json 'user_id:0,tribe_handler:tpp,pid:4' format
-		if ( false === strpos( $encoded, '{' ) && false === strpos( $encoded, '}' ) ) {
+		if ( false === json_decode( $encoded ) ) {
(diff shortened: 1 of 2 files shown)


src/Tribe/Commerce/PayPal/Custom_Argument.php
```diff
@@ -57,6 +57,20 @@ public static function decode( $encoded, $assoc_array = false ) {
 			$encoded = str_replace( '\"', '"', $encoded );
 		}
 
+		// in case we receive the param in non json 'user_id:0,tribe_handler:tpp,pid:4' format
+		if ( ! strpos( $encoded, '{' ) && ! strpos( $encoded, '}' ) ) {
```
mitogh: You might need to check here exactly for false instead of a falsy value as 0 might be returned from `strpos` and is false

src/Tribe/Commerce/PayPal/Custom_Argument.php
```diff
@@ -57,6 +57,20 @@ public static function decode( $encoded, $assoc_array = false ) {
...
+			// we create the json from that string
+			$encoded = explode( ',', $encoded );
+
+			$encoded = array_reduce( $encoded, function ( $array, $item ) {
```
mitogh: Is this PHP 5.2 compatible?

mitogh: The callback of `array_reduce`

juanfra: It seems: http://php.net/manual/en/function.array-reduce.php

Haven't tested it, though.

bordoni: Nope, callback needs to be setup into another method.

mitogh: Woops I mean the anonymous function not `array_reduce`

src/Tribe/Commerce/PayPal/Custom_Argument.php
```diff
@@ -57,6 +57,20 @@ public static function decode( $encoded, $assoc_array = false ) {
 			$encoded = str_replace( '\"', '"', $encoded );
 		}
 
+		// in case we receive the param in non json 'user_id:0,tribe_handler:tpp,pid:4' format
+		if ( false === strpos( $encoded, '{' ) && false === strpos( $encoded, '}' ) ) {
+			// we create the json from that string
+			$encoded = explode( ',', $encoded );
+
+			$encoded = array_reduce( $encoded, function ( $array, $item ) {
```
bordoni: We can't use anonymous functions on PHP 5.2. we need to setup a method and pass it as a `array()` callable here.

bordoni:
This seems to be breaking the tests I think. We should take a look at that.

juanfra:
https://github.com/moderntribe/event-tickets/blob/master/tests/wpunit/Tribe/Tickets/Commerce/PayPal/Custom_ArgumentTest.php#L79

The tests are not up to date with the changes. Paypal is not returning a valid JSON, and that's what I had to adapt.

bordoni:
So how should we change the tests?

src/Tribe/Commerce/PayPal/Custom_Argument.php
```diff
@@ -57,6 +77,18 @@ public static function decode( $encoded, $assoc_array = false ) {
 			$encoded = str_replace( '\"', '"', $encoded );
 		}
 
+		// in case we receive the param in non json 'user_id:0,tribe_handler:tpp,pid:4' format
+		if ( false === strpos( $encoded, '{' ) && false === strpos( $encoded, '}' ) ) {
```
lucatume: why not try to `json_decode` and check if it's false? The PHP implementation will make the same check you are doing at a much lower, and faster, level

src/Tribe/Commerce/PayPal/Custom_Argument.php
```diff
@@ -57,6 +77,18 @@ public static function decode( $encoded, $assoc_array = false ) {
...
+			// we create an array that string
+			$encoded = explode( ',', $encoded );
+
+			// Set the proper keys and values for the new array
+			$encoded = array_reduce( $encoded, 'self::array_fix_keys', array() );
```
lucatume: not really a fan as IDEs here will not recognize the use, please change to:
```php
$encoded = array_reduce(array(__CLASS__, 'array_fix_keys'), array());
```

src/Tribe/Commerce/PayPal/Custom_Argument.php
```diff
@@ -57,6 +77,19 @@ public static function decode( $encoded, $assoc_array = false ) {
 			$encoded = str_replace( '\"', '"', $encoded );
 		}
 
+
+		// in case we receive the param in non json 'user_id:0,tribe_handler:tpp,pid:4' format
+		if ( false === json_decode( $encoded ) ) {
```
lucatume: do not waste this, set the `$decoded` var here:
```php
if ( false === $decoded = json_decode...
```

lucatume:
Good to merge, thanks!

PR #785, opened 2018-06-19
Process and generate Paypal orders

juanfra:
https://central.tri.be/issues/108436

Add missing argument for the array_reduce callback. We really need to call the json_decode on line 93 with the new $encoded generated inside the IF statement


juanfra:
Checking the tests

tests/wpunit/Tribe/Tickets/Commerce/PayPal/Custom_ArgumentTest.php
```diff
@@ -68,7 +68,6 @@ public function should_allow_getting_the_original_custom_arguments() {
 		$encoded = urlencode_deep( json_encode( $args ) );
 
 		$this->assertEquals( (object) $args, Custom_Argument::decode( $encoded ) );
-		$this->assertEquals( $args, Custom_Argument::decode( $encoded, true ) );
```
mitogh: Why this one is removed?

juanfra: Actually, both have to be removed. Because the `decode` method has changed, given that paypal changed the way they return their data.

@lucatume can we get rid of those two assertions?

bordoni: I think we are ok on removing, since the other tests cover the bases, and the changes on the code above seems to be invalidating this test.

src/Tribe/Commerce/PayPal/Custom_Argument.php
```diff
@@ -77,19 +77,20 @@ public static function decode( $encoded, $assoc_array = false ) {
...
 
 			// Set the proper keys and values for the new array
-			$encoded = array_reduce( array( __CLASS__, 'array_fix_keys' ), array() );
+			$encoded = array_reduce( $encoded, array( __CLASS__, 'array_fix_keys' ), array() );
 
 			// we convert it into a json
 			$encoded = json_encode( $encoded );
 		}
 
+		$decoded = json_decode( urldecode_deep( $encoded ), $assoc_array );
```
borkweb: In the event that the initial `json_decode()` is not null, we'd end up doing `json_decode()` twice. Perhaps something like this would be more efficient?

```php
if ( null === $decoded = json_decode( urldecode_deep( $encoded ), $assoc_array ) ) {
  // do stuff with $encoded

  $decoded = json_decode( urldecode_deep( $encoded ), $assoc_array );
}
```

juanfra: The first `json_decode` is to check if it's a valid JSON, and if not (that's the genesis of this issue, PayPal changing the way they send back the data)we have to recreate it from the string we're receiving (i.e.:`user_id:0,tribe_handler:tpp,pid:4'`).

The second `json_decode` is to actually work around with the valid JSON (if it comes as a valid parameter or if we recreate it from what we get).

If I move the second one inside the IF, then if they send a valid JSON, the function won't be returning the decoded value.

juanfra: Already worked on something like that after the first CR here: https://github.com/moderntribe/event-tickets/pull/780/files and broke the patch

PR #797, opened 2018-06-25
Fix Issues with calculating and displaying ticketed events on the admin list

juanfra:
🎫 https://central.tri.be/issues/71122

Figured out that the "ticketed events" list was sometimes not including the "past" events.


juanfra:
Thank you, Sir!

PR #799, opened 2018-06-27
Do not enable tickets on events by default when activating TEC

juanfra:
🎫 https://central.tri.be/issues/73599

As requested on the ticket:

> The settings behavior should be consistent. Since the pages and posts options are not checked by default when you first activate Event Tickets by itself, the events option should also not be checked when The Events Calendar is first activated.


bordoni:
Check on tests, there seems to be a few failing.

PR #803, opened 2018-06-29
Spotfix: string didn't make it to POT

andrasguseo:
String was not picked up by the POT file, likely due to curly brackets.
https://central.tri.be/issues/109638

### DIFF
--- a/src/admin-views/tribe-options-tickets.php
+++ b/src/admin-views/tribe-options-tickets.php
@@ -272,8 +272,9 @@ $paypal_fields            = array(
 	'ticket-paypal-success-page'      => array(
 		'type'            => 'dropdown',
 		'label'           => esc_html__( 'Success page', 'event-tickets' ),
-		'tooltip'         => esc_html__( "After a successful PayPal order users will be redirected to this page; use the [{$tpp_success_shortcode}] shortcode to display the order confirmation to the user in the page content.",
-			'event-tickets' ),
+		'tooltip'         => sprintf(
+		                         esc_html__( "After a successful PayPal order users will be redirected to this page; use the %s shortcode to display the order confirmation to the user in the page content.", 'event-tickets' ),
+		                         '[$tpp_success_shortcode]' ),
 		'size'            => 'medium',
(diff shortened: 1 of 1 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | String &quot;After a successful PayPal order users will be redirected to this page; use the %s shortcode to display the order confirmation to the user in the page content.&quot; does not require double quotes; use single quotes instead | [tribe-options-tickets.php:276](https://github.com/moderntribe/event-tickets/blob/8921a9328b1049800b5975dbeaf023f4280badd5/src/admin-views/tribe-options-tickets.php#L276)

src/admin-views/tribe-options-tickets.php
```diff
@@ -272,8 +272,9 @@
 	'ticket-paypal-success-page'      => array(
 		'type'            => 'dropdown',
 		'label'           => esc_html__( 'Success page', 'event-tickets' ),
-		'tooltip'         => esc_html__( "After a successful PayPal order users will be redirected to this page; use the [{$tpp_success_shortcode}] shortcode to display the order confirmation to the user in the page content.",
-			'event-tickets' ),
+		'tooltip'         => sprintf(
```
mitogh: Thanks for this fix @andrasguseo 

Just a small thing regarding the formatting of your params in the function to make it more easier to spot what params are used. 

Would you mind update with this format instead.

```php
sprintf(
	esc_html__( 
		"text",
		'event-tickets'
	)
	'[$tpp_success_shortcode]'
)
```

Thank you

andrasguseo: Thanks! I'm always struggling with the tabs and spaces. :)

mitogh: Great let me know if I can help with anything but you might want to type it as if it might be wrong from GH so is easier if you use your tabs on your editor to update accordingly.

andrasguseo: Can you please check now @mitogh ?

PR #815, opened 2018-07-10
#110248 - fix tribe_is_event fatal if the events calendar is not active

jesseeproductions:
This fatal occurs when The Events Calendar is not active.


src/Tribe/REST/V1/Headers/Base.php
```diff
@@ -92,7 +92,7 @@ public function get_api_origin_meta_name() {
 	 * @return string
 	 */
 	public function get_rest_url() {
-		if ( is_single() && tribe_is_event() ) {
+		if ( is_single() && is_singular( 'tribe_events' ) ) {
```
Camwyn: Isn't this redundant? 
Not a real issue, since `tribe_is_event` is getting fixed (and I assume this will go away then) but thought I'd point it out?

PR #816, opened 2018-07-10
Implement 'show description' setting for tribe commerce tickets

niconerd:
https://central.tri.be/issues/100524


readme.txt
```diff
@@ -124,6 +124,8 @@ Currently, the following add-ons are available for Event Tickets:
 
 = [4.7.6] TBD =
 
+* Fix - Fixd the `Show description` setting for Tribe Commerce tickets in the backend and frontend [100524]
```
juanfra: Please use "Fixed". I would use double quotes for the "Show description" given that the other ones can convert into code in some previews.

niconerd: Fixed, thanks!

PR #822, opened 2018-07-16
Stop showing posts with "pending review" status in the blog page.

juanfra:
🎫 https://central.tri.be/issues/102184

I was hesitant at first about how to fix this one, and I ended up deciding to change the wording. The pending post status is reserved and we shouldn't mess with that. It's better to change it now and provide a snippet in case somebody has an issue with their paypal orders where the status was pending.


bordoni:
Merging latest release branch should fix the problem.

PR #824, opened 2018-07-17
Make sure we're calling the function with the correct param

juanfra:
🎫 https://central.tri.be/issues/110674

Fix for bug introduced in 🎫 https://central.tri.be/issues/107121

### DIFF
--- a/src/Tribe/Ticket_Object.php
+++ b/src/Tribe/Ticket_Object.php
@@ -322,7 +322,7 @@ if ( ! class_exists( 'Tribe__Tickets__Ticket_Object' ) ) {
 			}
 
 			// Bail if we don't have an end date and the event has passed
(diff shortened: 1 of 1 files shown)


src/Tribe/Ticket_Object.php
```diff
@@ -322,7 +322,7 @@ public function date_in_range( $datetime = 'now' ) {
 			}
 
 			// Bail if we don't have an end date and the event has passed
-			if ( empty( $end ) && tribe_is_past_event() ) {
+			if ( empty( $end ) && tribe_is_past_event( tribe_events_get_event( $this->event_id ) ) ) {
```
bordoni: Shouldn't we split this into a variable? Before passing it directly into another method? Maybe check the value?

juanfra: done

PR #831, opened 2018-07-23
Jentheo patch 82684

jentheo:
Adds start sale date and time (off by default) to tickets not yet available message, plus some filters to turn these on or off.


jentheo:
Ticket: 82684

PR #839, opened 2018-08-03
Merge MR 18.11 into bucket

lucatume:
Ticket: https://central.tri.be/issues/111522


lucatume:
tests will pass when common is merged

PR #840, opened 2018-08-03
Fixed datepicker formats validation error on ticket start sale date

juanfra:
🎫 https://central.tri.be/issues/102815

The origin of the problem is in the validation script in common. Bumping version and adding readme entry to log that we solved this one.

Related: https://github.com/moderntribe/tribe-common/pull/750

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -124,6 +124,8 @@ Currently, the following add-ons are available for Event Tickets:
 
 = [4.7.7] TBD =
 
+* Fix - Fixed datepicker formats YYYY.MM.DD, MM.DD.YYYY and DD.MM.YYYY validation error on ticket start sale date. Thanks @dmitry-zhuk, Albert and others for reporting this issue! [102815]
+
 = [4.7.6] 2018-08-01 =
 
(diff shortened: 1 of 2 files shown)


mitogh:
Changes seems okay however I would like @barryhughes to take another look as we have an update of min required version from `common` and if I recall correctly we flagged this type of changes based on last MR updates.

mitogh:
cc @juanfra ⬆️

juanfra:
Thanks @mitogh! Yeah, the version bump is basically because we need the changes on moderntribe/tribe-common#750 for this to work.

barryhughes:
> I would like @barryhughes to take another look as we have an update of min required version from common and if I recall correctly we flagged this type of changes based on last MR updates.

Thanks for flagging! This seems A-OK to me :-)

My thinking:

* A bump to the min required version of Tribe Common is fine here, because Event Tickets ships with it anyway (we can effectively guarantee that at least this version will be present and loaded)
* It's changes in the min required version of _other_ plugins that is more problematic during MRs (ie, if the min required version of TEC changes for PRO, because we can't guarantee that TEC will also be updated at the same time, etc)

PR #841, opened 2018-08-05
#107897 - get the query string without modifications to parse for act…

jesseeproductions:
…ive tab

_Ref:_ [C#107897](https://central.tri.be/issues/107897)

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -124,6 +124,8 @@ Currently, the following add-ons are available for Event Tickets:
 
 = [4.7.7] TBD =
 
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -124,6 +124,8 @@ Currently, the following add-ons are available for Event Tickets:
 
 = [4.7.7] TBD =
 
+* Fix - Fix the active tab logic for Attendees in Tribe Commerce, thanks Luc [107897]
```
bordoni: Should we avoid duplicating the word _Fix_?

PR #852, opened 2018-08-17
Do not send an invoice number, send it in the custom arguments

lucatume:
Ticket: https://central.tri.be/issues/109421

It looks like sending an `invoice` argument will block further payments; it's supposed to be a passthru argument for the benefit of the merchant but it appears not.
Since we use it, in reality, to identify the cart I've modified the code to send, and consume, the same information in the `custom` argument instead.


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *M18.13*, but this was pull requested against *release/M18.12*.

lucatume:
Travis CI tests keep failing due to a db timing issue and never in the same place.
The affected code runs correctly.

jesseeproductions:
Since Luca is out this week. I am going to merge.

PR #853, opened 2018-08-17
#112579 - rename method and keep protected method to prevent fatal er…

jesseeproductions:
…rors

### DIFF
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -643,12 +643,33 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 		}
 
 		/**
+		 * Get All Attendees by ticket/attendee ID
+		 *
+		 * @since TBD
+		 *
+		 * @param int $attendee_id
+		 * @return array
+		 */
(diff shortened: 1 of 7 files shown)


src/Tribe/Tickets.php
```diff
@@ -642,13 +642,34 @@ public function attendee_decreases_inventory( array $attendee ) {
...
+		 * Get All Attendees by ticket/attendee ID
+		 *
+		 * @since TBD
+		 *
+		 * @param int $attendee_id
+		 * @return array
+		 */
+		public function get_all_attendees_by_attendee_id( $attendee_id ) {
+			$attendees_query = new WP_Query( array(
+				'p'         => $attendee_id,
```
juanfra: Do we have some sort of "rule" on where to sanitize data when we make queries? Perhaps, taking into consideration that is a public method, we may want to sanitize the `$attendee_id` and be sure that is an int

jesseeproductions: Not sure we have a rule, but I think it is a good idea.

jesseeproductions:
This passed tests once, but now is failing after adding the sanitization. Looks like there are other issues with tests today. I think this is good to merge and get into QA for M18.12

PR #855, opened 2018-08-20
ticket_description cannot be null so default to an empty string

technosailor:
https://central.tri.be/issues/113038

Discovered on AAA project. When creating a ticket, if `ticket_description` is not set, then it defaults to null which WordPress rejects. So this changes the default to an empty string instead.


bordoni:
That is a good PR, we should get it into `release/M18.13`

juanfra:
Created ticket: https://central.tri.be/issues/113038

PR #862, opened 2018-09-03
Buy now button

Camwyn:
🎫 https://central.tri.be/issues/109730

This does a few things to the table, most notably using flex on smaller screens to fix the flow and get everything to fit properly. laso moved the inout down next to the button on smaller screens as it looked lonely up at the top?

screen shot 2018-09-03 at 5 12 15 pm


mitogh:
Just missing a changelog entry update.

PR #895, opened 2018-10-26
Attendee info template + assets

juanfra:
🎫 https://central.tri.be/issues/116709

The initial proposal we received was including calls to header and footer functions within the templates. We don't want to have it that way, in order to make it easier to be compatible with themes. As a result, we needed to include the views in the content of a phantom page. This PR includes the class where the phantom page is handled in order to have it working with the templates from `src/views`, hijacking the WP context. At the same time, imported the assets we done for this in gutenberg (the templates use the same classes), registered the asset and added enqueueing on demand.

**TLDR;**
* Create template logic so we use a phantom page with the attendee info populated by the templates in `src/views`
* Register the assets
* Import initial CSS structure
* Enqueue the assets on demand


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | TRUE, FALSE and NULL must be lowercase; expected &quot;null&quot; but found &quot;NULL&quot; | [Template.php:57](https://github.com/moderntribe/event-tickets/blob/43a54835c57ce51dc5588b60ba2dccf32268a876/src/Tribe/Attendee_Info/Template.php#L57)

tests/wpunit/Tribe/Tickets/Attendee_Info/ViewTest.php
```diff
@@ -1,73 +0,0 @@
-<?php
-
-namespace Tribe\Tickets\Attendee_Info;
```
mitogh: So this one is plan to be replaced or is no longer relevant?

src/Tribe/Attendee_Info/Template.php
```diff
@@ -0,0 +1,344 @@
...
+		'twentynineteen',
+		'twentyseventeen',
+		'twentysixteen',
+		'twentyfifteen',
+	);
+
+	/**
+	 * Initialize the template class
+	 */
+	public function init() {
```
mitogh: I think following the conventions this one should be called `hook`

src/Tribe/Attendee_Info/Template.php
```diff
@@ -0,0 +1,344 @@
...
+	public function setup_context( $posts ) {
+	    global $wp, $wp_query;
+
+	    // Bail if we're not on the attendee info page
+	    if ( empty( $wp->query_vars['attendee-info'] ) ) {
+			return $posts;
+		}
+
+		// Empty posts
+        $posts = null;
```
mitogh: Weird indentation on this method

src/Tribe/Attendee_Info/Template.php
```diff
@@ -0,0 +1,344 @@
...
+		// nor it loads the comments template
+		$wp_query->is_page     = false;
+        $wp_query->is_singular = false;
+        $wp_query->is_home     = false;
+        $wp_query->is_archive  = true;
+        $wp_query->is_category = false;
+        $wp_query->is_404      = false;
+        $wp_query->found_posts = 1;
+        $wp_query->posts_per_page = 1;
+
```
mitogh: Might be useful to set a custom variable like `tribe_is_past` here: 

- https://github.com/moderntribe/the-events-calendar/blob/1a7b164733efda167d769811006d38274e850641/src/Tribe/Query.php#L171

tests/wpunit/Tribe/Tickets/Attendee_Info/ViewTest.php
juanfra: No longer relevant because we're not using the method that was being used at first (`template_redirect`). That test was checking if the script exited or not. We're now using `template_include`, hijacking the context, creating a fake page, stealing the content and replacing that with what we get from the templates. Ideally the test to check this, should be different. I've been discussing with Luca about it.

And this one is not relevant any more.

src/Tribe/Attendee_Info/Template.php
juanfra: What would be the purpose of having that custom var?

src/Tribe/Attendee_Info/Template.php
juanfra: Thanks, changed my IDE recently. Will find a plugin to solve these ones.

src/Tribe/Attendee_Info/Template.php
mitogh: To access the property or logic to detect if `is_attendee_info_page` outside of this class by hooking into the `wp_query` instead. So extensibility mainly.

mitogh: Right however is part of the `query_vars` not of the query it self.

juanfra:
Closing in favor of https://github.com/moderntribe/event-tickets/pull/898

PR #900, opened 2018-11-07
Fix checkbox markup

niconerd:
https://central.tri.be/issues/117191

Fix markup for checkbox (not closing input tag), using the same format as radio template


src/views/registration/attendees/fields/checkbox.php
```diff
@@ -39,7 +39,7 @@ class="ticket-meta"
 					name="tribe-tickets-meta[<?php echo esc_attr( $attendee_id ) ?>][<?php echo esc_attr( $slug ); ?>]"
 					value="<?php echo esc_attr( $option ); ?>"
 					<?php echo $required ? 'required' : ''; ?>
-					<?php checked( true, in_array( $slug, $value ) ); ?>
+					<?php checked( true, in_array( $slug, $value ) ); ?>>
```
mitogh: Looks good however I would use explicit self closing to make a bit more clear and place it on the next line: 

```php
/>
```

PR #926, opened 2018-11-15
Render parts of the view conditionally if tickets+ is present

mitogh:
Ticket: https://central.tri.be/issues/117636


src/modules/blocks/tickets/settings/container.js
```diff
@@ -8,16 +8,20 @@ import { compose } from 'redux';
  * Internal dependencies
  */
 import TicketsSettingsDashboard from './template';
-
+import { plugins } from '@moderntribe/common/data/plugins';
```
nealfennimore: `import { plugins } from '@moderntribe/common/data';`

src/modules/data/sagas.js
```diff
@@ -6,8 +6,10 @@ import { sagas as RSVPSagas } from '@moderntribe/tickets/data/blocks/rsvp';
 import { sagas as TicketSagas } from '@moderntribe/tickets/data/blocks/ticket';
 import { sagas as AttendeesSagas } from '@moderntribe/tickets/data/blocks/attendees';
 
-[
-	RSVPSagas,
-	TicketSagas,
-	AttendeesSagas
-].forEach( sagas => store.run( sagas ) );
+export default () => {
```
nealfennimore: we should probably do the same pattern in our other plugins

mitogh: Yeah In this way we execute sagas only once, as right now `data` is included independently and then included in the blocks that triggers the `sagas` watchers to be executed twice.

PR #932, opened 2018-11-17
Report errors back to the client to avoid "silent failing"

mitogh:
Ticket: https://central.tri.be/issues/118118

### DIFF
--- a/src/modules/data/blocks/ticket/sagas.js
+++ b/src/modules/data/blocks/ticket/sagas.js
@@ -99,9 +99,7 @@ export function* removeActiveTicketBlock( action ) {
 		} );
 		yield put( actions.removeTicketBlock( blockId ) );
 	} catch ( e ) {
-		/**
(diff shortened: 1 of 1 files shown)


mitogh:
100%, however I think it should be executed only on a specific scenario as otherwise we might be shipping third party code into the plugin I would prefer to make it flexible enough so anyone can plug any logger system on top, right now I think it would be a great addition for us so we can debug easily or keep track of errors while we are running QA.

paulmskim:
Is this the best way of doing this, by showing an error in the console? Is there another way we can handle this instead?

mitogh:
I think main important take away of this is not handling the error is to report back to the client that there's an error happening so we don't have silent errors happening on catch blocks I agree that we should handle this but we should report back again so we know there's something wrong, on the other side is a `console.warn` which will be displayed if `warning` filter is displayed or all are displayed

PR #933, opened 2018-11-17
Add missing asssets from migration for FE view

mitogh:
Ticket: https://central.tri.be/issues/118344
Ticket: https://central.tri.be/issues/118342

### DIFF
--- a/src/resources/js/views/rsvp.js
+++ b/src/resources/js/views/rsvp.js
@@ -1,0 +1,183 @@
+(function( $, my ) {
+	'use strict';
+
+	// base elements
+	var tribe_rsvp = $( '.tribe-block__rsvp' );
+
+	// Bail if we don't have what we need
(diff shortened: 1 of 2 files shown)


src/resources/js/views/rsvp.js
```diff
@@ -0,0 +1,183 @@
+(function( $, my ) {
+	'use strict';
+
+	// base elements
+	var tribe_rsvp = $( '.tribe-block__rsvp' );
+
+	// Bail if we don't have what we need
+	if ( 0 === tribe_rsvp.length ) {
```
nealfennimore: Could simplify and just make `! tribe_rsvp.length`

mitogh: Yes I think that one would work in this case was a `cp` operation from `extension` as this files were not part of the merge but I think we should do a bit of review on them afterwards 👍

PR #935, opened 2018-11-19
Fix: Tickets & RSVP blocks > Attendee and Order links are underlined

juanfra:
🎫 https://central.tri.be/issues/118106

### DIFF
--- a/src/modules/elements/action-dashboard/style.pcss
+++ b/src/modules/elements/action-dashboard/style.pcss
@@ -27,12 +27,6 @@
 	flex: none;
 	margin-right: 25px;
 
-	.tribe-editor__link {
-		&, &:hover {
-			box-shadow: none;
-		}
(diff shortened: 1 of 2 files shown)


src/modules/elements/action-dashboard/style.pcss
```diff
@@ -27,6 +27,12 @@
 	flex: none;
 	margin-right: 25px;
 
+	.tribe-editor__link {
```
paulmskim: These styles should go in the action button instead, not in the action dashboard. Also need styles for `:focus`.

paulmskim: sure, no problem with the specificity. In the action button styles file, you can add another class specific for that style. but main thing is where the style lives so we're not putting styles for the same thing in different places.

PR #938, opened 2018-11-20
Fix/117446 rsvp date cutoff

paulmskim:
For ticket: https://central.tri.be/issues/117446

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -124,7 +124,7 @@ Currently, the following add-ons are available for Event Tickets:
 
 = [4.9] TBD =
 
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -124,7 +124,7 @@ Currently, the following add-ons are available for Event Tickets:
 
 = [4.9] TBD =
 
-
+* Tweak - Adjust rsvp duration picker wider to allow more of the date to be shown
```
mitogh: I'm not sure if we are adding individual change log entries now at the plugin level, any comments @bordoni ?

PR #942, opened 2018-11-21
Make sure header image is a string

mitogh:


paulmskim:
@mitogh could we also make sure that the header image id from the tickets block attributes and proptypes for the image upload are also strings?

mitogh:
Yes, however at this point on the component is not directly used by the Template is only used by the HOC that is in charge of saving the properties associated with the container. 

I'm adding a check anyways into the template. We already have one check in place where the ID is directly used: 

- https://github.com/moderntribe/event-tickets/blob/a7208c157469f2b6d80fb27894d6b6feccd0a8a0/src/modules/blocks/tickets/header-image/template.js#L46

PR #943, opened 2018-11-21
Specify the type of header on Template level

mitogh:
Followup: https://github.com/moderntribe/event-tickets/pull/942#issuecomment-440557455

### DIFF
--- a/src/modules/blocks/tickets/template.js
+++ b/src/modules/blocks/tickets/template.js
@@ -18,13 +18,14 @@ class Tickets extends PureComponent {
 		isSelected: PropTypes.bool,
 		isSettingsOpen: PropTypes.bool,
 		clientId: PropTypes.string,
+		header: PropTypes.string,
 	};
 
 	render() {
 		const {
 			isSelected,
 			isSettingsOpen,
-			clientId
+			clientId,
 		} = this.props;
 
 		return (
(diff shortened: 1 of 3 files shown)


src/modules/blocks/tickets/template.js
```diff
@@ -18,13 +18,14 @@ class Tickets extends PureComponent {
 		isSelected: PropTypes.bool,
 		isSettingsOpen: PropTypes.bool,
 		clientId: PropTypes.string,
+		header: PropTypes.string,
```
paulmskim: this is actually not what i meant in my previous comment, as we don't use the `header` prop in the template, I don't think it makes sense to have this here.

mitogh:
Well the only place where it's important to be a string is only on the Higher Component that is used on tickets as is the only place that is used to send back the ID into the `meta` part of the request and should remain as a string as the value is send back automatically on saving the post.

The rest of the links above seems fine as those access directly a number value as the ID is a number in the reducer, so I don't think there's seems to much benefit difference on changing those neither on this PR so I think closing this one might be better as the only place where it needs to be a string is here: 

- https://github.com/moderntribe/event-tickets/blob/a7208c157469f2b6d80fb27894d6b6feccd0a8a0/src/modules/blocks/tickets/container.js#L15

Let me know what do you think so we can decline and close this PR or update the rest. 

Thanks.

paulmskim:
Fair enough. However, wherever we use the selector `getTicketsHeaderImageId()` and the components that are used, the proptypes should be updated.

Also I realized the selector tests weren't updated, could you check that they're working?

mitogh:
Sure, I think we can create a specifc selector or do the string conversion into the `Container` level so all selectors return `Numbers` instead. 

Will look into the tests I don't think that one is in place on the travis integration.

mitogh:
@paulmskim changes has been applied, I'm checking tests in another PR to unblock this one.

PR #944, opened 2018-11-21
Display the ticket icon according to the date of the event 

juanfra:
🎫 https://central.tri.be/issues/115802

**📹 Screencast**
https://www.useloom.com/share/4f030baa968f44349483a8e29a5fa5e6

### DIFF
--- a/src/modules/blocks/ticket/container/container.js
+++ b/src/modules/blocks/ticket/container/container.js
@@ -13,6 +13,7 @@ import { selectors } from '@moderntribe/tickets/data/blocks/ticket';
 
 const mapStateToProps = ( state, ownProps ) => ( {
 	isDisabled: selectors.isTicketDisabled( state, ownProps ),
+	isFuture: selectors.isTicketFuture( state, ownProps ),
 } );
 
 export default compose(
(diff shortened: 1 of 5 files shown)


juanfra:
Thanks @paulmskim ! 

When you mention the requirements of the ticket you mean this? https://central.tri.be/issues/115802#note-22

I think, by definition a ticket will be scheduled if the start date of the ticket is actually > of the current moment. The icon is meant to display what the person is actually doing, if they change the date, and it's in the future, they see a clock. If the current date is within a ticket date params, then we need to show a ticket. Perhaps we can get some sort of clarification from strategy. If they set a ticket with a date from a day before, I assume the user would like to see a ticket and not a clock, and that's what it's doing.

Technically speaking, do you see something that's purely wrong? Or some code that can break stuff?

paulmskim:
Ah I did not see that note. However, that doesn't change my interpretation of the strategy. So how I'm understanding it is that "if the duration is manually adjusted", meaning if the start date, start time, end date, or end time is altered in any way, regardless of whether it's backwards or forwards, it should show the clock icon. It also requires the ticket to be created to display, which right now we do not have a check for. Would love to hop on a call to clarify this.

juanfra:
https://www.useloom.com/share/fefaecc6ecf84841af4be8d3468ada8c

juanfra:
Thanks @nealfennimore!

Ended up using `isMoment` as it's more reliable.

PR #946, opened 2018-11-22
Show attendees block subtitle in the BE preview

juanfra:
🎫 https://central.tri.be/issues/117039

Show the subtitle in the BE preview. Will log a ticket to actually show the same info as the FT

### DIFF
--- a/src/modules/blocks/attendees/template.js
+++ b/src/modules/blocks/attendees/template.js
@@ -24,6 +24,7 @@ import './style.pcss';
  */
 
 const placeholder = __( 'Who\'s Attending?', 'events-gutenberg' );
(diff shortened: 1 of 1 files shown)


src/modules/blocks/attendees/template.js
```diff
@@ -24,6 +24,7 @@ import './style.pcss';
  */
 
 const placeholder = __( 'Who\'s Attending?', 'events-gutenberg' );
+const subtitle    = __( 'How many people is attending this event', 'events-gutenberg' );
```
paulmskim: "are", not "is"?

juanfra: Thanks! yeah, didn't pay much attention to the message. Was more obsessed with trying to get the post title and number of attendees. Fixed 👍

PR #959, opened 2018-11-27
Prevent to include blocks on non event types

mitogh:


juanfra:
Shouldn't we get these from the options?

mitogh:
Thanks for the feedback @juanfra @paulmskim I did an adjustment

PR #966, opened 2018-11-28
Remove dependency from TEC for tickets

mitogh:
This will prevent to fail on cases where tickets or RSVP is required to
be injected into other types that are not events.

Ticket: https://central.tri.be/issues/117612

### DIFF
--- a/src/modules/data/blocks/rsvp/sagas.js
+++ b/src/modules/data/blocks/rsvp/sagas.js
@@ -1,6 +1,7 @@
 /**
  * External Dependencies
  */
+import moment from 'moment/moment';
 import { select as wpSelect, dispatch as wpDispatch } from '@wordpress/data';
 import { put, call, all, select, takeEvery } from 'redux-saga/effects';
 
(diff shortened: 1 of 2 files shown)


src/modules/data/blocks/rsvp/sagas.js
```diff
@@ -96,6 +102,7 @@ export function* initializeRSVP() {
 		] );
 	} catch ( err ) {
 		// ¯\_(ツ)_/¯
+		console.log( errr );
```
nealfennimore: Need to remove an `r` here

mitogh:
Oh! sorry I wasn't aware there was a doc for this, however when running and creating a ticket on a post type that is not an event `window.tribe.events.data.blocks.datetime.selectors.getStart` does not exists as the JS file is not loaded here, however I'm falling back to the default value used by the reducer that creates a new event to prevent an error being displayed on the screen. 

Let me know what do you think about this and if we should include strategy into this conversation.

Thanks.

nealfennimore:
@mitogh yes that sounds like a new case we need to solve for. If we put your changes in with how it is now, it'll end up changing how the default end date should be blank.

So the case we're trying to solve for is, creating ticket/RSVP on a non-event post type, that should have a default end date? I think we should sync up with Leah for this

nealfennimore:
Yes true, we need to update the date pickers to handle blanks eventually

PR #967, opened 2018-11-29
Add Block Ticket Item registration

mitogh:
Register the render mechanism for the Ticket Item, used to populate the
template argument "tickets" used to have access to the set of tickets used
to render the block.

So every ticket injected respects the order in which they are presented
into the BE.


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Each line in an array declaration must end in a comma | [Tickets_Item.php:49](https://github.com/moderntribe/event-tickets/blob/3b7783ada6f8b742b88f69500e6c785704a3cf8b/src/Tribe/Editor/Blocks/Tickets_Item.php#L49)

PR #973, opened 2018-11-29
RSVP: Remove dotted border when the icon wrapper takes the whole width

juanfra:
🎫 https://central.tri.be/issues/ 118952


src/styles/rsvp/frontend.pcss
```diff
@@ -36,6 +36,10 @@
 	svg {
 		width: 60px;
 	}
+
+	@media only screen and (max-width: 575px) {
```
paulmskim: better if you do mobile first for css, using `min-width` rather than the `max-width` media query.

juanfra: Ok, made the changes.

paulmskim: Awesome. Just one more thing to be aware of: when using `min-width`, in this case at `575px`, the styles in the media query take effect at and greater than `575px` screen width (so it includes `575px`. However, with how you had it at `max-width`, the styles take effect at and lesser than `575px` screen width. So in one case, the smaller viewport includes `575px`, the other case the larger viewport includes `575px`. I'd check to make sure that at `575px` screen width, the style is what you expect it to be.

juanfra:
Thanks, missed that.

PR #982, opened 2018-12-03
Add display Title/Subtitle functionality for the attendees block

juanfra:
🎫 https://central.tri.be/issues/ 117040


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *Sprint 6*, but this was pull requested against *release/M18.18*.

src/modules/blocks/attendees/template.js
```diff
@@ -82,27 +88,78 @@ const RenderSubtitle = () => (
...
+	render() {
+		return [
+			renderUI( this.props ),
+			renderControls( this.props ),
+		];
+	}
+
+}
+
 Attendees.propTypes = {
```
mitogh: Might be a good idea to add `onKeyDown` and `onClick` as prop types as well for this component.

src/modules/blocks/attendees/template.js
```diff
@@ -82,27 +88,78 @@ const RenderSubtitle = () => (
...
+	}
+
+	componentWillUnmount() {
+		const { onKeyDown, onClick } = this.props;
+		document.removeEventListener( 'keydown', onKeyDown );
+		document.removeEventListener( 'click', onClick );
+	}
+
+	render() {
+		return [
```
mitogh: Nothing to change here as it works at it's but I think it can make things a bit more _Reactish_, as the values from `renderUI` and `renderControls` are components you can refactor to change the name and the values here into something like.

```js
return [
	<UI {...props} />,
	<Controls {...props} />,
];
```

Just need to make sure `UI` and  `Controls` are capitalized also removing the `render` as they are components the `render on calling them is implicit.

src/modules/blocks/attendees/container.js
```diff
@@ -8,19 +8,33 @@ import { bindActionCreators, compose } from 'redux';
...
 const mapStateToProps = ( state ) => ( {
-	title: selectors.getTitle( state ),
+	title: UISelectors.getTitle( state ),
+	displayTitle: UISelectors.getDisplayTitle( state ),
+	displaySubtitle: UISelectors.getDisplaySubtitle( state ),
 } );
 
-const mapDispatchToProps = ( dispatch ) => bindActionCreators( actions, dispatch );
+const mapDispatchToProps = ( dispatch ) => ( {
+	...bindActionCreators( UIActions, dispatch ),
```
mitogh: Might be a good idea to specify the actions being passed here instead to make it clear which ones are the ones used into the component.

juanfra: I don't understand this one. Can you please give me an example? I assumed I'm already specifying the actions being passed here (at least the ones I've added). Thanks.

paulmskim: What @mitogh is talking about is something like https://github.com/moderntribe/event-tickets/blob/1cae571db0aa927d160deaf78d25090c3afe72e7/src/modules/blocks/rsvp/container-header/container.js#L29 that explicitly dispatches the actions. Using `bindActionCreators()` is concise, but it's not declarative and makes code difficult to follow.

Also, it looks like below you've got 2 dispatch props defined, which dispatch `UIActions`. This means you're double-declaring dispatch props, you'll have one called `onSetDisplayTitleChange` and another called `setDisplayTitle` which do the exact same thing (we're just passing unnecessary props). Best to declare your dispatch props declaratively rather than imperatively.

src/modules/data/blocks/attendees/sagas.js
```diff
@@ -13,9 +13,22 @@ import * as actions from './actions';
...
+			put( actions.setDisplaySubtitle( get( 'displaySubtitle', DEFAULT_STATE.displaySubtitle ) ) ),
+		] );
 }
 
 export default function* watchers() {
 	yield takeEvery( types.SET_ATTENDEES_INITIAL_STATE, setInitialState );
 }
+
+
+export function* onSetDisplayTitleChange( action ) {
```
paulmskim: this and `onSetDisplaySubtitleChange` below don't do anything here, you should remove them.

src/modules/blocks/attendees/container.js
mitogh: Yes, sorry for the confusion.

Instead of spreading all the properties from `bindActionCreators` you specify only the actions you need, like: 

https://github.com/moderntribe/event-tickets/blob/master/src/modules/blocks/rsvp/container.js#L64-L69

So it's clear which actions are passed in the components.

src/modules/blocks/attendees/template.js
```diff
@@ -82,27 +88,80 @@ const RenderSubtitle = () => (
...
 );
 
-const Attendees = ( props ) => {
-
-	const { isSelected, title } = props;
+const UI = ( props ) => {
+	const { isSelected, title, displayTitle, displaySubtitle } = props;
 	const blockTitle = ! ( isSelected || title )
 		? renderPlaceholder()
 		: [ renderLabelInput( props ) ];
```
paulmskim: this doesn't need the square brackets around it.

src/modules/blocks/attendees/template.js
```diff
@@ -82,27 +88,80 @@ const RenderSubtitle = () => (
 	</div>
 );
 
-const Attendees = ( props ) => {
-
-	const { isSelected, title } = props;
+const UI = ( props ) => {
+	const { isSelected, title, displayTitle, displaySubtitle } = props;
 	const blockTitle = ! ( isSelected || title )
```
paulmskim: I'm a little bit confused by the logic here. Is the `isSelected` actually necessary here? If there is no title, then we show a placeholder. If there is a title, we show the title in the input. Are we showing the placeholder if the block is not selected?

src/modules/blocks/attendees/container.js
```diff
@@ -8,19 +8,33 @@ import { bindActionCreators, compose } from 'redux';
  * Internal dependencies
  */
 import { withSaveData, withStore } from '@moderntribe/common/hoc';
-import * as actions from '@moderntribe/tickets/data/blocks/attendees/actions';
-import * as selectors from '@moderntribe/tickets/data/blocks/attendees/selectors';
+import * as UIActions from '@moderntribe/tickets/data/blocks/attendees/actions';
```
paulmskim: Are these actually `UIActions` and `UISelectors`? Since they're actions and selectors from the attendees block, wouldn't it make sense to just leave them as `actions` and `selectors`?

src/modules/blocks/attendees/template.js
```diff
@@ -82,27 +88,80 @@ const RenderSubtitle = () => (
...
+class Attendees extends Component {
+
+	componentDidMount() {
+		const { onKeyDown, onClick } = this.props;
+		document.addEventListener( 'keydown', onKeyDown );
+		document.addEventListener( 'click', onClick );
+	}
+
+	componentWillUnmount() {
+		const { onKeyDown, onClick } = this.props;
```
paulmskim: Where are the `onKeyDown` and `onClick` handlers coming from? I don't see them declared in the container nor do I see them in the actions.

src/modules/blocks/attendees/container.js
juanfra: Ok, I get it, that will make it more precise in terms of specificity, but I can't get it to work with the title change actions. If you want to enlighten me on how to do that I'll appreciate it.

paulmskim: for example you might have:
```js
const mapDispatchToProps = ( dispatch ) => ( {
    setDisplayTitle: ( checked ) => dispatch( actions.setDisplayTitle( checked ) );,
    ...
} );
```

juanfra: Yes, that will work with the set display title, but the block also has an input to edit the text of the title. That's what I mean.

juanfra: Besides, if I do that, the onchange for the toggle stop working as well.

src/modules/blocks/attendees/template.js
```diff
@@ -82,27 +88,80 @@ const RenderSubtitle = () => (
...
+	isSelected && (
+		<InspectorControls key="inspector">
+			<PanelBody title={ __( 'Attendees Settings', 'events-gutenberg' ) }>
+				<ToggleControl
+					label={ __( 'Display Title', 'events-gutenberg' ) }
+					checked={ displayTitle }
+					onChange={ onSetDisplayTitleChange }
+				/>
+				<ToggleControl
+					label={ __( 'Display Subtitle', 'events-gutenberg' ) }
```
bordoni: We need to use Tickets domain, here.

src/modules/blocks/attendees/template.js
```diff
@@ -82,27 +88,80 @@ const RenderSubtitle = () => (
...
+	displayTitle,
+	displaySubtitle,
+	onSetDisplayTitleChange,
+	onSetDisplaySubtitleChange,
+} ) => (
+	isSelected && (
+		<InspectorControls key="inspector">
+			<PanelBody title={ __( 'Attendees Settings', 'events-gutenberg' ) }>
+				<ToggleControl
+					label={ __( 'Display Title', 'events-gutenberg' ) }
```
bordoni: We need to use Tickets domain, here.

src/modules/blocks/attendees/template.js
```diff
@@ -82,27 +88,80 @@ const RenderSubtitle = () => (
...
+const Controls = ( {
+	isSelected,
+	displayTitle,
+	displaySubtitle,
+	onSetDisplayTitleChange,
+	onSetDisplaySubtitleChange,
+} ) => (
+	isSelected && (
+		<InspectorControls key="inspector">
+			<PanelBody title={ __( 'Attendees Settings', 'events-gutenberg' ) }>
```
bordoni: We need to use Tickets domain, here.

src/modules/blocks/attendees/container.js
paulmskim: I could write it out here, but maybe we should hop on a call to work this out. It might be faster.

juanfra: @paulmskim thanks for the clarification. I've just submitted the changes we chatted about. Please let me know how you see it

src/modules/blocks/attendees/template.js
juanfra: Yes, we're showing a default value for the title. This default value is the same that the block has on the FE.

src/modules/blocks/attendees/container.js
```diff
@@ -23,19 +23,13 @@ const mapStateToProps = ( state ) => ( {
 } );
 
 const mapDispatchToProps = ( dispatch ) => ( {
-	...bindActionCreators( actions, dispatch ),
-	onSetDisplayTitleChange: onSetDisplayTitleChange( dispatch ),
-	onSetDisplaySubtitleChange: onSetDisplaySubtitleChange( dispatch ),
+	dispatch,
```
paulmskim: let's remove this since we aren't using this in the template.

juanfra: Perfect, done.

paulmskim:
@juanfra I think you're good to go here. @bordoni your notes on tickets domain have been resolved.

juanfra:
Thanks!!!

PR #985, opened 2018-12-03
Fix/117627 new ticket issue

paulmskim:
For ticket: https://central.tri.be/issues/117627

### DIFF
--- a/src/modules/data/blocks/ticket/selectors.js
+++ b/src/modules/data/blocks/ticket/selectors.js
@@ -113,16 +113,6 @@ export const getTicketsCount = createSelector(
 	( allIds ) => allIds.length,
 );
 
-export const getTicketsIdsInBlocks = createSelector(
-	[ getTicketsArray ],
-	( tickets ) => tickets.reduce( ( accumulator, ticket ) => {
-		if ( ticket.ticketId !== 0 ) {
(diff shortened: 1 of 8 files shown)


tr1b0t:
Closing this Pull Request because it is pull requested against the wrong branch. The Central ticket is for *Sprint 6*, but this was pull requested against *release/M18.18*.

src/modules/data/blocks/ticket/selectors.js
```diff
@@ -113,21 +113,18 @@ export const getTicketsCount = createSelector(
...
-);
-
 export const hasTickets = createSelector(
 	[ getTicketsCount ],
 	( count ) => count > 0,
 );
 
+export const hasCreatedTickets = createSelector(
+	[ getTicketsArray ],
+	( tickets ) => tickets.reduce( ( hasCreated, ticket ) => (
```
nealfennimore: Are we trying to see if every item has been created, or just some of them? We can probably replace it with the `.every` or `.some` method if it's either of those cases

mitogh: We are trying to look if at least one is  `true`

paulmskim: ⬆️

PR #989, opened 2018-12-04
Load PUE when having invalid versions

bordoni:
_Ref:_ [C#119262](http://central.tri.be/issues/119262)
---

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -124,6 +124,8 @@ Currently, the following add-ons are available for Event Tickets:
 
 = [4.9.1] TBD =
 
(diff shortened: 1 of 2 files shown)


jesseeproductions:
Please add a readme entry as this will be the only change for ET in 18.17

juanfra:
Should we deprecate `maybe_include_et_plus_file` instead of just renaming the method? I see in the todo that is supposed to disappear. Just asking.

bordoni:
@juanfra  no need, it's a Private method.

juanfra:
Right, didn't see that. I'm burnt out!

PR #993, opened 2018-12-05
Fix > RSVP > Login is required for purchase not working> 

juanfra:
🎫 https://central.tri.be/issues/119270

I've found that we haven't updated the textdomain for the templates, also added that.

### DIFF
--- a/src/views/blocks/attendees/description.php
+++ b/src/views/blocks/attendees/description.php
@@ -10,12 +10,12 @@
  *
  * @link {INSERT_ARTCILE_LINK_HERE}
  *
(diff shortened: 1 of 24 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | A closing tag is not permitted at the end of a PHP file | [extra-available-quantity.php:21](https://github.com/moderntribe/event-tickets/blob/da911ed112fcbbcd4acd12ed805d252fc8dd2302/src/views/blocks/tickets/extra-available-quantity.php#L21)

PR #998, opened 2018-12-06
Fix/117428 time picker formats

paulmskim:
For ticket: https://central.tri.be/issues/117428


src/modules/elements/date-time-range-picker/element.js
```diff
@@ -202,32 +197,14 @@ class DateTimeRangePicker extends Component {
...
 			end: time.END_OF_DAY,
+			onBlur: onToTimePickerBlur,
 			onChange: onToTimePickerChange,
 			onClick: onToTimePickerClick,
+			onFocus: onToTimePickerFocus,
 			timeFormat: date.FORMATS.WP.time,
 			disabled: toTimeDisabled,
 		};
 
-		if ( isSameDay ) {
```
mitogh: 😍

PR #999, opened 2018-12-06
Feature/118337 rsvp default sale

nealfennimore:
https://central.tri.be/issues/118337

Allows RSVP end sale date/time to be kept in sync with event's start date/time, unless manually edited: http://p.tri.be/y17Kvl
- Fixes RSVP time not saving correctly 

On hold until https://github.com/moderntribe/event-tickets/pull/998 is merged in, as there is some overlapping changes

Also needs to add tests


src/modules/data/blocks/rsvp/sagas.js
```diff
@@ -125,9 +351,37 @@ export function* handleRSVPMove() {
 	}
 }
 
+/**
+ * Temporary bandaid until datepickers allow blank state
+ *
+ * @export
+ */
+export function* setNonEventPostTypeEndDate() {
+	if ( ! ( yield call( isTribeEventPostType ) ) ) {
```
mitogh: Can we invert logic here and return to make it instead.

```js
if ( yield call( isTribeEventPostType ) ) {
	return;
}
```

mitogh:
_Some conflicts to be resolved_

PR #1009, opened 2018-12-11
Sync MR 18.18 into MR.18.19

mitogh:


mitogh:
Hey @paulmskim @nealfennimore would you mind double check at the `diff` as there are some files that were creating conflicts so I took some parts from `18.18` and others from `18.19`

PR #1015, opened 2018-12-11
Feature/119049 ticket block default sale

nealfennimore:
https://central.tri.be/issues/119049

Allows ticket sale end date to be synced with event start date. Adds ticket saving when sale end date is synced and post is updated.

Adding tests but can be reviewed

### DIFF
--- a/src/modules/data/blocks/ticket/sagas.js
+++ b/src/modules/data/blocks/ticket/sagas.js
@@ -788,7 +788,7 @@ export function* syncTicketsSaleEndWithEventStart( prevStartDate ) {
  * @param {String} prevStartDate Previous start date before latest set date time changes
  * @export
  */
-export function* syncTicketSaleEndWithEventStart(prevStartDate, blockId){
+export function* syncTicketSaleEndWithEventStart( prevStartDate, blockId ){
 	try {
 		const tempEndMoment = yield select( selectors.getTicketTempEndDateMoment, { blockId } );
 		const endMoment = yield select( selectors.getTicketEndDateMoment, { blockId } );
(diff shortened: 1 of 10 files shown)


src/modules/data/blocks/ticket/sagas.js
```diff
@@ -737,6 +739,146 @@ export function* setTicketTempDetails( action ) {
...
+export function* handleEventStartDateChanges() {
+	try {
+		// Ensure we have a postType set before proceeding
+		const postTypeChannel = yield call( hasPostTypeChannel );
+		yield take( postTypeChannel );
+		yield call( [ postTypeChannel, 'close' ] );
+
+		const isEvent = yield call( isTribeEventPostType );
+		if ( isEvent && window.tribe.events ) {
+			const { SET_START_DATE_TIME, SET_START_TIME } = window.tribe.events.data.blocks.datetime.types;
```
mitogh: Looks good quick question about `window.tribe.events.data.blocks.datetime.types` as it might not be defined on post that are not events that has support for Tickets like `Post` or `Pages` or others.

nealfennimore: Yup it shouldn't work on those types, only on events

mitogh: Great sorry my question was if this code is executed there so in that case will trigger an error because `window.tribe.events.data.blocks.datetime` is not there.

nealfennimore: it's in a `try` block and there's a `window.tribe.events` check as well

nealfennimore: are you saying to throw an error?

mitogh: Oh that's right I overlooked that one 👍 thanks for answering my silly questions 💃 🕺

PR #1021, opened 2018-12-12
Set custom date format to date pickers

mitogh:
- Remove the need of having TEC variables created to setup date formats
on date pickers

Ticket: https://central.tri.be/issues/119356


src/resources/js/tickets.js
```diff
@@ -307,6 +306,23 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
 			var startofweek = 0;
 
+			/**
+			 * There might be cases when Tickets is used in isolation where TEC is not installed
+			 * for those cases tribe_datepicker_opts is undefined as is a variable defined by TEC. One of
+			 * the most important part of this variable is the dateFormat value, in this case we created
+			 * a new global variable so any other element that dependes on it has access to this value
+			 */
+			if ( typeof tribe_datepicker_opts === 'undefined' ) {
+				var dateFormatEl = document.querySelector( '[data-datepicker_format]' );
```
bordoni: If we have jQuery here, why are we mixing with Vanilla JS? Lets stick to one, for now, jQuery and we can make the decision to go Vanilla later.

mitogh: The decision was around two things.

1. Simplicity in code as we can just test if the variable is not `null` on the other hand on jQuery we need to test if the length of the variable is not `falsey` value
1. Related to the above but removes the need to create a jQuery object.

However I get the point here to keep it consistent as is mostly an old code base. 👍

PR #1031, opened 2018-12-14
Prevent the save of tickets as RSVPs

mitogh:
Ticket: https://central.tri.be/issues/

### DIFF
--- a/src/modules/data/blocks/ticket/selectors.js
+++ b/src/modules/data/blocks/ticket/selectors.js
@@ -578,6 +578,16 @@ export const getTicketProviders = () => {
 	return tickets.providers || [];
 };
 
+export const getDefaultProvider = () => {
+	const tickets = ticketsConfig();
+	return tickets.default_provider || '';
+};
+
+export const hasValidProvider = () => {
+	const provider = getDefaultProvider();
+	return provider !== '' && provider !== constants.RSVP_CLASS;
+};
+
 export const hasMultipleTicketProviders = createSelector(
 	[ getTicketProviders ],
 	( providers ) => providers.length > 1,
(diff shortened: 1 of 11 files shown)


src/modules/data/blocks/ticket/selectors.js
```diff
@@ -578,6 +578,16 @@ export const getTicketProviders = () => {
 	return tickets.providers || [];
 };
 
+export const getDefaultProvider = () => {
```
paulmskim: can we follow the naming convention here to stay consistent (`getDefaultTicketProvider`)

src/modules/data/blocks/ticket/selectors.js
```diff
@@ -578,6 +578,16 @@ export const getTicketProviders = () => {
...
+	const tickets = ticketsConfig();
+	return tickets.default_provider || '';
+};
+
+export const hasValidProvider = () => {
```
paulmskim: same here on naming consistency

src/modules/data/blocks/ticket/utils.js
```diff
@@ -4,12 +4,17 @@
 import { string, globals } from '@moderntribe/common/utils';
 
 const { settings, priceSettings, tickets: ticketsConfig } = globals;
+/**
+ * Internal dependencies
+ */
+import { getDefaultProvider, getTicketProviders } from '@moderntribe/tickets/data/blocks/ticket/selectors';
```
paulmskim: I personally feel that using selectors in utils is a bit of an anti-pattern, but since these selectors here are only grabbing straight from global configs, this is OK. Just something to be aware of in other cases.

PR #1039, opened 2018-12-22
Add helper text to form location settings

leahkoerper:
Clarifies that these two settings only matter for events made with the classic editor, not the block editor.

🎫 https://central.tri.be/issues/120963


sc0ttkclark:
There's also no ticket for this

jrserani:
_Ref:_ [C#120963](http://central.tri.be/issues/120963)

PR #1041, opened 2018-12-28
Remove the wp admin common requirement and test

Camwyn:
🎫 [C#119755](http://central.tri.be/issues/119755)


sc0ttkclark:
If https://github.com/moderntribe/event-tickets/blob/0a765c5ca571c5194e912ea818261016c088dd5e/src/Tribe/Main.php#L491 runs the admin enqueue method, should it really be named differently? I look at it and feel like there are things it loads that may only be needed in the WP dashboard, but yet it loads on front?

Camwyn:
`tribe_tickets_plugin_loaded` runs both the FE and admin enqueues - so maybe this does require more work to straighten out - but that may be digging into common?

sc0ttkclark:
I think it's worth one last look to confirm that we are enqueuing it correctly here. Are these styles supposed to all be enqueues on frontend AND admin?

Camwyn:
@sc0ttkclark 
> I think it's worth one last look to confirm that we are enqueuing it correctly here. Are these styles supposed to all be enqueues on frontend AND admin?

Yes, the styles are for the ticket interface - which is mainly used in the event admin, but is also used in the CT front-end ticket forms. So they are `admin` styles that are intentionally loaded on the FE.

At some point in the future we should convert CE & CT to use `tribe-assets` where we'll have some more fine-tuned control over when/where things get loaded. And/or create a FE stylesheet in ET for this. But those are a bigger lift than this ticket warrants, IMHO?

PR #1042, opened 2018-12-31
Add support for order in order_by method

lucatume:
Tickets:
* https://central.tri.be/issues/117811
* https://central.tri.be/issues/114674

Based on https://github.com/moderntribe/tribe-common/pull/904


lucatume:
The build passed before on the same exact code, CI keeps timing out so I'll merge this.

PR #1055, opened 2019-01-10
Feature/119466 block attendee registration

paulmskim:
For ticket: https://central.tri.be/issues/119466

Still a WIP, need to fix issue with REST API.


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | The method parameter $single is never used | [Compatibility.php:151](https://github.com/moderntribe/event-tickets/blob/73e2e1043e1d425e51a583eb94fe112db4f9f05c/src/Tribe/Editor/REST/Compatibility.php#L151)

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | The method parameter $single is never used | [Compatibility.php:155](https://github.com/moderntribe/event-tickets/blob/5eaba465ff175307f044ff0a2121815e2854af97/src/Tribe/Editor/REST/Compatibility.php#L155)

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | The method parameter $single is never used | [Compatibility.php:155](https://github.com/moderntribe/event-tickets/blob/4628ba5a89a42539dd6eeceecf90949c02064761/src/Tribe/Editor/REST/Compatibility.php#L155)

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | The method parameter $single is never used | [Compatibility.php:155](https://github.com/moderntribe/event-tickets/blob/7ff1fcfbe0ec8c498bca0b7ebb3ed628712add53/src/Tribe/Editor/REST/Compatibility.php#L155)

src/Tribe/Editor/Attendee_Registration.php
```diff
@@ -117,6 +117,10 @@ public function render() {
 		.accordion-header.tribe_attendee_meta {
 			display:none;
 		}
+
+		.accordion-content.tribe_attendee_meta {
+			display: block;
```
Camwyn: Um, why not remove the accordion, or at least just remove the `display:none` style above (L 117-119) instead of overriding it?

This is the issue - it's in an accordion and it shouldn't be anymore. Accordions hide their sections by default. This should really be a template change in the end - should have been when it got moved.

Camwyn: If we _have_ to reuse the same template, we should add the opening class (or remove the inline style?) so it's not collapsed.

src/Tribe/Editor/REST/Compatibility.php
```diff
@@ -134,4 +135,36 @@ public function filter_going_fields( $check, $object_id, $meta_key, $single ) {
...
+
+		$valid_keys = array(
+			'_tribe_ticket_has_attendee_info_fields',
+		);
+
+		if ( ! in_array( $meta_key, $valid_keys ) ) {
+			return $check;
+		}
+
+		$repository    = tribe( 'tickets.rest-v1.repository' );
```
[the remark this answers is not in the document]
paulmskim: Should I just remove the two filters above that are for `get_post_meta`? We really don't need them, correct?

src/Tribe/Editor/Attendee_Registration.php
paulmskim: Nope, I misread that. So the line I put is `accordion-content`, the line above is `accordion-header`. The call to render the accordion doesn't allow for a flag that sets the accordion as open or closed.

paulmskim: I agree on the template change though. I'd love to fix it, but will need a pass to make sure it doesn't affect other parts of the plugin and other plugins.

src/Tribe/Editor/REST/Compatibility.php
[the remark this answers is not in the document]
paulmskim: I've removed them.

src/Tribe/Editor/Attendee_Registration.php
paulmskim: So the accordion is still used in the classic editor. I pulled out the meta content into its own admin view and rendered that instead.

Camwyn: Nice!

src/Tribe/Editor/REST/Compatibility.php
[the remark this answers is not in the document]
Camwyn: Test that things work correctly on the FE without them? Those filter may have actually been turning those "fake" meta into real ones. If we're not using it as a "real" meta, or outside the api calls,  then removing them should be safe 👍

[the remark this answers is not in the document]
paulmskim: Seems to be working OK

PR #1070, opened 2019-01-16
Remove compiled assets

Camwyn:
Proposal for removing compiled files (and adding build to git checkout via [husky](https://github.com/typicode/husky))

Open to any and all suggestions - including using something other than husky, maybe [commitizen](https://github.com/commitizen/cz-cli)?

🎫 https://central.tri.be/issues/120854


Camwyn:
The purpose of the post-checkout hook is to ensure the build gets done since we're removing all the compiled files. That should cover anyone that clones the plugins rather than using the zips.

.huskyrc
```diff
@@ -0,0 +1,5 @@
+{
+  "hooks": {
```
bordoni: Can we remove this file?

package.json
```diff
@@ -74,6 +74,7 @@
     "gettext-parser": "1.3.0",
     "glob": "~4.0.5",
     "gulp": "^3.9.1",
+    "husky": "^1.3.1",
```
bordoni: Lets keep using `npm` scripts

.huskyrc
Camwyn: This is the config for husky, but we can put this in `package.json` instead. I'm just all for separation of concerns 😉

package.json
Camwyn: So, husky lets us hook `npm` scripts to git hooks - so we automate the build on each branch checkout (since we're removing the compiled files, we really need to do it every time to be sure we have the correct files in place)

.huskyrc
paulmskim: I'm indifferent about using husky, some people don't like it, others like it. But if we are use it, they should be in a separate file (much like our `.babelrc`, `.eslintrc`, etc.)

.gitignore
```diff
@@ -32,4 +32,9 @@ vendor/phpunit
 vendor/sebastian
 vendor/symfony
 vendor/autoload.php
+
 src/resources/**/app
+src/resources/css
+src/resources/js/**/*.min.js
```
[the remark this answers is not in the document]
Camwyn: `src/resources/**/app` on line 36?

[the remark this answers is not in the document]
Camwyn: My next task is to minify the `/views` js as well - and ignore it.

.huskyrc
nealfennimore: "Gotta keep em separated" - Offspring

Camwyn: I agree and will move it back.

Camwyn:
One remaining question: do we remove `package-lock.json` or leave it?

OK, two - is there anything else we should remove?

sc0ttkclark:
I think we need to leave package-lock.json so we an all use the same packages on install and the same ones get used by the jenkins packager.

sc0ttkclark:
We should probably hold on this until G19.04 mainly because I want to make sure product devs agree on everything here.

tr1b0t:
Status|Message|Line
:-----:|-----|-----
:warning: | Comment refers to a TODO task | [View.php:60](https://github.com/moderntribe/event-tickets/blob/8f3d0905fc066abd67d85fdc5eeaa0884b74178a/src/Tribe/Attendee_Registration/View.php#L60)

PR #1087, opened 2019-01-29
Fix/121592 attendee reg save data

paulmskim:
For ticket: https://central.tri.be/issues/121592


src/resources/postcss/tickets-registration-page.pcss
```diff
@@ -186,6 +186,31 @@ body.page-tribe-attendee-registration {
...
+.tribe-block__tickets__item__attendee__fields__success {
+	color: rgb(24, 217, 0);
+	display: none;
+	padding: 0 0 30px 0;
+}
+
+.tribe-block__tickets__item__attendee__fields__loader {
+	background: rgba( 255, 255, 255, .7 );
+	display: none;
+	position: absolute;
```
Camwyn: sorta nit-picky, but in products, we alphabetize properties (not going to repeat this for all the pcss :wink:)

src/views/registration/attendees/loader.php
```diff
@@ -0,0 +1,23 @@
...
+ * This template renders the attendee registration loader
+ *
+ * @since TBD
+ * @version TBD
+ *
+ */
+
+?>
+<div class="tribe-block__tickets__item__attendee__fields__loader">
+  <svg version="1.1" id="L4" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 100 100" enable-background="new 0 0 0 0" xml:space="preserve">
```
Camwyn: Q: Any reason to not use `<?php echo file_get_contents("loader.svg"); ?>` other than an additional HTTP request (which could be enough justification)? Just thinking of reusable components (specifically the svg itself), and that approach should give you the same inline experience/control.

Camwyn: Heck, add a filter for the file name and it allows customization for users as well...

paulmskim:
Closing in favor of https://github.com/moderntribe/event-tickets/pull/1102

PR #1104, opened 2019-02-06
Fix/119757 admin attendee reg with ct

paulmskim:
For ticket: https://central.tri.be/issues/119757


paulmskim:
Closing as this was fixed in an earlier release

PR #1105, opened 2019-02-06
Fix/122368 extensions register check with pdc

jesseeproductions:
_Ref:_ [C#122368](https://central.tri.be/issues/122368)


sc0ttkclark:
Needs common updated

PR #1115, opened 2019-02-13
Fixing text domain

andrasguseo:


paulmskim:
@sc0ttkclark this is the actual fix to https://github.com/moderntribe/event-tickets/pull/1103 so changelog and ticket should be the same.

paulmskim:
thanks @andrasguseo wasn't paying close enough attention here.

PR #1136, opened 2019-02-22
Add filter to show email form for non-admins

Camwyn:
🎫 https://central.tri.be/issues/123070

This is a compromise solution:

The user can activate the email form for non-admins via a filter as it works but it throws notices as it's still trying to load admin stuff on the FE. Once we have the notices corrected, we can add the filter as always-on in CT.

### DIFF
--- a/src/Tribe/Attendees_Table.php
+++ b/src/Tribe/Attendees_Table.php
@@ -540,9 +540,13 @@ class Tribe__Tickets__Attendees_Table extends WP_List_Table {
 
 		if ( $allow_fe && ! is_admin() ) {
 			$email_link = str_replace( '/wp-admin/', '/', $email_link );
-			$email_link = str_replace( 'page=tickets-attendees&', '', $email_link );
-
-			$email_link = substr( $email_link, 0, stripos( $email_link, '&post_type' ) );
+			$email_link = add_query_arg(
+				array(
+					'page' => null,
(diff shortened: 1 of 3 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | Expected 1 space after &quot;=&quot;; 2 found | [Attendees_Table.php:528](https://github.com/moderntribe/event-tickets/blob/c819d06264c911174175a9a8fa0dccb006dbcd7c/src/Tribe/Attendees_Table.php#L528)

src/Tribe/Attendees_Table.php
```diff
@@ -513,25 +513,51 @@ public function extra_tablenav( $which ) {
...
+		 * Filtert to show email form for non-admins.
+		 *
+		 * @since TBD
+		 *
+		 * @param boolean
+		 */
+		$allow_fe = apply_filters( 'tribe_allow_admin_on_frontend', false );
+		if ( $allow_fe && ! is_admin() ) {
+			global $wp;
+			$parent = $wp->request . '/';
```
sc0ttkclark: Should there be a untrailingslash() no the request property? Just asking since I don't use that property much.

Camwyn: I don't believe it needs it but it wouldn't hurt.

PR #1146, opened 2019-02-28
Show RSVP in list view when it's the only ticket

Camwyn:
🎫 https://central.tri.be/issues/123124

### DIFF
--- a/src/Tribe/RSVP.php
+++ b/src/Tribe/RSVP.php
@@ -193,7 +193,7 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 		add_action( 'rsvp_checkin', array( $this, 'purge_attendees_transient' ) );
 		add_action( 'rsvp_uncheckin', array( $this, 'purge_attendees_transient' ) );
 		add_action( 'tribe_events_tickets_attendees_event_details_top', array( $this, 'setup_attendance_totals' ) );
-		add_filter( 'tribe_get_cost', array( $this, 'tribe_trigger_get_cost'), 10, 3 );
+		add_filter( 'tribe_get_cost', array( $this, 'tribe_trigger_get_cost' ), 10, 3 );
 		add_filter(
 			'event_tickets_attendees_rsvp_checkin_stati',
 			array( $this, 'filter_event_tickets_attendees_rsvp_checkin_stati' )
@@ -1010,17 +1010,17 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 	 *
 	 * @param string $cost
 	 * @param int $post_id
-	 * @param boolean $with_currency_symbol
+	 * @param boolean $unused_with_currency_symbol
 	 *
 	 * @return string $cost
 	 */
-	public function tribe_trigger_get_cost( $cost, $post_id, $with_currency_symbol ) {
+	public function tribe_trigger_get_cost( $cost, $post_id, $unused_with_currency_symbol ) {
 
 		if (
 			empty( $cost )
 			&& tribe_events_has_tickets( get_post( $post_id ) )
 		) {
-			$cost = 'Free';
+			$cost = esc_html__( 'Free', 'event-tickets' );
 		}
 
 		return $cost;
--- a/src/Tribe/RSVP.php
+++ b/src/Tribe/RSVP.php
@@ -1006,6 +1006,7 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 
 	/**
 	 * Trigger for tribe_get_cost if there are only RSVPs
+	 *
 	 * @since TBD
 	 *
 	 * @param string $cost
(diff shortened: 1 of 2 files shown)


tr1b0t:
Status|Message|Line
:-----:|-----|-----
:x: | There must be a single space before the last parenthesis | [RSVP.php:196](https://github.com/moderntribe/event-tickets/blob/131ca1cc7bd53c461676e75f5286a4d713b7cfdf/src/Tribe/RSVP.php#L196)
:x: | There must be a single space before the last parenthesis | [RSVP.php:196](https://github.com/moderntribe/event-tickets/blob/131ca1cc7bd53c461676e75f5286a4d713b7cfdf/src/Tribe/RSVP.php#L196)
:x: | Missing space before array closer. | [RSVP.php:196](https://github.com/moderntribe/event-tickets/blob/131ca1cc7bd53c461676e75f5286a4d713b7cfdf/src/Tribe/RSVP.php#L196)
:warning: | The method parameter $with_currency_symbol is never used | [RSVP.php:1017](https://github.com/moderntribe/event-tickets/blob/131ca1cc7bd53c461676e75f5286a4d713b7cfdf/src/Tribe/RSVP.php#L1017)

src/Tribe/RSVP.php
```diff
@@ -1003,6 +1004,28 @@ public function delete_ticket( $event_id, $ticket_id ) {
...
+	 *
+	 * @return string $cost
+	 */
+	public function tribe_trigger_get_cost( $cost, $post_id, $with_currency_symbol ) {
+
+		if (
+			empty( $cost )
+			&& tribe_events_has_tickets( get_post( $post_id ) )
+		) {
+			$cost = 'Free';
```
bordoni: Do we need to translate? If not explain the reason so nobody in the future changes that.

src/Tribe/RSVP.php
```diff
@@ -1003,6 +1004,28 @@ public function delete_ticket( $event_id, $ticket_id ) {
...
+	 * Trigger for tribe_get_cost if there are only RSVPs
+	 * @since TBD
+	 *
+	 * @param string $cost
+	 * @param int $post_id
+	 * @param boolean $with_currency_symbol
+	 *
+	 * @return string $cost
+	 */
+	public function tribe_trigger_get_cost( $cost, $post_id, $with_currency_symbol ) {
```
bordoni: Why namespace it in our own class?

src/Tribe/RSVP.php
```diff
@@ -1003,6 +1004,28 @@ public function delete_ticket( $event_id, $ticket_id ) {
 		return true;
 	}
 
+	/**
+	 * Trigger for tribe_get_cost if there are only RSVPs
+	 * @since TBD
```
bordoni: Spacing before `@since`

src/Tribe/RSVP.php
```diff
@@ -193,6 +193,7 @@ public function hooks() {
 		add_action( 'rsvp_checkin', array( $this, 'purge_attendees_transient' ) );
 		add_action( 'rsvp_uncheckin', array( $this, 'purge_attendees_transient' ) );
 		add_action( 'tribe_events_tickets_attendees_event_details_top', array( $this, 'setup_attendance_totals' ) );
+		add_filter( 'tribe_get_cost', array( $this, 'tribe_trigger_get_cost'), 10, 3 );
```
bordoni: Odd spacing, we should also start using `[]` notation.

src/Tribe/RSVP.php
Camwyn: How is it odd? Looks right to me

sc0ttkclark: We usually add an extra line between.

src/Tribe/RSVP.php
```diff
@@ -1003,6 +1004,28 @@ public function delete_ticket( $event_id, $ticket_id ) {
...
+	 *
+	 * @return string $cost
+	 */
+	public function trigger_get_cost( $cost, $post_id, $unused_with_currency_symbol ) {
+
+		if (
+			empty( $cost )
+			&& tribe_events_has_tickets( get_post( $post_id ) )
+		) {
+			$cost = esc_html__( 'Free', 'event-tickets' );
```
sc0ttkclark: Just a question since I’m not at the computer, does this filter require you to escape the return? Wondering if this should be a plain __()

Camwyn: It's a bit of a tossup, I've removed it because it should be text only and it's a hard-coded (though translated) string. It bails if there is anything but an empty string/value (when it should display "Free". The filter itself could likely use some escaping, it's not escaped before returning?

Camwyn: Filter is here: https://github.com/moderntribe/the-events-calendar/blob/master/src/functions/template-tags/general.php#L830

sc0ttkclark: tribe_get_cost() usage elsewhere is probably / should be escaped, so yeah we can remove that escaping here and be good

PR #1147, opened 2019-03-01
Swap optional to the end

Camwyn:
🎫  https://central.tri.be/issues/123608

### DIFF
--- a/src/Tribe/Attendees.php
+++ b/src/Tribe/Attendees.php
@@ -800,8 +800,8 @@ class Tribe__Tickets__Attendees {
 		 * @since 4.10.1
 		 *
 		 * @param bool $user_can return value, user can or can't
-		 * @param int $user_id id of the user we're checking
 		 * @param int $event_id id of the event we're checking (matter for checks on event authorship)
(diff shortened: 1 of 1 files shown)


src/Tribe/Attendees.php
```diff
@@ -803,7 +803,7 @@ public function user_can_manage_attendees( $user_id = 0, $event_id = '' ) {
 		 * @param int $user_id id of the user we're checking
 		 * @param int $event_id id of the event we're checking (matter for checks on event authorship)
```
sc0ttkclark: Change the param order too

sc0ttkclark:
This filter is already out there, should we stick with the param order and fix usage in a hotfix for CT?

Camwyn:
We can, the filter was just added last sprint.

I hate changing signatures too, but we treat `$user_id` as an optional, optionals should go to the end, and `$event_id` is _not_ optional?

sc0ttkclark:
Need to discuss this further on Monday

Camwyn:
Closing due to change in code strategy

PR #1160, opened 2019-03-06
Add try/catch for IE/Edge

Camwyn:
🎫 https://central.tri.be/issues/121073


src/resources/js/tickets-block.js
```diff
@@ -45,16 +45,48 @@ tribe.tickets.block = {
 		'.tribe-block__tickets__item__quantity__remove, .tribe-block__tickets__item__quantity__add',
 		function( e ) {
 			e.preventDefault();
-			var input  = $( this ).parent().find( 'input[type="number"]' );
-			var add    = $( this ).hasClass( 'tribe-block__tickets__item__quantity__add' );
+			var input = $( this ).parent().find( 'input[type="number"]' );
+			var add = $( this ).hasClass( 'tribe-block__tickets__item__quantity__add' );
+			var step = ( undefined !== Number( input[ 0 ].step ) ) ? Number( input[ 0 ].step ) : 1;
```
paulmskim: do we support step for the input field, and are we using it?

src/resources/js/tickets-block.js
```diff
@@ -45,16 +45,48 @@ tribe.tickets.block = {
...
 
-			// stepUp or stepDown the input according to the button that was clicked
-			add ? input[0].stepUp() : input[0].stepDown();
 
-			// Trigger the on Change for the input as it's not handled via stepUp() || stepDown()
-			input.trigger( 'change' );
+			// stepUp or stepDown the input according to the button that was clicked
+			// handle IE/Edge
+			if ( add ) {
+				var max = ( undefined !== Number( input[ 0 ].min ) ) ? Number( input[ 0 ].min ) : 0;
```
paulmskim: should this be `input[ 0 ].max`?

src/resources/js/tickets-block.js
```diff
@@ -45,16 +45,48 @@ tribe.tickets.block = {
...
+					} catch ( ex ) {
+						if ( 0 === max || max > input[ 0 ].value ) {
+							input[ 0 ].value = Number( input[ 0 ].value ) + step;
+						} else {
+							input[ 0 ].value = max;
+						}
+					}
+				}
+			} else if ( typeof input[ 0 ].stepDown === 'function' ) {
+				var min = ( undefined !== Number( input[ 0 ].max ) ) ? Number( input[ 0 ].max ) : 0;
```
paulmskim: should this be `input[ 0 ].min`?

src/resources/js/tickets-block.js
```diff
@@ -45,16 +45,48 @@ tribe.tickets.block = {
...
-			add ? input[0].stepUp() : input[0].stepDown();
 
-			// Trigger the on Change for the input as it's not handled via stepUp() || stepDown()
-			input.trigger( 'change' );
+			// stepUp or stepDown the input according to the button that was clicked
+			// handle IE/Edge
+			if ( add ) {
+				var max = ( undefined !== Number( input[ 0 ].min ) ) ? Number( input[ 0 ].min ) : 0;
+
+				if ( typeof input[ 0 ].stepUp === 'function' ) {
```
paulmskim: does this work in IE and Edge? because IE and Edge do not support `stepUp()` and `stepDown()`, if they are not functions, would the ticket quantities ever change in these browsers?

src/resources/js/tickets-block.js
Camwyn: https://github.com/moderntribe/event-tickets/blob/master/src/views/blocks/tickets/quantity-number.php#L28

src/resources/js/tickets-block.js
Camwyn: Ah, but they do, the support is just buggy and throws an InvalidState error.

So yes, it works, but there should be an `else` that handles edge cases (no pun intended) where they aren't functions. So an addition is in order...

src/resources/js/tickets-block.js
Camwyn: Yes bad copy pasta

src/resources/js/tickets-block.js
Camwyn: Yes bad copy pasta - apparently I swapped them when I moved them :/ (they were originally in the block at the top)

src/resources/js/tickets-block.js
paulmskim: gotcha, instead of checking specifically for `undefined` here we can do the following instead: `var step = input[ 0 ].step ? Number( input [ 0 ].step ) : 1` to catch any other cases that might be falsey.

src/resources/js/tickets-block.js
```diff
@@ -45,16 +45,46 @@ tribe.tickets.block = {
...
 
-			// stepUp or stepDown the input according to the button that was clicked
-			add ? input[0].stepUp() : input[0].stepDown();
 
-			// Trigger the on Change for the input as it's not handled via stepUp() || stepDown()
-			input.trigger( 'change' );
+			// stepUp or stepDown the input according to the button that was clicked
+			// handle IE/Edge
+			if ( add ) {
+				var max = ( undefined !== Number( input[ 0 ].max ) ) ? Number( input[ 0 ].max ) : 0;
```
paulmskim: same comment here on the `undefined`

paulmskim: Also, why are we setting `max` to `0` if it doesn't exist? If the RSVP was unlimited, would this not increase?

src/resources/js/tickets-block.js
```diff
@@ -45,16 +45,46 @@ tribe.tickets.block = {
...
+
+				if ( typeof input[ 0 ].stepUp === 'function' ) {
+					try {
+						input[ 0 ].stepUp();
+					} catch ( ex ) {
+						input[ 0 ].value = ( 0 === max || max > input[ 0 ].value ) ? originalValue + step : max;
```
paulmskim: in the case `step > 1`:

```suggestion
						input[ 0 ].value = ( max > originalValue + step ) ? originalValue + step : max;
```

src/resources/js/tickets-block.js
```diff
@@ -45,16 +45,46 @@ tribe.tickets.block = {
...
+					}
+				} else {
+					input[ 0 ].value = ( 0 === max || max > input[ 0 ].value ) ? originalValue + step : max;
```
paulmskim: same thing here

src/resources/js/tickets-block.js
```diff
@@ -45,16 +45,46 @@ tribe.tickets.block = {
...
+				}
+			} else {
+				var min = ( undefined !== Number( input[ 0 ].min ) ) ? Number( input[ 0 ].min ) : 0;
```
paulmskim: same comment on `undefined`

src/resources/js/tickets-block.js
```diff
@@ -45,16 +45,46 @@ tribe.tickets.block = {
...
+
+				if ( typeof input[ 0 ].stepDown === 'function' ) {
+					try {
+						input[ 0 ].stepDown();
+					} catch ( ex ) {
+						input[ 0 ].value = ( min < input[ 0 ].value ) ? originalValue - step : min;
```
paulmskim: in the case `step > 1`:

```suggestion
						input[ 0 ].value = ( min < originalValue - step ) ? originalValue - step : min;
```

src/resources/js/tickets-block.js
```diff
@@ -45,16 +45,46 @@ tribe.tickets.block = {
...
+					}
+				} else {
+					input[ 0 ].value = ( min < input[ 0 ].value ) ? originalValue - step : min;
```
paulmskim: same thing here

src/resources/js/tickets-block.js
Camwyn: hm... `max >= originalValue + step`?

src/resources/js/tickets-block.js
Camwyn: It's a placeholder for unlimited.
A max of 0 makes no sense - that would mean the ticket is sold out, in which case the input shouldn't show.
We can't arbitrarily set a max - some user will come along that wants to make it higher, so we use 0 as a shortcut - essentially `max(0) == unlimited` which is a variable test vs. checking the input value?

src/resources/js/tickets-block.js
paulmskim: either or, both work.

PR #1162, opened 2019-03-07
Attendee Registration: fix getting theme page templates

eri-trabiccolo:
the current code assumes the array returned by wp_get_theme()->get_templates()
is in the form {template_name} => {filename} while actually it's the opposite.


sc0ttkclark:
Fix put into #1164 which is already scheduled for release since we're at the end of dev cut off and this is an easy fix.

PR #1163, opened 2019-03-08
Fix styles, remove committed css files

Camwyn:
Fix button style for Safari. (wrap text)
Remove some css files that got committed.

:ticket: https://central.tri.be/issues/70618

No new changelog, as this is just a tweak to https://github.com/moderntribe/event-tickets/pull/1161


Camwyn:
> Needs changelog

I don't think so - it's not a separate change from #1161, just an extension of it?

PR #1180, opened 2019-03-14
Fix (remove) float around submit button

Camwyn:
🎫 https://central.tri.be/issues/123044#note-25


Camwyn:
Testing something...will reopen.

PR #1181, opened 2019-03-14
fix(attendee reg): Remove float on submit button.

Camwyn:
Fix styling around AR submit button - remove class that adds float so that other blocks don't
overlap/collapse around it.

:ticket: https://central.tri.be/issues/123044#note-25


Camwyn:
Note this PR was used as a test case for [commitizen](https://github.com/commitizen/cz-cli) created commits

PR #1189, opened 2019-03-27
fix(RSVP): Check for block content and don't show classic form if present.

Camwyn:
Test the post content for blocks and for the tribe-block class before allowing the RSVP classic form to display.

:ticket: https://central.tri.be/issues/124394

### DIFF
--- a/src/Tribe/RSVP.php
+++ b/src/Tribe/RSVP.php
@@ -1074,7 +1074,7 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 			return $content;
 		}
 
-		// test for blocks in content, but usually called after the blocks have been convereted
+		// test for blocks in content, but usually called after the blocks have been converted
 		if (
 			has_blocks( $content )
 			|| false !== strpos( (string) $content, 'tribe-block ' )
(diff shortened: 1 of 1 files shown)


src/Tribe/RSVP.php
```diff
@@ -1071,18 +1071,13 @@ public function front_end_tickets_form( $content ) {
...
-
+		// test for blocks in content, but usually called after the blocks have been convereted
 		if (
-			tribe_get_option( 'toggle_blocks_editor' )
-			&& (
-				has_blocks( $post->ID )
-				|| ! $template_overwrite->has_classic_editor( $post->ID )
-			)
+			has_blocks( $content )
+			|| false !== strpos( (string) $content, 'tribe-block' )
```
sc0ttkclark: Will it ever not have blocks in $content but still have tribe-block?

Camwyn: I don't believe so - that is the class we put on blocks.

Camwyn: that class does appear in other pages (I'm looking at you AR) but I'm not seeing any issues in my tests.

Camwyn: By changing it to `tribe-block ` (with a space at the end) we're specifically targeting blocks and avoiding tripping on things like `tribe-block__rsvp__ticket tribe-block__rsvp__ticket--inactive`

PR #1191, opened 2019-03-29
 fix(capacity): Add a check for unlimited tickets 

Camwyn:
Add a check to tribe_events_count_available_tickets() for unlimited tickets so it returns -1 when
one is found.

🎫 https://central.tri.be/issues/119844

### DIFF
--- a/src/template-tags/tickets.php
+++ b/src/template-tags/tickets.php
@@ -155,7 +155,7 @@ if ( ! function_exists( 'tribe_events_count_available_tickets' ) ) {
 			$stock_level = $global_stock_mode === Tribe__Tickets__Global_Stock::CAPPED_STOCK_MODE ? $ticket->global_stock_cap : $ticket->stock;
 
 			// If we find an unlimted ticket, just return unlimited so we don't use -1 as a real stock
-			if ( -1 === $stock_level ) {
+			if ( -1 === $stock_level || Tribe__Tickets__Ticket_Object::UNLIMITED_STOCK === $stock_level ) {
 				return $stock_level;
 			}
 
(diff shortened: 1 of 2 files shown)


src/template-tags/tickets.php
```diff
@@ -153,6 +153,12 @@ function tribe_events_count_available_tickets( $event = null ) {
 			}
 
 			$stock_level = $global_stock_mode === Tribe__Tickets__Global_Stock::CAPPED_STOCK_MODE ? $ticket->global_stock_cap : $ticket->stock;
```
sc0ttkclark: Should we `(int)` cast here instead of below the new conditional, so it also makes it through the `-1` conditional below?

src/template-tags/tickets.php
```diff
@@ -153,6 +153,12 @@ function tribe_events_count_available_tickets( $event = null ) {
...
+
+			// If we find an unlimted ticket, just return unlimited so we don't use -1 as a real stock
+			if ( -1 === $stock_level ) {
+				return $stock_level;
+			}
+
 			$count += (int) $stock_level; // Explicit cast needed because it's possible $stock_level will be an empty string (unlimited stock)
```
```suggestion
			$count += $stock_level;
```

src/template-tags/tickets.php
sc0ttkclark: Also double check that empty string is meant as unlimited like that old comment said below, if so, we need to return early in the new conditional for that too, right?

sc0ttkclark:
```suggestion
			 // Explicit cast needed because it's possible $stock_level will be an empty string (unlimited stock)
			$stock_level = (int) $ticket->stock;

			if ( $global_stock_mode === Tribe__Tickets__Global_Stock::CAPPED_STOCK_MODE ) {
				$stock_level = (int) $ticket->global_stock_cap;
			}
```

Camwyn: Good call. Was just going back over it for gotchas like this 👍

sc0ttkclark: If we are checking against `UNLIMITED_STOCK` then we should probably not cast here and instead check before we cast?

Camwyn: yes - that's why i didn't just commit your suggestion - I'll commit something shortly (after testing locally)

Camwyn: This thread has got a bit long/confusing so here's a summary of my approach: (code is actually below)

- Don't cast until after we test against `UNLIMITED_STOCK` as that constant is an empty string, not an int.
- Once we've tested for the string, test as an int (against -1) and then cast as an int when we add it to the running total.
- also cast in `tribe_events_has_unlimited_stock_tickets` when we test against -1 (the lack of this test is why we were getting false negatives from this function, which dominoed to other functions that called on it.)

sc0ttkclark: 👍

PR #1199, opened 2019-04-03
feat(attendee registration shortcode): Change the shortcode and option to use page ID

Camwyn:
...instead of page slug.

Change the option to use page ID to identify the page set as folks can change the page slug later.

:ticket: https://central.tri.be/issues/124997

Requires: https://github.com/moderntribe/event-tickets-plus/pull/760


src/Tribe/Attendee_Registration/Main.php
```diff
@@ -29,11 +29,18 @@ class Tribe__Tickets__Attendee_Registration__Main {
...
-		$page = get_page_by_path( $slug );
+		$id = Tribe__Settings_Manager::get_option( 'ticket-attendee-page-id', false );
+
+		// check for old saved slug - since TBD
+		if ( ! empty( $id ) ) {
+			$page = get_page( $id );
+			$slug = $page->post_name;
+		} else {
+			$slug = Tribe__Settings_Manager::get_option( 'ticket-attendee-page-slug', false );
+			$page = get_page_by_path( $slug );
```
sc0ttkclark: Probably shouldn’t call this function if it doesn’t have a slug

src/Tribe/Attendee_Registration/Main.php
```diff
@@ -29,11 +29,18 @@ class Tribe__Tickets__Attendee_Registration__Main {
...
+		}
 
-		if ( empty( $slug ) || ! has_shortcode( $page->post_content, 'tribe_attendee_registration' ) ) {
+		if ( empty( $slug ) || empty( $page ) || ! has_shortcode( $page->post_content, 'tribe_attendee_registration' ) ) {
```
sc0ttkclark: Will there be a case where the isn’t both or none: slug/page?

Camwyn: eh, trying to avoid calling `$page->post_content` if we don't have a `$page` - let me rework that.

sc0ttkclark:
Let's chat a little more about backcompat

PR #1200, opened 2019-04-04
fix(attendee registration): Add some cache-busting to the AR page setup/load.

Camwyn:
Add `DONTOCACHEPAGE` and `LSCACHE_NO_CACHE` as well as a cookie to help circumvent caching on the attendee registration page(s).

:ticket: https://central.tri.be/issues/123980


src/Tribe/Attendee_Registration/Template.php
```diff
@@ -18,36 +18,58 @@ class Tribe__Tickets__Attendee_Registration__Template extends Tribe__Templates {
...
+		add_filter( 'get_the_archive_title', [ $this, 'modify_archive_title' ], 1000 );
+	}
+
+	public function cache_busters() {
+		// We only want to do this for the AR page(s)
+		if ( ! $this->is_on_ar_page() ) {
+			return;
+		}
+
+		// For caching plugins that respect the constant
```
Camwyn: Possible upgrade: use a filterable array of constant keys and loop through them?

src/Tribe/Attendee_Registration/Template.php
```diff
@@ -18,36 +18,58 @@ class Tribe__Tickets__Attendee_Registration__Template extends Tribe__Templates {
...
+			define('DONOTCACHEPAGE', true );
+		}
+
+		// LightSpeed Cache
+		if ( ! defined( 'LSCACHE_NO_CACHE' ) ) {
+			define('LSCACHE_NO_CACHE', true);
+		}
+
+		// set a cookie to (hopefully) tell Varnish to not cache (at least on WPE)
+		setcookie('wordpress_tribe_cachebuster', microtime(true), time()+600, '/');
```
sc0ttkclark: What about adding `nocache_headers()` perhaps?

src/Tribe/Attendee_Registration/Template.php
```diff
@@ -18,36 +18,58 @@ class Tribe__Tickets__Attendee_Registration__Template extends Tribe__Templates {
...
 		// Modify the page title
-		add_filter( 'document_title_parts', array( $this, 'modify_page_title' ), 1000 );
-		add_filter( 'get_the_archive_title', array( $this, 'modify_archive_title' ), 1000 );
+		add_filter( 'document_title_parts', [ $this, 'modify_page_title' ], 1000 );
+		add_filter( 'get_the_archive_title', [ $this, 'modify_archive_title' ], 1000 );
+	}
+
+	public function cache_busters() {
+		// We only want to do this for the AR page(s)
+		if ( ! $this->is_on_ar_page() ) {
```
Camwyn: This actually doesn't work because it's too early!

src/Tribe/Attendee_Registration/Template.php
sc0ttkclark:
```suggestion
		setcookie('wordpress_tribe_cachebuster', microtime(true), time()+600, '/');

		add_action( 'send_headers', 'nocache_headers' );
```

src/Tribe/Attendee_Registration/Template.php
sc0ttkclark: Would be even better if this cache buster functionality could be added to common and get called from ET, since we may need this elsewhere for other places

src/Tribe/Attendee_Registration/Template.php
rafsuntaskin: Maybe `wp` hook will work properly. Check this out: https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-cache-helper.php

src/Tribe/Attendee_Registration/Template.php
Camwyn: I like this but we're up against the timing issue: the earliest I can check the query (`$wp_query`) is `parse_query` - which comes _after_ `send_headers` so we'll get "headers already sent" warnings/errors?

sc0ttkclark: Yeah valid point

sc0ttkclark:
We should try and get this resolved in the next sprint

Camwyn:
I'm going to close this as there just isn't a good solution yet. This isn't moving forward right now and  it's cluttering the PR list 😉

PR #1203, opened 2019-04-05
QR Code Complete Check in Status

jesseeproductions:
_Ref:_ [C#122458](https://central.tri.be/issues/122458)

Updates so that the QR Check in only does it for complete status attendees.


src/Tribe/Status/Manager.php
```diff
@@ -206,19 +206,23 @@ public function get_statuses_by_action( $action, $commerce, $operator = 'AND' )
...
 				$action => true,
-			) );
+			] );
 		}
 
 		foreach ( $filtered_statuses as $status ) {
 			$trigger_statuses[] = $status->provider_name;
+
+			if ( ! empty( $status->additional_names ) ) {
+				$trigger_statuses = $this->add_additional_names_to_array( $trigger_statuses, $status->additional_names );
```
sc0ttkclark: What's the difference between this and array_merge()?

src/Tribe/Status/Manager.php
```diff
@@ -292,10 +296,71 @@ public function get_status_options( $commerce ) {
...
+		$filtered_statuses = wp_list_filter( $this->statuses[ $abbvetied_name ]->statuses, [
+			'count_completed' => true,
+		] );
+
+
+		foreach ( $filtered_statuses as $status ) {
+			$trigger_statuses[] = $status->provider_name;
+
+			if ( ! empty( $status->additional_names ) ) {
+				$trigger_statuses = $this->add_additional_names_to_array( $trigger_statuses, $status->additional_names );
```
sc0ttkclark: What's the difference between this and array_merge()?

jesseeproductions: From what I read it is best to avoid using array_merge in loops. Not sure if that would matter here, but it is habit.

PR #1204, opened 2019-04-05
fix(tribe-user-event-confirmations): Modify query to exclude deleted attendees.

Camwyn:
Modify the query - needed for backward compat. We should also consider doing something to remove this data from the dB?

:ticket: https://central.tri.be/issues/123609

Props to Dan Bramer for his help with the query!


sc0ttkclark:
As discussed, this will be handled by the ORM usage work I'm doing, so no worries!

PR #1210, opened 2019-04-09
feat(tickets table): Add error rows and move errors to the row so they are clickable.

Camwyn:
Add error rows to the tickets table below each ticket row. Move errors/notifications to the error row. Add clickable links to the errors for capacity mismatch and WC manage stock. Style the rows to hide by default and add js to show on clicking the icons.

Screen Shot 2019-04-08 at 5 57 23 PM

:ticket: https://central.tri.be/issues/94274#note-25


src/admin-views/editor/list-row.php
```diff
@@ -115,3 +117,19 @@ class="tribe-ticket-field-order"
 		?>
 	</td>
 </tr>
+<?php if ( $needs_warning ) : ?>
+		<tr class="warning_row">
```
paulmskim: Does this work if you have more than one ticket and you try to drag and drop tickets around with ET+?

src/resources/js/tickets.js
```diff
@@ -768,6 +768,25 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
 			$( document.getElementById( 'tribe_ticket_header_image_id' ) ).val( '' );
 		} );
 
+		/* Toggle ticket table warninfg row */
```
paulmskim: spelling

src/admin-views/editor/list-row.php
Camwyn: Oh, good catch - probably not. OK, time to re-strategize!

Camwyn:
Closing this, replaced by https://github.com/moderntribe/event-tickets/pull/1215

PR #1214, opened 2019-04-10
Fix/121703 tickets block order

paulmskim:
For ticket: https://central.tri.be/issues/121703

### DIFF
--- a/src/modules/data/blocks/ticket/sagas.js
+++ b/src/modules/data/blocks/ticket/sagas.js
@@ -204,7 +204,7 @@ export function* setTicketInitialState( action ) {
 	if ( ticketId !== 0 ) {
 		yield all( [
 			put( actions.setTicketId( clientId, ticketId ) ),
-			put( actions.fetchTicket( clientId, ticketId ) ),
+			call( fetchTicket, { payload: { clientId, ticketId } } ),
 		] );
 	}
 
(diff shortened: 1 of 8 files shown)


src/modules/data/blocks/ticket/sagas.js
```diff
@@ -204,20 +204,22 @@ export function* setTicketInitialState( action ) {
 	if ( ticketId !== 0 ) {
 		yield all( [
 			put( actions.setTicketId( clientId, ticketId ) ),
-			put( actions.fetchTicket( clientId, ticketId ) ),
+			call( fetchTicket, { payload: { clientId, ticketId } } ),
```
nealfennimore: I do not think this will proceed with that new `while` loop in the `fetchTicket` saga, is that your intention?

paulmskim: The `while` loop is not in the `fetchTicket` saga, it's in the `saveTicketWithPostSave` saga, which is `fork`ed after this. I changed this to a `call` because the `put` would just dispatch the action and not wait for the `fetch` to finish. Then the `handleTicketDurationError` below runs before the `fetch` isn't even finished, which is not what was intended. That's not the important part though, I actually need the `saveTicketWithPostSave` just below that to run after the `fetch`.

paulmskim: What's happening here now is all tickets are saving on post save. Before, you had set it up so it save only if it was synced. But because we need to save the order of the tickets, we need to save the tickets on each post save.

nealfennimore: Oh I see, sorry about that. I thought I expanded the diff all the way - carry on :)

paulmskim:
on hold until I can confirm one detail in https://central.tri.be/issues/121703#note-30

PR #1216, opened 2019-04-11
Validate variable, might be uninitialized

stian-overasen:
The `$html` variable might be uninitialized if the template file does not exists. ([tribe-ext-pdf-tickets](https://github.com/mt-support/tribe-ext-pdf-tickets/blob/master/tribe-ext-pdf-tickets.php#L750) looks for a custom pdf-tickets.php template file which does not exist by default)

Screenshot 2019-04-11 at 11 12 26


src/template-tags/tickets.php
```diff
@@ -644,7 +644,7 @@ function tribe_tickets_get_template_part( $slug, $name = null, array $data = nul
 
 	if ( ! $echo ) {
 		// Return should come at the end
-		return $html;
+		return isset( $html ) ? $html : '';
```
bordoni: Hi @stian-overasen,

Thanks for contributing with this, would you mind making a change on the code here, to improve reliability moving forward?

Just after the function starting here we should have `$html = null;` which would prevent this notice in general.

stian-overasen: I agree. That would be better. Updated now.

jrserani:
_Ref:_ [C#125913](http://central.tri.be/issues/125913)

@stian-overasen thank you for submitting this - we're taking a look and will let you know if we have any questions or comments :-)

readme.txt
```diff
@@ -132,6 +132,7 @@ Currently, the following add-ons are available for Event Tickets:
 * Fix - Prevent potential PayPal issues by not allowing $0 tickets in the block editor for Tribe Commerce [123835]
 * Fix - When moving an attendee prevent shared capacity from being enabled on the receiving event [120727]
 * Fix - Tidy attendee list print styles [12599]
+* Fix - Prevent PHP notices when looking for a template that does not exist in `tribe_tickets_get_template_part()` [125913]
```
```suggestion
* Fix - Prevent PHP notices when looking for a template that does not exist in `tribe_tickets_get_template_part()` (props @stian-overasen) [125913]
```

PR #1223, opened 2019-04-16
fix(csv/rsvp import): Change the string used to grab rsvp imports.

Camwyn:
Change the string to grab RSVP imports to the default as the mapping is not working correctly.

:ticket: https://central.tri.be/issues/124052


readme.txt
```diff
@@ -129,6 +129,7 @@ Currently, the following add-ons are available for Event Tickets:
 * Tweak - Attendee registration fields configuration for block editor opens up in modal [123818]
 * Tweak - Move IPN settings together in ticket settings tab [122333]
 * Tweak - Change attendee registration page shortcode to use ID instead of page slug, add function for backward compatibility [124997]
+* Fix - Change rsvp imports so they give the correct response info [124052]
```
```suggestion
* Fix - Change RSVP import identifier in CSV importer so they provide the correct response messages [124052]
```

PR #1232, opened 2019-04-22
fix(attendee registration shortcode): Use tribe context for cron discovery.

Camwyn:
Use the `doing_cron` function from tribe.context to be backward compatible with WP versions before
4.8

:ticket: https://central.tri.be/issues/126111

see: https://github.com/moderntribe/tribe-common/blob/f8f3ec8aca567b84a37629d036a8181cd9e2dc22/src/Tribe/Context.php#L242


sc0ttkclark:
Needs Changelog

PR #1233, opened 2019-04-22
Feature/126229 add template tag tests

Camwyn:
Draft PR for https://central.tri.be/issues/126229

Will continue to add tests, some are not passing, there seems to be an issue with detecting/counting tribe-commerce tickets currently.

Note: stock tests are only testing `own` stock right now


src/template-tags/tickets.php
```diff
@@ -14,7 +14,7 @@
 	/**
 	 * Returns the current post object that can have tickets attached to it
 	 *
-	 * Optionally the post object or ID of a ticket post can be passed in and, again, the
+	 * Optionally the post object or ID of a ticketed post can be passed in and, again, the
```
Camwyn: I found this docblock a bit unclear (expected a different outcome) so I changed it to make it a bit more clear.

PR #1239, opened 2019-04-25
#126226 - move report totals to own method, add tests for qty and lin…

jesseeproductions:
Originally this was to test abstract methods that count qty and line items with RSVP. 

But RSVP does not have orders and that is what those methods are designed to do.

I changed this to instead test with TPP. I added a test support class to generate orders and moved the coding that does the counting for the report to its own method so it could be tested. 

I believe this test now provides coverage for the counting abstract methods in Tribe__Tickets__Status__Abstract_Commerce

As well as counts on the TPP Report Page.

_Ref:_ [C#126226](https://central.tri.be/issues/126226)


sc0ttkclark:
Sorry about the PR close here, can you reopen this against release/G19.09?

PR #1245, opened 2019-04-30
#126749 - add a filter to remove product and download from the saved …

jesseeproductions:
_Ref:_ [C#126749](https://central.tri.be/issues/126749)

A customer upgraded from ET 4.4 to 4.10 and their saved supported post types included product and downloads. When trying to save it was creating a recursion with wp_update_post() trying to save the ticket order. 

- I attempted to make this work using the filter tribe_tickets_post_types in Woo and EDD, but it would run to late to change it. 
- I tried numerous places to get it to run and it would always be late due to when Ticket Handler gets called. 
- I also attempted to run it from a theme function.php, but again it was not working. It finally worked in a plugin that runs before ET. 
- Considering that I added some code to remove it directly in the method. The filter could be used if someone desires to crash their site from a plugin to add them back. ;)

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -139,7 +139,6 @@ Currently, the following add-ons are available for Event Tickets:
 * Fix - Correct escaping on attendee registration shortcode [125964]
 * Fix - Fix error with creating new ticket in block editor [126266]
 
-
 = [4.10.4.3] 2019-04-26 =
 
 * Fix - Prevent Composer autoloader from throwing Fatal due to unexistent `setClassMapAuthoritative()` method [126590]
--- a/readme.txt
+++ b/readme.txt
@@ -129,7 +129,7 @@ Currently, the following add-ons are available for Event Tickets:
 * Tweak - Update status manage to accept provider names or abbreviations [120862]
 * Tweak - In the Ticket Block add link to EDD Orders Page [121440]
 * Tweak - Change "Attendee Registration" to "Attendee Information" in several locations [126038]
-* Tweak - Exclude WooCommerce Product and EDD Downloads as supported post types for tickets to prevent recursion errors [126749]
+* Tweak - Exclude WooCommerce Product and EDD Downloads as supported post types when saving for tickets to prevent recursion errors, in case they were previously saved before we removed them from the options list [126749]
 * Fix - Add checks to `tribe_events_count_available_tickets()` and `tribe_events_has_unlimited_stock_tickets()` to properly detect unlimited tickets. [119844]
 * Fix - Change `inventory` to compare the correct ticket when checking event shared capacity [119844]
 * Fix - Make Attendees Report match the order report, specifically "Total Tickets Issued" should not include cancelled tickets [69823]
--- a/readme.txt
+++ b/readme.txt
@@ -129,6 +129,7 @@ Currently, the following add-ons are available for Event Tickets:
 * Tweak - Update status manage to accept provider names or abbreviations [120862]
 * Tweak - In the Ticket Block add link to EDD Orders Page [121440]
 * Tweak - Change "Attendee Registration" to "Attendee Information" in several locations [126038]
+* Tweak - Supported post types for tickets by removing product and download from saved options to prevent recursion errors [126749]
 * Fix - Add checks to `tribe_events_count_available_tickets()` and `tribe_events_has_unlimited_stock_tickets()` to properly detect unlimited tickets. [119844]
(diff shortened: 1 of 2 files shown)


src/Tribe/Main.php
```diff
@@ -760,12 +760,15 @@ public function post_types() {
 			tribe_update_option( 'ticket-enabled-post-types', $defaults );
 		}
 
+		// removes WooCommerce Product and EDD Download Post Type to Prevent Recursion Fatal Error on Saving
+		$filtered_post_types = array_diff( $options['ticket-enabled-post-types'], array( 'product', 'download' ) );
```
sc0ttkclark:
```suggestion
		$filtered_post_types = array_diff( $options['ticket-enabled-post-types'], [ 'product', 'download' ] );
```

src/Tribe/Main.php
```diff
@@ -760,12 +760,15 @@ public function post_types() {
 			tribe_update_option( 'ticket-enabled-post-types', $defaults );
 		}
 
+		// removes WooCommerce Product and EDD Download Post Type to Prevent Recursion Fatal Error on Saving
```
sc0ttkclark:
```suggestion
		// Remove WooCommerce Product and EDD post types to prevent recursion fatal error on save.
```

readme.txt
```diff
@@ -129,6 +129,7 @@ Currently, the following add-ons are available for Event Tickets:
 * Tweak - Update status manage to accept provider names or abbreviations [120862]
 * Tweak - In the Ticket Block add link to EDD Orders Page [121440]
 * Tweak - Change "Attendee Registration" to "Attendee Information" in several locations [126038]
+* Tweak - Supported post types for tickets by removing product and download from saved options to prevent recursion errors [126749]
```
```suggestion
* Tweak - Exclude WooCommerce Product and EDD Downloads as supported post types for tickets to prevent recursion errors [126749]
```

sc0ttkclark: Is this fix only on save or does it no longer show them in the form either now?

jesseeproductions: They were not showing in the form. They upgraded and the values were still saved from an older version.

src/Tribe/Main.php
```diff
@@ -760,12 +760,15 @@ public function post_types() {
 			tribe_update_option( 'ticket-enabled-post-types', $defaults );
 		}
 
+		// Remove WooCommerce Product and EDD post types to prevent recursion fatal error on save.
+		$filtered_post_types = array_diff( $options['ticket-enabled-post-types'], [ 'product', 'download' ] );
```
barryhughes: Hey just a thought ... 

It caught my eye that we do an explicit cast to array on line 721. Should we that here, too? 

```suggestion
		$filtered_post_types = array_diff( (array) $options['ticket-enabled-post-types'], [ 'product', 'download' ] );
```

1. We could then drop the explicit cast on line 771
2. It would avoid a PHP warning if `$options['ticket-enabled-post-types']` is null _(unlikely_ for sure, but possible if a user deselects all post types from the relevant setting)

src/Tribe/Main.php
```diff
@@ -760,12 +760,15 @@ public function post_types() {
...
+		// Remove WooCommerce Product and EDD post types to prevent recursion fatal error on save.
+		$filtered_post_types = array_diff( (array) $options['ticket-enabled-post-types'], [ 'product', 'download' ] );
+
 		/**
 		 * Filters the list of post types that support tickets
 		 *
 		 * @param array $post_types Array of post types
 		 */
-		return apply_filters( 'tribe_tickets_post_types', (array) $options['ticket-enabled-post-types'] );
+		return apply_filters( 'tribe_tickets_post_types', (array) $filtered_post_types );
```
sc0ttkclark:
```suggestion
		return apply_filters( 'tribe_tickets_post_types', $filtered_post_types );
```

readme.txt
```diff
@@ -129,6 +129,7 @@ Currently, the following add-ons are available for Event Tickets:
 * Tweak - Update status manage to accept provider names or abbreviations [120862]
 * Tweak - In the Ticket Block add link to EDD Orders Page [121440]
 * Tweak - Change "Attendee Registration" to "Attendee Information" in several locations [126038]
+* Tweak - Exclude WooCommerce Product and EDD Downloads as supported post types for tickets to prevent recursion errors [126749]
```
```suggestion
* Tweak - Exclude WooCommerce Product and EDD Downloads as supported post types when saving for tickets to prevent recursion errors, in case they were previously saved before we removed them from the options list [126749]
```

PR #1256, opened 2019-05-09
fix(Main): fix autoloader related issues

lucatume:
This fix makes sure that, even when the plugin is not loaded due to a missing requirement or
incompatibility, its source paths are registered, appending them, in the Autoloader. This makes it
so that the paths registered by the plugin owning Common will be searced **first** using the "safe"
version of the classes; if the classes are not found in the version of Common owned by the plugin
loading it then the additional plugin paths will be searched avoiding fatals caused by some plugin
functionalities relying on some classes that do not exist at all in the loaded (and probably older)
version of Common.

re https://central.tri.be/issues/127173

related to: https://github.com/moderntribe/the-events-calendar/pull/2493


lucatume:
I need to work on this to fix the tests.

lucatume:
Merging away to let Green Team deal with fixing the currently not passing 4 tests.

bordoni:
We need a changelog for this.

sc0ttkclark:
@bordoni Added changelog at bottom of fixes here: https://github.com/moderntribe/event-tickets/blob/release/G19.06/readme.txt

PR #1269, opened 2019-05-17
feat(admin tickets editor): Add class to wrapper div.

Camwyn:
Adds a class to the wrapper div for ticket controls (add ticket/rsvp, settings) in classic editor
and Community Tickets FE display. Specifically for better style control in FE CE display.

Needed for styles added in https://github.com/moderntribe/events-community/pull/392

🎫 https://central.tri.be/issues/127193 is the ticket for the CE PR.


sc0ttkclark:
Is this a problem if this is not in ET until G19.09?

Camwyn:
> Is this a problem if this is not in ET until G19.09?

No, the styles just won't be applied if the class isn't present.

PR #1271, opened 2019-05-20
Fix RSVP queries

sc0ttkclark:
https://central.tri.be/issues/128071

* Fixes rsvp_status queries
* Sets proper orm_provider for RSVP class
* Only show attendance totals in Attendees admin view for RSVPs if post has RSVP tickets OR has RSVP counts to show (don't show if they never had RSVP tickets OR never had any attendees RSVP)

### DIFF
--- a/src/Tribe/REST/V1/Endpoints/Attendee_Archive.php
+++ b/src/Tribe/REST/V1/Endpoints/Attendee_Archive.php
@@ -108,7 +108,7 @@ class Tribe__Tickets__REST__V1__Endpoints__Attendee_Archive
 			'status'                         => 'post_status',
 			'order_status'                   => 'order_status',
 			'checkedin'                      => 'checkedin',
-			'rsvp_going_status'              => 'rsvp_status',
+			'rsvp_going_status'              => 'rsvp_status__or_none',
 			'price_min'                      => 'price_min',
 			'price_max'                      => 'price_max',
 			'attendee_information_available' => 'has_attendee_meta',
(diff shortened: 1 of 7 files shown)


src/Tribe/REST/V1/Endpoints/Attendee_Archive.php
```diff
@@ -108,7 +108,7 @@ public function get( WP_REST_Request $request ) {
 			'status'                         => 'post_status',
 			'order_status'                   => 'order_status',
 			'checkedin'                      => 'checkedin',
-			'rsvp_going_status'              => 'rsvp_status',
+			'rsvp_going_status'              => 'rsvp_status__or_none',
```
mitogh: Just curios here on the reason on the amount of `_` for `rsvp_status__or_none` as it does not seems to follow a pattern, asking if is intentional?

sc0ttkclark: This change is to ensure backwards compatibility with previous REST API requests that included the RSVP status as an optional / match value. The normal pattern for filtering exact matches wasn't followed originally so this resolves that.

mitogh: Oh! My question was more about the usage of double underscore to separate one word and one for another. 

`rsvp_status` `__` `or_none`

sc0ttkclark: Ah, yes the double underscore is intentional, although I've pinged Gustavo/Luca about this because I misinterpreted our position on following the WordPress usage.

sc0ttkclark: I'm going to say we keep this since ET has already established this pattern for itself https://github.com/moderntribe/event-tickets/blob/master/src/Tribe/Attendee_Repository.php#L45

PR #1263, opened 2019-05-16
 Fixed domain in string to be translated

cfaria:
Select an option can't be translated due to wrong domain. Fixed that.

PR #1275, opened 2019-06-03
fix(translations): Fixed domain in string to be translated.

Camwyn:
Correct two places where we had `events-tickets` instead of `event-tickets`, preventing correct
string translations.

:ticket: https://central.tri.be/issues/128193

Ref: https://github.com/moderntribe/event-tickets/pull/1263


readme.txt
```diff
@@ -122,6 +122,7 @@ Currently, the following add-ons are available for Event Tickets:
 * Tweak - Add class to wrapper div around ticket controls in admin [127193]
 * Tweak - Reduced file size by removing .po files and directing anyone creating or editing local translations to translate.wordpress.org
 * Tweak - Modify methods to check for a post id of 0 to prevent PHP notices [128346]
+* Fix - Correct two places where the translation domain was incorrect. Thanks to @cfaria for the catch! []
```
```suggestion
* Fix - Correct two places where the translation domain was incorrect. Thanks to @cfaria for the catch! [128193]
```

PR #1276, opened 2019-06-06
Don't assume a `wp_` prefix for db tables | spotfix

barryhughes:
Fixes a hardcoded reference to `wp_posts` (otherwise, for installs that use a different prefix, the query will break):

> Unknown column 'wp_posts.ID' in 'on clause'

Targeting `master` since I wasn't sure which branch would be best given where we are in the release cycle, feel free to update as needed.

https://central.tri.be/issues/129053

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -120,7 +120,7 @@ Currently, the following add-ons are available for Event Tickets:
 = [4.10.6.1] 2019-06-11 =
 
 * Tweak - Adjust newsletter signup submission destination [129034]
-* Fix - Resolve hardcoded reference to `wp_posts` table in optout ORM queries []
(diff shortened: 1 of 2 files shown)


Camwyn:
Nice catch!

readme.txt
```diff
@@ -120,6 +120,7 @@ Currently, the following add-ons are available for Event Tickets:
 = [4.10.6.1] 2019-06-11 =
 
 * Tweak - Adjust newsletter signup submission destination [129034]
+* Fix - Resolve hardcoded reference to `wp_posts` table in optout ORM queries []
```
```suggestion
* Fix - Resolve hardcoded reference to `wp_posts` table in optout ORM queries [129053]
```

PR #1277, opened 2019-06-07
Update embedded newsletter signup form | #129034

barryhughes:
Parallels what is essentially the [same change but for TEC](https://github.com/moderntribe/the-events-calendar/pull/2551): instead of submitting directly to createsend.com, the newsletter signup form will now submit to our own service.

:warning: Important! As with [TEC PR 2551]((https://github.com/moderntribe/the-events-calendar/pull/2551)) this change depends on [Support API PR 28](https://github.com/moderntribe/support-api.tri.be/pull/28), which introduces the endpoint we need for this to work. That change must be approved and be deployed before this work ships.

:ticket: [♯129034](https://central.tri.be/issues/129034)

### DIFF
--- a/event-tickets.php
+++ b/event-tickets.php
@@ -3,7 +3,7 @@
 Plugin Name: Event Tickets
 Plugin URI:  http://m.tri.be/1acb
 Description: Event Tickets allows you to sell basic tickets and collect RSVPs from any post, page, or event.
-Version: 4.10.6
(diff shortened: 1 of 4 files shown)


barryhughes:
:memo: Tests are failing, but that is in relation to the attendee and ticker archive tests. The fails don't relate to this changeset.

PR #1279, opened 2019-06-09
Use $wpdb->prefix in filter_by_optout()

justlevine:
`filter_by_optout()` is using the hardcoded `wp_posts` for the query, which causes issues for any installations with a custom database prefix.

**The solution:**
On line 263, switch `wp_posts` with `{$wpdb->prefix}posts`, so the query works correctly with custom database prefixes.


sc0ttkclark:
Merged via https://github.com/moderntribe/event-tickets/pull/1276 already, thanks for the fix though, it will be in the next release!

PR #1288, opened 2019-06-17
Bail out of bootstrapping non-destructively if older TEC is present

borkweb:
**Background**

The `maybe_set_common_lib_info()` method was in `plugins_loaded()` inappropriately. Once relocated _out_ of `plugins_loaded()` and back to `__construct()` where it belongs, a problem reared its head with regards to plugin validation and bailing. 

Before the Plugin Dependency work, we never checked for out-of-date versions from _both_ TEC and ET. When Plugin Dependency was completed, we added a new check to support this desired validation at the expense of moving the common bootstrap to the wrong place.

**The fix**

In a previous PR for this hotfix, `maybe_set_common_lib_info()` was relocated back to `__construct()`. This PR follows along with some more changes of note:

1. There is now a class property that holds a boolean that allows plugins_loaded() to bail if we determine the plugin should not initialize & bootstrap.
2. I've added two hooks to the `plugins_loaded` action at `-1` priority to ensure it runs _before_ we attempt to initialize and bootstrap.
    1. The first hook validates that the other core plugin–TEC if executing from within ET, or ET if executing from within TEC–is of the appropriate version. If not, it _re-points_ common _back_ to the older plugin.
    2. The second hook validates that the appropriate WP or PHP version is running. If not, it _re-points_ common _back_ to the older plugin.
3. If the above hooks determine that the plugin should not init/bootstrap, it doesn't.

See: https://central.tri.be/issues/129478 and https://central.tri.be/issues/129479

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -242,7 +242,7 @@ class Tribe__Tickets__Main {
 		add_action( 'network_admin_notices', [ $this, 'tec_compatibility_notice' ] );
 		add_action( 'tribe_plugins_loaded', [ $this, 'remove_exts' ], 0 );
 		/*
-		* After common was loaded by another source (e.g. Event Tickets) let's append this plugin source files
+		* After common was loaded by another source (e.g. The Events Calendar) let's append this plugin source files
 		* to the ones the Autoloader will search. Since we're appending them the ones registered by the plugin
 		* "owning" common will be searched first.
 		*/
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -101,6 +101,11 @@ class Tribe__Tickets__Main {
 	private $should_prevent_autoload_init = false;
 
 	/**
+	 * @var string tribe-common VERSION regex
+	 */
+	private $common_version_regex = "/const\s+VERSION\s*=\s*'([^']+)'/m";
+
+	/**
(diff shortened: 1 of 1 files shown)


src/Tribe/Main.php
```diff
@@ -189,57 +201,97 @@ public function maybe_set_common_lib_info() {
...
-		if ( ! self::supported_version( 'wordpress' ) || ! self::supported_version( 'php' ) ) {
-			add_action( 'admin_notices', array( $this, 'not_supported_error' ) );
-			add_action( 'network_admin_notices', array( $this, 'not_supported_error' ) );
+		$this->should_prevent_autoload_init = true;
+
+		add_action( 'admin_notices', [ $this, 'tec_compatibility_notice' ] );
+		add_action( 'network_admin_notices', [ $this, 'tec_compatibility_notice' ] );
+		add_action( 'tribe_plugins_loaded', [ $this, 'remove_exts' ], 0 );
+		/*
+		* After common was loaded by another source (e.g. Event Tickets) let's append this plugin source files
```
barryhughes: Minor, but, should this be "_After common was loaded by another source (e.g. **The Events Calendar**)_ ?

barryhughes: ...not worth blocking the change, though.

PR #1290, opened 2019-06-19
Update wrong text domains

rafsuntaskin:
https://central.tri.be/issues/129677


sc0ttkclark:
It appears this has already been fixed in G19.09

PR #1297, opened 2019-07-11
Removed hooking into nonexistent filter

cliffordp:
https://central.tri.be/issues/107997#note-19


sc0ttkclark:
This looks right but could use an additional eye on this

cliffordp:
@sc0ttkclark, I didn't think this one needed it because it was a stray line of code that never did anything. Please confirm.

PR #1337, opened 2019-09-03
Prevent formulas from being exported when exporting attendees to CSV

sc0ttkclark:
https://central.tri.be/issues/133550

### DIFF
--- a/src/Tribe/Attendees.php
+++ b/src/Tribe/Attendees.php
@@ -554,6 +554,49 @@ class Tribe__Tickets__Attendees {
 	}
 
 	/**
+	 * Sanitize rows for CSV usage.
+	 *
+	 * @since 4.10.7.2
+	 *
+	 * @param array $rows Rows to be sanitized.
+	 *
+	 * @return array Sanitized rows.
+	 */
+	public function sanitize_csv_rows( array $rows ) {
+		foreach ( $rows as &$row ) {
+			$row = array_map( [ $this, 'sanitize_csv_value' ], $row );
+		}
+
+		return $rows;
+	}
+
+	/**
+	 * Sanitize a value for CSV usage.
+	 *
+	 * @since 4.10.7.2
+	 *
+	 * @param mixed $value Value to be sanitized.
+	 *
+	 * @return string Sanitized value.
+	 */
+	public function sanitize_csv_value( $value ) {
+		if (
+			0 === mb_strpos( $value, '=' )
+			|| 0 === mb_strpos( $value, '+' )
+			|| 0 === mb_strpos( $value, '-' )
+			|| 0 === mb_strpos( $value, '@' )
+		) {
+			// Remove the character from the start of the value.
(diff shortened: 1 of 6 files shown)


src/Tribe/Attendees.php
```diff
@@ -565,18 +608,36 @@ public function maybe_generate_csv() {
...
+			|| ! $this->user_can( 'edit_posts', $event_id )
+		) {
+			return;
+		}
+
+		// Generate filtereed list of attendees.
+		$items = $this->generate_filtered_list( $event_id );
+
+		// Sanitize items for CSV usage.
+		$items = array_map( [ $this, 'sanitize_csv_rows' ], $items );
```
zbtirrell: I wonder if we should sanitize after the filter?

src/Tribe/Attendees.php
```diff
@@ -553,6 +553,49 @@ private function generate_filtered_list( $event_id ) {
...
+	 * @param mixed $value Value to be sanitized.
+	 *
+	 * @return string Sanitized value.
+	 */
+	public function sanitize_csv_value( $value ) {
+		if (
+			0 === mb_strpos( $value, '=' )
+			|| 0 === mb_strpos( $value, '+' )
+			|| 0 === mb_strpos( $value, '-' )
+			|| 0 === mb_strpos( $value, '@' )
```
Camwyn: Maybe add pipe?

src/Tribe/Attendees.php
Camwyn: +1

sc0ttkclark: I left this before the filter to allow people to allow customized formulas for their attendees exports. I can put it after the filter if you think we shouldn't allow that at all.

src/Tribe/Attendees.php
sc0ttkclark: This only covers starting characters, but specifically these characters that we're looking for: https://www.owasp.org/index.php/CSV_Injection

src/Tribe/Attendees.php
```diff
@@ -553,6 +553,49 @@ private function generate_filtered_list( $event_id ) {
...
+	/**
+	 * Sanitize a value for CSV usage.
+	 *
+	 * @since 4.10.7.2
+	 *
+	 * @param mixed $value Value to be sanitized.
+	 *
+	 * @return string Sanitized value.
+	 */
+	public function sanitize_csv_value( $value ) {
```
nickpelton: Hey @sc0ttkclark, this looks good at first glance. Looks like what you are doing is checking for the injection char, and if that character is the fist part of a rows string, you remove it, then recursively hit the row again to check the next char until it passes the test. After a little review, this looks like an effective solution. Possible to throw a few unit tests at this to confirm?

sc0ttkclark: Yep, doing that next.

I just changed the functionality to prefix instead of remove, it looks like it's the recommended solution from Zach and I's research.

PR #1340, opened 2019-09-05
#132279 - disable ar redirects if modal is active

jesseeproductions:
_Ref:_ [C#132279](https://central.tri.be/issues/132279)

I am not sure this is required, but I thought I would put it out here to review and for information at this point and not to merge.  

This method was redirecting me away from the EDD checkout when I had AR fields.

The AR it not complete so that is the reason now. My coding disables it if the modal is active. 

We might want to keep the coding as is and not merge this if it can detect if AR fields are complete for EDD tickets, but otherwise  not redirect.


sc0ttkclark:
Solved in another PR that was merged for AR modal

PR #1351, opened 2019-09-12
#134358 - Fixed capitalization

andrasguseo:
Due to having both words capitalized 'Not Going', the string would not be translated, even if it was translated in the .po file because the .po/.mo file has the string like 'Not going' (small g). This fixes it.
Screen rec of the issue:
https://1drv.ms/v/s!Aqh6lbOjUW1Tgb4SjrYBOgniyxVRug?e=FYVq16

_Ref:_ [C#134358](https://central.tri.be/issues/134358)


jesseeproductions:
Readme and Ticket Added along with Tests passing, this is ready for a review.

PR #1384, opened 2019-10-03
Fix/134108 orders info capacity reached

cliffordp:
https://central.tri.be/issues/134108


sc0ttkclark:
Holding on merge for official sprint planning

PR #1418, opened 2019-10-25
On older (4.7) WP tribe() fails here - let's not use it.

Camwyn:
This works with latest core as well.

🎫 https://central.tri.be/issues/136322

📷 http://p.tri.be/vZBA7Q (no error)

### DIFF
--- a/src/Tribe/Tickets_View.php
+++ b/src/Tribe/Tickets_View.php
@@ -963,10 +963,10 @@ class Tribe__Tickets__Tickets_View {
 		$provider    = call_user_func( [ $provider_id, 'get_instance' ] );
 
 		/** @var \Tribe__Tickets__Editor__Template $template */
-		$template       = tribe( 'tickets.editor.template' );
+		$template       = new \Tribe__Tickets__Editor__Template;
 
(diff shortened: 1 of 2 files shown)


src/Tribe/Tickets_View.php
```diff
@@ -963,10 +963,10 @@ public function get_tickets_block( $post ) {
 		$provider    = call_user_func( [ $provider_id, 'get_instance' ] );
 
 		/** @var \Tribe__Tickets__Editor__Template $template */
-		$template       = tribe( 'tickets.editor.template' );
+		$template       = new \Tribe__Tickets__Editor__Template;
```
borkweb: Is this the correct solution–a shift from DI52 for this raises a small flag for me? Does this fail due to the fact that the singleton is [declared in a spot](https://github.com/moderntribe/event-tickets/blob/1f2f1251b5a7372045af192b2f991353293aa31c/src/Tribe/Editor/Provider.php#L22) that doesn't work well for FE?

Camwyn: I'm willing to dig some more - but why only on older (pre-block? 💡 ) versions of WP?

PR #1422, opened 2019-10-28
A few spacing tweaks via mattcr

Camwyn:
🎫 https://central.tri.be/issues/129452

📷 http://p.tri.be/srxHdm
📷 http://p.tri.be/MPJuWg
📸 http://p.tri.be/jlTAc9

### DIFF
--- a/src/resources/postcss/tickets.pcss
+++ b/src/resources/postcss/tickets.pcss
@@ -413,10 +413,9 @@
 
 		&.tribe-common-c-btn {
 			white-space: nowrap;
-			margin-left: 8px;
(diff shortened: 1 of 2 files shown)


borkweb:
A couple of screenshots as opposed to screencasts would do nicely for the artifacts here!

Camwyn:
> In the third screenshot, however, the text after the radio buttons is further to the right than the text after the checkboxes. 👈 That should be ticketed separately and doesn't need to be handled in this PR (or even in this release, tbh).

https://central.tri.be/issues/136513

PR #1424, opened 2019-10-29
Twak label alignment

Camwyn:
🎫 https://central.tri.be/issues/136513

📷 (before) http://p.tri.be/jlTAc9
📸  (after) http://p.tri.be/u8cKDa


borkweb:
Mergeable once tests pass.

PR #1430, opened 2019-11-02
2nd param shouldn't be duplicate of 1st

cliffordp:
https://central.tri.be/issues/73766 (but not really)

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -118,7 +118,9 @@ Currently, the following add-ons are available for Event Tickets:
 == Changelog ==
 
 = [TBD] TBD =
+
 * Tweak - Add a check for empty tickets to `ticket_availability()` method to avoid notices. [122334]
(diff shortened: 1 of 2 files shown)


sc0ttkclark:
@cliffordp I'm not sure we should do this only because the use case was the filter would let you change the value but still be able to reference that second filter argument as the original value.

Do you need to reference the values you're passing now for other code? If so, we should probably add them as an additional argument.

cliffordp:
@sc0ttkclark I just recognized this as a bug along the way of investigating all the other things. I don't think I've ever seen (and still don't understand) why you'd want duplicate params, since only the first can be returned anyway... and I thought it best to have other information to reference while filtering. Example: If I want to filter `total_rsvps`, maybe I want to force it to display as `total_going` value instead. It's up to you, but, again, I don't know why you'd want duplicate params on any filter.

sc0ttkclark:
I wouldn't have done it that way (I would have passed the second argument as an array of counts available) but for backwards compatibility we'll have to keep the argument as is since the use case of "original value" seems valid enough that someone could have started using it for that purpose.

PR #1456, opened 2019-11-13
VR > ET Integration

juanfra:
🎫 https://central.tri.be/issues/137260
🎥 https://www.loom.com/share/162f79190b514bfc827b8456c027fe7e

### DIFF
--- a/src/views/v2/day/event/cost.php
+++ b/src/views/v2/day/event/cost.php
@@ -9,6 +9,7 @@
  *
  * @link {INSERT_ARTCILE_LINK_HERE}
  *
+ * @since   4.10.9
  * @version TBD
  *
  * @var WP_Post $event The event post object with properties added by the `tribe_get_event` function.
--- a/src/views/v2/day/event/cost.php
+++ b/src/views/v2/day/event/cost.php
@@ -9,7 +9,7 @@
  *
  * @link {INSERT_ARTCILE_LINK_HERE}
  *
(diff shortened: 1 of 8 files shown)


src/views/v2/day/event/cost.php
```diff
@@ -9,7 +9,7 @@
  *
  * @link {INSERT_ARTCILE_LINK_HERE}
  *
- * @version 4.10.9
+ * @version TBD
```
sc0ttkclark: These should all be

```suggestion
 * @since 4.10.9
 * @version TBD
```

juanfra: Oh cool! didn't know we implemented the `@since` for templates. The `@version` has a key role when letting our users know, programatically, that they need to update their implementations (if any).

Is there any particular goodie with `@since` besides our internal docs?

Camwyn: Specifies when it was introduced - that helps them as well as us.
We could but don't yet leverage it beyond that I believe?

sc0ttkclark: On the green team we use `@since` and `@version` to keep track of when the view was introduced and what changes each version were. It's been helpful for debugging and from the customer side as we can tell them when a specific view was introduced by looking at the code.

juanfra: Thanks!

PR #1478, opened 2019-11-21
Whitespace fail.

Camwyn:
Unintended whitespace in a regex = failing regex.

🎫 https://central.tri.be/issues/133179

📷 http://p.tri.be/0N0fOQ


borkweb:
I'm assuming you have an artifact incoming?

PR #1491, opened 2019-11-26
fix(API security): Don't allow viewing "private" attendees if user can't

Camwyn:
… edit private posts.

Add a filter to the Repository (done in tribe-common) and utilize it when making API calls - if the
user cannot edit private posts then they can only view attendees from event that have the "Show
attendees list on event page" checkbox checked.

Requires: https://github.com/moderntribe/tribe-common/pull/1222 (included in PR)

:ticket: https://central.tri.be/issues/133333

📷 http://p.tri.be/8hslf3 (as admin, 72 is "private")
📸 http://p.tri.be/gWpFmE (as non-admin)


src/Tribe/Attendee_Repository.php
```diff
@@ -395,6 +396,23 @@ public function filter_by_event_status( $event_status ) {
...
 	}
 
+	/**
+	 * Filters attendee to only get those related to posts with "Show attendees list on event page" set to true.
+	 *
+	 *
+	 * @since TBD
+	 */
+	public function filter_by_show_attendees() {
+		$can_read_private_posts = current_user_can( 'read_private_posts' );
```
sc0ttkclark: This var is setup but not used below

Camwyn:
@sc0ttkclark Yeah, so those REST tests failed big-time... 😢

sc0ttkclark:
I have one concern about this I'd like to talk further about

sc0ttkclark:
As per our discussion, we should update this code to check for hide = 1 OR hide IS NULL (for cases where the hide is not set at all)

PR #1500, opened 2019-12-05
Fix/124403 asset loading

cliffordp:
https://central.tri.be/issues/124403

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -317,13 +317,11 @@ class Tribe__Tickets__Main {
 	 * Load Text Domain on tribe_common_loaded as it requires common
 	 *
 	 * @since 4.10
-	 *
 	 */
 	public function bootstrap() {
-
 		Tribe__Main::instance( $this )->load_text_domain( 'event-tickets', $this->plugin_dir . 'lang/' );
 
-		// Intialize the Service Provider for Tickets
+		// Initialize the Service Provider for Tickets
 		tribe_register_provider( 'Tribe__Tickets__Service_Provider' );
 
 		$this->hooks();
@@ -339,13 +337,13 @@ class Tribe__Tickets__Main {
 		Tribe__Tickets__JSON_LD__Order::hook();
 		Tribe__Tickets__JSON_LD__Type::hook();
 
+		/** @var Tribe__Tickets__Privacy */
 		tribe( 'tickets.privacy' );
 
 		/**
 		 * Fires once Event Tickets has completed basic setup.
 		 */
 		do_action( 'tribe_tickets_plugin_loaded' );
-
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


src/Tribe/Main.php
```diff
@@ -575,11 +573,17 @@ public function hooks() {
 			add_filter( 'tribe_event_import_rsvp_column_names', [ Tribe__Tickets__CSV_Importer__Column_Names::instance(), 'filter_rsvp_column_names' ] );
 		}
 
-		// Load our assets
-		add_action( 'tribe_tickets_plugin_loaded', tribe_callback( 'tickets.assets', 'enqueue_scripts' ) );
-		add_action( 'tribe_tickets_plugin_loaded', tribe_callback( 'tickets.assets', 'admin_enqueue_scripts' ) );
-		add_action( 'admin_enqueue_scripts', tribe_callback( 'tickets.assets', 'enqueue_editor_scripts' ) );
-		add_filter( 'tribe_asset_data_add_object_tribe_l10n_datatables', tribe_callback( 'tickets.assets', 'add_data_strings' ) );
+		/**
```
sc0ttkclark: Did we need this change?

src/template-tags/tickets.php
```diff
@@ -1324,6 +1325,60 @@ function tribe_tickets_is_event_page( $post = null ) {
...
+
+		// Front-end archive
+		if ( is_post_type_archive( $post_types ) ) {
+			return true;
+		}
+
+		/**
+		 * Whether or not we are in tickets-enabled context, such as determining if we should load plugin assets.
+		 *
+		 * @param bool           $result
```
```suggestion
		 * @since TBD
		 *
		 * @param bool           $result
```

src/Tribe/Main.php
cliffordp: _need?_ no. helps IDE though.

sc0ttkclark: IDE maybe but the difference between tribe() and tribe_callback() is one sets up the object, one is a callback that sets up the object and calls the method. I could be wrong and maybe the assets object gets setup ahead of time so this is negligible but the replacement of tribe_callback() is something we'll want to weigh carefully.

cliffordp: @sc0ttkclark 
Isn't that what `$assets = tribe( 'tickets.assets' );` is doing (just below this line in GitHub's inline commenting)?

**Are you saying these 2 are slightly different?**

Before:
Screen Shot 2019-12-08 at 12 31 57 PM


After:
image

sc0ttkclark: Yes, tribe_callback() returns a temporary callback reference, it does not setup tribe() for that associated container until the callback function is ultimately called via PHP when the action itself runs.

cliffordp: tyvm for clarifying

PR #1507, opened 2019-12-09
Update ticket maker for PayPal to set ticket provider field on post

sc0ttkclark:
https://central.tri.be/issues/138103


Camwyn:
Update and merge?

PR #1516, opened 2019-12-14
Handle special characters for Event Ticket field labels like single quotes and colons so they don't break the saving

sc0ttkclark:
https://central.tri.be/issues/136451

Screencast: https://p199.p4.n0.cdn.getcloudapp.com/items/P8uYRnn5/Screen+Recording+2019-12-14+at+12.26+PM.mov

Tested against:
```
Johnny's `~!@#$%^&*()_+-={}|[]:";'<>?,.\/ yes
```

Only character not saving correctly is backslash but it gets stripped out and does not effect the rest of the text (previously single quotes would break everything on save).


sc0ttkclark:
Because only two characters (from what I can tell) affect php's ability to parse the string properly, everything else comes through fine. Backslashes are OK but they just don't stick around when doing further PHP processing. They won't break like the `:` or `'` were breaking entirely.

PR #1518, opened 2019-12-16
Enforce string when getting capacity amount and fix threshold restriction

sc0ttkclark:
https://central.tri.be/issues/137875


src/views/blocks/rsvp/details/availability.php
```diff
@@ -54,7 +54,7 @@
 		<?php if ( $show_unlimited) : ?>
 			<span class="tribe-block__rsvp__unlimited"><?php echo esc_html( $handler->unlimited_term ); ?></span>
 		<?php endif; ?>
-	<?php elseif ( $remaining_tickets <= $threshold ) : ?>
+	<?php elseif ( 0 === $threshold || $remaining_tickets <= $threshold ) : ?>
```
Camwyn: I think I see what you are doing here - "set it to 0 and it never shows"?

sc0ttkclark: We absint() the threshold now, so 0 or blank in the TEC settings wouldn't work

PR #1529, opened 2020-01-07
ET-587 - add check for classic editor option to load assets with tick…

jesseeproductions:
…et block

_Ref:_ [C#ET-587](https://moderntribe.atlassian.net/browse/ET-587)

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -117,7 +117,7 @@ Currently, the following add-ons are available for Event Tickets:
 
 == Changelog ==
 
-= [4.11.3] TBD =
+= [4.11.2] TBD =
 
 * Fix - Load JavaScript assets with Ticket Block when using Classic Editor [ET-587]
 
--- a/readme.txt
+++ b/readme.txt
@@ -117,6 +117,10 @@ Currently, the following add-ons are available for Event Tickets:
 
 == Changelog ==
 
+= [4.11.3] TBD =
+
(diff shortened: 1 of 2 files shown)


src/Tribe/Tickets_View.php
```diff
@@ -979,6 +979,12 @@ public function get_tickets_block( $post ) {
 		/** @var \Tribe__Tickets__Editor__Blocks__Tickets $blocks_tickets */
 		$blocks_tickets = tribe( 'tickets.editor.blocks.tickets' );
 
+		// load assets if classic editor is active
+		$option = get_option( 'classic-editor-replace' );
```
jesseeproductions: - I am not sure if this check is required.
- We could always load the assets here as there are other plugins to disable gutenberg besides the classic editor. 
- Not sure if there is a scenario where we load assets twice and it creates a performance issue.

readme.txt
```diff
@@ -117,6 +117,10 @@ Currently, the following add-ons are available for Event Tickets:
 
 == Changelog ==
 
+= [4.11.3] TBD =
```
```suggestion
= [4.11.2] TBD =
```

src/Tribe/Tickets_View.php
sc0ttkclark: It sounds like we should always load the assets (once) here. Doesn't the assets loader only ever load it once?

jesseeproductions: I believe so. I updated the PR to always load them. Ready for another review.

PR #1530, opened 2020-01-07
build(husky): Add a script to run a simple test on branch name pre-push.

Camwyn:
Adds an example pre-push hook to `.huskyrc.sample` and corresponding script in `dev/husky/`.

The script tests branch name for the presence of a proper prefix and Jira ticket number in the format `[[:alpha:]]{2,}-[[:digit:]]{2,}`.

🎥 http://p.tri.be/EOglOX

**NOTE** for this to work locally, you will have to copy the `.huskyrc.sample` content to `.huskyrc` (which is currently git-ignored)

Probably worth noting that the above "Jira ticket" in the branch name is a work-around as the branch could not be pushed without the `ET-00` (not a real ticket)


Camwyn:
Oh, I can add it back in to the `.huskyrc.sample` as well! 
AFAIK no-one other than me has been using Husky (yet) ☹️

Camwyn:
@sc0ttkclark do we want to sneak this into G20.01? It requires no QA and doesn't affect the live code at all...

Camwyn:
@bordoni what do you think of this - I can create a version for TEC where I can focus on it more if it causes issues (I don't expect it would!)

.huskyrc.sample
```diff
@@ -1,5 +1,6 @@
 {
   "hooks": {
-    "post-checkout": "npm run rebuild"
+    "pre-push": "sh dev/husky/pre-push.sh",
+		"post-checkout": "npm run rebuild"
```
sc0ttkclark, line 4:
```suggestion
    "pre-push": "sh dev/husky/pre-push.sh",
    "post-checkout": "npm run rebuild"
```

PR #1531, opened 2020-01-08
fix(ticket header image): Allow adding ticket header image on non-event posts.

Camwyn:
The block editor always used the event REST API for adding ticket(and RSVP) images - this fails if we're not editing an event. 

This code changes it to detect the post type and use the appropriate endpoint (`tribe_events` for events, `posts` otherwise).

Code changes are only to update and delete for both RSVP and Ticket.

:ticket: [ET-54]
📸   http://p.tri.be/E6sYH4

### DIFF
--- a/src/modules/data/blocks/rsvp/sagas.js
+++ b/src/modules/data/blocks/rsvp/sagas.js
@@ -458,7 +458,19 @@ export function* updateRSVPHeaderImage( action ) {
 		yield put( ticketActions.setTicketsIsSettingsLoading( true ) );
 
 		const post_type = wpSelect('core/editor').getCurrentPostType();
-		const rest_base = ( 'tribe_events' === post_type ) ? 'tribe_events' : 'posts';
+		let rest_base = post_type;
+
+		switch ( post_type ) {
+			case 'post':
+				rest_base = 'posts';
+				break;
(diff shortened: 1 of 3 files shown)


src/modules/data/blocks/rsvp/sagas.js
```diff
@@ -456,8 +456,12 @@ export function* updateRSVPHeaderImage( action ) {
 		 */
 		yield put( actions.setRSVPIsSettingsLoading( true ) );
 		yield put( ticketActions.setTicketsIsSettingsLoading( true ) );
+
+		const post_type = wpSelect('core/editor').getCurrentPostType();
+		const rest_base = ( 'tribe_events' === post_type ) ? 'tribe_events' : 'posts';
```
paulmskim: One concern here is if the user puts this block on any other post type other than events or posts (pages or other custom post types), this will break. Can we make the `rest_base` depend on the post type completely?

Camwyn: Hm, good point!

Camwyn:
@paulmskim changed it up - good catch.
If there is a better way to get the endpoint, please let me know. This works but feels a bit much?

PR #1533, opened 2020-01-10
ETP-51 fix missing rsvp now

cliffordp:
Redo of https://github.com/moderntribe/event-tickets/pull/1523, which was based on https://github.com/moderntribe/event-tickets/tree/release/B20.01

New 1min video demo for _this_ PR: https://share.getcloudapp.com/geurRryA

[ETP-51]

[ETP-51]: https://moderntribe.atlassian.net/browse/ETP-51

PR #1532, opened 2020-01-09
Revert "[ETP-51] 138725 work even if post only has RSVPs"

sc0ttkclark:
Reverts moderntribe/event-tickets#1523

Merged into the wrong branch

PR #1523, opened 2019-12-26
[ETP-51] 138725 work even if post only has RSVPs

cliffordp:
https://central.tri.be/issues/138725

4min demo video: https://share.getcloudapp.com/6quLQloN

*See question comments, below...*

[ETP-51]

[ETP-51]: https://moderntribe.atlassian.net/browse/ETP-51

PR #1543, opened 2020-01-13
Fix/etp 51 redo

cliffordp:
[ETP-51]

[ETP-51]: https://moderntribe.atlassian.net/browse/ETP-51

### DIFF
--- a/src/template-tags/tickets.php
+++ b/src/template-tags/tickets.php
@@ -284,7 +284,7 @@ if ( ! function_exists( 'tribe_tickets_buy_button' ) ) {
 				}
 
 				$button = sprintf(
-					'<div class="tribe-common"><a class="tribe-common-c-btn" href="%1$s"><button class="tribe-common-c-btn">%2$s</button></a></div>',
+					'<div class="tribe-common"><a class="tribe-common-c-btn" href="%1$s">%2$s</a></div>',
 					esc_url( get_the_permalink( $event_id ) . $button_anchor ),
 					esc_html( $button_label )
 				);
--- a/src/template-tags/tickets.php
+++ b/src/template-tags/tickets.php
@@ -192,9 +192,13 @@ if ( ! function_exists( 'tribe_tickets_buy_button' ) ) {
 	function tribe_tickets_buy_button( $echo = true ) {
 		$event_id = get_the_ID();
 
-		// check if there are any tickets on sale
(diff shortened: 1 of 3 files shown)


src/template-tags/tickets.php
```diff
@@ -403,6 +396,48 @@ function tribe_events_ticket_is_on_sale( Tribe__Tickets__Ticket_Object $ticket )
...
+
+			// Skip tickets that are for a different provider than the event provider.
+			if (
+				$default_provider !== $ticket_provider->class_name
+				&& Tribe__Tickets__RSVP::class !== $ticket_provider->class_name
+			) {
+				continue;
+			}
+
+			$has_tickets_available = ( $has_tickets_available || tribe_events_ticket_is_on_sale( $ticket ) );
```
Camwyn: This is going to run through all the tickets no matter what, but it looks like you can break out once this is set to true?

Camwyn:
```suggestion
			if ( tribe_events_ticket_is_on_sale( $ticket ) ) {
			    $has_tickets_available = true;
			    break;
			}
```

You could even shorten it a bit more by leaving out `$has_tickets_available` and just setting `$ticket_availability[ $post_id ]` ?

Camwyn:
Can we get a link to the original PR in the description - assuming there is more info there?
Or a description, etc.
Just a ticket number link is never enough 😞

cliffordp:
@Camwyn well Jira auto-connects them all so I thought the info there would be fine. Nevertheless, for easier reference ;)
* https://github.com/moderntribe/event-tickets/pull/1523
* https://github.com/moderntribe/event-tickets/pull/1532
* https://github.com/moderntribe/event-tickets/pull/1533

src/template-tags/tickets.php
cliffordp: `tribe_tickets_is_current_time_in_date_window()` came from @borkweb in https://github.com/moderntribe/event-tickets/pull/1533/commits/e21e6c2fd8769983ddb72959b1030784c469003a so I didn't want to affect it for eventual merging with B20.01/G20.01

But good suggestions so let's do it anyway :)

sc0ttkclark: I added a note for myself to come back to this function in my optimization tasks. We can keep as is for now.

PR #1545, opened 2020-01-14
Fix/etp 53 thousand as decimal display

jesseeproductions:
_Ref:_ [ETP-53](https://moderntribe.atlassian.net/browse/ETP-53)

Fixes the thousand as decimal display. Since we always convert to use the period as decimal for calculations when we format the price to displace we should account for that. 

Screenshots of correct display:
https://cloudup.com/ci4J9ISUxFl

Incorrect screenshots are attached to the ticket.


readme.txt
```diff
@@ -119,6 +119,7 @@ Currently, the following add-ons are available for Event Tickets:
 
 = [4.11.2] TBD =
 
+* Fix - Price formatting method to prevent incorrect display when a comma is used as the decimal separator [ETP-53]
```
```suggestion
* Fix - Price formatting method now correctly prevents incorrect display when a comma is used as the decimal separator. [ETP-53]
```

src/resources/js/tickets-block.js
```diff
@@ -694,6 +694,12 @@ tribe.tickets.block = {
 
 		const s = ( prec ? toFixedFix( n, prec ) : Math.round( n ) ).toString().split( dec );
```
sc0ttkclark:
```suggestion
		let s = ( prec ? toFixedFix( n, prec ) : Math.round( n ) ).toString().split( dec );
```

jesseeproductions: good catch for some reason I glossed over the const part.....

PR #1546, opened 2020-01-14
Fix/etp 41 woo widget checkout link

jesseeproductions:
_Ref:_ [ETP-41](https://moderntribe.atlassian.net/browse/ETP-41)

Modify the Attendee Registration Shortcode so that is loads with ajax calls.
This prevents the has_shortcode from failing to find it when updating the checkout link in the WooCommerce Mini Cart Widget.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -119,7 +119,7 @@ Currently, the following add-ons are available for Event Tickets:
 
 = [4.11.2] TBD =
 
-* Fix - WooCommerce Mini-Cart widget checkout link to use the custom page for attendee registration if it is setup [ETP-41]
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -119,6 +119,7 @@ Currently, the following add-ons are available for Event Tickets:
 
 = [4.11.2] TBD =
 
+* Fix - WooCommerce Mini-Cart widget checkout link to use the custom page for attendee registration if it is setup [ETP-41]
```
```suggestion
* Fix - Override checkout link in WooCommerce Mini-Cart widget so it uses the custom page for attendee registration if it is setup. [ETP-41]
```

PR #1547, opened 2020-01-14
ET-596 - Initial removal of frontend-tickets script

jesseeproductions:
_Ref:_ [ET-596](https://moderntribe.atlassian.net/browse/ET-596)

Related: https://github.com/moderntribe/event-tickets-plus/pull/941

No screenshots or videos as the functionality remains the same.


src/Tribe/Tickets.php
```diff
@@ -1832,6 +1769,71 @@ public static function tickets_own_stock( $post_id ) {
...
+			}
+
+			$plugin = Tribe__Tickets__Main::instance();
+
+			wp_enqueue_script( 'wp-util' );
+
+			tribe_asset(
+				$plugin,
+				'tribe_tickets_frontend_tickets',
+				'frontend-ticket-form.js',
```
Camwyn: This file is deleted - so it'll never load? Are we trying to use this just for the localized variables? (I'm fairly sure they are all loaded in `event-tickets/src/Tribe/Editor/Blocks/Tickets.php->assets()`

Or was this supposed to change to load `tickets-block.js`?

src/Tribe/Tickets.php
```diff
@@ -1832,6 +1769,71 @@ public static function tickets_own_stock( $post_id ) {
 			return true;
 		}
 
+		/**
```
Camwyn: If we've removed all calls to this function, why aren't we removing it?
The file it registers is also removed.

jesseeproductions: If you would have an older version of ET+ running it could cause a fatal.  We could completely remove and then in the dependency manager update the required version. This way seems a bit safer.

src/Tribe/Tickets.php
jesseeproductions: From what I saw tickets-block has its own localize script. My QA of this did not produce any errors that it was missing something. 

I was wondering if the js file should remain until this deprecated method is removed?

jesseeproductions: I think I should add the deprecated method to the top of the function and remove the rest of the contents so it does not try to load the missing file.

src/Tribe/Tickets.php
Camwyn: Gotcha, makes sense!

src/Tribe/Tickets.php
Camwyn: Yeah, `tickets-block.js` does have it's own localize.

I'm down with the shift so the file doesn't load - maybe as part of deprecating it we just call for `tickets-block.js` instead?

Camwyn: If you make any change, go ahead and take off the Q label 😄

readme.txt
```diff
@@ -119,6 +119,7 @@ Currently, the following add-ons are available for Event Tickets:
 
 = [4.11.2] TBD =
 
+* Tweak - Remove duplicate ticket script to prevent conflicts  [ET-596]
```
```suggestion
* Tweak - Remove duplicate ticket script to prevent conflicts. [ET-596]
```

readme.txt
```diff
@@ -119,6 +119,7 @@ Currently, the following add-ons are available for Event Tickets:
...
 * Fix - Override checkout link in WooCommerce Mini-Cart widget so it uses the custom page for attendee registration if it is setup. [ETP-41]
 * Fix - Ensure that attendee images display horizontally in the frontend for Twenty Nineteen and Twenty Twenty themes. [ET-590]
 * Fix - Load JavaScript assets with Ticket Block when using Classic Editor [ET-587]
```
```suggestion
* Fix - Load JavaScript assets with Ticket Block when using Classic Editor. [ET-587]
```

sc0ttkclark:
Due to the `const` issue in JS, should we revisit the way this happens?

jesseeproductions:
Not sure I follow why the const is an issue for this coding? Is that more for the formatting PR?

sc0ttkclark:
Because the `tickets-block.js` sets a `const tribe` on first load, future loads could conflict there since it's set at the global level correct?

sc0ttkclark:
Fixed via #1554, we can move forward here

jesseeproductions:
Great, I was going to suggest another change, but that check in the script works for me.

PR #1549, opened 2020-01-14
Add ET Freemius integration

sc0ttkclark:
[ET-595]

## Screencasts

* TEC + ET opt-ins: https://share.getcloudapp.com/E0uEe57W
* ET only: https://share.getcloudapp.com/5zuJKZ5k
* TEC already opted-in and activated ET: https://share.getcloudapp.com/xQu0oKgE

[ET-595]: https://moderntribe.atlassian.net/browse/ET-595


src/Tribe/Integrations/Freemius.php
```diff
@@ -0,0 +1,307 @@
...
+			'has_addons'     => false,
+			'has_paid_plans' => false,
+		] );
+
+		$this->instance->add_filter( 'connect_url', [ $this, 'redirect_settings_url' ] );
+		$this->instance->add_filter( 'after_skip_url', [ $this, 'redirect_settings_url' ] );
+		$this->instance->add_filter( 'after_connect_url', [ $this, 'redirect_settings_url' ] );
+		$this->instance->add_filter( 'after_pending_connect_url', [ $this, 'redirect_settings_url' ] );
+
+		tribe_asset( $this->object_class, 'tribe-tickets-freemius', 'freemius.css', [], 'admin_enqueue_scripts' );
```
borkweb: I've found a hacky way around the plugin icon issue (and it relates the the freemius.css enqueue. Rather than the above line, we can do:

```php
		add_filter( 'fs_plugin_icon_event-tickets', function( $icon ) {
			add_action( 'admin_footer', function() {
				echo '<link rel="stylesheet" href="' . Tribe__Tickets__Main::instance()->plugin_url . '/src/resources/css/freemius.css"/>';
			});

			return $icon;
		} );
```

When the plugin icon template is being output, it is specific to the plugin being loaded in the moment. We can then include a stylesheet in the admin_footer that provides the CSS file with the correct icon. A similar change would be needed in TEC like so:

```php
		add_filter( 'fs_plugin_icon_the-events-calendar', function( $icon ) {
			add_action( 'admin_footer', function() {
				echo '<link rel="stylesheet" href="' . Tribe__Events__Main::instance()->plugin_url . '/src/resources/css/freemius.css"/>';
			});

			return $icon;
		} );
```

borkweb: Now, if the freemius.css file needs to be included all the time, we can break the icon-specific rule into its own CSS file. Then we'd have something like:

```php
tribe_asset( $this->object_class, 'tribe-tickets-freemius', 'freemius.css', [], 'admin_enqueue_scripts' );

add_filter( 'fs_plugin_icon_event-tickets', function( $icon ) {
	add_action( 'admin_footer', function() {
		echo '<link rel="stylesheet" href="' . Tribe__Tickets__Main::instance()->plugin_url . '/src/resources/css/freemius-site-icon.css"/>';
	});

	return $icon;
} );
```

sc0ttkclark:
```php
$this->instance->add_filter( 'plugin_icon', static function( $icon ) use ( $object_class ) {
	wp_enqueue_style( 'tribe-tickets-freemius', Tribe__Tickets__Main::instance()->plugin_url . '/src/resources/css/freemius.css' );

	return $icon;
}, 11, 6 );
```

PR #1551, opened 2020-01-15
fix(global stock): Incorrect global stock on new tickets.

Camwyn:
When saving tickets initially, we weren't updating the event meta value we check on the front end,
resulting in new shared tickets appearing sold out. This adds that update.

🐫 http://p.tri.be/iZdLbf
📷 http://p.tri.be/6geO8s

Requires: https://github.com/moderntribe/event-tickets-plus/pull/944

[ET-614]

[ET-614]: https://moderntribe.atlassian.net/browse/ET-614


src/Tribe/Ticket_Object.php
```diff
@@ -748,6 +748,10 @@ public function stock( $value = null ) {
 				&& isset( $this->get_event()->ID )
 			) {
 				$stock[] = (int) get_post_meta( $this->get_event()->ID, Tribe__Tickets__Global_Stock::GLOBAL_STOCK_LEVEL, true );
+				bdump(
```
sc0ttkclark: Maybe not

PR #1561, opened 2020-01-22
Move Woo-specific to ET+

cliffordp:
[ET-620]

[ET-620]: https://moderntribe.atlassian.net/browse/ET-620


cliffordp:
@sc0ttkclark I see the `.tribe-amount` wrapper around `1.00`: https://share.getcloudapp.com/yAuLeow1

Should we force wrapping in something like `.tribe-amount-suffix` _if $suffix is not empty_?

PR #1563, opened 2020-01-23
fix(optout logic): Correct the logic for hiding the optout.

Camwyn:
~Change `&&` to `||` - as the conditions are mutually exclusive.~
Remove the checks for the list - we always show it now.

Caught while doing testing instructions for the ticket.

🎫 [ET-615]

[ET-615]: https://moderntribe.atlassian.net/browse/ET-615


Camwyn:
Force push was to remove the unintentional common changes.

PR #1567, opened 2020-01-24
fix(global stock): Ensure we update the event when we add our first global stock ticket.

Camwyn:
In some cases, adding a global stock ticket did not update the event meta that indicated it was
using global stock. Now, in the ticket save() function where we deal with global stock, if we have
not enabled global stock already, we make sure to update the event to indicate it is enabled.

:ticket: [ET-614]

[ET-614]: https://moderntribe.atlassian.net/browse/ET-614


src/views/blocks/tickets/opt-out-hidden.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Block: RSVP
+ * Block: Tickets
  * Form Opt-Out
  *
  * Override this template in your own theme by creating a file at:
```
sc0ttkclark: We'll need to set the `@version TBD` here

PR #1580, opened 2020-02-03
fix(event date checks): Don't test a post as an event!

Camwyn:
At one point in the process, the post gets tested for event start/end dates
They don't exists if we're not on an event.

So let's not check them if they aren't there.

[ETP-159]

[ETP-159]: https://moderntribe.atlassian.net/browse/ETP-159


src/Tribe/Ticket_Object.php
```diff
@@ -350,6 +350,7 @@ public function date_in_range( $datetime = null ) {
 		 */
 		public function get_date( $date = '', $as_timestamp = true ) {
 			if ( '' === $date ) {
+				bdump('bail on no end date');
```
sc0ttkclark:
```suggestion
```

PR #1602, opened 2020-02-19
Save initial shared capacity value for global stock correctly on first Tribe Commerce ticket so availability shows as expected instead of zero

sc0ttkclark:
[ET-737]

Screencast: https://share.getcloudapp.com/NQuZbz5R

[ET-737]: https://moderntribe.atlassian.net/browse/ET-737


bordoni:
@sc0ttkclark did you do any testing of this behavior on Blocks editor?

sc0ttkclark:
Tested against Blocks vs Classic Editor and confirmed it works properly. There was one compatibility issue with Blocks enabled where it would save global stock correctly once and rerun the save for global stock and mess it up. I added code to remove the hooks before the saving to prevent that. It only happens for Blocks, but it should future proof that code a bit better.

PR #1625, opened 2020-03-24
Tickets > Move action to proper place.

juanfra:
🎫 https://moderntribe.atlassian.net/browse/ETP-302

Move action after the ticket was created, so it works also on ticket creation.

### DIFF
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -2842,7 +2842,7 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 				$ticket->end_date = date( Tribe__Date_Utils::DBDATEFORMAT, strtotime( $end_datetime ) );
 			}
 
-			// Pass the control to the child object
+			// Pass the control to the child object.
 			$save_ticket = $this->save_ticket( $post_id, $ticket, $data );
 
 			/**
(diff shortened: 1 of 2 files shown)


src/Tribe/Tickets.php
```diff
@@ -2842,6 +2842,9 @@ public function ticket_add( $post_id, $data ) {
 				$ticket->end_date = date( Tribe__Date_Utils::DBDATEFORMAT, strtotime( $end_datetime ) );
 			}
 
+			// Pass the control to the child object
```
tr1b0t: :no_entry_sign: Inline comments must end in full-stops, exclamation marks, or question marks

<sub>hash: 1e0a8762b1c11a6aae6293cca2eb0fe9</sub>

bordoni:
That is a good fix! Would you mind laying out steps that QA could test this? Would make it easier to make sure we can get it into the next release.

juanfra:
Well, there's nothing really to test from a QA perspective, it's just moving the action so the ticket is created when it's being triggered.

Maybe adding this to their functions.php file and checking that the ID is in place with the fix:

`add_action( 'tribe_tickets_ticket_add', 'check_ticket_data_on_ticket_add', 10, 3 );

function check_ticket_data_on_ticket_add( $post_id, $ticket, $data ) {

	error_log( var_export( $ticket->ID, true ) );

}`

sc0ttkclark:
Please update ticket with testing instructions for QA to use

PR #1654, opened 2020-05-06
Remove incompatible usage of func_get_args for PHP 7.0

bordoni:
🎟 [TEC-3104]
---
No changes in behavior just add compatibility for version PHP 7.0 and above. 
Originally added here: https://github.com/moderntribe/the-events-calendar/pull/2750

[TEC-3104]: https://moderntribe.atlassian.net/browse/TEC-3104


Luc45:
Hey Gustavo, I tried to reproduce an incompatibility of `func_get_args()` on PHP 7.0+ but I couldn't.

Is there any specific message showing?

This is how I tried to reproduce an error:

```php
<?php

class Foo {
    public static function bar($a, $b) {
        var_dump( func_get_args() );
    }
}

Foo::bar('arg1','arg2');
```

Output for PHP 5.0.0 until PHP 7.4.5:

```php
array(2) {
  [0]=>
  string(4) "arg1"
  [1]=>
  string(4) "arg2"
}
```
Which seems expected.

Here's a live sandbox if you want to play with it: https://3v4l.org/8QTcK

Luc45:
Ah, I see it now. https://3v4l.org/9vAIX

I'm not sure about the implications of this behavior in that context, but thanks, man!!

src/Tribe/Templates.php
```diff
@@ -16,8 +16,8 @@ class Tribe__Tickets__Templates extends Tribe__Templates {
 	 **/
 	public static function get_template_hierarchy( $template, $args = array() ) {
 		if ( ! is_array( $args ) ) {
+			$passed        = [ $template, $args ];
```
sc0ttkclark, line 19: This doesn't seem backcompat with the `$backwards_map`, am I wrong?

sc0ttkclark: Original TEC PR has the same backcompat breakage: https://github.com/moderntribe/the-events-calendar/pull/2750/files

sc0ttkclark:
```suggestion
			$passed        = func_get_args();
```

PR #1663, opened 2020-05-14
move non-ORM-specific bits to parent class

sc0ttkclark:
To support tests for [ETP-313]

[ETP-313]: https://moderntribe.atlassian.net/browse/ETP-313


borkweb:
This is way too old to be useful. Closing.

PR #1668, opened 2020-05-19
Tests/page objects

Luc45:
This is a sample implementation of the  Page Object approach to write tests driven by a browser.

webdriver

In this example it creates an event and tries to confirm a RSVP from the front-end.

These tests take time to write, and they take time to run, but they seem like a good choice to write end-to-end tests to cover some of the main behaviors of the plugin.

The idea of this PR is to briefly discuss  the idea of Page Objects, and whether we want to write more of those tests or not.

Ticket https://moderntribe.atlassian.net/browse/ETP-319


bordoni:
I really like this idea, we should continue to expand on this. Lets get these tests passing and start working towards merging.

PR #1670, opened 2020-05-25
[ET-844] Feature(virtual events): Add hook for VE in email template.

Camwyn:
This adds a hook that Virtual Events depends on to add the virtual link/zoom info to the ticket emails.

[VE-46]
[ET-844]

:camera: http://p.tri.be/EsQ6Ly

RE: https://github.com/moderntribe/events-virtual/pull/59

[VE-46]: https://moderntribe.atlassian.net/browse/VE-46

[ET-844]: https://moderntribe.atlassian.net/browse/ET-844

### DIFF
--- a/src/views/tickets/email.php
+++ b/src/views/tickets/email.php
@@ -440,8 +440,8 @@
 										 *
 										 * @since TBD
 										 *
-										 * @param array $ticket Ticket information
-										 * @param WP_Post $event Event post object.
+										 * @param array   $ticket Ticket information.
+										 * @param WP_Post $event  Event post object.
 										 */
 										do_action( 'tribe_tickets_ticket_email_after_details', $ticket, $event );
 										?>
--- a/src/views/tickets/email.php
+++ b/src/views/tickets/email.php
@@ -435,6 +435,23 @@
 											</tr>
 										</table>
 										<?php
+										/**
+										 * Allows inserting content after the "ticket details" section.
+										 *
+										 * @since TBD
+										 */
+										do_action( 'tribe_tickets_ticket_email_after_details', $ticket, $event );
+										?>
+										<?php if ( has_action( 'tribe_tickets_ticket_email_after_details' ) ) : ?>
+											<table class="whiteSpace" border="0" cellpadding="0" cellspacing="0" width="100%">
+												<tr>
+													<td valign="top" align="left" width="100%" height="30" style="height:30px; background:#f7f7f7; padding: 0 !important; margin:0 !important;">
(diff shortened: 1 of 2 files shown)


src/views/tickets/email.php
```diff
@@ -435,6 +435,23 @@
 											</tr>
 										</table>
 										<?php
+										/**
+										 * Allows inserting content after the "ticket details" section.
+										 *
```
juanfra, line 440: Let's add the params to the docblock.

src/views/tickets/email.php
```diff
@@ -435,6 +435,23 @@
...
+										 * @since TBD
+										 */
+										do_action( 'tribe_tickets_ticket_email_after_details', $ticket, $event );
+										?>
+										<?php if ( has_action( 'tribe_tickets_ticket_email_after_details' ) ) : ?>
```
juanfra, line 445: Not sure if I'm getting the logic here. Wouldn't `has_action` return always true given that you defined the action some lines above?

Camwyn: It should only return true if something actually hooks to it.

Camwyn: Going to re-test and make sure it's doing what I expect.

juanfra: Right, what I mean is that if anything else is added, even if the action hooked doesn't print things, we're adding more markup. What I meant is that maybe this markup could be part of what's inserting and not be here.

sc0ttkclark: Should the hook adding content be in charge of adding the whitespace markup? Since multiple things hooking in could add multiple renders?

Camwyn: Yep - moving the addition to the templates in VE.

src/views/tickets/email.php
```diff
@@ -435,6 +435,17 @@
 											</tr>
 										</table>
 										<?php
+										/**
+										 * Allows inserting content after the "ticket details" section.
+										 *
+										 * @since TBD
+										 *
+										 * @param array $ticket Ticket information
+										 * @param WP_Post $event Event post object.
```
sc0ttkclark, line 444:
```suggestion
										 * @param array   $ticket Ticket information.
										 * @param WP_Post $event  Event post object.
```

Camwyn:
@sc0ttkclark 
> Looks good pending release branch/ticket

Does this need a green ticket or is the blue one good enough? (added to the description)

sc0ttkclark:
This will need a ticket for ET specifically and planned for in the next sprint. cc @jrserani

PR #1673, opened 2020-05-26
[ET-1343] Only use wp_cache_flush() if an external object cache plugin is not being used

r-a-y:
### 🎫 Ticket

[ET-1343] 

### 🗒️ Description

It's generally bad practice to use `wp_cache_flush()` especially when an object cache plugin is in use.

This PR only calls `wp_cache_flush()` if an external object cache is not in use.

### 🎥 Artifacts 

N/A

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1343]: https://theeventscalendar.atlassian.net/browse/ET-1343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Luc45:
Looking at the git history of that specific line, it seems it was introduced precisely to accommodate compatibility with Object Cache, as to clear object cache as well when clearing transients.

If that's so, we might indeed remove `wp_flush_cache()`, as it clears ALL the object cache, and look into a way of clearing object cache just for what we want to reset, as the original transient behavior does with `delete_transient( __CLASS__ . $key );`.

r-a-y:
Thanks for the quick review!

Sounds good to remove the `wp_cache_flush()` call. Not sure why you guys were clearing the entire object cache, but I wanted to ensure that it was still called if you needed it to :)

bordoni:
First I wanted to say, thanks for putting the Pull Request to solve that problem, it's def something that we need to tackle.

One of the major problems with caching expiring is that there isn't a way to just delete a group of cache keys, and it would be a lot of custom code to get to that point when it comes to the individual object cache pieces we need to expire. 

I am spinning up a ticket on our internal Tracker for us to figure out the history of that particular piece of code.

boonebgorges:
Many thanks to the team for your quick attention to this issue.

> One of the major problems with caching expiring is that there isn't a way to just delete a group of cache keys, and it would be a lot of custom code to get to that point when it comes to the individual object cache pieces we need to expire.

I'd recommend considering the strategy of concatenating cache keys using an incrementor. The WP convention is to use `last_changed`. So:

```php
$last_changed = wp_cache_get_last_changed( 'event-tickets' );
$cache_key = 'some_string' . $last_changed;
$cached = wp_cache_get( $cache_key, 'event-tickets' );
if ( false === $cached ) {
    $cached = perform_a_query();
    wp_cache_set( $cache_key, $cached, 'events-tickets' );
}
```

Then, you can invalidate every one of these items by invalidating the incrementor:

```php
wp_cache_delete( 'last_changed', 'event-tickets' );
```

Luc45:
Needs detailed automated tests, but this should delete just the relevant keys from object cache:

```suggestion
if ( ! wp_using_ext_object_cache() ) {
	foreach ( $this->keys as $key ) {
		delete_transient( __CLASS__ . $key );
		delete_transient( __CLASS__ . $key . $hash );
	}
} else {
	foreach ( $this->keys as $key ) {
		wp_cache_delete( __CLASS__ . $key );
		wp_cache_delete( __CLASS__ . $key . $hash );
	}
}

```

Luc45:
I wonder if @lucatume or @sc0ttkclark could share some insight in regards to @boonebgorges 's suggestion?

borkweb:
Howdy, @r-a-y! Sorry this took so long to wrap up. It got lost in the shuffle of a _lot_ of activity. Luckily, @codingmusician reviewed the thread and took over the work to get it merged in via #2215. Thank you for the PR, the conversation, and the poke!

PR #1676, opened 2020-06-03
Fixes an notice for undefined variable in the birth template

Luc45:


src/views/registration-js/attendees/fields/birth.php
```diff
@@ -28,7 +29,7 @@
 	<div <?php tribe_classes( $classes ); ?> >
 		<label
 			class="tribe-common-b2--min-medium tribe-tickets-meta-label"
-			for="<?php echo esc_attr( $birth_field->month_id ); ?>"
+			for="<?php echo esc_attr( $field_name ); ?>"
```
juanfra, line 32: The value here should be `$option_id`

src/views/registration-js/attendees/fields/birth.php
```diff
@@ -13,6 +13,7 @@
 
 $required    = isset( $field->required ) && 'on' === $field->required ? true : false;
```
juanfra, line 14: We don't need a ternary to set `true` or `false` here. We can skip that by using the logic as the bool.

juanfra, line 14:
```suggestion
$required    = isset( $field->required ) && 'on' === $field->required;
```

src/views/registration-js/attendees/fields/birth.php
juanfra: https://www.w3schools.com/tags/att_label_for.asp

src/views/registration-js/attendees/fields/birth.php
Luc45: Good catch man! Thanks!

src/views/registration-js/attendees/fields/birth.php
```diff
@@ -11,8 +11,9 @@
  * @see     Tribe__Tickets_Plus__Meta__Field__Birth
  */
 
-$required    = isset( $field->required ) && 'on' === $field->required ? true : false;
+$required    = isset( $field->required ) && 'on' === $field->required;
```
tr1b0t, line 14: :warning: Equals sign not aligned with surrounding assignments; expected 2 spaces but found 4 spaces

<sub>hash: e7c2c977c2ceea34fff7b220d220367c</sub>

src/views/registration-js/attendees/fields/birth.php
```diff
@@ -11,8 +11,9 @@
  * @see     Tribe__Tickets_Plus__Meta__Field__Birth
```
juanfra, line 11: Every time we introduce changes to a template file we need to also document this with something like:
`@since VERSION Descriptive text explaining the changes`.

Also this template is missing the`@version`. Each time we make changes to a template, we need to set the `@version TBD`.

Luc45: Added. Thanks. https://github.com/moderntribe/event-tickets/pull/1676/commits/c5b33a5f81300426a1a5443f2a2a56b4b7a47c05

src/Tribe/Events/Views/V2/Models/Tickets.php
```diff
@@ -189,7 +189,7 @@ public function fetch_data() {
 							$text = _n( '%1$s spot left', '%1$s spots left', $stock, 'event-tickets' );
 						} else {
 							/* translators: %1$s: Number of stock, %2$s: Ticket label, %3$s: Tickets label */
-							$text = _n( "%1$s %2$s left", "%1$s %3$s left", $stock, 'event-tickets' ); // phpcs:ignore
+							$text = _n( '%1$s %2$s left', '%1$s %3$s left', $stock, 'event-tickets' );
```
tr1b0t, line 192: :warning: Mismatched placeholders is probably an error

<sub>hash: 8a757c1133745b7f32a5854c84259b37</sub>

src/views/registration-js/attendees/fields/birth.php
```diff
@@ -6,13 +6,17 @@
...
+ * @version 4.12.1
+ *
+ * @since 4.12.1 Introduced template.
+ * @since TBD Fix PHP notice for undefined variable.
  *
  * @see     Tribe__Tickets_Plus__Meta__Field__Birth
  */
 
-$required    = isset( $field->required ) && 'on' === $field->required ? true : false;
+$required    = isset( $field->required ) && 'on' === $field->required;
```
tr1b0t, line 17: :warning: Equals sign not aligned with surrounding assignments; expected 2 spaces but found 4 spaces

<sub>hash: c4cfe3f588068a6ad31c641ea85f94d3</sub>

src/views/registration-js/attendees/fields/birth.php
Luc45: Nice catch man, thanks! https://github.com/moderntribe/event-tickets/pull/1676/commits/5456e6aefbb3fb43f7b744febebdab0eac9310b6

src/views/registration-js/attendees/fields/birth.php
```diff
@@ -6,13 +6,17 @@
  * Override this template in your own theme by creating a file at:
  * [your-theme]/tribe/tickets/registration-js/attendees/fields/birth.php
  *
- * @since   4.12.1
+ * @version 4.12.1
```
juanfra, line 9:
```suggestion
 * @version TBD
```

Nice! One small thing, When we work on templates we should set the version as TBD, so when we pack things for the release with https://github.com/moderntribe/tribe-product-utils we can set the correct version.

Luc45: Ah! So:

- `@since` is used to inform when that template was introduced.
- `@version` is used to inform what changed between each version

Is that accurate?

juanfra: So:

-  `@since` is to state in which version the template was introduced. And then to list the changes introduced `@since` version.
- `@version` is the version of the template. If we make changes on a template, it means that the version should be updated to the release version. By leaving it as TBD, when we make the package and release we ensure that's being set like it should.

This info is used, for example, in "Events > Help" to let users know if we introduced changes to templates or something. And ideally if they're using their own templates, we could warn them when they are out of sync with our own versions.

To set an example, let's say a template was introduced in `1.2.3` and we're adding changes to... add bold to a title. So, the changes would be:

```
@since 1.2.3
@since TBD We are adding bold to a title.

@version TBD
```

Then when we do the release tasks, the TBD will be replaced with the number we're using for the release.

Luc45: Awesome, thanks for the explanation! :100:

readme.txt
```diff
@@ -117,6 +117,10 @@ Currently, the following add-ons are available for Event Tickets:
 
 == Changelog ==
 
+= [TBD] =
+
+* Fix - Resolve a PHP notice when rendering the Birth date field in Event Tickets Plus [ETP-330]
```
sc0ttkclark, line 122:
```suggestion
* Fix - Resolve a PHP notice when rendering the Event Tickets Plus field for Birth date [ETP-330]
```

PR #1683, opened 2020-06-11
[ET-796] [ET-797] Handle RSVP admin notice, constant, setup opt-in setting, and start handling RSVP rendering

sc0ttkclark:
[ET-796]
[ET-797]

This PR accomplishes:

- New opt-in setting in admin area (not shown unless dev constant is enabled)
- New opt-in setting method check with support for our dev constant `TRIBE_TICKETS_RSVP_NEW_VIEWS`
- New opt-in notice (commented out from being called for now as it will ship enabled in G20.07)
- RSVP classic now renders new abstracted block render function (if opt-in enabled)
- RSVP block now renders new abstracted block render function
- RSVP abstracted block render function will render the new `v2/rsvp` view if opt-in enabled, otherwise it will render the old `blocks/rsvp` view
- RSVP abstracted block render function now supports new temporary preview constant `TRIBE_TICKETS_RSVP_NEW_VIEWS_PREVIEW` which will show static kitchen sink for themers and customers to use as a guide when preparing their view customizations

[ET-796]: https://moderntribe.atlassian.net/browse/ET-796
[ET-797]: https://moderntribe.atlassian.net/browse/ET-797

### DIFF
--- a/src/Tribe/Admin/Notices.php
+++ b/src/Tribe/Admin/Notices.php
@@ -50,7 +50,7 @@ class Tribe__Tickets__Admin__Notices {
 		/** @var Tribe__Settings $settings */
 		$settings = tribe( 'settings' );
 
-		// Bail if user cannot change settings
+		// Bail if user cannot change settings.
 		if ( ! current_user_can( $settings->requiredCap ) ) {
 			return;
 		}
--- a/src/Tribe/Admin/Notices.php
+++ b/src/Tribe/Admin/Notices.php
@@ -68,11 +68,8 @@ class Tribe__Tickets__Admin__Notices {
 			return;
 		}
 
-		/** @var Tribe__Tickets__RSVP $rsvp */
-		$rsvp = tribe( 'tickets.rsvp' );
-
 		// Bail if the option is already in use.
-		if ( $rsvp->use_new_views() ) {
+		if ( tribe_tickets_rsvp_new_views_is_enabled() ) {
 			return;
 		}
 
--- a/src/Tribe/Admin/Notices.php
+++ b/src/Tribe/Admin/Notices.php
@@ -13,7 +13,104 @@ class Tribe__Tickets__Admin__Notices {
 	 * @since 4.7
 	 */
 	public function hook() {
-		add_action( 'plugins_loaded', array( $this, 'maybe_display_plus_commerce_notice' ) );
+		add_action( 'admin_init', array( $this, 'maybe_display_notices' ) );
+	}
+
+	/**
+	 * Maybe display admin notices.
+	 *
+	 * @since TBD
+	 */
+	public function maybe_display_notices() {
+		// Bail on the unexpected
+		if (
+			! function_exists( 'tribe_installed_before' )
+			|| ! class_exists( 'Tribe__Admin__Notices' )
+		) {
+			return;
+		}
+
+		$this->maybe_display_plus_commerce_notice();
+
+		/** @var Tribe__Tickets__RSVP $rsvp */
+		$rsvp = tribe( 'tickets.rsvp' );
+
+		// @todo Remove this early return when we are ready to show the opt-in notice in G20.07.
+		if ( ! $rsvp->use_new_views() ) {
+			return;
+		}
+
+		$this->maybe_display_rsvp_new_views_options_notice();
+	}
+
+	/**
+	 * Display dismissible notice about new RSVP view settings.
+	 *
+	 * @since TBD
+	 */
+	public function maybe_display_rsvp_new_views_options_notice() {
+		// Bail if previously dismissed this notice.
+		if ( Tribe__Admin__Notices::instance()->has_user_dimissed( __FUNCTION__ ) ) {
+			return;
+		}
+
+		/** @var Tribe__Settings $settings */
+		$settings = tribe( 'settings' );
+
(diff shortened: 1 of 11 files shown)


src/Tribe/Admin/Notices.php
```diff
@@ -13,7 +13,105 @@ class Tribe__Tickets__Admin__Notices {
...
+	public function maybe_display_rsvp_new_views_options_notice() {
+		// Bail if previously dismissed this notice.
+		if ( Tribe__Admin__Notices::instance()->has_user_dimissed( __FUNCTION__ ) ) {
+			return;
+		}
+
+		/** @var Tribe__Settings $settings */
+		$settings = tribe( 'settings' );
+
+		// Bail if user cannot change settings
```
juanfra, line 53:
```suggestion
		// Bail if user cannot change settings.
```

src/Tribe/Editor/Blocks/Rsvp.php
```diff
@@ -198,6 +185,30 @@ public function assets() {
...
+			array( 'jquery' ),
+			null,
+			array(
+				'localize'     => array(
+					'name' => 'TribeRsvp',
+					'data' => array(
+						'ajaxurl' => admin_url( 'admin-ajax.php', ( is_ssl() ? 'https' : 'http' ) ),
+					),
+				),
+			)
```
juanfra, line 201: Let's use the array short syntax here.

src/Tribe/Editor/Blocks/Rsvp.php
```diff
@@ -198,6 +185,30 @@ public function assets() {
...
+		);
+
+		tribe_asset(
+			$plugin,
+			'tribe-tickets-rsvp-style',
+			'app/v2/rsvp/frontend.css',
+			array(),
```
juanfra, line 208:
```suggestion
			[],
```

src/Tribe/RSVP.php
```diff
@@ -1207,7 +1212,27 @@ public function front_end_tickets_form( $content ) {
...
+	public function use_new_views() {
+		// Determine if ET was installed at version 4.12.2+.
+		$should_default_to_on = ! tribe_installed_before( 'Tribe__Tickets__Main', '4.12.2' );
+
+		return (
+			(
+				defined( 'TRIBE_TICKETS_RSVP_NEW_VIEWS' )
+				&& TRIBE_TICKETS_RSVP_NEW_VIEWS
+			)
+			|| (boolean) tribe_get_option( 'rsvp_use_new_views', $should_default_to_on )
```
juanfra, line 1234: Maybe we can prefix the option? Not sure how we do it normally, but I feel we should prefix it.

src/Tribe/RSVP.php
```diff
@@ -1207,7 +1212,27 @@ public function front_end_tickets_form( $content ) {
...
+	}
+
+	/**
+	 * Determine whether to use new RSVP views.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool Whether to use new RSVP views.
+	 */
+	public function use_new_views() {
```
bordoni, line 1225: Can we make this into a template tag to have more explicit visibility on it? Similar to what we did on Views V2?

src/Tribe/RSVP.php
bordoni: It's inside of our own Option so it would be safe I think.

src/Tribe/RSVP.php
```diff
@@ -1207,7 +1212,27 @@ public function front_end_tickets_form( $content ) {
...
+		// Determine if ET was installed at version 4.12.2+.
+		$should_default_to_on = ! tribe_installed_before( 'Tribe__Tickets__Main', '4.12.2' );
+
+		return (
+			(
+				defined( 'TRIBE_TICKETS_RSVP_NEW_VIEWS' )
```
bordoni, line 1231: I would love for us to use the structure from what we did on Views V2 for this one, where we check each one and return right after so one can overwrite the other, with also a filter for those using the option.

src/Tribe/Editor/Blocks/Rsvp.php
```diff
@@ -178,23 +165,45 @@ public function assets() {
...
+			[ 'jquery' ],
+			null,
+			[
+				'localize' => [
+					'name' => 'TribeRsvp',
+					'data' => [
+						'ajaxurl' => admin_url( 'admin-ajax.php', ( is_ssl() ? 'https' : 'http' ) ),
+					],
+				],
+			] );
```
bordoni, line 200:
```suggestion
			]
		);
```

src/Tribe/RSVP.php
sc0ttkclark: Do you have a reference I can use for that?

src/Tribe/RSVP.php
sc0ttkclark: A global function that has the logic in it? Or a global function that calls this method?

src/Tribe/RSVP.php
bordoni: https://github.com/moderntribe/the-events-calendar/blob/master/src/functions/views/provider.php#L14

src/Tribe/RSVP.php
bordoni: Global with the logic in it.

PR #1729, opened 2020-07-23
[ET-880] Check against the class name like we do everywhere else

sc0ttkclark:
[ET-880]

Fixes usage of the provider class var and checking. No changelog needed as it fixes an item that already had a changelog. Also, it's worth noting that the screenshot is bugged with a style issue we're investigating separately.

**Original problem where minicart was empty:** https://share.getcloudapp.com/bLuj4jkG

**Screenshot:**

image

[ET-880]: https://moderntribe.atlassian.net/browse/ET-880


bordoni:
No need for changelog since it's something we already have a changelog for.

PR #1716, opened 2020-07-14
[ET-857] [ET-858] [ET-859] Add AJAX action and initial handling

sc0ttkclark:
[ET-857] + [ET-858] + [ET-859]

[ET-857]: https://moderntribe.atlassian.net/browse/ET-857

[ET-858]: https://moderntribe.atlassian.net/browse/ET-858

[ET-859]: https://moderntribe.atlassian.net/browse/ET-859

Screengif of the AJAX handling in action: https://share.skc.dev/yAubWrZw

Going to follow up with another PR to handle the final processing of RSVP and ARI separately from this.

Related: moderntribe/event-tickets-plus#1013

PR #1738, opened 2020-08-04
Fix/et 833 opt in out js

juanfra:
🎫 https://moderntribe.atlassian.net/browse/ET-833
🎥 https://www.loom.com/share/7bd62a9b692e4cd3a581bf3ff1ae25c1

This one goes hand in hand with https://github.com/moderntribe/event-tickets/pull/1716 to complete the functionality. This is the implementation of the JS call, according to the data attributes defined on the other PR.

### DIFF
--- a/src/resources/js/v2/rsvp-block.js
+++ b/src/resources/js/v2/rsvp-block.js
@@ -46,6 +46,7 @@ tribe.tickets.rsvp.block = {};
 		cancelButton: '.tribe-tickets__rsvp-form-button--cancel',
 		errorMessage: '.tribe-tickets__form-message--error',
 		hiddenElement: '.tribe-common-a11y-hidden',
+		displayToggle: '.tribe-tickets__rsvp-actions-success-going-toggle-input',
 	};
 
 	/**
@@ -139,6 +140,35 @@ tribe.tickets.rsvp.block = {};
 	};
 
 	/**
+	 * Handle the RSVP toggle for listing in public attendee list.
+	 *
+	 * @since TBD
+	 *
+	 * @param {event} e submission event
+	 */
+	obj.handleDisplayToggle = function( e ) {
+		e.preventDefault();
+
+		const $input = $( e.target );
+		const rsvpId = $input.data( 'rsvp-id' );
+		const checked = $input.prop( 'checked' );
+		const attendeeIds = $input.data( 'attendee-ids' );
+		const nonce = $input.data( 'opt-in-nonce' );
+		const $container = e.data.container;
+
+		var data = {
+			action: 'tribe_tickets_rsvp_handle',
+			ticket_id: rsvpId,
+			step: 'opt-in',
+			opt_in: checked,
+			opt_in_none: nonce,
+			attendee_ids: attendeeIds,
+		};
+
+		tribe.tickets.rsvp.manager.request( data, $container );
+	};
+
+	/**
 	 * Handle the RSVP form submission
 	 *
 	 * @since TBD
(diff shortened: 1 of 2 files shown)


src/resources/js/v2/rsvp-block.js
```diff
@@ -138,6 +135,35 @@ tribe.tickets.rsvp.block = {};
...
 	};
 
+	/**
+	 * Handle the RSVP toggle for listing in public attendee list.
+	 *
+	 * @since TBD
+	 *
+	 * @param {event} e submission event
+	 */
+	obj.handleDisplayToggle = function( e ) {
```
bordoni, line 145: since we are using the variable, let us call it `event`.

src/resources/js/v2/rsvp-block.js
```diff
@@ -138,6 +135,35 @@ tribe.tickets.rsvp.block = {};
...
+
+		const $input = $( e.target );
+		const rsvpId = $input.data( 'rsvp-id' );
+		const checked = $input.prop( 'checked' );
+		const attendeeIds = $input.data( 'attendee-ids' );
+		const nonce = $input.data( 'opt-in-nonce' );
+		const $container = e.data.container;
+
+		var data = {
+			action: 'tribe_tickets_rsvp_handle',
```
bordoni, line 156: Wonder if we could store this action name in an attribute of this object, similar to how we approach our PHP.

src/resources/js/v2/rsvp-block.js
```diff
@@ -138,6 +135,35 @@ tribe.tickets.rsvp.block = {};
...
+	 */
+	obj.handleDisplayToggle = function( e ) {
+		e.preventDefault();
+
+		const $input = $( e.target );
+		const rsvpId = $input.data( 'rsvp-id' );
+		const checked = $input.prop( 'checked' );
+		const attendeeIds = $input.data( 'attendee-ids' );
+		const nonce = $input.data( 'opt-in-nonce' );
+		const $container = e.data.container;
```
bordoni, line 153: Should we save before using it or use it directly? (not sure)

src/resources/js/v2/rsvp-block.js
```diff
@@ -138,6 +135,35 @@ tribe.tickets.rsvp.block = {};
...
+
+		var data = {
+			action: 'tribe_tickets_rsvp_handle',
+			ticket_id: rsvpId,
+			step: 'opt-in',
+			opt_in: checked,
+			opt_in_none: nonce,
```
paulmskim, line 160: `opt_in_none`? Is this how it is handled in the PHP or is this a typo on the JS side?

juanfra: No, that's a typo haha🤦‍♂️

src/resources/js/v2/rsvp-block.js
juanfra: I don't understand what you mean. What are you suggesting to save?

src/resources/js/v2/rsvp-block.js
juanfra: We already have it this way in all our different JS files across projects. I feel we should keep consistency with how we have the JS files and not PHP.

sc0ttkclark: I'm also okay with leaving this as it is for now, but if there's anything we need to do to improve this later across our JS that'd be a cool ticket to work on

src/resources/js/v2/rsvp-block.js
sc0ttkclark: I'm good with leaving it set here

PR #1732, opened 2020-07-30
RSVP > ARI

juanfra:
🎫 https://moderntribe.atlassian.net/browse/ET-866
🎥 https://www.loom.com/share/a98290516a634611b698d0b484de774f

Require changes from ET+

PR #1739, opened 2020-08-05
Fix/et 866 a11y remove tabindex

juanfra:
🎫 https://moderntribe.atlassian.net/browse/ET-866
📹 not required.


juanfra:
Follow up of https://github.com/moderntribe/event-tickets/pull/1732#pullrequestreview-461343846

PR #1756, opened 2020-08-10
Fix/et 825 not going text

juanfra:
🎫 ET-825
🎥 https://www.loom.com/share/0c1f808079284564a8ab6a1373de21a2

### DIFF
--- a/src/resources/postcss/rsvp/_actions.pcss
+++ b/src/resources/postcss/rsvp/_actions.pcss
@@ -104,7 +104,7 @@
 
 	.tribe-tickets__rsvp-actions-success-going-text {
 		margin-left: var(--spacer-1);
-		tex-transform: capitalize;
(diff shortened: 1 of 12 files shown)


src/resources/postcss/rsvp/_actions.pcss
```diff
@@ -104,6 +104,7 @@
 
 	.tribe-tickets__rsvp-actions-success-going-text {
 		margin-left: var(--spacer-1);
+		tex-transform: capitalize;
```
paulmskim, line 107: Was this meant to be capitalized?

```suggestion
		text-transform: capitalize;
```

juanfra: Exactly! Thanks for the heads up. Butterfly keyboard was the worse decision apple ever made.

PR #1768, opened 2020-08-20
[ET-898] Fix the Attendees list output so it will echo or return properly

sc0ttkclark:
[ET-898]

The issue was preventing the classic Attendees List from working properly in the context of echo.

Screenshot: https://p199.p4.n0.cdn.getcloudapp.com/items/jkuZPpYY/Screen%20Shot%202020-08-20%20at%205.25.29%20PM.png?v=317133c23c465e6c3b7b2a984b94b4d9

### DIFF
--- a/src/Tribe/Tickets_View.php
+++ b/src/Tribe/Tickets_View.php
@@ -1132,6 +1132,13 @@ class Tribe__Tickets__Tickets_View {
 
 		$before_content = ob_get_clean();
 
+		// Maybe echo the content from the action.
+		if ( $echo ) {
+			echo $before_content;
+
+			$before_content = '';
+		}
+
 		// Maybe render the new views.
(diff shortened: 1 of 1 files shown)


src/Tribe/Tickets_View.php
```diff
@@ -1132,6 +1132,13 @@ public function get_rsvp_block( $post, $echo = true ) {
 
 		$before_content = ob_get_clean();
 
+		// Maybe echo the content from the action.
+		if ( $echo ) {
+			echo $before_content;
```
tr1b0t, line 1137: * :warning: Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$before_content'
* :no_entry_sign: All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks), found '$before_content'.

<sub>hash: 39d003e13b1f68d90f180b29f819d0bd</sub>

PR #1774, opened 2020-08-31
[ETP-455] ET JS Normalization > Componetize and use loader library.

juanfra:
🎫 ETP-455
🎥 https://www.loom.com/share/536e9d639b084a53ab8581660967e335 

- Add loader library.
- Use the componetized view that works with the library. 
- Enqueue assets by groups and fix dependencies.

The tickets block will be added independently because I need to add more things that wouldn't be relevant to this PR.


src/views/registration-js/content.php
```diff
@@ -187,18 +193,20 @@ class="tribe-tickets__item__attendee__fields__form <?php echo sanitize_html_clas
 			'components/notice',
 			[
 				'notice_classes'  => $notice_classes,
-				'content' => sprintf(
+				'content'         => sprintf(
```
tr1b0t, line 196: :warning: Array double arrow not aligned correctly; expected 8 space(s) between "'content'" and double arrow, but found 9.

<sub>hash: 6a7d3e8ce89f7bb5953dc66ecf7aad14</sub>

src/resources/postcss/tickets-registration-page.pcss
```diff
@@ -36,6 +36,13 @@ body.page-tribe-attendee-registration {
 		}
 	}
 
+	/* -----------------------------------------------------------------------------
+	 * Hide from screenreaders & browsers
+	 * ----------------------------------------------------------------------------- */
+	.tribe-common-a11y-hidden {
```
paulmskim, line 42: does this not already come from common? or is common CSS not loaded on the tickets registration page?

src/views/registration-js/content.php
```diff
@@ -82,12 +87,13 @@
...
 			'components/notice',
 			[
-				'id' => 'tribe-tickets__notice__attendee-registration',
+				'id'             => 'tribe-tickets__notice__attendee-registration',
 				'notice_classes' => [
 					'tribe-tickets__notice--error',
 					'tribe-tickets__validation-notice',
 				],
 				'content' => sprintf(
+					// Translators: % the tickets number.
```
paulmskim, line 96:
```suggestion
					// Translators: %s the tickets number.
```

paulmskim (commented):
some lint issues that need to be fixed, also add `/* global tribe */` to the top of the JS files to get rid of most of the lint errors.

src/resources/postcss/tickets-registration-page.pcss
juanfra: Specifically in the AR page the loader is outside of the wrapper, and it needs this class to be hidden.

src/views/registration-js/content.php
juanfra: Done. Thanks.

src/views/registration-js/content.php
juanfra: Done.

juanfra:
Thank you, Paul!

src/views/registration-js/attendees/content.php
```diff
@@ -10,6 +10,7 @@
  * @version 4.11.0
```
sc0ttkclark, line 10: Needs a since even for whitespace changes, so we should remove them or add the since

src/resources/postcss/tickets-registration-page.pcss
sc0ttkclark: @paulmskim any further concern here?

src/resources/js/tickets-registration-page.js
```diff
@@ -1,24 +1,29 @@
...
-if ( 'undefined' === typeof window.tribe ) {
-	window.tribe = {};
-}
-
-if ( 'undefined' === typeof window.tribe.tickets ) {
-	window.tribe.tickets = {};
-}
-
-window.tribe.tickets.registration = {};
+/* global tribe */
```
sc0ttkclark, line 1: Shouldn't much of these changes be in a v2 file?

src/Tribe/Assets.php
```diff
@@ -43,21 +43,52 @@ public function enqueue_scripts() {
 		// Tickets registration page styles.
 		tribe_asset(
```
sc0ttkclark, line 44: This should be behind the opt-in check

sc0ttkclark (changes requested):
Some tweaks probably here

src/resources/postcss/tickets-registration-page.pcss
paulmskim: no concerns. this is good.

src/resources/js/v2/tickets-loader.js
```diff
@@ -0,0 +1,93 @@
...
+	 * @since TBD
+	 *
+	 * @return {void}
+	 */
+	obj.ready = function() {
+		// Silence is golden.
+	};
+
+	// Configure on document ready.
+	$document.ready( obj.ready );
```
paulmskim, line 92: we don't' need to run empty functions if we don't need it.

src/resources/js/v2/tickets-loader.js
```diff
@@ -0,0 +1,93 @@
...
+tribe.tickets = tribe.tickets || {};
+
+/**
+ * Configures ET Loader Object in the Global Tribe variable
+ *
+ * @since TBD
+ *
+ * @type   {PlainObject}
+ */
+tribe.tickets.loader = {};
```
paulmskim, line 18: from a bigger perspective, I wonder if having a common JS for the loader component is a good idea.

src/resources/js/v2/tickets-loader.js
sc0ttkclark: I'm pretty sure this is code that Juan will be working on. But this would be a good candidate for a `@todo` marker explaining your intent of what you need to do. The function has that, but it's not clear that it wasn't done. Leaving `@todo` comments is super helpful to look back on as a dev working on stuff -- but even more helpful for everyone else :)

src/resources/js/v2/tickets-loader.js
juanfra: Yes, that would be an option. I feel it could become tricky when some plugin changes the loader templates (each plugin has their own), depending on their needs. This could be abstracted in common if we make sure the main template structure doesn't change.

juanfra:
@sc0ttkclark @paulmskim thanks for the reviews. <3

This code was pushed before we decided to have views v2 for ET/ET+, hence part of this code will be ditched. The loader implementation will be handled in a different PR, once the views v2 are approved and merged into the bucket.

PR #1776, opened 2020-09-01
[ET-899] bail if valid Post object with ID of zero

cliffordp:
🎫 [ET-899] bail if valid Post object with ID of zero

No test because only edited a non-public method.

1min demo: https://share.getcloudapp.com/jkuZAD15

[ET-899]: https://moderntribe.atlassian.net/browse/ET-899

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -120,7 +120,7 @@ Currently, the following add-ons are available for Event Tickets:
 
 = [5.0.1] TBD =
 
-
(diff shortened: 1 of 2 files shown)


cliffordp:
@sc0ttkclark I'm guessing the `it_should_render_rsvp_step` failure must be from something else on G20.09

sc0ttkclark (approved):

PR #1779, opened 2020-09-01
[ET-904] Add HTML `id=` to the RSVP block.

andrasguseo:
🎫 https://moderntribe.atlassian.net/browse/ET-904

Added the missing ID to the new RSVP block.


src/Tribe/Tickets_View.php
```diff
@@ -1080,6 +1080,13 @@ public function get_rsvp_block( $post, $echo = true ) {
 		/** @var Tribe__Tickets__RSVP $rsvp */
 		$rsvp = tribe( 'tickets.rsvp' );
 
+		// Check if the call is coming from a shortcode.
+		$doing_shortcode = tribe_is_truthy( array_search( 'do_shortcode', array_column( debug_backtrace(), 'function' ) ) );
```
tr1b0t, line 1084: :warning: debug_backtrace() found. Debug code should not normally be used in production.

<sub>hash: 3bfee8bd910c9851aa9046f0f2a78aa5</sub>

juanfra: @sc0ttkclark do you have any idea on how we could be more tidy about detecting if it's doing shortcode?

sc0ttkclark: We could hook into `pre_do_shortcode_tag` to mark that we are "doing" a specific shortcode. Then hook into `do_shortcode_tag` to mark that we are done with that shortcode.

Maybe something for Tribe Common perhaps?

src/Tribe/Tickets_View.php
```diff
@@ -1080,6 +1080,13 @@ public function get_rsvp_block( $post, $echo = true ) {
...
+
+		// Get the RSVP block HTML ID.
+		$block_html_id  = 'rsvp-now';
+		$block_html_id .= $doing_shortcode ? '-' . uniqid() : '';
```
sc0ttkclark, line 1088: Could this just use the `$post_id`? Is there any possible way there would be multiple RSVP forms listed for the same post?

sc0ttkclark: We would also just need to adjust our rsvp-now links to use the post ID it's for.

sc0ttkclark (changes requested):
Questions on approach

src/Tribe/Tickets_View.php
juanfra: > Could this just use the $post_id? Is there any possible way there would be multiple RSVP forms listed for the same post? 
If we use the post ID, for the cases where they use the shortcode on an event (or CPT, where they created the RSVP) it'll be the same as having `rsvp-now`, because it'll be fixed. 

If they have a sales page with a builder, and then they have the RSVP block in different places, then they'll also have the duplicate ID if we rely on the `$post_id`.

So there are several cases where using `$post_id` would be the same as having a fixed string. I used `uniqueid()` so if it's coming from a shortcode, it always have a unique ID and that way we make sure that we don't have a duplicate ID. And have the fixed string `rsvp-now` for the cases in which we're including the block into the enabled CPTs (event, page, post, etc).

src/Tribe/Tickets_View.php
juanfra: How would that be? I'm not sure if I'm understanding. 

This is serving the purpose perfectly way. It's a perfect way of checking if it's being called from the shortcode or not. I feel it's something we should also have for the tickets block and the other shortcodes (Some people asked about it on the forums, so they can do different styling or even hiding if it's not coming from shortcode).

sc0ttkclark: Too bad we don't have this https://core.trac.wordpress.org/ticket/47616

I will lean on judgement from @lucatume on whether this is acceptable use of debug_backtrace() on production code.

src/Tribe/Tickets_View.php
sc0ttkclark: I was just thinking over whether we could find a solution outside of needing to find out if we're doing the shortcode. I think we can move forward as-is if Luca thinks the `debug_backtrace()` is okay.

src/Tribe/Tickets_View.php
lucatume: `debug_backtrace` is computationally very intensive and should not be used in production code, I would use a different approach.
@sc0ttkclark suggestion to hook on the shortcode filter to detect if we're in shortcode context or not is the way I would take. Ideally make that something that lives in `common` to be able to reuse it across plugins, the code provided in the patch linked by Scott is a good starting point for this.

lucatume (changes requested):
Do not use `debug_backtrace` and hook on the shortcode filters to know when a shortcode is being done. Create a `tribe_doing_shortcode` function in `common` and move the logic there to make this re-usable; add tests for it.

src/Tribe/Tickets_View.php
juanfra: Cool, thanks! here's an implementation for that https://github.com/moderntribe/tribe-common/pull/1437

src/Tribe/Tickets_View.php
sc0ttkclark: ✅

sc0ttkclark (dismissed):

src/Tribe/Tickets_View.php
sc0ttkclark: ✅

sc0ttkclark (approved):

PR #1800, opened 2020-10-07
[ETP-460][V2] Tickets block, and tickets utils.

juanfra:
🎫 ETP-460
🎥 not yet available.

🗒️ **Notes**

- PR against a branch where we took V1 of `tickets-block.js` as a starting point so we can see what we inherited.
- The `v2/tickets-block.js` is a cleanup of V1, with triggers in place so we can extend things for the modal in ET+. V1 for tickets block was a mix of functionalities, utils and more.
- `v2/tickets-block.js` is container based, and we are taking data from the HTML attributes, which ideally will allow us to have different tickets blocks in a same page (via shortcode, for example) and they should work.
- `v2/tickets-block.js` has 1,000+ lines of code less. I've been moving things and creating libraries as we needed.
- `v2/tickets-utils.js` is a library containing utils functions I found in V1 of tickets block. These utils were used in different files and we they were duplicated everywhere.

📆 07/10 

- ~I still have to check a couple things, but this is more or less the new library.~

📆 12/10 

- We need to review if we want to rename CSS classes as part of views v2 (My intuition says yes, it's the time to do it). This would be worked separately.

📆 15/10 

- This is ready for review.


sc0ttkclark (commented):
Looking pretty good so far

sc0ttkclark (approved):

sc0ttkclark (dismissed):

sc0ttkclark (approved):

PR #1801, opened 2020-10-08
[ETP-460] Add copies of JS for v2 purposes

sc0ttkclark:
[ETP-460]

Note: These are just copies of the existing files, there are no reviews necessary here.

[ETP-460]: https://moderntribe.atlassian.net/browse/ETP-460


sc0ttkclark (commented):
✅

PR #1814, opened 2020-10-21
[ETP-459] Update CSS class naming for V2.

juanfra:
🎫 ETP-459
🎥 https://www.loom.com/share/79eaf664ded84fc49d4867351918c578

**🗒️ NOTES**

- We're updating the class names for the views V2 to make it more BEM friendly.
- The [snapshot updates were be PRed against the head ref of this PR](https://github.com/moderntribe/event-tickets/pull/1815). So we can have smaller PRs.
- The "attendee fields" will be moved to ET+, and componentized as part of a different task.

PR #1815, opened 2020-10-22
[ETP-459] Update CSS class naming for V2 > Snapshots

juanfra:
🎫 ETP-459
🎥 https://www.loom.com/share/79eaf664ded84fc49d4867351918c578

Updating snapshots for ET. Some were needed for the tickets block v1. But the main changes are the updates of the snapshots for https://github.com/moderntribe/event-tickets/pull/1814


src/views/blocks/tickets.php
```diff
@@ -133,7 +133,19 @@
...
+	ob_start();
+	/**
+	 * Allows filtering of extra classes used on the tickets-block loader
+	 *
+	 * @since  4.11.0
+	 *
+	 * @param  array $classes The array of classes that will be filtered.
+	 */
+	$loader_classes = apply_filters( 'tribe_tickets_block_loader_classes', [ 'tribe-tickets-loader__tickets-block' ] );
+	include Tribe__Tickets__Templates::get_template_hierarchy( 'components/loader.php' );
```
tr1b0t, line 145: :warning: File is being unconditionally included; use "require" instead

<sub>hash: 06736d8d3900eee8dbe48bc0cdd83a33</sub>

src/views/blocks/tickets.php
```diff
@@ -133,7 +133,19 @@
...
+	$html = ob_get_contents();
+	ob_end_clean();
+	echo $html;
```
tr1b0t, line 148: * :warning: Expected next thing to be an escaping function (see Codex for 'Data Validation'), not '$html'
* :no_entry_sign: All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks), found '$html'.

<sub>hash: 0b431934e1ffb22c4ae66c5a9fdc2414</sub>

tr1b0t (commented):
:no_entry_sign: 1 error(s)
:warning: 2 warning(s)

src/views/blocks/tickets.php
sc0ttkclark, line 145: Why this change?

juanfra: I'm just rolling back to what we had on tickets v1. This is the template of tickets v1.

The inclusion of the loader component v2 here was merged when we didn't decide to split v1/v2 yet. So, at some point we changed this on the bucket branch, but to me it doesn't make sense to move into the loader v2 in the tickets view v1. That's why I'm reverting this. And also because it affects the snapshots (the loader in v1 includes an extra class, coming from the `tribe_tickets_block_loader_classes ` filter).

sc0ttkclark: 👍

sc0ttkclark (approved):

PR #1819, opened 2020-10-23
Revert changed class name to keep compatibility with V1 templates

rafsuntaskin:
[ET-916]

### **Before the fix:**
Markup 2020-10-23 at 23 06 52

### After the fix:

Markup 2020-10-23 at 22 50 44

[ET-916]: https://moderntribe.atlassian.net/browse/ET-916

### DIFF
--- a/src/resources/postcss/tickets-registration-page.pcss
+++ b/src/resources/postcss/tickets-registration-page.pcss
@@ -186,7 +186,7 @@ body.page-tribe-attendee-registration {
 				-ms-grid-columns: 6fr 1fr 2fr 2fr;
 				grid-template-columns: 6fr 1fr 2fr 2fr;
 
-				&.tribe-tickets__item--price-suffix {
+				&.tribe-tickets__item--price-sufix {
 					-ms-grid-columns: 4fr 3fr 2fr 2fr;
 					grid-template-columns: 4fr 3fr 2fr 2fr;
 				}


juanfra (approved):
🚀

src/resources/postcss/tickets-registration-page.pcss
```diff
@@ -186,7 +186,7 @@ body.page-tribe-attendee-registration {
 				-ms-grid-columns: 6fr 1fr 2fr 2fr;
 				grid-template-columns: 6fr 1fr 2fr 2fr;
 
-				&.tribe-tickets__item--price-suffix {
+				&.tribe-tickets__item--price-sufix {
```
sc0ttkclark, line 189: Not necessary since this is effectively the v2 version of the CSS, the v1 is a different stylesheet so it's still compatible

juanfra: Hey Scott, maybe there's a bit of confusion but the new version of the stylesheet[ lives now in ET+](https://github.com/moderntribe/event-tickets-plus/blob/bucket/ari-js-normalization/src/resources/postcss/tickets-registration-page.pcss). 

This revert, is part of the follow ups of the "Big PR". More precisely, [this comment](https://github.com/moderntribe/event-tickets/pull/1784#discussion_r490684192). So, it's basically reverting and leaving things as they were for the v1 of the AR page.

sc0ttkclark (approved):

PR #1856, opened 2020-11-06
pass context when not from global (e.g. ETP cart)

cliffordp:
[ETP-472]

[ETP-472]: https://moderntribe.atlassian.net/browse/ETP-472


src/views/v2/tickets/item/content.php
```diff
@@ -50,8 +50,16 @@
...
+	[
+		'ticket'  => $ticket,
+		'post_id' => $post_id,
+	]
+);
 
-$this->template( 'v2/tickets/item/extra' );
+$this->template( 'v2/tickets/item/content/description', $context );
+
+$this->template( 'v2/tickets/item/extra', $context );
```
sc0ttkclark, line 65: This shouldn't be necessary, if it is, we should fix it in a future sprint as part of a Common thing. This workaround shouldn't be needed/used.

src/views/v2/tickets/item/extra.php
```diff
@@ -76,13 +76,14 @@
...
 <div <?php tribe_classes( $classes ); ?>>
 
-	<?php $this->template( 'v2/tickets/item/extra/price' ); ?>
+	<?php $this->template( 'v2/tickets/item/extra/price', $context ); ?>
 
-	<?php $this->template( 'v2/tickets/item/extra/available' ); ?>
+	<?php $this->template( 'v2/tickets/item/extra/available', $context ); ?>
 
-	<?php $this->template( 'v2/tickets/item/extra/description-toggle' ); ?>
+	<?php $this->template( 'v2/tickets/item/extra/description-toggle', $context ); ?>
```
sc0ttkclark, line 87: This shouldn't be necessary, if it is, we should fix it in a future sprint as part of a Common thing. This workaround shouldn't be needed/used.

sc0ttkclark (changes requested):
src/views/v2/tickets/item/extra/available.php is good here, but the other two files changed includes code that shouldn't be necessary/used.

cliffordp:
@sc0ttkclark both ET and ETP tests passing after removing what you said (this PR's most recent commit). There have been many changes to base branch and others' PRs so maybe the reason I did it resolved itself another way 👍 

ET passing:
image

ETP passing
image

**FYI: https://github.com/moderntribe/event-tickets-plus/pull/1087#issuecomment-726312170 was prior to this most recent commit. So both that PR and this PR should be good to merge.**

src/views/v2/tickets/item/extra/available.php
```diff
@@ -53,6 +53,7 @@
```
sc0ttkclark, line 53: Needs since/version TBD update

sc0ttkclark (changes requested):
Needs since/version TBD update

PR #1870, opened 2020-11-19
[ET-648]Tweak/ Show delete confirmation dialogue on delete

rafsuntaskin:
[ET-648]

Show a confirmation dialogue while deleting Attendees from the attendee report.

Screenshot: 
Screen Capture on 2020-11-26 at 16-45-22

[ET-648]: https://moderntribe.atlassian.net/browse/ET-648


src/resources/js/tickets-attendees.js
```diff
@@ -58,6 +58,9 @@ var tribe_event_tickets_attendees = tribe_event_tickets_attendees || {};
 			}
 		});
 
+		$( '.trash' ).click( function ( e ) {
+			return confirm( 'Please confirm you would like to delete this attendee.' );
```
sc0ttkclark, line 62: To be translatable

sc0ttkclark (changes requested):
I know it's WIP, just reviewing for now so I'm ahead

src/Tribe/Attendees.php
```diff
@@ -320,9 +320,19 @@ public function enqueue_assets( $hook ) {
 				'check'     => wp_create_nonce( 'move_tickets' ),
 				'TB_iframe' => 'true',
 			) ),
+			'confirmation'  => esc_html__( 'Please confirm that you would like to delete this attendee.', 'event-tickets' ),
```
tr1b0t, line 323: :warning: Array double arrow not aligned correctly; expected 4 space(s) between "'confirmation'" and double arrow, but found 2.

<sub>hash: 52ed871a67a9d669e8b2155fd81d1eb0</sub>

tr1b0t (commented):
:warning: 1 warning(s)

src/resources/js/tickets-attendees.js
```diff
@@ -58,6 +58,9 @@ var tribe_event_tickets_attendees = tribe_event_tickets_attendees || {};
 			}
 		});
 
+		$( '.trash' ).click( function ( e ) {
+			return confirm( Attendees.confirmation );
+		});
```
paulmskim, line 63: Is the `.trash` element the link? If so, shouldn't it be `preventDefault()` first and check that the confirmation dialogue was confirmed before allowing the user to continue? What happens if you click on the Trash button but don't touch the confirmation dialogue for 10 seconds?

sc0ttkclark (changes requested):
JS question from Paul and needs changelog

src/Tribe/Attendees.php
```diff
@@ -307,22 +307,32 @@ public function enqueue_assets( $hook ) {
...
+		$config_data = [
 			'nonce'           => wp_create_nonce( 'email-attendee-list' ),
 			'required'        => esc_html__( 'You need to select a user or type a valid email address', 'event-tickets' ),
 			'sending'         => esc_html__( 'Sending...', 'event-tickets' ),
 			'ajaxurl'         => admin_url( 'admin-ajax.php' ),
 			'checkin_nonce'   => wp_create_nonce( 'checkin' ),
 			'uncheckin_nonce' => wp_create_nonce( 'uncheckin' ),
 			'cannot_move'     => esc_html__( 'You must first select one or more tickets before you can move them!', 'event-tickets' ),
-			'move_url'        => add_query_arg( array(
+			'move_url'        => add_query_arg( [
```
tr1b0t, line 318: :no_entry_sign: Opening parenthesis of a multi-line function call must be the last content on the line

<sub>hash: b639e2122be2e115f15fadf86518f531</sub>

src/Tribe/Attendees.php
```diff
@@ -307,22 +307,32 @@ public function enqueue_assets( $hook ) {
...
 				'dialog'    => Tribe__Tickets__Main::instance()->move_tickets()->dialog_name(),
 				'check'     => wp_create_nonce( 'move_tickets' ),
 				'TB_iframe' => 'true',
-			) ),
-		);
+			] ),
```
tr1b0t, line 322: :no_entry_sign: Closing parenthesis of a multi-line function call must be on a line by itself

<sub>hash: 6470831bd29e2b2f70ee457b59d881eb</sub>

tr1b0t (commented):
:no_entry_sign: 2 error(s)

src/resources/js/tickets-attendees.js
rafsuntaskin: .trash is the span element around the link: http://p.tri.be/Ax9C1C

We don't need `preventDefault()` here as we are using the native JS confirm that works same without using that. 

It won't continue if you don't click 'Ok', clicking cancel just cancels the action and the user is not redirected.

I have tried the 10-second delay and it won't do anything until an action button is pressed.

paulmskim: So from the admin side, it may not be as important as the FE, but we still want to consider users that are not mouse users. So this would work if the user uses their mouse to click, but if the user were a keyboard user, this would not work.

Let's target the link instead to capture all the possible ways of interacting with the trash link.

rafsuntaskin: Thanks, @paulmskim. I have updated the target to link as you suggested. 

Before that, I tried to delete using the keyboard and it worked fine for me with Chrome using `esc` to cancel and `Enter` to confirm: http://p.tri.be/eeD8nM

I am not sure if you were referring to a similar use case like mine.

sc0ttkclark (approved):

PR #1900, opened 2020-12-10
[ET-947] Resolve CSS display conflict with Tribe Dependency for Warning notice

sc0ttkclark:
[ET-947]

The warning notice has a class that uses `display:flex` but it gets a bit messy when involving Tribe dependency and AJAX loading too. This resolves those issues.

Dev screencast: https://share.skc.dev/OAuWgl20

[ET-947]: https://moderntribe.atlassian.net/browse/ET-947

### DIFF
--- a/src/resources/postcss/tickets-admin.pcss
+++ b/src/resources/postcss/tickets-admin.pcss
@@ -626,6 +626,14 @@ p.description {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


rafsuntaskin (approved):
Looks great. Nice catch 🥇

bordoni (approved):

PR #1903, opened 2020-12-11
Set body class for when we're on the custom page with the shortcode for AR.

alliuca:
**Description:** Adds an additional class to the body to target the custom AR page (with shortcode)

**Ticket:** https://moderntribe.atlassian.net/browse/TEC-3700


juanfra (approved):
Code is looking good to me! I assume that you'll be adding readme entries on the bucket branch? It'd be nice to note that we're adding this body class, somewhere.

alliuca:
thanks @juanfra yes I'll deal with changelog entries in the bucket branch

PR #1912, opened 2020-12-22
Fix / TEC-3564 datepicker text wrapping

alliuca:
Description: Just updates the common hash for https://github.com/moderntribe/tribe-common/pull/1524

Ticket: https://moderntribe.atlassian.net/browse/TEC-3564


paulmskim (dismissed):

sc0ttkclark (changes requested):
Common conflict 😿

alliuca:
hey @sc0ttkclark the build currently fails because of the issue with the docker pull limit, but tests seem to be failing locally (using `tric run`). Here's an example of what I get for the integration suite

Screen Shot 2021-01-12 at 9 59 55 AM

how can we solve that?

PR #1945, opened 2021-02-10
Fix acceptance tests on GHA

juanfra:
🎫 N/A
📷 N/A


.github/workflows/tests-php.yml
```diff
@@ -86,6 +86,8 @@ jobs:
           ${TRIC_BIN} debug on
           ${TRIC_BIN} info
           ${TRIC_BIN} config
+      - name: Start the Chrome container
+        run: ${TRIC_BIN} up chrome
```
sc0ttkclark, line 90: I wonder if we can only run this if the matrix suite is acceptance

.github/workflows/tests-php.yml
```diff
@@ -157,7 +159,30 @@ jobs:
...
       - name: Init the WordPress container
-        run: ${TRIC_BIN} up wordpress
+        run: |
+          ${TRIC_BIN} up wordpress
+          ${TRIC_BIN} site-cli core version
+      # ------------------------------------------------------------------------------
+      # Install and activate TwentyTwenty
+      # ------------------------------------------------------------------------------
+      - name: Install and activate TwentyTwenty
+        run:  ${TRIC_BIN} site-cli theme install twentytwenty --activate
```
sc0ttkclark, line 173: I wonder if we can only run this if the matrix suite is acceptance

sc0ttkclark (approved):

sc0ttkclark (approved):

PR #1966, opened 2021-03-09
[ET-951] Move duration field out of advanced section on ET blocks

juanfra:
🎫 [ET-951]
🎥 https://www.loom.com/share/4d3318a4313b4deba2de641e39784d76

[ET-951]: https://theeventscalendar.atlassian.net/browse/ET-951


sc0ttkclark (approved):

src/modules/blocks/rsvp/advanced-options/template.js
```diff
@@ -33,10 +33,6 @@ class RSVPAdvancedOptions extends PureComponent {
 
 	getContent = () => (
```
paulmskim, line 34: Are we using this AdvancedOptions component anywhere? If not, should we delete it?

juanfra: Kept it because of [this comment from Gustavo](https://theeventscalendar.atlassian.net/browse/ET-951?focusedCommentId=91917), and in case we wanted to bring it back again at some point. But I can totally delete it.

paulmskim: OK fair. Either way, we have the history of the component so we can easily reimplement it if we remove it.

paulmskim (approved):

sc0ttkclark (changes requested):
Needs changelog

sc0ttkclark (dismissed):

PR #1971, opened 2021-03-16
[ET-981] Add confirmation dialog for bulk deletion of attendees.

rafsuntaskin:
[ET-981]

Screencast: https://d.pr/v/kq1v2m

[ET-981]: https://theeventscalendar.atlassian.net/browse/ET-981

### DIFF
--- a/src/admin-views/attendees.php
+++ b/src/admin-views/attendees.php
@@ -162,7 +162,7 @@ $show_title = apply_filters( 'tribe_tickets_attendees_show_title', is_admin(), $
 	</div>
 	<?php do_action( 'tribe_events_tickets_attendees_event_summary_table_after', $event_id ); ?>
 
-	<form id="attendees-admin-form" class="topics-filter attendees-admin-form" method="post">
+	<form id="event-tickets__attendees-admin-form" class="topics-filter event-tickets__attendees-admin-form" method="post">
 		<input type="hidden" name="<?php echo esc_attr( is_admin() ? 'page' : 'tribe[page]' ); ?>" value="<?php echo esc_attr( isset( $_GET['page'] ) ? $_GET['page'] : '' ); ?>" />
 		<input type="hidden" name="<?php echo esc_attr( is_admin() ? 'event_id' : 'tribe[event_id]' ); ?>" id="event_id" value="<?php echo esc_attr( $event_id ); ?>" />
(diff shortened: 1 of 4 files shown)


src/resources/js/tickets-attendees.js
```diff
@@ -63,6 +63,21 @@ var tribe_event_tickets_attendees = tribe_event_tickets_attendees || {};
...
 
+		$( '.attendees-admin-form' ).on( 'submit', function ( e ) {
+
+			// If not the delete action, return.
+			if ( 'delete_attendee' !== $( '#bulk-action-selector-top' ).val() ) {
+				return;
+			}
+
+			// If no attendee was selected, bail out.
+			if ( ! $( this ).serialize().includes( '&attendee' ) )  {
```
paulmskim, line 74: Do we still support IE11? If we do, the `.includes()` function will not work on IE11. Instead, we'll want to use `.indexOf()`.

If we don't care about IE11, then disregard this comment ;)

paulmskim (commented):
a couple suggestions

src/resources/js/tickets-attendees.js
juanfra: We're no longer supporting IE11 https://docs.theeventscalendar.com/wiki/view/Crossbrowser_Testing

juanfra (approved):

juanfra:
Let's address Paul's suggestion and we're good to go 🎉

src/admin-views/attendees.php
```diff
@@ -162,7 +162,7 @@
 	</div>
 	<?php do_action( 'tribe_events_tickets_attendees_event_summary_table_after', $event_id ); ?>
 
-	<form id="topics-filter" class="topics-filter" method="post">
+	<form id="attendees-admin-form" class="topics-filter attendees-admin-form" method="post">
```
sc0ttkclark, line 165: If we're introducing ids/classes, we should namespace them at minimal

src/resources/js/tickets-attendees.js
```diff
@@ -63,6 +63,21 @@ var tribe_event_tickets_attendees = tribe_event_tickets_attendees || {};
 			return confirm( Attendees.confirmation );
 		});
 
+		$( '.attendees-admin-form' ).on( 'submit', function ( e ) {
```
sc0ttkclark, line 66: Namespace class

sc0ttkclark (changes requested):
Needs class/id namespacing

sc0ttkclark (approved):

PR #1980, opened 2021-04-01
[ET-1049] Moving Ticket banner to notice class

rafsuntaskin:
### 🎫 Ticket

[ET-1049]

### 🗒️ Description

Add Collapse & expand for Settings > Tickets > Help Banner by making use of Notices Class.

### 🎥 Artifacts
📷 screenshot(s):
Screen Shot 2021-04-01 at 10 03 40 AM

### ✔️ Checklist
- [x] My code is tested.

[ET-1049]: https://theeventscalendar.atlassian.net/browse/ET-1049

### DIFF
--- a/src/Tribe/Admin/Notices.php
+++ b/src/Tribe/Admin/Notices.php
@@ -35,6 +35,7 @@ class Tribe__Tickets__Admin__Notices {
 		$this->maybe_display_rsvp_new_views_options_notice();
 		$this->maybe_display_classic_editor_ecp_recurring_tickets_notice();
 		$this->maybe_display_plus_commerce_notice();
+		$this->maybe_display_ticket_settings_help_banner();
 	}
 
 	/**
@@ -236,5 +237,37 @@ class Tribe__Tickets__Admin__Notices {
 
 			tribe_notice( "event-tickets-plus-missing-{$provider}-support", $message, 'dismiss=1&type=warning' );
 		}
+	}
+
+	/**
+	 * Display Settings help banner on Event Ticket's Settings.
+	 *
+	 * @since TBD
+	 */
+	public function maybe_display_ticket_settings_help_banner() {
+
+		// Bail if we aren't in Events > Settings.
+		if ( 'tribe-common' !== tribe_get_request_var( 'page' ) ) {
+			return;
+		}
+
+		// Bail if not at wp-admin > Events > Settings > Tickets tab.
+		if ( 'event-tickets' !== tribe_get_request_var( 'tab' ) ) {
+			return;
+		}
+
+		/** @var \Tribe\Tickets\Admin\Settings\Service_Provider $settings */
+		$settings = tribe( 'tickets.admin.settings' );
+
+		echo $settings->render_settings_banner();
+
+//		tribe_notice(
+//			'test-99',
+//			"sample-notice",
+//			[
+//				'dismiss' => true,
+//				'type'    => 'success',
+//			]
+//		);
 	}
 }
--- a/src/Tribe/Admin/Settings/Service_Provider.php
+++ b/src/Tribe/Admin/Settings/Service_Provider.php
@@ -33,7 +33,7 @@ class Service_Provider extends tad_DI52_ServiceProvider {
 			return;
 		}
 
-		add_action( 'tribe_settings_before_content_tab_event-tickets', [ $this, 'render_settings_banner' ] );
+//		add_action( 'admin_notices', [ $this, 'render_settings_banner' ], 99 );
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +238,36 @@ public function maybe_display_plus_commerce_notice() {
...
+		if ( 'event-tickets' !== tribe_get_request_var( 'tab' ) ) {
+			return;
+		}
+
+		/** @var \Tribe\Tickets\Admin\Settings\Service_Provider $settings */
+		$settings = tribe( 'tickets.admin.settings' );
+
+		echo $settings->render_settings_banner();
+
+//		tribe_notice(
```
rafsuntaskin, line 264: Using the `tribe_notice` method will require us to change the template a lot and adjust CSS again to make it compatible.

src/Tribe/Admin/Settings/Service_Provider.php
```diff
@@ -33,7 +33,7 @@ protected function hooks() {
 			return;
 		}
 
-		add_action( 'tribe_settings_before_content_tab_event-tickets', [ $this, 'render_settings_banner' ] );
+//		add_action( 'admin_notices', [ $this, 'render_settings_banner' ], 99 );
```
rafsuntaskin, line 36: If we can't make use of `tribe_notice` does it make more sense of keeping the code within this class and making use of `admin_notices` ?

Also, we can control the position of notice by using higher priority here which is not possible with `tribe_notice`

src/Tribe/Admin/Notices.php
juanfra: With the relocation using this method is looking broken on load: https://www.loom.com/share/a20208ce334c45f793e7e0ee3b09a75c

If you check it's re-positioned and it's not a great experience.

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +238,36 @@ public function maybe_display_plus_commerce_notice() {
 			tribe_notice( "event-tickets-plus-missing-{$provider}-support", $message, 'dismiss=1&type=warning' );
 		}
 	}
+
+	/**
+	 * Display Settings help banner on Event Ticket's Settings.
+	 *
+	 * @since TBD
+	 */
+	public function maybe_display_ticket_settings_help_banner() {
```
juanfra, line 247: We should use tickets in plural so it's more obvious that's the tickets plugin settings and not a particular ticket.

```suggestion
	public function maybe_display_tickets_settings_help_banner() {
```

src/Tribe/Admin/Notices.php
```diff
@@ -35,6 +35,7 @@ public function maybe_display_notices() {
 		$this->maybe_display_rsvp_new_views_options_notice();
 		$this->maybe_display_classic_editor_ecp_recurring_tickets_notice();
 		$this->maybe_display_plus_commerce_notice();
+		$this->maybe_display_ticket_settings_help_banner();
```
juanfra, line 38:
```suggestion
		$this->maybe_display_tickets_settings_help_banner();
```

src/Tribe/Admin/Settings/Service_Provider.php
```diff
@@ -105,9 +105,9 @@ public function render_settings_banner() {
 			'etp_resource_links' => $etp_resource_links,
 		];
 
-		/** @var Tribe__Tickets__Admin__Views $admin_views */
+		/** @var \Tribe__Tickets__Admin__Views $admin_views */
 		$admin_views = tribe( 'tickets.admin.views' );
 
-		return $admin_views->template( 'settings/getting-started', $context );
+		return $admin_views->template( 'settings/getting-started', $context, false );
```
juanfra, line 111: Here you could do something like this:

```$html = $settings->render_settings_banner();

		tribe_notice(
			'event-tickets-settings-getting-started',
			$html,
			[
				'dismiss' => true,
				'type'    => 'info',
			]
		);```

File: src/admin-views/settings/getting-started.php [hunk 6]
```diff
@@ -14,7 +14,7 @@
 	: __( 'Thank you for using Event Tickets! We recommend looking through the settings below so that you can fine tune your specific ticketing needs. Here are some resources that can help.', 'event-tickets' );
 
 ?>
-<div class="event-tickets__admin-banner">
+<div class="event-tickets__admin-banner notice is-dismissible">
```
juanfra, line 17: And here remove the div wrapper. Then you can apply the particular padding and styles to match the design to this class `.tribe-notice-event-tickets-settings-getting-started`

juanfra (changes requested):
I left some comments. Please, also that we need to do is add the `?utm_source=event-tickets-settings-getting-started` parameter to the links that are part of this notice to eventually track if people are clicking these links or not.

PR #2018, opened 2021-06-09
Update package file to use ltest task master

mitogh:
### 🗒️ Description

This PR makes sure the `pacakge.json` file contains the latest version of ticket master in order to build the assets as expected and format the file using the formatted from NVM

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,10 +10,7 @@
   "_glotPressUrl": "https://translate.wordpress.org",
   "_glotPressSlug": "wp-plugins/event-tickets/stable",
   "_glotPressFileFormat": "%textdomain%-%wp_locale%.%format%",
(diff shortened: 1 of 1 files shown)


paulmskim (changes requested):
Is there a reason why you're making this change now? Do you need the product taskmaster for something specific?

There are a few things we need to do here to get this to work properly:
1. remove the old product taskmaster from the package.json
2. update the package-lock so it contains the latest product taskmaster
3. update any uses of product taskmaster to use the scoped version

Currently, it's still a work in progress to get the latest product taskmaster into the main plugins.

mitogh:
The main reason is: 

Running `npm ci` creates this file when building the plugin for the first time,

PR #2019, opened 2021-06-09
Prevent PHP warnings on tribe endpoints.

mitogh:
### 🗒️ Description

Make sure empty strings are not compared with `stripos` which can
trigger a PHP warning that affect the JSON generated for the
rest endpoints creating an invalid JSON response.

If the value is empty instead just assume the result should be the
default value.

Requires: https://github.com/the-events-calendar/event-tickets/pull/2018


paulmskim (changes requested):
Can we remove the product taskmaster work from here and address the PHP warnings separately?

mitogh:
Thanks PR has been isolated.

PR #2055, opened 2021-08-17
[ET-1162] Remove rsvp kitchen sink

codingmusician:
### 🎫 Ticket

[ET-1162] 

### 🗒️ Description

You can view the current kitchen sink by creating an RSVP block within an event (in V2 views) and then adding the following code snippet to your theme's functions.php file:

`define('TRIBE_TICKETS_RSVP_NEW_VIEWS_PREVIEW', true);`

We are removing this functionality and we are removing the template files so that there is no confusion among users that want to customize their template files.

### 🎥 Artifacts 

https://youtu.be/w8Ck8RF1DR8

### ✔️ Checklist
- [x] My code is tested. 

[ET-1162]: https://theeventscalendar.atlassian.net/browse/ET-1162


readme.txt
```diff
@@ -178,6 +178,10 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 == Changelog ==
 
+= [TBD] TBD =
+
+* Fix - Removed RSVP kitchen sink template and functionality. [ET-1162]
```
rafsuntaskin, line 183: Not sure if `Kitchen sink` is something that we would want to include in the public readme. 

Maybe we can say something like:

```Removed unused RSVP V2 preview templates. ```

juanfra: I'd avoid having `unused`, but that's a good point. We can use "RSVP V2 preview templates" so that's more understandable what we're doing.

codingmusician: Thanks @rafsuntaskin and @juanfra !

juanfra (commented):
Nice @codingmusician! You need to fix a conflict on the readme and this one is good to go!

rafsuntaskin (approved):
Looks good, solving the conflict is the last thing remaining.

rafsuntaskin (approved):
🚀

PR #2067, opened 2021-08-27
Attendee generation is happening + Inventory calculation

bordoni:
Things I've handled in this Particular PR:
- Attendee Generation without Attendee data, just to test Stock/Inventory
- Added more data to the Order Meta, specially dates when an order was moved from one status to another and when certain flags were triggered.

### DIFF
--- a/src/Tickets/Commerce/Flag_Actions/Flag_Action_Abstract.php
+++ b/src/Tickets/Commerce/Flag_Actions/Flag_Action_Abstract.php
@@ -43,8 +43,19 @@ abstract class Flag_Action_Abstract implements Flag_Action_Interface {
 	/**
 	 * {@inheritDoc}
 	 */
-	public function get_flags() {
-		return $this->flags;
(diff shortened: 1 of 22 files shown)


src/Tickets/Commerce/Flag_Actions/Flag_Action_Abstract.php
```diff
@@ -65,7 +97,7 @@ public function get_post_types() {
 	 * {@inheritDoc}
 	 */
 	public function should_trigger( Status_Interface $new_status, $old_status, $post ) {
-		if ( ! $this->has_flags( $new_status ) ) {
+		if ( ! $this->has_flags( $new_status, $post ) ) {
```
paulmskim, line 100: not sure if you've addressed this yet, but it looks like this function call doesn't match the signature.

bordoni: I think it got fixed in another PR later.

PR #2070, opened 2021-08-31
[ET-1179] Removed style causing overlapping text

redscar:
### 🎫 Ticket

[GTRIA-655]  

### 🗒️ Description
Removed extra CSS that were causing text to overlap when the description was show in the AR modal
`
		@media (--viewport-medium) {
			-ms-grid-row: 3;
			grid-row: 3;
		}

### 🎥 Artifacts 

image
image

--- Screenshot with Ticket Blocks ---
image
image

### ✔️ Checklist
- [x] My code is tested. 

[GTRIA-655]: https://theeventscalendar.atlassian.net/browse/GTRIA-655


readme.txt
```diff
@@ -178,6 +178,10 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 == Changelog ==
 
+= [TBD] TBD =
+
+* Fix - Fixed text overlapping description in the ticket AR modal.
```
rafsuntaskin, line 183:
```suggestion
* Fix - Fixed text overlapping description in the ticket AR modal. [ET-1179]
```

rafsuntaskin (changes requested):
Missing ticket entry for readme. Nice work 🥇

bordoni (dismissed):

juanfra (commented):
The code we're removing is related to the tickets block, for the responsive styles particularly. Have we checked the effect of this on the responsive view for the tickets block? 

I'm not sure that by removing this we're gonna be removing something that we don't want to remove from the tickets block. In any case, if we need a special treatment for the `tribe-tickets-attendees-list-optout--wrapper` inside the modal, we could be adding the exception on the ET+ stylesheet.

juanfra:
@redscar thanks for checking into that! What I mean with the tickets block is this: https://cloudup.com/cAWraJz__Hq - We're touching code that's specifically for that component, to fix something that we identified on ET+ (the modal).

So, what I was saying is that we should be checking the tickets block with these changes because possibly the way that the checkbox looks there could be affected without that CSS. It's a possibility, but I'm definitely not certain. If the changes we're adding here affect the tickets block we should roll back and add some CSS exception for the modal in ET+.

redscar:
@juanfra Thanks for clarifying.  I just checked again locally and everything looks good.

image


image

juanfra:
@redscar thanks for sharing those! All good then

juanfra (approved):
⚡

PR #2079, opened 2021-09-08
[ET-1186] Enhancement - Add loader/spinner on Ticket Commerce Checkout page

rafsuntaskin:
### 🎫 Ticket

[ET-1186] 

### 🗒️ Description

* Added a loader component on the Ticket Commerce JS.
* Created custom events that can be fired from the Gateway scripts to show/hide the loader.
* Added loader triggers to PayPal checkout JS to show and hide loader view.
* Load PayPal script in async.
* Add Ticket Commerce JS as a dependency for PayPal checkout JS.

### 🎥 Artifacts 
🎥 screencast: 
 * ~~Take 1 : https://d.pr/v/92H5qx~~
 * Take 2 : https://d.pr/v/vQPThq ( latest with JS trigger )

### ✔️ Checklist
- [ ] ~~I've included a changelog entry both in readme.txt and changelog.txt files.~~ Not required 
- [x] My code is tested. 

[ET-1186]: https://theeventscalendar.atlassian.net/browse/ET-1186


src/resources/js/v2/tickets-commerce.js
```diff
@@ -136,9 +148,42 @@ tribe.tickets.commerce = {};
...
+	/**
+	 * Hide the loader/spinner.
+	 *
+	 * @since TBD
+	 */
+	obj.loaderHide = function() {
+		const loadClass = obj.selectors.loader;
+		$( loadClass ).addClass( obj.selectors.hidden );
+	};
+
```
juanfra, line 186: We already have the implementation of `tickets-loader.js` that we can make use of. Here's the library: https://github.com/the-events-calendar/event-tickets/blob/master/src/resources/js/v2/tickets-loader.js

And here's an example of how this can be used from other places: https://github.com/the-events-calendar/event-tickets/blob/master/src/resources/js/v2/rsvp-manager.js#L178

We need to be sure that the tickets commerce script uses tickets loader as a dependency and then we can make use of the library instead of re-implementing the loader.

src/resources/js/v2/tickets-commerce.js
```diff
@@ -30,6 +30,16 @@ tribe.tickets.commerce = {};
...
 	const $document = $( document );
 
+	/**
+	 * Ticket Commerce custom Events.
+	 *
+	 * @since TBD
+	 */
+	obj.customEvents = {
+		showLoader : 'tribeTicketsCommerceCheckout.showLoader',
+		hideLoader : 'tribeTicketsCommerceCheckout.hideLoader',
```
bordoni, line 40: Let's make sure we use `tecTicketsCommerce` no need for `Checkout` there too.
For jQuery events, the namespace comes after the event name. So it would be :
* `'showLoader.tecTicketsCommerce'`

src/resources/js/commerce/gateway/paypal/checkout.js
```diff
@@ -350,6 +350,20 @@ tribe.tickets.commerce.gateway.paypal.checkout = {};
...
+	 *
+	 * @since TBD
+	 *
+	 * @return {void}
+	 */
+	obj.setupLoader = function() {
+		// Hide loader when Paypal buttons are added.
+		$document.on( 'DOMNodeInserted', obj.selectors.buttons, function() {
+			$document.trigger( tribe.tickets.commerce.customEvents.hideLoader );
+		});
```
bordoni, line 364:
```suggestion
		$document.on( 'DOMNodeInserted', obj.selectors.buttons, function() {
			$document.trigger( tribe.tickets.commerce.customEvents.hideLoader );
		} );
```

src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
```diff
@@ -53,6 +53,7 @@ public function setup_template_vars() {
 			'must_login'       => ! is_user_logged_in() && tribe( Module::class )->login_required(),
 			'login_url'        => tribe( Checkout::class )->get_login_url(),
 			'registration_url' => tribe( Checkout::class )->get_registration_url(),
+			'et_template'      => tribe( 'tickets.editor.template' ),
```
bordoni, line 56: No need to pass it down on this case, use it directly on the template.

bordoni (changes requested):

src/views/v2/components/loader/loader.php
```diff
@@ -14,14 +14,18 @@
...
  */
 
-$classes = $this->get( 'classes' ) ?: [];
+$classes        = $this->get( 'classes' ) ?: [];
+$visible        = tribe_is_truthy( $this->get( 'visible' ) ) || false;
 
 $spinner_classes = [
 	'tribe-tickets-loader__dots',
 	'tribe-common-c-loader',
-	'tribe-common-a11y-hidden',
```
juanfra, line 22: Why are we modifying the default behavior of the loader? The norm is that by default it's hidden. This change can potentially cause issues in other places where we use the loader. 

I feel that if we actually need this to be visible by default in a particular scenario we can do it with the JS method directly.

src/views/v2/components/loader/loader.php
```diff
@@ -14,14 +14,18 @@
...
 ];
 
+if ( ! $visible ) {
+	$spinner_classes[] = 'tribe-common-a11y-hidden';
```
juanfra, line 26: We could avoid the if by doing something like this:

```
$spinner_classes = [
	'tribe-common-a11y-hidden => empty( $visible ),
]
```

(as part of the definition of the `$spinner_classes` above).

src/views/v2/components/loader/loader.php
```diff
@@ -14,14 +14,18 @@
  * @version 5.0.4
  */
 
-$classes = $this->get( 'classes' ) ?: [];
+$classes        = $this->get( 'classes' ) ?: [];
+$visible        = tribe_is_truthy( $this->get( 'visible' ) ) || false;
```
juanfra, line 18:
```suggestion
$visible        = tribe_is_truthy( $this->get( 'visible' ) );
```
This method already returns a bool, I think there's no need to have the OR false.

src/resources/js/v2/tickets-commerce.js
```diff
@@ -170,8 +168,7 @@ tribe.tickets.commerce = {};
 	 * @since TBD
 	 */
 	obj.loaderShow = function() {
-		const loadClass = obj.selectors.loader;
-		$( loadClass ).removeClass( obj.selectors.hidden );
+		tribe.tickets.loader.show( $( obj.selectors.checkoutContainer ) );
```
juanfra, line 171: Nice use of the library! 🎉

src/resources/js/commerce/gateway/paypal/checkout.js
```diff
@@ -358,10 +358,6 @@ tribe.tickets.commerce.gateway.paypal.checkout = {};
 	 * @return {void}
 	 */
 	obj.setupLoader = function() {
-
-		// Show loader.
-		$document.trigger( tribe.tickets.commerce.customEvents.showLoader );
```
juanfra: If we keep the loader template as it was, we can do a `show` here. That's how we do it for the other places where we make use of the loader.

juanfra (commented):
Great use of the loader library! A couple of questions regarding the modification of a lib that can have effects on other places.

src/resources/js/commerce/gateway/paypal/checkout.js
rafsuntaskin: Yes, that is what I thought until I found that we are not getting into the `ready` state while the `PayPal Checkout` script is getting loaded. So, it keeps showing blank and by the time we trigger the loader to show the buttons are already rendered and the loader doesn't help at all.

So, I have modified the loader to be visible at page render immediately while the JS scripts get loaded.

src/views/v2/components/loader/loader.php
rafsuntaskin: We are not getting into the ready state while the PayPal Checkout script is getting loaded. So, it keeps showing blank and by the time we trigger the loader to show, the Paypal buttons are already rendered and the loader doesn't help at all.

So, I have modified the loader to be visible at page render immediately while the JS scripts get loaded.

The loader template should be unaffected as we are changing the classes only if the `visible` variable is set.

bordoni: I would still avoid changing the default template and make the changes by overwriting this from the Usage of the component

src/resources/js/v2/tickets-commerce.js
```diff
@@ -136,9 +146,40 @@ tribe.tickets.commerce = {};
...
+	};
+
+	/**
+	 * Hide the loader/spinner.
+	 *
+	 * @since TBD
+	 */
+	obj.loaderHide = function() {
+		tribe.tickets.loader.hide( $( obj.selectors.checkoutContainer ) );
+	};
```
paulmskim, line 181: let's move these above so that we define first and use them after. `bindLoaderEvents` uses `loaderShow` and `loaderHide`, so those should be defined first, and `bindCheckoutEvents` uses `bindLoaderEvents`, so we should move `bindLoaderEvents` above.

src/views/v2/components/loader/loader.php
paulmskim: This change shouldn't affect the other uses of the loader already? Since we are not passing in a `visible` property to the other loaders, it should still print the hidden class.

However, I do think that adding the hidden class by default is not the right call here, we should allow users of this template to pass it in on their own (there is now no way to show by default). Is there a way we can change that in the places we need to? Or does that touch too many pieces?

juanfra: > Is there a way we can change that in the places we need to? Or does that touch too many pieces?

At least the tickets block, RSVP, AR page, AR modal, Manual Attendees.

With regards to the default behavior (hidden), ET is following [what TEC does](https://github.com/the-events-calendar/the-events-calendar/blob/master/src/views/v2/components/loader.php#L17), so that's kind of in-line.

paulmskim: Yeah, not the best implementation :(, we hadn't thought about that.

paulmskim: The other thing we could do is in the JS, on ready, we can remove the `hidden` class.

rafsuntaskin:
I have updated the scripts to show the loader using JS now. Opted out from the default HTML show.

* Load PayPal script in async.
* Add Ticket Commerce JS as a dependency for PayPal checkout JS.

Once I get approval for the new changes, then I will revert the changes in the `Loader` component that was made for showing loader immediately in HTML.

src/resources/js/v2/tickets-commerce.js
```diff
@@ -131,12 +169,15 @@ tribe.tickets.commerce = {};
 	 * @return {void}
 	 */
 	obj.bindCheckoutEvents = function( $container ) {
-		$document.trigger( 'beforeSetup.tribeTicketsCommerceCheckout', [ $container ] );
+		$document.trigger( 'beforeSetup.TicketsCommerceCheckout', [ $container ] );
```
bordoni, line 172:
```suggestion
		$document.trigger( 'beforeSetup.tecTicketsCommerce', [ $container ] );
```

src/resources/js/v2/tickets-commerce.js
```diff
@@ -131,12 +169,15 @@ tribe.tickets.commerce = {};
...
 
 		// Bind container based events.
 		obj.bindCheckoutItemDescriptionToggle( $container );
 
-		$document.trigger( 'afterSetup.tribeTicketsCommerceCheckout', [ $container ] );
+		// Bind loader visibility.
+		obj.bindLoaderEvents();
+
+		$document.trigger( 'afterSetup.TicketsCommerceCheckout', [ $container ] );
```
bordoni, line 180:
```suggestion
		$document.trigger( 'afterSetup.tecTicketsCommerce', [ $container ] );
```

bordoni (changes requested):

src/resources/js/v2/tickets-commerce.js
bordoni: When we change these particular filters we modify `commerce/gateway/paypal/checkout.js` that make usage of these.

src/resources/js/v2/tickets-commerce.js
bordoni: When we change these particular filters we modify `commerce/gateway/paypal/checkout.js` that make usage of these.

src/resources/js/commerce/gateway/paypal/checkout.js
```diff
@@ -358,9 +374,13 @@ tribe.tickets.commerce.gateway.paypal.checkout = {};
 	 * @return {void}
 	 */
 	obj.ready = function () {
-		$document.on( 'afterSetup.tribeTicketsCommerceCheckout', obj.setupButtons );
+		obj.setupLoader();
+
+		window.onload = (event) => {
+			obj.setupButtons( event, $( tribe.tickets.commerce.selectors.checkoutContainer ) );
+		};
```
paulmskim, line 381: curious why this needs to be done on `onload` rather than just in the `ready` function?

paulmskim, line 381: Should this check whether the `checkoutContainer` exists before running any of this code? Saves us from running code that doesn't need to be run in the admin.

src/resources/js/commerce/gateway/paypal/checkout.js
```diff
@@ -350,6 +350,22 @@ tribe.tickets.commerce.gateway.paypal.checkout = {};
...
+	 *
+	 * @return {void}
+	 */
+	obj.setupLoader = function() {
+		$document.trigger( tribe.tickets.commerce.customEvents.showLoader );
+
+		// Hide loader when Paypal buttons are added.
+		$document.on( 'DOMNodeInserted', obj.selectors.buttons, function() {
+			$document.trigger( tribe.tickets.commerce.customEvents.hideLoader );
+		} );
```
paulmskim, line 366: make the function a separate named function on the `obj` and remove the `DOMNodeInserted` event listener after triggering the `hideLoader` event.

src/resources/js/commerce/gateway/paypal/checkout.js
rafsuntaskin: We are loading the external PayPal script using the template: https://github.com/the-events-calendar/event-tickets/blob/1a756c95dedbebcdb72bbd0b1e81302d7276382c/src/views/v2/commerce/gateway/paypal/checkout-script.php#L29-L28

So, the `ready` method triggers before PayPal is available causing an error.

rafsuntaskin: the `tribe.tickets.commerce` JS is a dependency for this script. So, that gets loaded before this `ready` is fired.

We are only using this script in FE, so I am not sure where it can run in admin.

paulmskim: gotcha. could we maybe target that specific element for the `onload` event instead of the entire window?

src/views/v2/components/loader/loader.php
juanfra: Yep, that's what I suggested initially, and here's what Rafsun found: https://github.com/the-events-calendar/event-tickets/pull/2079#discussion_r705233563

bordoni (approved):

src/resources/js/v2/tickets-commerce.js
```diff
@@ -131,12 +169,15 @@ tribe.tickets.commerce = {};
...
+		$document.trigger( 'beforeSetup.tecTicketsCommerce', [ $container ] );
 
 		// Bind container based events.
 		obj.bindCheckoutItemDescriptionToggle( $container );
 
-		$document.trigger( 'afterSetup.tribeTicketsCommerceCheckout', [ $container ] );
+		// Bind loader visibility.
+		obj.bindLoaderEvents();
+
+		$document.trigger( 'afterSetup.tecTicketsCommerce', [ $container ] );
```
juanfra, line 180: Maybe as a suggestion, to add more clarity, and to make it explicit that the trigger corresponds to the generic before and after setup, I'd create a method called `bindEvents` with the two triggers, and calling `bindLoaderEvents` and `bindCheckoutEvents` and just keeping `bindCheckoutItemDescriptionToggle` in `bindCheckoutEvents`.

juanfra (approved):
One small suggestion, other than that it's looking great. Good job!!! ⚡

src/resources/js/v2/tickets-commerce.js
rafsuntaskin: @juanfra Sorry, I am not clear about this one. `bindLoaderEvents` is part of the checkout and that's why it's within the `bindCheckoutEvents `. Which generic triggers you are referring to?

juanfra (approved):

PR #2081, opened 2021-09-08
[ET-1197] Fix price formatting for ticket totals

codingmusician:
### 🎫 Ticket

[ET-1197] 

### 🗒️ Description

The formatting of the ticket totals for non-standard thousands and decimal separators was messed up. This PR fixes those.

### 🎥 Artifacts 

Before changes: https://youtu.be/7KIgSbid-eY
After changes: https://youtu.be/01Xn5--EKYk

### ✔️ Checklist
- [x] My code is tested. 

[ETP-648]: https://theeventscalendar.atlassian.net/browse/ETP-648

[ET-1197]: https://theeventscalendar.atlassian.net/browse/ET-1197


src/resources/js/v2/tickets-utils.js
```diff
@@ -159,13 +159,7 @@ tribe.tickets.utils = {};
...
 		};
 
-		let s = ( prec ? toFixedFix( n, prec ) : Math.round( n ) ).toString().split( dec );
-
-		// if period is the thousands_sep we have to spilt using the decimal and not the comma as we work
-		// with numbers using the period as the decimal in JavaScript.
-		if ( '.' === format.thousands_sep ) {
-			s = ( prec ? toFixedFix( n, prec ) : Math.round( n ) ).toString().split( '.' );
-		}
+		let s = ( prec ? toFixedFix( n, prec ) : Math.round( n ) ).toString().split( '.' );
```
bordoni, line 162: That seems like it would break thing when using `,` as the separator.

bordoni (changes requested):

src/resources/js/v2/tickets-utils.js
codingmusician: The `numberFormat()` method won't be receiving a formatted price, though. It will always be a regular decimal (.) with no thousands separator. This is because we are either reading straight from the `data-ticket-price` attribute, which is a regular float value with a decimal (.), or we're running the price text through the `cleanNumber()` method to convert it to a regular float with a decimal (.).

readme.txt
```diff
@@ -178,6 +178,10 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 == Changelog ==
 
+= [TBD] TBD =
+
+* Fix - Fixed ticket total formatting when using non-standard thousands and decimal separators. [ET-1197]
```
rafsuntaskin, line 183:
```suggestion
* Fix - Fixed ticket total formatting when using custom thousands and decimal separators. [ET-1197]
```
It's actually standard in Germany to use `,` as a decimal separator ;)

codingmusician: Good call and great suggestion!

rafsuntaskin:
@codingmusician The artifact is not showing the fixed version in action, it would be great if you had shown the fixed version as well.

codingmusician:
Added an artifact to show what the fixed version looks like.

bordoni (approved):

rafsuntaskin:
The same calculation is available in both AR Modal and AR Page as well, are those working as expected too?

PR #2086, opened 2021-09-13
Tickets Commerce PayPal Webhooks

bordoni:
### 🎫 Ticket

[ET-1164]

### 🗒️ Description
https://www.loom.com/i/1e6124bb62084b6dafa6e7edc33b5574

No video Artifacts on this one yet, since @juanfra will likely handle the UI to display errors and the current status on the settings page later on. Webhooks are configured locally and on PayPal, on register and via a button o the settings page.

[ET-1164]: https://theeventscalendar.atlassian.net/browse/ET-1164


src/Tickets/Commerce/Gateways/PayPal/Client.php
```diff
@@ -131,19 +137,52 @@ public function get( $endpoint, array $query_args = [], array $request_arguments
 				'Content-Type'  => 'application/json',
 			]
 		];
+
+		if ( 'GET' !== $method ) {
+			$default_arguments['body'] = [];
+		}
```
juanfra, line 143: We're having the same conditional in the same function with a few lines difference. I assume this is to initialize the $default_arguments or something. Can we possibly leave some inline documentation to justify that we're doing this so that's not removed by some other dev in the future?

src/Tickets/Commerce/Gateways/PayPal/Client.php
```diff
@@ -131,19 +137,52 @@ public function get( $endpoint, array $query_args = [], array $request_arguments
...
+
+			if (
+				! empty( $request_arguments['body'] )
+				&& 'application/json' === strtolower( $content_type )
+			) {
+				$request_arguments['body'] = wp_json_encode( $request_arguments[ $key ] );
+			}
+		}
+
+		if ( 'GET' === $method ) {
```
juanfra, line 163:
```suggestion
		} elseif ( 'GET' === $method ) {
```

src/Tickets/Commerce/Gateways/PayPal/Client.php
```diff
@@ -494,5 +531,292 @@ public function refund_payment( $capture_id ) {
 		return $response;
 	}
 
+	/**
+	 * This uses the links property of the payment to retrieve the Parent Payment ID from PayPal
```
juanfra, line 535:
```suggestion
	 * This uses the links property of the payment to retrieve the Parent Payment ID from PayPal.
```

src/Tickets/Commerce/Gateways/PayPal/Client.php
```diff
@@ -494,5 +531,292 @@ public function refund_payment( $capture_id ) {
...
+	 * @since TBD
+	 *
+	 * @param string $webhook_id The webhook ID.
+	 *
+	 * @return array|null The PayPal webhook data.
+	 */
+	public function get_webhook( $webhook_id ) {
+		$query_args = [];
+		$body       = [];
+		$args       = [
```
juanfra, line 646: Do we maybe want to add these as optional parameters and use `wp_parse_args` with the defaults? Not sure if we need the flexibility to do it, now or in the future. As a dev I always appreciate when the code is flexible and I can easily use things for particular edge scenarios without modifying too much code.

src/Tickets/Commerce/Gateways/PayPal/Client.php
```diff
@@ -494,5 +531,292 @@ public function refund_payment( $capture_id ) {
...
+	 * @since TBD
+	 *
+	 * @return array|\WP_Error
+	 */
+	public function create_webhook() {
+		$events      = tribe( Events::class )->get_registered_events();
+		$webhook_url = tribe( Webhook_Endpoint::class )->get_route_url();
+
+		$query_args = [];
+		$body       = [
```
juanfra, line 689: The same question here about making these as parameters with defaults.

src/Tickets/Commerce/Gateways/PayPal/Hooks.php
```diff
@@ -167,6 +168,17 @@ public function handle_action_refresh_user_info() {
...
 
+	/**
+	 * Handles the refreshing of the webhook on PayPal for this site/merchant.
+	 *
+	 * @todo  Display some message when refreshing user info.
+	 * @since TBD
+	 *
+	 */
+	public function handle_action_refresh_webhook() {
+		$updated = $this->container->make( Webhooks::class )->create_or_update_existing();
```
juanfra, line 179: Do we want to return this?

src/Tickets/Commerce/Gateways/PayPal/On_Boarding_Redirect_Handler.php
```diff
@@ -16,6 +16,32 @@
...
+		$action_list = sprintf(
+			'<ol><li>%1$s</li><li>%2$s</li><li>%3$s %4$s</li></ol>',
+			esc_html__( 'Make sure to complete the entire PayPal process. Do not close the window you have finished the process.', 'event-tickets' ),
+			esc_html__( 'The last screen of the PayPal connect process includes a button to be sent back to your site. It is important you click this and do not close the window yourself.', 'event-tickets' ),
+			esc_html__( 'If you’re still having problems connecting:', 'event-tickets' ),
+			$this->settings->get_guidance_html()
+		);
+
+		$standard_error = sprintf(
+			'<div id="give-paypal-onboarding-trouble-notice" class="tribe-common-a11y-hidden"><p class="error-message">%1$s</p><p>%2$s</p></div>',
```
juanfra, line 37: Let's remove the Give reference.

src/Tickets/Commerce/Gateways/PayPal/Webhooks/Events.php
```diff
@@ -0,0 +1,120 @@
...
+	 * Webhook Event name for a capture of reversed payment.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	const PAYMENT_CAPTURE_REVERSED = 'PAYMENT.CAPTURE.REVERSED';
+
+	/**
+	 * Default mapping from PayPal Status to Tickets Commerce
```
juanfra, line 52:
```suggestion
	 * Default mapping from PayPal Status to Tickets Commerce.
```

src/Tickets/Commerce/Gateways/PayPal/Webhooks/Events.php
```diff
@@ -0,0 +1,120 @@
...
+
+	/**
+	 * Gets the valid mapping of the webhook events.
+	 *
+	 * @since TBD
+	 *
+	 * @return array
+	 */
+	public function get_valid() {
+		return $this->default_map;
```
juanfra, line 73: Do we want to add a filter here? Thinking of third parties/extensions wanting to extend to a new event or something?

juanfra (commented):
❤️‍🔥 great stuff here! I left a couple questions, but things are looking good!

src/Tickets/Commerce/Gateways/PayPal/Client.php
paulmskim: I think this might be intentional, otherwise the `elseif` with the `'POST'` would never run

src/Tickets/Commerce/Gateways/PayPal/Client.php
```diff
@@ -494,5 +531,292 @@ public function refund_payment( $capture_id ) {
...
+			'url'         => $webhook_url,
+			'event_types' => array_map(
+				static function ( $eventType ) {
+					return [
+						'name' => $eventType,
```
paulmskim, line 694:
```suggestion
				static function ( $event_type ) {
					return [
						'name' => $event_type,
```

paulmskim (commented):
Code looks great, though I'll defer the approval to someone else. Just one suggestion though:

src/Tickets/Commerce/Gateways/PayPal/On_Boarding_Redirect_Handler.php
bordoni: None of the methods on this file will stay. They are there just for reference. Error handling examples.

src/Tickets/Commerce/Gateways/PayPal/Webhooks/Events.php
bordoni: Probably a good idea.

src/Tickets/Commerce/Gateways/PayPal/Hooks.php
bordoni: No in this case it's an action URL so no need to return.

src/Tickets/Commerce/Gateways/PayPal/Client.php
bordoni: Exatcly.

src/Tickets/Commerce/Gateways/PayPal/Client.php
bordoni: Not specifically for the connection to PayPal. These are super "strict" methods, mostly for internal usage.

src/Tickets/Commerce/Gateways/PayPal/Client.php
bordoni: WIll make sure to add some comments, I like the idea.

PR #2091, opened 2021-09-13
[ET-1205] Fix ticket settings update breaking provider meta data.

rafsuntaskin:
### 🎫 Ticket

[ET-1205] 

### 🗒️ Description

* When the Ticket Settings update is clicked we trigger the `save_form_settings` method that makes use of `update_post_meta` for saving the provider data.
* For the new Ticket Commerce our provider class is namespaced which should be saved like this as meta: `TEC\\Tickets\\Commerce\\Module` but as it is running through `wp_unslash` it's getting saved as `TECTicketsCommerceModule`.

### 🎥 Artifacts 
🎥 screencast(s): https://d.pr/v/kOMioC

### ✔️ Checklist
- [ ] ~~I've included a changelog entry both in readme.txt.~~ ( Not required )
- [x] My code is tested. 

[ET-1205]: https://theeventscalendar.atlassian.net/browse/ET-1205

### DIFF
--- a/src/Tickets/Commerce/Hooks.php
+++ b/src/Tickets/Commerce/Hooks.php
@@ -440,7 +440,7 @@ class Hooks extends tad_DI52_ServiceProvider {
 		 */
 		$ticket_handler = tribe( 'tickets.handler' );
 
-		add_filter( "sanitize_post_meta_{$ticket_handler->key_provider_field}" , [ $this, 'filter_modify_sanitization_provider_meta' ], 10, 2 );
+		add_filter( "sanitize_post_meta_{$ticket_handler->key_provider_field}" , [ $this, 'filter_modify_sanitization_provider_meta' ] );
 	}
 
 	/**
@@ -448,12 +448,11 @@ class Hooks extends tad_DI52_ServiceProvider {
 	 *
 	 * @since TBD
 	 *
-	 * @param string $meta_key   Metadata key.
 	 * @param mixed  $meta_value Metadata value.
 	 *
 	 * @return string
 	 */
-	public function filter_modify_sanitization_provider_meta( $meta_value, $meta_key ) {
-		return tribe( Settings::class )->skip_sanitization( $meta_value, $meta_key );
+	public function filter_modify_sanitization_provider_meta( $meta_value ) {
+		return tribe( Settings::class )->skip_sanitization( $meta_value );
 	}
 }
--- a/src/Tickets/Commerce/Hooks.php
+++ b/src/Tickets/Commerce/Hooks.php
@@ -94,9 +94,9 @@ class Hooks extends tad_DI52_ServiceProvider {
 		// Add a post display state for special Event Tickets pages.
 		add_filter( 'display_post_states', [ $this, 'add_display_post_states' ], 10, 2 );
 
-		// Todo make the following filters dynamic.
-		add_filter( 'sanitize_post_meta__tribe_default_ticket_provider_for_tribe_events' , [ $this, 'skip_sanitization' ], 10, 2 );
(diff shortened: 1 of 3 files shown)


src/Tickets/Commerce/Hooks.php
```diff
@@ -416,4 +420,12 @@ public function add_display_post_states( $post_states, $post ) {
 		return $post_states;
 	}
 
+	public function skip_sanitization( $meta_value, $meta_key ) {
+
+		if ( $meta_value != wp_unslash( tribe( Module::class )->class_name ) ) {
+			return $meta_value;
+		}
+
+		return tribe( Module::class )->class_name;
```
bordoni, line 429: `tribe( Module::class )->class_name` holds the same value as: `Module::class`

src/Tickets/Commerce/Hooks.php
```diff
@@ -416,4 +420,12 @@ public function add_display_post_states( $post_states, $post ) {
 		return $post_states;
 	}
 
+	public function skip_sanitization( $meta_value, $meta_key ) {
+
+		if ( $meta_value != wp_unslash( tribe( Module::class )->class_name ) ) {
```
bordoni, line 425: `tribe( Module::class )->class_name` holds the same value as: `Module::class`

src/Tickets/Commerce/Hooks.php
```diff
@@ -416,4 +420,12 @@ public function add_display_post_states( $post_states, $post ) {
 		return $post_states;
 	}
 
+	public function skip_sanitization( $meta_value, $meta_key ) {
```
bordoni, line 423: Doc block. functionality should not live in the Hooks class, so create a method on the `Settings` and use a filter here to point to it. See the method `filter_registration_cart_provider`

bordoni (changes requested):

src/Tickets/Commerce/Hooks.php
```diff
@@ -93,6 +93,10 @@ protected function add_filters() {
 
 		// Add a post display state for special Event Tickets pages.
 		add_filter( 'display_post_states', [ $this, 'add_display_post_states' ], 10, 2 );
+
+		// @todo make the following filters dynamic.
```
paulmskim, line 97: could we add a person responsible for this + a jira ticket?

src/Tickets/Commerce/Settings.php
```diff
@@ -397,4 +397,23 @@ public function apply_commerce_enabled_conditional( $settings ) {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param string $meta_key   Metadata key.
+	 * @param mixed  $meta_value Metadata value.
+	 *
+	 * @return string
+	 */
+	public function skip_sanitization( $meta_value, $meta_key ) {
+
```
paulmskim, line 411: no need to pad functions with empty lines

src/Tickets/Commerce/Hooks.php
```diff
@@ -416,4 +420,17 @@ public function add_display_post_states( $post_states, $post ) {
...
+	 * Handle saving of Ticket provider meta data.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $meta_key   Metadata key.
+	 * @param mixed  $meta_value Metadata value.
+	 *
+	 * @return string
+	 */
+	public function handle_provider_meta( $meta_value, $meta_key ) {
```
bordoni, line 433: Since this is a filter, we should prefix it. `filter_modify_sanitization_provider_meta`

bordoni (changes requested):

src/Tickets/Commerce/Hooks.php
bordoni: Actually can we make it into a dynamic thing now? It shouldn't be hard.

bordoni (approved):

PR #2108, opened 2021-09-25
[ET-1188] Avoid TicketCommerce duplicate orders on Checkout

rafsuntaskin:
### 🎫 Ticket

[ET-1188] 

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1188]: https://theeventscalendar.atlassian.net/browse/ET-1188


bordoni:
I am not sure I agree this is something we should do. I think new orders need to be created every time. Not entirely sure tho.

rafsuntaskin:
As @bordoni  mentioned, we need to keep all the raw data related to orders as much as we can. This will allow us to create more detailed reports if needed.

So, the number of canceled and not completed orders should be tracked instead of avoiding them.

However, we do have our eyes on the possible overload of unused/orphaned rows this might create. We will be moving towards a custom table solution in the future or maybe implement a cron Job or action that trims the data based on time or quantity.

PR #2122, opened 2021-10-06
[ET-1214] AR Fields are not being saved with Ticket Commerce tickets.

rafsuntaskin:
### 🎫 Ticket

[ET-1214] 

### 🗒️ Description

* We were not using the proper action hook for saving ticket data that triggers the saving of Ticket Plus Meta.

### 🎥 Artifacts 
🎥 screencast: https://d.pr/i/Z7lrJb

### ✔️ Checklist
- [ ] ~~I've included a changelog entry both in readme.txt and changelog.txt files.~~ 
- [x] My code is tested. 

[ET-1214]: https://theeventscalendar.atlassian.net/browse/ET-1214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Ticket.php
+++ b/src/Tickets/Commerce/Ticket.php
@@ -630,16 +630,16 @@ class Ticket {
 		do_action( "tec_tickets_commerce_after_{$save_type}_ticket", $post_id, $ticket, $raw_data, static::class );
 
 		/**
-		 * Generic action fired after saving a ticket
+		 * Generic action fired after saving a ticket.
 		 *
-		 * @since 4.7
+		 * @since TBD
 		 *
 		 * @param int                            $post_id  Post ID of post the ticket is tied to
 		 * @param \Tribe__Tickets__Ticket_Object $ticket   Ticket that was just saved
 		 * @param array                          $raw_data Ticket data
 		 * @param string                         $class    Commerce engine class
 		 */
-		do_action( 'tec_tickets_commerce_after_save_ticket', $post_id, $ticket, $raw_data, static::class );
+		do_action( 'event_tickets_after_save_ticket', $post_id, $ticket, $raw_data, static::class );
 
 		return $ticket->ID;
 	}


src/Tickets/Commerce/Ticket.php
```diff
@@ -630,16 +630,16 @@ public function save( $post_id, $ticket, $raw_data = [] ) {
...
-		 * @since 4.7
+		 * @since TBD
 		 *
 		 * @param int                            $post_id  Post ID of post the ticket is tied to
 		 * @param \Tribe__Tickets__Ticket_Object $ticket   Ticket that was just saved
 		 * @param array                          $raw_data Ticket data
 		 * @param string                         $class    Commerce engine class
 		 */
-		do_action( 'tec_tickets_commerce_after_save_ticket', $post_id, $ticket, $raw_data, static::class );
+		do_action( 'event_tickets_after_save_ticket', $post_id, $ticket, $raw_data, static::class );
```
juanfra, line 642: Do we maybe keep the `tec_tickets_` one for the sake of consistency here, and then add the proper add/remove_actions on ET+ to the new?

Like, for example, going here https://github.com/the-events-calendar/event-tickets-plus/blob/1a52f7f67e75bbae3318d9273e6f0b8af1c42b73/src/Tribe/Attendee_Registration/Service_Provider.php#L99 and having the `add_action` for both

rafsuntaskin: We already have another similar hook for consistency: https://github.com/the-events-calendar/event-tickets/blob/d759c39a3456ca1a296e03ce8d4e339631239cf0/src/Tickets/Commerce/Ticket.php#L630

We need to split this method into `update` and `create` methods in the future, while doing that we can add more specific actions. 

For now, I was trying to avoid further confusion by adding the new hook again. Ideally, we should only have the `do_action` hooks in the base class.

PR #2145, opened 2021-10-18
Fix - My Tickets page not working for TicketsCommerce without ETP

rafsuntaskin:
### 🎫 Ticket

[TICKET_ID] - Ticket ID, if there's any put it between brackets -->

* **Needs Ticket for QA.**

### 🗒️ Description

* Added a new V2 template for the My Tickets section of TicketsCommerc tickets.
* We can make use of this template in the future to unify other providers, this template is brought in from ETP but it should have been in the ET in the first place. We need to convert the `My Tickets` templates to V2 in the future and this can be a good base for that.

### 🎥 Artifacts 

| Before | After |
| - | - |
Screen Shot 2021-10-19 at 3 36 58 AM | Screen Shot 2021-10-19 at 3 36 18 AM

### ✔️ Checklist
- [ ] My code is tested. 

### DIFF
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -20,6 +20,7 @@ namespace PHPSTORM_META {
 		'tickets.rest-v1.system'       => \Tribe__Tickets__REST__V1__System::class,
 		'tickets.rest-v1.validator'    => \Tribe__Tickets__REST__V1__Validator__Base::class,
 		'tickets.rest-v1.repository'   => \Tribe__Tickets__REST__V1__Post_Repository::class,
+
 	];
 
 	// Allow PhpStorm IDE to resolve return types when calling tribe( Object_Type::class ) or tribe( `Object_Type` )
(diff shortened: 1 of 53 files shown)


juanfra:
@bordoni @rafsuntaskin I'm not sure about these changes. We're introducing new templates that don't have to do with how we do new templates. We're introducing new CSS that doesn't have to do with how we're doing CSS these days. Once these things get into production it's really hard to deprecate and move to something that we call good.

My feeling is that we need to implement the new templates right now. Or at least some sort of version that's not what we get from the old code. Maybe, given that it's a more FE thing, I can tackle the rest of what's missing here so that Rafsun can continue with more BEish tickets.

paulmskim:
I agree with Juan here, unless there's a good reason why we're keeping the old markup + CSS, if we can update this to an updated markup it would be better for us going forward.

If we do anything, we should at least move all the logic out of the template so we have the separation of logic.

PR #2157, opened 2021-10-26
[ET-1239] Remove constant dependency and update logic for `TribeCommerce` visibility.

rafsuntaskin:
### 🎫 Ticket

[ET-1239] 

### 🗒️ Description

This is the base PR that is controlling our TribeCommerce to TicketsCommerce upgrade flows.

### 🎥 Artifacts 
🎥 screencast(s):
New install check: https://d.pr/v/QIz21v

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1239]: https://theeventscalendar.atlassian.net/browse/ET-1239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/functions/commerce/provider.php
```diff
@@ -93,12 +87,67 @@ function tec_tribe_commerce_is_available() {
...
+
+	if ( ! $force_check ) {
+		// If we don't find any cached value, we keep things activated.
+		return (bool) get_option( $cache_key, true );
+	}
+
+	$has_active_tickets = (bool) tribe_tickets()->by( 'post_type', 'tribe_tpp_tickets' )->where( 'is_active' )->count();
+
+	// Cache the data.
+	update_option( $cache_key, $has_active_tickets );
```
bordoni, line 150: use a transient with a 24h expiration here.

bordoni (changes requested):

src/functions/commerce/provider.php
rafsuntaskin: I have avoided using transient intentionally.

We are setting this value only when the customer is visiting the `Tickets` settings page. So, this is getting updated whenever the admin is actively using the website or trying to change the settings, which makes it realtime.

Using transient will not make it realtime and we might end up not showing the notices even if there are no active tickets.

rafsuntaskin: Also, `get_transient` doesn't have any default value option like `get_option`. So, changing to that will require some logic updates around `tec_tribe_commerce_should_be_active`

bordoni: This is a transient piece of data, meaning it changes once certain piece of criteria is met. The same things you do with `update_option` and `get_option` can be used with the transient, the diff is lets say our QA needs to reset this check for some reason, they can just use any simple transient manager Like they are already used to.

bordoni (approved):

PR #2179, opened 2021-11-01
Flush transients whens saving plugin options

moraleida:
### 🎫 Ticket

[ET-1287] 

### 🗒️ Description

Flushes the transients whenever the Sandbox option is toggled on/off.

### 🎥 Artifacts 

Peek 2021-11-01 17-41

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1287]: https://theeventscalendar.atlassian.net/browse/ET-1287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Hooks.php
+++ b/src/Tickets/Commerce/Hooks.php
@@ -122,7 +122,6 @@ class Hooks extends tad_DI52_ServiceProvider {
 
 		add_filter( 'tec_tickets_commerce_payments_tab_settings', [ $this, 'filter_payments_tab_settings' ] );
 
-		add_filter( 'tribe-events-save-options', [ $this, 'flush_transients_when_toggling_sandbox_mode' ] );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


src/Tickets/Commerce/Hooks.php
```diff
@@ -701,4 +703,26 @@ public function register_event_compatibility_hooks() {
...
+	public function flush_transients_when_toggling_sandbox_mode( $options ) {
+
+		if ( isset( $options[ Settings::$option_sandbox ] ) &&
+			 tec_tickets_commerce_is_sandbox_mode() !== $options[ Settings::$option_sandbox ]
+		) {
+			$signup = tribe( Base_Commerce\Gateways\PayPal\Signup::class );
+			$signup->delete_transient_data();
+		}
+
+		return $options;
```
bordoni, line 726: Invert the logic. to avoid indentation.

bordoni (changes requested):

rafsuntaskin:
As this is related to PayPal gateway specifically. We should move the method into: `wp-content/plugins/event-tickets/src/Tickets/Commerce/Gateways/PayPal/Settings.php` and then Hook from `wp-content/plugins/event-tickets/src/Tickets/Commerce/Gateways/PayPal/Hooks.php`.

src/Tickets/Commerce/Hooks.php
moraleida: done.

moraleida:
> Maybe as this is related to PayPal gateway specifically. We should move the method into: `wp-content/plugins/event-tickets/src/Tickets/Commerce/Gateways/PayPal/Settings.php` and then Hook from `wp-content/plugins/event-tickets/src/Tickets/Commerce/Gateways/PayPal/Hooks.php`.

God thinkning. I've moved the method into `Gateway\PayPal\Hooks.php`. Not sure if we need a new method in `Settings` though because all this filter is doing is calling a method of `Signup`, so no real functionality implemented in the `Hooks` file. Let me know what you think @rafsuntaskin

bordoni:
I would rather see it on the Signup class instead of the settings one.

bordoni (approved):

PR #2184, opened 2021-11-02
Include the assets via a hook into the shortcode template with print …

bordoni:
### 🎫 Ticket

[ET-1269] 

### 🎥 Artifacts 
https://www.loom.com/share/17dd9810830746cd9280b2d1100173b0

[ET-1269]: https://theeventscalendar.atlassian.net/browse/ET-1269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Hooks.php
+++ b/src/Tickets/Commerce/Hooks.php
@@ -705,12 +705,11 @@ class Hooks extends tad_DI52_ServiceProvider {
 	}
 
 	/**
-	 * Includes the Assets to the checkout page.
+	 * Includes the Assets to the checkout page shortcode.
 	 *
 	 * @since TBD
 	 */
 	public function include_assets_checkout_shortcode() {
-		// Enqueue assets.
-		tribe_asset_enqueue_group( 'tribe-tickets-commerce-checkout' );
+		Shortcodes\Checkout_Shortcode::enqueue_assets();
 	}
 }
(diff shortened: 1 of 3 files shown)


src/Tickets/Commerce/Hooks.php
```diff
@@ -701,4 +703,14 @@ public function register_event_compatibility_hooks() {
 
 		add_filter( 'wp_redirect', [ tribe( Compatibility\Events::class ), 'prevent_filter_redirect_canonical' ], 1, 2 );
 	}
+
+	/**
+	 * Includes the Assets to the checkout page.
+	 *
+	 * @since TBD
+	 */
+	public function include_assets_checkout_shortcode() {
```
juanfra, line 712: Should we maybe move this to the checkout shortcode class?

juanfra: It's just a line, but as we do it in other places, I think we may want to do the same.

juanfra (approved):
Looking good!

src/Tickets/Commerce/Hooks.php
bordoni: Good point moving.

juanfra (approved):

PR #2189, opened 2021-11-03
Move duplicated pending orders to Not Completed once one of them is Completed

bordoni:
See conversation with Sydney here:
https://lw.slack.com/archives/C01T9UJGE2U/p1635885014370700

### DIFF
--- a/src/Tickets/Commerce/Flag_Actions/Duplicated_Pending_Orders.php
+++ b/src/Tickets/Commerce/Flag_Actions/Duplicated_Pending_Orders.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace TEC\Tickets\Commerce\Flag_Actions;
+
+use TEC\Tickets\Commerce\Order;
+use TEC\Tickets\Commerce\Status\Not_Completed;
+use TEC\Tickets\Commerce\Status\Pending;
+use TEC\Tickets\Commerce\Status\Status_Interface;
+use Tribe__Utils__Array as Arr;
+
+/**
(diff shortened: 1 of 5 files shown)


src/Tickets/Commerce/Flag_Actions/Duplicated_Pending_Orders.php
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Class Duplicated_Pending_Orders, normally triggered when a given order is completed it will modify the status of all the
+ * other orders created with the same Hash cart key, which prevents leaving pending orders open when one was completed.
+ *
+ * @since    TBD
+ *
+ * @package  TEC\Tickets\Commerce\Flag_Actions
+ */
+class Duplicated_Pending_Orders extends Flag_Action_Abstract {
```
borkweb, line 19: I don't think this class name makes sense for what it is doing. Perhaps a better name would be more verb-like such a the following: `End_Duplicate_Pending_Orders`

src/Tickets/Commerce/Flag_Actions/Duplicated_Pending_Orders.php
```diff
@@ -0,0 +1,64 @@
...
+
+	/**
+	 * {@inheritDoc}
+	 *
+	 * @since TBD
+	 */
+	protected $flags = [
+		'duplicated_pending_orders',
```
borkweb, line 27: If the class name changes, this will need to change as well.

borkweb (changes requested):

PR #2190, opened 2021-11-03
workaround for empty emails_in_order when sending emails from a manua…

moraleida:
### 🎫 Ticket

[ET-1307] 

### 🗒️ Description

Temporary workaround for empty array when sending emails to manual attendees

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1307]: https://theeventscalendar.atlassian.net/browse/ET-1307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Flag_Actions/Send_Email.php
+++ b/src/Tickets/Commerce/Flag_Actions/Send_Email.php
@@ -34,6 +34,14 @@ class Send_Email extends Flag_Action_Abstract {
 	 * {@inheritDoc}
 	 */
 	public function handle( Status_Interface $new_status, $old_status, \WP_Post $order ) {
+
+		// temporary fix for manual attendees first email
+		// @todo backend review this logic
+		if ( ! empty( $order->gateway ) && 'manual' === $order->gateway && empty( $order->events_in_order ) ) {
+			$order->events_in_order[] = $order;
+		}
+
+
 		if ( empty( $order->events_in_order ) || ! is_array( $order->events_in_order ) ) {
 			return;
 		}


src/Tickets/Commerce/Flag_Actions/Send_Email.php
```diff
@@ -34,6 +34,14 @@ class Send_Email extends Flag_Action_Abstract {
 	 * {@inheritDoc}
 	 */
 	public function handle( Status_Interface $new_status, $old_status, \WP_Post $order ) {
+
+		// temporary fix for manual attendees first email
+		// @todo backend review this logic
+		if ( ! empty( $order->gateway ) && 'manual' === $order->gateway && empty( $order->events_in_order ) ) {
+			$order->events_in_order[] = $order;
```
bordoni, line 41: Why are we adding the order to the events array?

bordoni (changes requested):

src/Tickets/Commerce/Flag_Actions/Send_Email.php
moraleida: This is a dirty workaround.

That said, it works because when adding attendees manually the Order already contains all the necessary information to send the email. In this case, Order ID and Event ID are identical when calling `send_tickets_email` but all the email information is re-fetched from the Order later, including which ticket and event it belongs to, so all information coming out in the email is correct.

bordoni: Ohh I see. Thansk a lot

bordoni (approved):

PR #2210, opened 2021-11-11
Fix operations with non-US currencies and formats

moraleida:
### 🎫 Ticket

[ET-1330] 

### 🎥 Artifacts 

Peek 2021-11-12 17-12

### ✔️ Checklist
- [x] My code is tested. 

[ET-1330]: https://theeventscalendar.atlassian.net/browse/ET-1330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Commerce/Currency.php
```diff
@@ -11,4 +11,32 @@
...
+	 *       were replaced by this method. We need to cleanup and have the entire codebase
+	 *       read currency codes from this class in the future. Once all currency operations are fixed to use this method,
+	 *       modify this method to delete the old keys from the settings array.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+
+	public static function get_currency() {
```
codingmusician, line 27: We might want to be more specific and call this `get_currency_code` so there's no confusion with things like 'currency symbol' or whatever else might come up in the future.

moraleida: good thinking

src/Tickets/Commerce/Currency.php
```diff
@@ -11,4 +11,32 @@
...
+	 *
+	 * @return string
+	 */
+
+	public static function get_currency_code() {
+		// New key
+		$currency = tribe_get_option( 'tickets-commerce-currency-code' );
+
+		if ( ! $currency ) {
+			// Old key
```
juanfra, line 32:
```suggestion
			// Old key.
```

src/Tickets/Commerce/Currency.php
```diff
@@ -11,4 +11,32 @@
...
+			$currency = tribe_get_option( 'ticket-commerce-currency-code', 'USD' );
+
+			// Duplicate the currency code in the new key
+			tribe_update_option( 'tickets-commerce-currency-code', $currency );
```
juanfra, line 36: Do we want to have this key in a class var?

src/Tickets/Commerce/Currency.php
```diff
@@ -11,4 +11,32 @@
...
+
+	public static function get_currency_code() {
+		// New key
+		$currency = tribe_get_option( 'tickets-commerce-currency-code' );
+
+		if ( ! $currency ) {
+			// Old key
+			$currency = tribe_get_option( 'ticket-commerce-currency-code', 'USD' );
+
+			// Duplicate the currency code in the new key
```
juanfra, line 35:
```suggestion
			// Duplicate the currency code in the new key.
```

juanfra (commented):
Code is looking great! cand we add a readme entry for this and artifacts?

src/Tickets/Commerce/Currency.php
```diff
@@ -11,4 +11,32 @@
...
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+
+	public static function get_currency_code() {
+		// New key
+		$currency = tribe_get_option( 'tickets-commerce-currency-code' );
+
+		if ( ! $currency ) {
```
juanfra, line 31: maybe use `empty()`?

moraleida:
Some tests are not passing, I'll fix those.

src/resources/js/v2/tickets-utils.js
```diff
@@ -84,48 +84,65 @@ tribe.tickets.utils = {};
...
+	 * Determines if a given separator acts as a decimal separator or something else in a string.
+	 *
+	 * The rule to determine a decimal is straightforward. It needs to exist only once
+	 * in the string and the piece of the string after the separator cannot be longer
+	 * than 2 digits. Anything else is serving another purpose.
+	 *
+	 * @since TBD
+	 *
+	 * @param {string} separator a separator token, like . or ,
+	 * @param {number} number    the number
```
redscar, line 132:
```suggestion
	 * @param {number} number    the number.
```

src/resources/js/v2/tickets-utils.js
```diff
@@ -84,48 +84,65 @@ tribe.tickets.utils = {};
 	 * Removes separator characters and converts decimal character to '.'
 	 * So they play nice with other functions.
 	 *
+	 * @since TBD major refactoring of the internal logic
```
redscar, line 87:
```suggestion
	 * @since TBD major refactoring of the internal logic.
```

src/resources/js/v2/tickets-utils.js
```diff
@@ -84,48 +84,65 @@ tribe.tickets.utils = {};
...
 	 * @since 5.0.3
 	 *
+	 * @todo clean up all calls to obj.cleanNumber to do not send a provider
```
redscar, line 90:
```suggestion
	 * @todo clean up all calls to obj.cleanNumber to do not send a provider.
```

src/Tickets/Commerce/Currency.php
```diff
@@ -11,4 +11,41 @@
...
+	 *       and have the entire codebase read currency codes from this class in the future. Once all currency
+	 *       operations are fixed to use this class, modify this method to delete the old keys from the settings array.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+
+	public static function get_currency_code() {
+		// New key
```
juanfra, line 37:
```suggestion
		// New key.
```

src/resources/js/v2/tickets-utils.js
```diff
@@ -84,48 +84,67 @@ tribe.tickets.utils = {};
...
+	 *
 	 * @param {number} passedNumber The number to clean.
 	 * @param {string} provider The provider.
 	 *
 	 * @returns {string} The cleaned number.
 	 */
-	obj.cleanNumber = function( passedNumber, provider ) {
-		let number = passedNumber;
-		const format = obj.getCurrencyFormatting( provider );
+	obj.cleanNumber = function( passedNumber, provider ) { // eslint-disable-line no-unused-vars
```
juanfra, line 97: This should also work with the other providers (Woo for example), and the info about the format is here `const format = obj.getCurrencyFormatting( provider );`. That's not only here but in other places and my fear is that by modifying/removing this we can break how things work for the other providers. Why is it that we're modifying this?

This piece of code is super important to work with the different formatting that people use around the world.

juanfra (commented):
Question about modifying the JS

src/Tickets/Commerce/Currency.php
```diff
@@ -11,4 +11,41 @@
...
+	 * @return string
+	 */
+
+	public static function get_currency_code() {
+		// New key.
+		$currency = tribe_get_option( static::$currency_code_option );
+
+		if ( empty( $currency ) ) {
+			// Old key.
+			$currency = tribe_get_option( 'ticket-commerce-currency-code', 'USD' );
```
bordoni, line 42: Comment here should make reference to Tribe Commerce specifically, not just old way.
Also before doing the update to a empty check on the value in case it's set as 0 or '', which will not return the USD value for default. Old code that is behind `tribe_get_option` being weird.

src/Tickets/Commerce/Utils/Price.php
```diff
@@ -42,8 +42,8 @@ public static function sub_total( $value, $quantity, $decimal = null, $thousand_
 		$decimal      = ! is_null( $decimal ) ? $decimal : tribe( \Tribe__Tickets__Commerce__Currency::class )->get_currency_locale( 'decimal_point' );
 		$thousand_sep = ! is_null( $thousand_sep ) ? $thousand_sep : tribe( \Tribe__Tickets__Commerce__Currency::class )->get_currency_locale( 'thousands_sep' );
```
bordoni, line 43: If we need to modify these two to make sure we dont make changes to old  Tribe Commerce code, please lets do that.

bordoni (changes requested):
The most important piece for this particular piece of code is that the solution should only impact Tickets Commerce files.

`src/Tribe/Commerce/Currency.php` and `src/resources/js/v2/tickets-utils.js` should not have any changes to them.

src/resources/js/v2/tickets-utils.js
moraleida: @juanfra @bordoni the reason for this change is that the current `cleanNumber'` method fails if it gets passed a value that is formatted differently from what the `provider` says it should. For TC, this means that any values that use dot for thousands and comma for decimals fail. 

You can see this if you apply the patch I suggested here https://gist.github.com/moraleida/4a01dee2a0859421318a649a507e6d39 and try to make a purchase using the Attendee Registration Modal.

Any values from currencies that use dot as the thousands separator get multiplied by 100 in the modal because 
in TC, the values that get passed to this method are already formatted floats. This way, `cleanNumber` tries to remove a thousands separator and ends up removing the decimal separator instead like so:

```
number = '1034.89';
number.split( format.thousands_sep ).join( '' ); // number = '103489'
```

The item price is `R$ 1.034,89`  and the number that shows up on the modal is `R$ 103.489,00`

The logic I applied to this change is the same format-agnostic one we already have in `Price::to_numeric()`, that is: if a non-digit is acting as a decimal, treat it as such, regardless of what the provider says. The only instance in which it would fail is if we're dealing with currencies that use 3-digit decimals, but I don't think we currently support those anyway.

-----

With all that said:
- this change only impacts the attendee registration modal.
- we can modify TC to stop sending formatted floats to the modal, but my understanding is that this would be a larger change that we probably can't fit into gobi. Probably a good change include ET-1331.
- if we're ok with telling non-USD folks to not use the Attendee Registration modal for now, we can remove this piece and merge only the PHP code, and it will work properly with the Attendee Registration page.

src/Tickets/Commerce/Currency.php
moraleida: Not sure I understand what you mean. This piece is here to say "if the proper TC option is not set, get the TPP option, annd if that is also not set, fall back to USD". This allows us to change `Tribe/Commerce/Currrency.php` keeping the TPP and TC functionalities working together.

With ET-1331 we should probably move away from `Tribe/Commerce/Currency.php` entirely, but I guess that's off the table for this release.

src/Tickets/Commerce/Utils/Price.php
moraleida: This will go away in ET-1331. Both `total` and `sub_total` will become pure methods and not deal with formatting at all.

src/resources/js/v2/tickets-utils.js
moraleida: No idea what it was, but it looks like some change from Gobi had already dealt with this issue, so none of the JS code is needed anymore.

bordoni (approved):

bordoni (approved):

bordoni:
Before merging lets do some in branch QA.

bordoni (approved):

PR #2215, opened 2021-11-30
[ET-1343] remove wp cache flush

codingmusician:
### 🎫 Ticket

[ET-1343] 

### 🗒️ Description

Remove use of `wp_cache_flush()`. Instead, if using an external object cache plugin, delete keys individually.

### 🎥 Artifacts 

N/A

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1343]: https://theeventscalendar.atlassian.net/browse/ET-1343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -182,6 +182,7 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 = [TBD] TBD =
 
 * Fix - Remove use of `wp_cache_flush()` and use conditional when using an external object cache. (props to @r-a-y for this change!) [ET-1343]
+* Fix - Fixes error being caused by an endless loop when currency settings are saved. [ET-1344]
 
 = [5.2.1] 2021-11-17 =
 
--- a/readme.txt
+++ b/readme.txt
@@ -179,6 +179,10 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 == Changelog ==
 
+= [TBD] TBD =
+
+* Fix - Remove use of `wp_cache_flush()` and use conditional when using an external object cache. [ET-1343]
+
 = [5.2.1] 2021-11-17 =
 
 * Enhancement - Auto generate checkout page when enabling Tickets Commerce. [ET-1232]
--- a/readme.txt
+++ b/readme.txt
@@ -181,7 +181,7 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 = [TBD] TBD =
 
-* Fix - Remove use of `wp_cache_flush()` and use conditional when using an external object cache. [ET-1343]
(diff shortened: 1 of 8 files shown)


readme.txt
```diff
@@ -179,6 +179,10 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 == Changelog ==
 
+= [TBD] TBD =
+
+* Fix - Remove use of `wp_cache_flush()` and use conditional when using an external object cache. [ET-1343]
```
borkweb, line 184:
```suggestion
* Fix - Remove use of `wp_cache_flush()` and use conditional when using an external object cache. (props to @r-a-y for this change!) [ET-1343]
```
```suggestion
* Fix - Remove use of `wp_cache_flush()` and use conditional when using an external object cache. [ET-1343]
```

src/Tribe/Cache/Transient_Cache.php
```diff
@@ -25,13 +25,16 @@ public function reset_all( $post_id = 0 ) {
...
+		$is_using_object_cache = wp_using_ext_object_cache();
+
+		array_walk( $this->keys, function( $key ) use ( $hash, $is_using_object_cache ) {
 			delete_transient( __CLASS__ . $key );
 			delete_transient( __CLASS__ . $key . $hash );
-		}
-
-		// Flush WordPress cache for cache compatibility
-		wp_cache_flush();
+			if ( $is_using_object_cache ) {
```
moraleida, line 33: @codingmusician this is not flushing the cache unless you have a caching plugin, I'm wondering if we even need to make this check? As long as we're only deleting keys we have created ourselves like you're doing below, I don't see the need for this distinction.

codingmusician: I was basing this solution on what was [mentioned here](https://github.com/the-events-calendar/event-tickets/pull/1673#issuecomment-634336682). Basically, `wp_cache_flush()` was [added](https://github.com/the-events-calendar/event-tickets/commit/ec3ca2e799dd7c9ca39f99355ca456fb38fa6d86) to ensure compatibility with object caching plugins.

codingmusician: I supposed it would be safe just to go ahead and `wp_cache_delete()` all the keys, though. Is that what you're recommending @moraleida?

moraleida: yes @codingmusician. calling `wp_cache_delete()` on our keys will not cause issues if caching drop-ins are being used and it prevents weird issues with in-memory cache. We had just cast a net too wide with the earlier fix.

bordoni (approved):

PR #2219, opened 2021-12-02
New Price work implementation

moraleida:
### 🎫 Ticket

[ET-1331] 

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1331]: https://theeventscalendar.atlassian.net/browse/ET-1331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Commerce/Module.php
```diff
@@ -449,20 +449,43 @@ public function supports_global_stock() {
...
+		 * Allow filtering of the Price HTML
+		 *
+		 * @since TBD
+		 *
+		 * @param string $price_html
+		 * @param mixed  $product
+		 * @param mixed  $attendee
+		 *
+		 */
+		return apply_filters( 'tec_tickets_commerce_ticket_price_html', $price_html, $product, $attendee );
```
bordoni, line 466: We 100% want to avoid adding anything to the Module file as aligned originally on the Tickets Commerce document we are creating.

This needs to live in the Ticket Class.

src/Tickets/Commerce/Module.php
```diff
@@ -449,20 +449,43 @@ public function supports_global_stock() {
...
 	 */
 	public function get_price_value( $product ) {
-		return tribe( Ticket::class )->get_price_value( $product );
+		$ticket = Commerce\Models\Ticket_Model::from_post( $product );
+
+		if ( ! is_a( $ticket, 'TEC\Tickets\Commerce\Models\Ticket_Model' ) ) {
+			return;
+		}
+
+		return $ticket->get_value();
```
bordoni, line 488: We 100% want to avoid adding anything to the Module file as aligned originally on the Tickets Commerce document we are creating.

This needs to live in the Ticket Class.

bordoni (changes requested):

src/Tickets/Commerce/Module.php
moraleida: ok, just so I understand. This method was moved here from the Ticket class because of the `@todo` here: https://github.com/the-events-calendar/event-tickets/pull/2219/files#diff-960d6af64b1cae0ab20d33a9293a9aa3ae975a624399a539da3c819f31d6089bL869.

We should keep calling the methods from Tickets, then, and just move the logic in them to Ticket_Model. Is that it?

moraleida:
@bordoni approach to Module/Tickets changed, you can review again

src/Tickets/Commerce/Cart.php
```diff
@@ -339,8 +338,11 @@ public function get_items_in_cart( $full_item_params = false ) {
 					return null;
 				}
 
+				$sub_total_value = new Commerce\Utils\Value();
+				$sub_total_value->set_value( $item['obj']->price );
```
bordoni, line 342: We should create a static method that returns an instance of Value. A `make` of sorts.

bordoni (changes requested):

src/Tickets/Commerce/Utils/Value.php
```diff
@@ -0,0 +1,89 @@
...
+	public function get_currency_separator_thousands() {
+		return $this->currency_separator_thousands;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function set_class_name() {
+		$this->class_name = 'tickets_commerce_value';
+	}
```
bordoni, line 81: Using `class_name` is confusing as this is no longer the name of the class at all, can we move toward something like `type`, `id`, `slug`. 

Lets avoid `set_class_name`, just use an abstract `get_id()` that pulls from `$this->id` or what ever name we give this prop. And to avoid problems where devs didn't set an ID throw an error.

src/Tickets/Commerce/Utils/Value.php
```diff
@@ -0,0 +1,89 @@
...
+	public static function build_list( $values ) {
+		return array_map( function( $value ) {
+
+			if ( \is_a(  $value,'Tribe\Values\Abstract_Currency' ) ) {
+				return $value;
+			}
+
+			return new self( $value );
+		}, $values );
+	}
```
bordoni, line 39: I am confused into what this method is used for.

src/Tickets/Commerce/Ticket.php
```diff
@@ -843,30 +872,29 @@ public function decrease_ticket_sales_by( $ticket_id, $quantity = 1, $shared_cap
...
+	 * @version TBD
+	 *
 	 */
 	public function get_price_value( $product ) {
-		$product = get_post( $product );
+		$ticket = Models\Ticket_Model::from_post( $product );
 
-		if ( ! $product instanceof \WP_Post ) {
-			return false;
+		if ( ! is_a( $ticket, 'TEC\Tickets\Commerce\Models\Ticket_Model' ) ) {
```
bordoni, line 889: Avoid `is_a`  use `instanceof`

src/Tickets/Commerce/Ticket.php
```diff
@@ -843,30 +872,29 @@ public function decrease_ticket_sales_by( $ticket_id, $quantity = 1, $shared_cap
...
 	 * @param int|\WP_Post $product
 	 *
-	 * @return string
+	 * @return Commerce\Utils\Value;
+	 * @version TBD
+	 *
 	 */
 	public function get_price_value( $product ) {
-		$product = get_post( $product );
+		$ticket = Models\Ticket_Model::from_post( $product );
```
bordoni, line 887: Use `tec_tc_get_ticket`

src/Tickets/Commerce/Utils/Value.php
```diff
@@ -0,0 +1,89 @@
...
+	public function get_currency_separator_decimal() {
+		return $this->currency_separator_decimal;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function get_currency_separator_thousands() {
+		return $this->currency_separator_thousands;
+	}
```
bordoni, line 74: All of these should be part of the abstract so we can have filtering.

bordoni (changes requested):

bordoni:
Re-Pull Release branch into your branch so you can have all tests working again.

bordoni:
Lets get this merged into Hadrosaurus

src/Tickets/Commerce/Ticket.php
moraleida: @bordoni `tec_tc_get_ticket`  retuns the `WP_Post` object of the ticket, right? We need the model instance here, not the post object, so we can create a new Value object from it.

src/Tickets/Commerce/Utils/Value.php
moraleida: This method takes a list of numeric values and returns a list of Value objects instantiated from those values. This is a helper to facilitate limiting the scope of methods such as `Value->total()` that take a list of objects, instead of a list of primitives.

src/Tribe/Commerce/Currency.php
```diff
@@ -192,159 +192,7 @@ public function format_currency( $cost, $post_id = null ) {
...
-				'thousands_sep' => ',',
-			),
-			'USD' => array(
-				'name'   => __( 'U.S. Dollar (USD)', 'event-tickets' ),
-				'symbol' => '&#x24;',
-				'decimal_point' => '.',
-				'thousands_sep' => ',',
-			),
-		);
+		$default_map = \TEC\Tickets\Commerce\Utils\Currency::get_default_currency_map();
```
bordoni, line 195: Lets keep the old map to avoid future problems.

bordoni (changes requested):

bordoni (approved):

PR #2237, opened 2022-01-10
Improve get_ticket() usage and fallbacks

sc0ttkclark:
### 🎫 Ticket

[TBD] 

### 🗒️ Description

This PR addresses issues when forcing a ticket object as null in Tribe__Tickets__Tickets::get_ticket() usage on each provider. This does not address usage issues on ET+ which will have to be done separately.

I am able to use various get_ticket hooks (per provider) to return null if the ticket should NOT be visible. It checks for the object/null cases in many other areas of the code but there are a few stragglers where it doesn't check yet.

The use case that prompted the need for this is to allow plugins like Paid Memberships Pro to control visibility of individual event tickets based on membership level.

Here's the working snippet I have for this which I tested my changes against to verify no notices were showing up: https://gist.github.com/sc0ttkclark/f561793c30855805c326990efc75707a

### 🎥 Artifacts 

As user who does NOT have a specific membership level: https://share.skc.dev/z8ukx4qZ

As a user who DOES have a specific membership level: https://share.skc.dev/OAu4bWx7

### ✔️ Checklist
- [x] My code is tested. 


bordoni:
Thanks so much for the PR @sc0ttkclark, let me get a ticket logged for Syd to test this soonish. We are over the cut-off for the next release but I will figure that out tomorrow.

rafsuntaskin:
Moved the changes to #2294 to make it work with GHA and update to proper branch.

PR #2253, opened 2022-01-27
Fix TicketsCommerce not sending out ticket emails properly.

rafsuntaskin:
### 🎫 Ticket

[GTRIA-730] 

### 🗒️ Description

https://lw.slack.com/archives/C01T0TSD4SC/p1643235063082800

### ✔️ Checklist
- [x] My code is tested. 

[GTRIA-730]: https://theeventscalendar.atlassian.net/browse/GTRIA-730?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bordoni:
@moraleida check this out we talked about it, and @rafsuntaskin found the issue here. I wanted to make sure we have a solution where priority is part of the equation, not for stripe but for afterwards.

src/Tickets/Commerce/Flag_Actions/Generate_Attendees.php
```diff
@@ -68,6 +68,13 @@ public function get_status_when_to_trigger() {
 	 * @return string[]
 	 */
 	public function modify_status_with_attendee_generation_flag( $flags, $post, $status ) {
+		$found = array_search( 'send_email', $flags, true );
```
moraleida, line 71: @rafsuntaskin let's move forward with this for now, but I think we should approach this issue from another angle. Documenting below so we can discuss later:

instead of rearranging the flags to send the email like we're doing here, we should have the `generate_attendees` and `send_email` flags in both `Pending` and `Complete` statuses. 

And inside the handle method in `Send_Email` we should check whether we are ready to send them. i.e: if we're in the proper status (as defined by the user) and if attendees were already generated. Then proceed.

I beleive this will give us more flexibility to add other statuses in which a user can choose to generate their attendees and send their emails in the future.

moraleida (approved):

src/Tickets/Commerce/Flag_Actions/Generate_Attendees.php
rafsuntaskin: I have created a ticket for this, so that we can work on this later: https://theeventscalendar.atlassian.net/browse/GTRIA-771

PR #2263, opened 2022-02-02
[ET-1385] Enabled default will be true if Gateway is connected

codingmusician:
### 🎫 Ticket

[ET-1385] 

### 🗒️ Description

This applies to PayPal, now, but could apply to other gateways in the future, if/when we add more. Users that are already using PayPal should automatically see that PayPal is "enabled" as long as they are connected to PayPal already.

### 🎥 Artifacts 

https://youtu.be/2Pu8E5PSm4s

### ✔️ Checklist
- [x] My code is tested. 

[ET-1385]: https://theeventscalendar.atlassian.net/browse/ET-1385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bordoni (dismissed):

codingmusician:
I had to make a change. We only want PayPal to look at the "connected" state when "enabled" isn't explicitly set.

src/Tickets/Commerce/Gateways/PayPal/Gateway.php
```diff
@@ -128,4 +128,21 @@ public function get_logo_url() {
...
+	
+	/**
+	 * @inheritDoc
+	 */
+	public static function is_enabled() {
+		if ( ! static::should_show() ) {
+			return false;
+		}
+		
+		$option_value = tribe_get_option( static::get_enabled_option_key() );
```
juanfra, line 140: Should we have the default as false?

juanfra: I mean, what would happen with a site that is installing ET for the first time?

src/Tickets/Commerce/Gateways/PayPal/Gateway.php
```diff
@@ -128,4 +128,21 @@ public function get_logo_url() {
...
+		if ( '' !== $option_value ) {
```
moraleida, line 141: this is a boolean setting, so if it's not set, shouldn't we return false instead of checking for a connection?

src/Tickets/Commerce/Gateways/PayPal/Gateway.php
codingmusician: The default is '' when you don't pass one, so I check for '' explicitly in the next line. I'm actually looking for the scenario of when the `enabled` flag isn't explicitly set. That would include first installs and upgrades to the newest version, since that flag didn't exist. I guess checking `false === $option` would serve the same purpose, but I just chose the one that had less code.

src/Tickets/Commerce/Gateways/PayPal/Gateway.php
codingmusician: The purpose of this is to serve those that were already using PayPal in Tickets Commerce prior to having an `enabled` flag to set, so we're setting the default to rely on the connection status. When they upgrade to the newest version of ET, we don't want their sales to be disrupted.

codingmusician: If the `enabled` flag IS set, it returns either a 1 or 0. That's why I'm looking explicitly for ''

src/Tickets/Commerce/Gateways/PayPal/Gateway.php
juanfra: Perfect!

juanfra (approved):

PR #2261, opened 2022-02-02
Fix/stripe minor fixes

moraleida:
### 🎫 Ticket

[ET-1371] 

### ✔️ Checklist
- [ ] I've included a changelog entry in the readme.txt file. 
- [ ] My code is tested. 
- [ ] My code has [proper inline documentation](https://the-events-calendar.github.io/products-engineering/docs/code-standards/).

[ET-1371]: https://theeventscalendar.atlassian.net/browse/ET-1371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2264, opened 2022-02-02
Feature/stripe tests and error handling

moraleida:
### 🎫 Ticket

[ET-1371] 

### 🗒️ Description

contains #2261 

### ✔️ Checklist
- [ ] My code is tested. 


src/Tickets/Commerce/Gateways/Stripe/Client.php
```diff
@@ -34,6 +35,15 @@ class Client {
...
 	public $payment_intent_transient_name;
 
+	/**
+	 * The percentage applied to Stripe transactions. Currently set at 2%.
+	 *
+	 * @since TBD
+	 *
+	 * @var float
+	 */
+	public $application_fee_percentage = 0.02;
```
bordoni, line 45: Can we move this to static protected value and use a static public method to pull it?

src/Tickets/Commerce/Gateways/Stripe/Client.php
```diff
@@ -232,9 +247,7 @@ public function get_payment_intent_transient() {
 	 * @param array $payment_intent payment intent data from Stripe
 	 */
 	public function store_payment_intent( $payment_intent ) {
-		if ( ! empty( $payment_intent['client_secret'] ) ) {
-			set_transient( $this->get_payment_intent_transient_name(), $payment_intent, 6 * HOUR_IN_SECONDS );
-		}
+		set_transient( $this->get_payment_intent_transient_name(), $payment_intent, 6 * HOUR_IN_SECONDS );
```
bordoni, line 250: I am confused, i though we spoke about moving this to the Order so we can avoid saving both the PaymentIntent and the Cart Hash. If we create an order as soon as we create the payment intent we shouldn't have a problem fetching the data using the ID.

src/Tickets/Commerce/Gateways/Stripe/Client.php
```diff
@@ -499,4 +508,81 @@ public function delete( $endpoint, array $query_args = [], array $request_argume
...
+				if ( ! empty( $body->error ) ) {
+					return new \WP_Error( $response['response']['code'], $body->error->message, $body->error );
+				}
+			}
+		}
+
+		return $response;
+	}
+
+	public function prepare_errors_to_display( \WP_Error $errors ) {
```
bordoni, line 541: Docblock.

src/Tickets/Commerce/Gateways/Stripe/Hooks.php
```diff
@@ -114,7 +115,7 @@ public function include_payment_buttons( $file, $name, $template ) {
 	 */
 	public function maybe_create_stripe_payment_intent() {
 
-		if ( ! tribe( Merchant::class )->is_connected() ) {
+		if ( ! tribe( Merchant::class )->is_connected() && tribe( Module::class )->is_checkout_page() ) {
```
bordoni, line 118: We should probably create the order here and the payment intent from the cart and avoid refreshing it.

bordoni (changes requested):

src/Tickets/Commerce/Gateways/Stripe/Client.php
moraleida: @bordoni you are correct, however, during implementation I faced an issue of Orders being too tightly coupled with the gateway provider - in that, to create an order without specifying a provider, we would need to refactor the way orders work. I think it's still worth pursuing, but not in the scope of this work.

Instead, what I did was create a payment intent on page-load and reload by the ID received from the front-end when creating the order, instead of relying only on the transients and naming structure on the back-end. This way, if the user pays with Stripe, they will always use the initial payment intent we created, so we won't be getting as many unused PIs as we were before. We still need a strategy to cancel these, but it's no longer as big of a deal as it was when we discussed this.

src/Tickets/Commerce/Gateways/Stripe/Hooks.php
moraleida: See my comment above. To avoid refactoring the Orders, we're creating the payment intent here and then retrieving the same PI from the Stripe API when creating the Order.

rafsuntaskin, line 118: @moraleida I assume you wanted to skip all other pages but the checkout page, so the logic should be: ` ! tribe( Module::class )->is_checkout_page() `

Otherwise, I am getting this error on all pages:
```[08-Feb-2022 10:30:55 UTC] PHP Notice:  Undefined property: stdClass::$code in /app/public/wp-content/plugins/event-tickets/src/Tickets/Commerce/Gateways/Stripe/Client.php on line 562
[08-Feb-2022 10:30:55 UTC] PHP Stack trace:
[08-Feb-2022 10:30:55 UTC] PHP   1. {main}() /app/public/index.php:0
[08-Feb-2022 10:30:55 UTC] PHP   2. require() /app/public/index.php:17
[08-Feb-2022 10:30:55 UTC] PHP   3. wp() /app/public/wp-blog-header.php:16
[08-Feb-2022 10:30:55 UTC] PHP   4. WP->main() /app/public/wp-includes/functions.php:1310
[08-Feb-2022 10:30:55 UTC] PHP   5. do_action_ref_array() /app/public/wp-includes/class-wp.php:771
[08-Feb-2022 10:30:55 UTC] PHP   6. WP_Hook->do_action() /app/public/wp-includes/plugin.php:522
[08-Feb-2022 10:30:55 UTC] PHP   7. WP_Hook->apply_filters() /app/public/wp-includes/class-wp-hook.php:331
[08-Feb-2022 10:30:55 UTC] PHP   8. TEC\Tickets\Commerce\Gateways\Stripe\Hooks->maybe_create_stripe_payment_intent() /app/public/wp-includes/class-wp-hook.php:307
[08-Feb-2022 10:30:55 UTC] PHP   9. TEC\Tickets\Commerce\Gateways\Stripe\Client->create_payment_intent() /app/public/wp-content/plugins/event-tickets/src/Tickets/Commerce/Gateways/Stripe/Hooks.php:108
[08-Feb-2022 10:30:55 UTC] PHP  10. TEC\Tickets\Commerce\Gateways\Stripe\Client->post() /app/public/wp-content/plugins/event-tickets/src/Tickets/Commerce/Gateways/Stripe/Client.php:143
[08-Feb-2022 10:30:55 UTC] PHP  11. TEC\Tickets\Commerce\Gateways\Stripe\Client->request() /app/public/wp-content/plugins/event-tickets/src/Tickets/Commerce/Gateways/Stripe/Client.php:480
[08-Feb-2022 10:30:55 UTC] PHP  12. TEC\Tickets\Commerce\Gateways\Stripe\Client->prepare_errors_to_display() /app/public/wp-content/plugins/event-tickets/src/Tickets/Commerce/Gateways/Stripe/Client.php:413```

PR #2265, opened 2022-02-03
Modify the Stripe JS to better adhere to our Standards

bordoni:
### 🗒️ Description

Took a stab at the Code from to make the JS follow the ES6 standards that we are trying to move to and some small fixes that were triggering Notices.

### DIFF
--- a/src/resources/js/commerce/gateway/stripe/checkout.js
+++ b/src/resources/js/commerce/gateway/stripe/checkout.js
@@ -1,16 +1,6 @@
 /* global tribe, jQuery, Stripe, tecTicketsCommerceGatewayStripeCheckout */
 
 /**
- * Minified by jsDelivr using Terser v5.3.5.
- * Original file: /npm/ky@0.27.0/index.js
- *
- * Do NOT use SRI with dynamically generated files! More information: https://www.jsdelivr.com/using-sri-with-dynamic-files
- */
-/*! MIT License © Sindre Sorhus */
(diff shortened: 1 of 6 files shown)


src/resources/js/commerce/gateway/stripe/checkout.js
```diff
@@ -206,60 +202,84 @@ tribe.tickets.commerce.gateway.stripe.checkout = {};
...
+	 * @since TBD
+	 *
+	 * @param result
+	 */
+	obj.handleConfirmPayment = ( result ) => {
+		obj.submitButton( true );
+		if ( result.error ) {
+			obj.handlePaymentError( result );
+		} else {
+			// @todo feels like we are missing a error handling here.
```
moraleida, line 240: no exactly error handling, but some cases in which the payment method returns a partial success, like a status of `pending_confirm` and some others.

moraleida (approved):
Great stuff. Some of these changes are also in #2264 so it would be good to merge this before that one, I think

src/resources/js/commerce/gateway/stripe/checkout.js
```diff
@@ -377,4 +424,4 @@ tribe.tickets.commerce.gateway.stripe.checkout = {};
 	};
 
 	$( obj.ready );
-})( jQuery, tribe.tickets.commerce, Stripe, ky );
+})( jQuery, tribe.tickets.commerce, Stripe, tribe.ky );
```
juanfra, line 427: The object here should be `tribe.tickets.commerce.gateway.stripe.checkout`

juanfra (commented):
Changes look good!

src/resources/js/commerce/gateway/stripe/checkout.js
moraleida: @juanfra this fix was edded to #2264

PR #2272, opened 2022-02-04
Setup stripe fees

moraleida:
### 🎫 Ticket

[ET-1380] 

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1380]: https://theeventscalendar.atlassian.net/browse/ET-1380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Gateways/Stripe/Settings.php
+++ b/src/Tickets/Commerce/Gateways/Stripe/Settings.php
@@ -92,20 +92,20 @@ class Settings extends Abstract_Settings {
 	 */
 	public function get_settings() {
 		$settings = [
-			'tickets-commerce-stripe-commerce-configure'        => [
+			'tickets-commerce-stripe-commerce-configure'             => [
 				'type'            => 'wrapped_html',
 				'html'            => $this->get_connection_settings_html(),
 				'validation_type' => 'html',
 			],
-			'tickets-commerce-stripe-settings-heading'          => [
+			'tickets-commerce-stripe-settings-heading'               => [
 				'type' => 'html',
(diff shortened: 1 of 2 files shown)


src/Tickets/Commerce/Gateways/Stripe/Settings.php
```diff
@@ -331,8 +331,29 @@ public function get_connection_settings_html() {
...
+	 *
+	 * @return bool
+	 */
+	public function should_apply_stripe_fee( $revalidate = false ) {
+
+		if ( class_exists( 'Tribe__Tickets_Plus__PUE' ) ) {
+			if ( tribe( \Tribe__Tickets_Plus__PUE::class )->is_current_license_valid( $revalidate ) ) {
+				return false;
+			}
+		}
```
juanfra, line 355: We should try to avoid having ET+ code on the free version. I feel the way to go here would be having a filter and move this method to ET+, hooking here.

moraleida: The problem with setting a filter here is that any user would be able to hook into it from their theme and return false, and then remove the fees without having a license.

bordoni: I agree with @moraleida in this case it's ok to have it not be a filter. But lets invert the method here. So it's not a nestted conditional.

juanfra: If they do that they deserve not to pay fees, lol! Your point is valid.

moraleida:
Closing this in favor of #2273

PR #2288, opened 2022-02-08
[ET-1369] Checkout Toggle: Some JS Tweaks

codingmusician:
### 🎫 Ticket

[ET-1369] 

### 🗒️ Description

Made some changes for accessibility and removed the animation.

### 🎥 Artifacts 

https://i.imgur.com/jK5ea2A.gif

### ✔️ Checklist
- [x] My code is tested. 

[ET-1369]: https://theeventscalendar.atlassian.net/browse/ET-1369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/resources/js/commerce/gateway/toggler.js
```diff
@@ -122,7 +113,7 @@ tribe.tickets.commerce.gateway = tribe.tickets.commerce.gateway || {};
 	 * @param {Element} gateway Gateway element to show.
 	 */
 	obj.toggler.showGateway = gateway => {
-		$( gateway ).show( obj.toggler.toggleDuration );
+		$( gateway ).show().attr('aria-expanded', 'true');
```
bordoni, line 116:
```suggestion
		$( gateway ).show().attr( 'aria-expanded', 'true' );
```

src/resources/js/commerce/gateway/toggler.js
```diff
@@ -133,7 +124,7 @@ tribe.tickets.commerce.gateway = tribe.tickets.commerce.gateway || {};
 	 * @param {Element} gateway Gateway element to hide.
 	 */
 	obj.toggler.hideGateway = gateway => {
-		$( gateway ).hide( obj.toggler.toggleDuration );
+		$( gateway ).hide().attr('aria-expanded', 'false');
```
bordoni, line 127:
```suggestion
		$( gateway ).hide().attr( 'aria-expanded', 'false' );
```

src/resources/js/commerce/gateway/toggler.js
```diff
@@ -167,7 +158,7 @@ tribe.tickets.commerce.gateway = tribe.tickets.commerce.gateway || {};
 			obj.toggler.showGateway( gateway );
 		});
 		obj.toggler.toggles.additional.addClass(obj.toggler.classes.toggleOpen);
-		obj.toggler.toggles.default.removeClass(obj.toggler.classes.toggleOpen).show();
+		obj.toggler.toggles.default.removeClass(obj.toggler.classes.toggleOpen).show().attr('aria-hidden', 'false');
```
bordoni, line 161:
```suggestion
		obj.toggler.toggles.default.removeClass(obj.toggler.classes.toggleOpen).show().attr( 'aria-hidden', 'false' );
```

src/resources/js/commerce/gateway/toggler.js
```diff
@@ -149,7 +140,7 @@ tribe.tickets.commerce.gateway = tribe.tickets.commerce.gateway || {};
 			}
 			obj.toggler.hideGateway( gateway );
 		});
-		obj.toggler.toggles.default.addClass(obj.toggler.classes.toggleOpen).hide();
+		obj.toggler.toggles.default.addClass(obj.toggler.classes.toggleOpen).hide().attr('aria-hidden', 'true');
```
bordoni, line 143:
```suggestion
		obj.toggler.toggles.default.addClass(obj.toggler.classes.toggleOpen).hide().attr( 'aria-hidden', 'true' );
```

bordoni (changes requested):

juanfra (commented):
Looking good! I'd try to also use the `aria-selected` on the "toggler buttons".

I'd also try to toggle the a11y classes we have (`.tribe-common-a11y-hidden`), apart from using the show/hide. Because, with how things are now, the container can be not visible in the browser but it's still visible for screen readers.

bordoni (changes requested):
Spacing is missing spaces on `()`

src/resources/js/commerce/gateway/toggler.js
```diff
@@ -72,6 +72,7 @@ tribe.tickets.commerce.gateway = tribe.tickets.commerce.gateway || {};
 		toggle: 'tribe-tickets__commerce-checkout-gateway-toggle',
 		toggleOpen: 'tribe-tickets__commerce-checkout-gateway-toggle--open',
 		toggleButton: 'tribe-tickets__commerce-checkout-gateway-toggle-button',
+		toggleHidden: 'tribe-common-a11y-hidden',
```
juanfra, line 75: Normally, [when we use class names or selectors](https://github.com/the-events-calendar/event-tickets/blob/master/src/resources/js/v2/tickets-block.js#L41) we include the dot `.`. And then we have a prototype function which is `.className()` to get it without the dot.

i.e.: `obj.toggler.classes.toggleHidden.className()`.

I think the best idea would be to have consistency with this and keep doing things like we're doing it on the other libraries. (also the key that it's normally used for the a11y class is `hiddenElement` to make it more abstract and not only related to the toggle.)

src/resources/js/commerce/gateway/toggler.js
```diff
@@ -113,7 +114,10 @@ tribe.tickets.commerce.gateway = tribe.tickets.commerce.gateway || {};
 	 * @param {Element} gateway Gateway element to show.
 	 */
 	obj.toggler.showGateway = gateway => {
```
juanfra, line 116: I'm not entirely sure why we use the `obj.toggler` prefix. We could remove the `toggler` part here, and addit in the obj (the last line, make it `})( jQuery, tribe.tickets.commerce.gateway.toggler );`)

We'll also need to define the toggler beforehand, up in the code: `tribe.tickets.commerce.gateway.toggler = tribe.tickets.commerce.gateway.toggler || {};`

I think we should keep doing that for consistency between libraries.

src/resources/js/commerce/gateway/toggler.js
```diff
@@ -141,7 +148,7 @@ tribe.tickets.commerce.gateway = tribe.tickets.commerce.gateway || {};
 			obj.toggler.hideGateway( gateway );
 		});
 		obj.toggler.toggles.default.addClass(obj.toggler.classes.toggleOpen).hide().attr( 'aria-hidden', 'true' );
-		obj.toggler.toggles.additional.removeClass(obj.toggler.classes.toggleOpen);
+		obj.toggler.toggles.additional.removeClass(obj.toggler.classes.toggleOpen).attr( 'aria-selected', 'false' );
```
juanfra: If I'm not mistaken we're applying this to the `containers` of all the additional gateways which are not the default, no?

The aria-selected should be applied to the "opener" and not the container.

src/resources/js/commerce/gateway/toggler.js
```diff
@@ -157,7 +164,7 @@ tribe.tickets.commerce.gateway = tribe.tickets.commerce.gateway || {};
 			}
 			obj.toggler.showGateway( gateway );
 		});
-		obj.toggler.toggles.additional.addClass(obj.toggler.classes.toggleOpen);
+		obj.toggler.toggles.additional.addClass(obj.toggler.classes.toggleOpen).attr( 'aria-selected', 'true' );
```
juanfra: If I'm not mistaken we're applying this to the `containers` of all the additional gateways which are not the default, no? I may be reading this the wrong way.  The aria-selected should be applied to the "opener" and not the container.

src/resources/js/commerce/gateway/toggler.js
codingmusician: Nice! Didn't know about that. Will implement it.

src/resources/js/commerce/gateway/toggler.js
codingmusician: Good idea. Thanks!

src/resources/js/commerce/gateway/toggler.js
codingmusician: This is actually for the toggle button itself. The toggle button is the `opener`

src/resources/js/commerce/gateway/toggler.js
codingmusician: This is actually for the toggle button itself. The toggle button is the `opener`

bordoni (approved):

PR #2295, opened 2022-02-09
[ET-1400] Remove usd from klarna

juanfra:
### 🎫 Ticket

[ET-1400] 

### 🗒️ Description

- [Stripe's documentation](https://stripe.com/docs/payments/payment-methods/integration-options) is wrong. They're listing that klarna is available for USD, but when testing we faced the following error: 

Screenshot on 2022-02-08 at 11-07-13

### 🎥 Artifacts 

N/A

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1400]: https://theeventscalendar.atlassian.net/browse/ET-1400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Gateways/Stripe/Settings.php
+++ b/src/Tickets/Commerce/Gateways/Stripe/Settings.php
@@ -337,7 +337,7 @@ class Settings extends Abstract_Settings {
 				'label'      => esc_html__( 'Giropay', 'event-tickets' ),
 			],
 			'klarna'            => [
-				'currencies' => [ 'DKK', 'EUR', 'GBP', 'NOK', 'SEK', 'USD' ],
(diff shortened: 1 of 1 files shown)


moraleida (approved):
I've seen documentation errors like this a few times now. We need to get a better list of supported methods and currencies than what is available in https://stripe.com/docs/payments/payment-methods/integration-options#payment-method-product-support

We should also add another check to make sure the merchant country matches the list in the link above.

juanfra:
Yes, it's a bit frightening that we can't rely on their documentation and as a result, we may experience the paymentElement not being created.

PR #2320, opened 2022-02-14
Spotfix - Updated Notice component

rafsuntaskin:
### 🎫 Ticket

[TICKET_ID] 

### 🗒️ Description
* make use of `tribe-common-a11y-hidden` instead of jQuery `show` & `hide`

### 🎥 Artifacts 
🎥 screencast(s): https://d.pr/v/SKSWkq

### ✔️ Checklist
- [x] My code is tested. 

### DIFF
--- a/src/resources/js/commerce/notice.js
+++ b/src/resources/js/commerce/notice.js
@@ -67,7 +67,7 @@ tribe.tickets.commerce.notice = tribe.tickets.commerce.notice || {};
 
 		$item.trigger( 'beforeShowNotice.tecTicketsCommerce', [ $container ] );
 
-		$item.show();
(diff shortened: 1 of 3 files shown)


juanfra:
❤️ Thanks so much for opening this one! It's looking great @rafsuntaskin 

The thing I'm thinking now is that I'm not sure if it'd be risky to include this one just now without testing other places where we make use of the notice script. Maybe we have to update the scripts for AR related stuff? I think it'd be best if we take care of this one on the next sprint, being sure that everything is working on every place where we have it. What do you think?

rafsuntaskin:
> 

I totally agree with you on this, I was thinking the same. I just had it prepared while we discussed the changes in the other PR. Let's hold it for now.

PR #2329, opened 2022-02-15
Spotfix - Stripe checkout - fix loader show and hide timing

rafsuntaskin:
### 🎫 Ticket

[TICKET_ID] 

### 🗒️ Description

* Loader was shown on click but got hidden 1-2s before the redirect is taking place which might be confusing.
* Hide the Purchase tickets button if the payment intent has errors.

### ✔️ Checklist
- [ ] My code is tested. 

### DIFF
--- a/src/resources/js/commerce/gateway/stripe/checkout.js
+++ b/src/resources/js/commerce/gateway/stripe/checkout.js
@@ -46,7 +46,8 @@ tribe.tickets.commerce.gateway.stripe.checkout = {};
 		cardErrors: '#tec-tc-gateway-stripe-errors',
 		paymentElement: '#tec-tc-gateway-stripe-payment-element',
 		paymentMessage: '#tec-tc-gateway-stripe-payment-message',
-		submitButton: '#tec-tc-gateway-stripe-checkout-button'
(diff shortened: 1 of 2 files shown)


bordoni (dismissed):

src/resources/js/commerce/gateway/stripe/checkout.js
```diff
@@ -66,6 +67,15 @@ tribe.tickets.commerce.gateway.stripe.checkout = {};
...
 	obj.stripeElements = null;
+	
+	/**
+	 * Loader container.
+	 *
+	 * @since TBD
+	 *
+	 * @type {Object|null}
+	 */
+	obj.loaderContainer = null;
```
juanfra, line 78: For the loader we're showing, we're using the checkout container as the container. I think in order to be more clear this should be called `checkoutContainer`. But also, I'm not sure why is that we're saving this here rather than getting the container on the fly?

rafsuntaskin: This is to just avoid slow jQuery execution if the page has a lot of HTML elements. We are just looking for the same element each time, so caching it should not be a problem.

juanfra (approved):

PR #2346, opened 2022-02-21
[ET-1418] TicketsCommerce Checkout unavailable notice logic fixed for 0 available gateways.

rafsuntaskin:
### 🎫 Ticket

[ET-1418] 

### 🗒️ Description

* `gateways_enabled` logic was flawed as it was only counting the enabled gateways irrespective of the individual gateway.
* We had the generic `Checkout Unavailable` notice already in place but it was not showing for the following case:
If you have Stripe _Connected_ but _not_ _enabled_ and PayPal _not connected_ but _enabled_ then the `gateways_enabled` was returning `1` which should have been 0.
* So, I renamed the `get_gateways_connected` method to `get_available_gateways` for clarity and modified the logic to check for `is_enabled` as well.
* Remove the redundant `gateways_enabled` argument.

### ✔️ Checklist
- [x] My code is tested. 

[ET-1418]: https://theeventscalendar.atlassian.net/browse/ET-1418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
+++ b/src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
@@ -60,7 +60,7 @@ class Checkout_Shortcode extends Shortcode_Abstract {
 			'is_tec_active'      => defined( 'TRIBE_EVENTS_FILE' ) && class_exists( 'Tribe__Events__Main' ),
 			'gateways'           => $gateways,
 			'gateways_active'    => $this->get_gateways_active(),
-			'gateways_connected' => $this->get_gateways_available(),
+			'gateways_connected' => $this->get_gateways_connected(),
 		];
 
 		$this->template_vars = $args;
--- a/src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
+++ b/src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
@@ -101,9 +101,21 @@ class Checkout_Shortcode extends Shortcode_Abstract {
 	}
 
 	/**
-	 * Get the number of available gateways that can be used for checkout. A gateway is only available if it's properly connected and enabled.
+	 * Get the number of connected gateways.
 	 *
 	 * @since 5.2.0
+	 *
+	 * @return int The number of connected gateways.
+	 */
+	public function get_gateways_connected() {
+		_deprecated_function( __METHOD__, 'TBD', __CLASS__ . '::get_gateways_available' );
+		return $this->get_gateways_available();
+	}
+
+	/**
+	 * Get the number of available gateways that can be used for checkout. A gateway is only available if it's properly connected and enabled.
+	 *
+	 * @since TBD
 	 *
 	 * @return int The number of available gateways.
 	 */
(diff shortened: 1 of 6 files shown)


src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
```diff
@@ -60,15 +60,8 @@ public function setup_template_vars() {
 			'is_tec_active'      => defined( 'TRIBE_EVENTS_FILE' ) && class_exists( 'Tribe__Events__Main' ),
 			'gateways'           => $gateways,
 			'gateways_active'    => $this->get_gateways_active(),
-			'gateways_connected' => $this->get_gateways_connected(),
-			'gateways_enabled' => 0,
+			'gateways_connected' => $this->get_available_gateways(),
```
juanfra, line 63: Not sure if we can just completely remove a public method. It'd also be better if we're consistent with the naming. If we're using `get_gateways_*()`, let's keep on using that format instead of switching one to `get_*_gateways()`

juanfra (commented):
Minor comments, looking good!

src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
```diff
@@ -108,17 +101,17 @@ public function get_gateways_active() {
...
-	 * Get the number of connected gateways.
+	 * Get the number of available gateways that can be used for checkout. A gateway is only available if it's properly connected and enabled.
 	 *
 	 * @since 5.2.0
 	 *
-	 * @return int The number of connected gateways.
+	 * @return int The number of available gateways.
 	 */
-	public function get_gateways_connected() {
+	public function get_gateways_available() {
```
juanfra, line 110: I am not sure if we should get rid of a public method that was in place, as it can be already being used by other people. Maybe we should create a new one for available and keep having the old one in place.

src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
```diff
@@ -115,10 +108,22 @@ public function get_gateways_active() {
 	 * @return int The number of connected gateways.
 	 */
 	public function get_gateways_connected() {
+		_deprecated_function( __METHOD__, 'TBD', __CLASS__ . '::get_gateways_available' );
+		return $this->get_gateways_available();
```
juanfra, line 112: Let's not deprecate this as we may want to keep the method to know how many gateways are connected. Just have the available in a new one and that's it.

src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
```diff
@@ -115,10 +108,21 @@ public function get_gateways_active() {
 	 * @return int The number of connected gateways.
 	 */
 	public function get_gateways_connected() {
+		return $this->get_gateways_available();
```
juanfra, line 111: @rafsuntaskin please keep the old code of gateways connected here. We handle the available, but we may use the number of connected gateways as well. So we should keep the logic of the connected gateways in place.

juanfra (approved):

PR #2365, opened 2022-03-17
[ET-1426] Tickets Commerce - Fix JS Errors on Checkout due to Assets Loading Improperly

codingmusician:
### 🎫 Ticket

[ET-1426] 

### 🗒️ Description

It was found that the PayPal and Stripe JS assets were loading on the checkout page even when they were disabled or disconnected. It was found that loading assets using `tribe_asset_enqueue_group()` was the culprit and was bypassing the `conditionals` parameter. Using an action filter and adding an `is_enabled()` check in the `conditionals` function solved the issue. 

### 🎥 Artifacts 

https://i.imgur.com/F5B5rnA.png

### ✔️ Checklist
- [x] My code is tested. 

[ET-1426]: https://theeventscalendar.atlassian.net/browse/ET-1426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
+++ b/src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
@@ -124,6 +124,7 @@ class Checkout_Shortcode extends Shortcode_Abstract {
 	 */
 	public static function enqueue_assets() {
 		// Enqueue assets.
+		do_action( 'tec-tickets-commerce-checkout-shortcode-assets' );
 		tribe_asset_enqueue_group( 'tribe-tickets-commerce-checkout' );
 		tribe_asset_enqueue( 'tribe-tickets-forms-style' );
 	}
--- a/readme.txt
+++ b/readme.txt
@@ -192,6 +192,10 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 == Changelog ==
 
+= [TBD] TBD =
+
+* Fix - Fixed JS assets loading and causing errors on checkout page for Tickets Commerce. [ET-1426]
+
 = [5.3.1] 2022-03-15 =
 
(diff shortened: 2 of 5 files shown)


src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
```diff
@@ -124,6 +124,7 @@ public function get_gateways_connected() {
 	 */
 	public static function enqueue_assets() {
 		// Enqueue assets.
+		do_action( 'tec-tickets-commerce-checkout-shortcode-assets' );
```
juanfra, line 127: I'm curious why are we adding a `do_action` here?

codingmusician: By removing the asset from the group and adding an action to load the assets instead, the 'conditionals' param is called.

juanfra: Thanks for the clarification. Yep, I understand that. I'm curious why is that we're having the new action there, instead of using `tribe_asset_enqueue()`.

codingmusician: Because we are loading (possibly) several assets at once. In the future, when/if we add more gateways, having this hook will allow the gateways to tie into the hook easily, rather than us manually calling an asset for each gateway and having to update the code each time.

juanfra: Makes sense!

juanfra (dismissed):

PR #2371, opened 2022-03-23
Validate webhook transactions before erroring out

moraleida:
### 🎫 Ticket

[ET-1474] 

### 🗒️ Description

Before discarding a PI validation as an error, check that it was indeed created in ET so we don't throw errors for PIs created in other applications.

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1474]: https://theeventscalendar.atlassian.net/browse/ET-1474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Gateways/Stripe/Webhooks/Payment_Intent_Webhook.php
+++ b/src/Tickets/Commerce/Gateways/Stripe/Webhooks/Payment_Intent_Webhook.php
@@ -34,6 +34,18 @@ class Payment_Intent_Webhook implements Webhook_Event_Interface {
 
 		if ( empty( $order ) ) {
 
+			if ( empty( $payment_intent['metadata'][ Payment_Intent::$tc_metadata_identifier ] ) ) {
+				$response->set_status( 200 );
+				$response->set_data( sprintf(
+				// Translators: %1$s is the event id and %2$s is the event type name.
+					__( 'Event %1$s was received and will not be handled because the Payment Intent %2$s does not refer to an Event Tickets transaction.', 'event-tickets' ),
+					esc_html( Arr::get( $event, 'id' ) ),
+					esc_html( $payment_intent_id )
+				) );
+
+				return $response;
+			}
+
 			if ( ! empty( $payment_intent['metadata'][ Payment_Intent::$test_metadata_key ] ) ) {
 				$response->set_status( 200 );
 				$response->set_data(
@@ -59,7 +71,7 @@ class Payment_Intent_Webhook implements Webhook_Event_Interface {
 					esc_html( $payment_intent_id )
 				)
 			);
-			
+
 			return $response;
 		}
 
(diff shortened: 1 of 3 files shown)


src/Tickets/Commerce/Gateways/Stripe/Webhooks/Payment_Intent_Webhook.php
```diff
@@ -34,6 +34,18 @@ public static function handle( array $event, Status_Interface $new_status, \WP_R
 
 		if ( empty( $order ) ) {
 
+			if ( empty( $payment_intent['metadata'][ Payment_Intent::$tc_metadata_identifier ] ) ) {
```
rafsuntaskin, line 37: We should check for this at the very beginning of the handler method. If this meta doesn't exist then it doesn't make sense to call `$order = tec_tc_get_order( $payment_intent['metadata']['order_id'] );` and `$order = tribe( Order::class )->get_from_gateway_order_id( $payment_intent_id );`

Also, are we sure we want to send $response back with `200` ? I thought in that case we are going to return `null`.

rafsuntaskin (changes requested):

src/Tickets/Commerce/Gateways/Stripe/Webhooks/Payment_Intent_Webhook.php
```diff
@@ -34,6 +34,18 @@ public static function handle( array $event, Status_Interface $new_status, \WP_R
...
+				$response->set_status( 200 );
+				$response->set_data( sprintf(
+				// Translators: %1$s is the event id and %2$s is the event type name.
+					__( 'Event %1$s was received and will not be handled because the Payment Intent %2$s does not refer to an Event Tickets transaction.', 'event-tickets' ),
+					esc_html( Arr::get( $event, 'id' ) ),
+					esc_html( $payment_intent_id )
+				) );
+
+				return $response;
```
rafsuntaskin, line 46: Also, are we sure we want to send $response back with `200` ? I thought in that case we are going to return `null`.

moraleida: Yes. If we're not handling we need to return 200 otherwise Stripe will keep marking it as a failed event.

src/Tickets/Commerce/Gateways/Stripe/Webhooks/Payment_Intent_Webhook.php
moraleida: This causes an issue when we release because any PI created before this version will not have the proper metadata and will be rejected and clients will lose order updates for those. We can improve on this logic in a later release, when we can be sure all in-flux PIs already have the identifier key.

rafsuntaskin (approved):

PR #2390, opened 2022-04-25
[ET-1492] Upsell Notice - Manual Attendees

codingmusician:
### 🎫 Ticket

[ET-1492] 

### 🗒️ Description

Added a Manual Attendees upsell notice on the attendees page in the admin area.

### 🎥 Artifacts 

https://i.imgur.com/F0YYYwK.png

### ✔️ Checklist
- [x] My code is tested. 

[ET-1492]: https://theeventscalendar.atlassian.net/browse/ET-1492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Admin/Upsell.php
+++ b/src/Tickets/Admin/Upsell.php
@@ -38,7 +38,11 @@ class Upsell {
 			'classes' => [
 				'tec-admin__upsell-event_tickets-capacity_arf'
 			],
-			'text'    => esc_html__( 'Get individual information collection from each attendee and advanced capacity options with' , 'event-tickets' ),
+			'text'    => sprintf(
+				// Translators: %s: Name of "Event Tickets Plus" plugin.
+				esc_html__( 'Get individual information collection from each attendee and advanced capacity options with %s' , 'event-tickets' ),
+				''
+			),
 			'link'    => [
 				'classes' => [
 					'tec-admin__upsell-link--underlined'
--- a/src/Tickets/Admin/Upsell.php
+++ b/src/Tickets/Admin/Upsell.php
@@ -70,7 +70,7 @@ class Upsell {
 				'tec-admin__upsell-tec-tickets-manual-attendees'
 			],
 			'text'    => sprintf(
-				// Translators: %s name of "Event Tickets Plus" plugin.
+				// Translators: %s: Link to "Event Tickets Plus" plugin.
 				esc_html__( 'Manually add attendees with %s' , 'event-tickets' ),
 				''
 			),
--- a/src/Tickets/Admin/Upsell.php
+++ b/src/Tickets/Admin/Upsell.php
@@ -20,6 +20,7 @@ class Upsell {
 	 */
 	public function hooks() {
 		add_action( 'tribe_events_tickets_pre_edit', [ $this, 'maybe_show_capacity_arf' ] );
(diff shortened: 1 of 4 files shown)


codingmusician:
Note: I do have this PR based to the ET-1493 branch because it requires that branch to be merged first.

src/Tickets/Admin/Upsell.php
```diff
@@ -20,6 +20,7 @@ class Upsell {
 	 */
 	public function hooks() {
 		add_action( 'tribe_events_tickets_pre_edit', [ $this, 'maybe_show_capacity_arf' ] );
+		add_action( 'tec_events_tickets_attendees_event_summary_table_extra', [ $this, 'maybe_show_manual_attendees' ] );
```
juanfra, line 23: We need to use `tec_tickets` as the prefix.

src/Tickets/Admin/Upsell.php
```diff
@@ -47,5 +48,33 @@ public function maybe_show_capacity_arf() {
...
+	public function maybe_show_manual_attendees() {
+		// If they already have ET+ activated, then bail.
+		if ( class_exists( 'Tribe__Tickets_Plus__Main' ) ) {
+			return;
+		}
+
+		echo '<div class="welcome-panel-column welcome-panel-extra">';
+		tribe( Upsell_Notice\Main::class )->render( [
+			'classes' => [
+				'tec-admin__upsell-event_tickets-manual_attendees'
```
juanfra, line 66:
```suggestion
				'tec-admin__upsell-tec-tickets-manual-attendees'
```

src/Tickets/Admin/Upsell.php
```diff
@@ -47,5 +48,33 @@ public function maybe_show_capacity_arf() {
...
+			],
+			'text'    => esc_html__( 'Manually add attendees with' , 'event-tickets' ),
```
juanfra, line 68: Maybe this is more related to the component itself, but is there any way that we can make the link a placeholder of the string? Not sure if it'll play well with all languages if we go this route.

src/admin-views/attendees.php
```diff
@@ -161,6 +161,14 @@
 					do_action( 'tribe_events_tickets_attendees_totals_bottom', $event_id );
 					?>
 				</div>
+				<?php 
+				/**
+				 * Fires after the last column so that "extra" content can be displayed.
+				 *
+				 * @param int $event_id
+				 */
+				do_action( 'tec_events_tickets_attendees_event_summary_table_extra', $event_id );
```
juanfra, line 170: The prefix should be `tec_tickets`. We need to add a comment to the `$event_id` param and add a `@since` as we're introducing this action.

juanfra (commented):
Looking good!

src/Tickets/Admin/Upsell.php
codingmusician: Yeah. The component treats the link as a totally different element. Would it be better to maybe use `sprint()` in this case, using 'Manually add attendees with %s', with the string just being a blank string, but a note for translators that `%s` is the name of the Event Tickets Plus plugin?

juanfra: Thanks! Yes, I think it can be better for i18n purposes.

PR #2408, opened 2022-05-17
[ET-1229] JEST Tests - tickets-utils.js

rafsuntaskin:
### 🎫 Ticket

[ET-1229] 

### 🗒️ Description

* Added a new test file for testing `ticket-utils.js`
* updated `package.json` to change the `jest` source directory to scan all `__tests__` directories.
* updated `eslint` config to ignore the test JS.
* Added Tests to make sure the utility methods are formatting and parsing the price numbers as expected.

### ✔️ Checklist
- [x] ~I've included a changelog entry in the readme.txt file.~ Not required. 
- [x] My code is tested. 

[ET-1229]: https://theeventscalendar.atlassian.net/browse/ET-1229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bordoni (dismissed):

package.json
```diff
@@ -36,7 +36,7 @@
       "!src/resources/postcss/tickets-report.pcss",
       "!src/resources/postcss/tickets-v1.pcss"
     ],
-    "jest": ["src/modules/**/__tests__/**/*.js"]
+    "jest": ["src/**/__tests__/**/*.js"]
```
paulmskim, line 39: You can set this to the same paths as what you have in the github workflow:

```suggestion
    "jest": [
      "src/modules/**/__tests__/**/*.js",
      "src/resources/**/__tests__/**/*.js"
    ]
```

src/resources/js/.eslintrc
```diff
@@ -27,5 +27,6 @@
     "tecTicketsCommerceData": true,
     "tecTicketsCommerceGatewayPayPalCheckout": true,
     "paypal": true
-  }
+  },
+  "ignorePatterns": ["**/*.test.js"]
```
paulmskim, line 31: What is the reason for this? If it's for the `tickets-utils.test.js`, I'd suggest either fixing the lint issues, disabling the eslint for the specific rows where we can't fix it, or disabling it for the entire file if there are too many that you cannot fix.

src/resources/js/v2/__tests__/tickets-utils.test.js
```diff
@@ -0,0 +1,117 @@
+global.TribeCartEndpoint = { url : 'test' };
```
paulmskim, line 1: There are a number of small formatting things in this file that may be causing eslint issues.

paulmskim (commented):
The tests generally look good, just a few small things

src/resources/js/.eslintrc
rafsuntaskin: There were a lot of linting errors that were showing up. Many of them were just for using arrow functions. So, not sure what linting config I need to change.

I have disabled the ignore pattern and the errors show up in the lint action: https://github.com/the-events-calendar/event-tickets/runs/6485383345?check_suite_focus=true

Let me know what changes I may make around the config or code to resolve these. Thanks for the help.

paulmskim: For the `no-undef` error on globals, see how it's done here: https://github.com/the-events-calendar/event-tickets/blob/fd5054e198c58a7d26d6e4a129ed976b4a891b4b/src/resources/js/tickets-attendees.js#L1
For the `es5/no-arrow-functions` error, it looks like the file you are testing is ES5? I'd say determine whether you write tests in ES5 or ES6 for files that are written in ES5, and if you do decide to continue writing tests in ES6, write on the top of the file `/* eslint-disable es5/no-arrow-functions */`.
Same thing with the `es5/no-destructuring` and `es5/no-shorthand-properties` errors.

rafsuntaskin: I tried adding the global comment previously as you have referred but that results in more errors of `no-redeclare` and `unused-vars`: https://d.pr/i/S6ZUFA

Now, I just ended up suppressing all the rules that came up which looks weird on top, but I am thinking it is much better to ignore the whole test file from the `.eslintrc` config instead.

The only confusion I now have is, why is it still triggering the `es5` rules? is it the `require` causing this?
I want to write tests only using `es6` as going forward, we are only going to follow `es6` on every new JS file.

rafsuntaskin: @paulmskim 
Also, the lint is failing for `async` keyword here: https://github.com/the-events-calendar/event-tickets/blob/07eb414c803b45f154449e74ff265b35f8ce9b64/src/resources/js/admin/gateway/stripe/webhooks.js#L102

I tried changing configs but I was not able to resolve this.

paulmskim: The reason why you're getting these lint errors is because for `src/resources/js/**/*.js` files, it uses a different config from `src/modules/**/*.js` files. If the consensus is to move from es5 to es6 in `src/resources/js/**/*.js` files, then it may be worth changing the config files in product taskmaster so that it allows es6 syntax.

For the async stuff, you could probably tackle it in another PR, since it doesn't really affect this task.

src/modules/blocks/ticket/container-content/attendee-collection/iac-setting/__tests__/template.test.js
```diff
@@ -6,7 +6,8 @@ import React from 'react';
 import IACSetting from './../template';
 
 describe( 'IACSetting', () => {
-	test( 'Render the component with no errors', () => {
+	//Todo @juanfra @rafsuntaskin fix this test from failing.
```
juanfra, line 9:
```suggestion
	// @todo @juanfra @rafsuntaskin fix this test from failing.
```

juanfra (approved):
🚀

PR #2431, opened 2022-06-14
[ET-1528] Don't Promoter ET+ if it's Already Installed

codingmusician:
### 🎫 Ticket

[ET-1528] 

### 🗒️ Description

On the Tickets -> Settings page, we are promoting ET+ even when it's already installed. We're adding in logic to remove the mention of ET+ when it's installed and activated already.

### 🎥 Artifacts 

image

### ✔️ Checklist
- [x] My code is tested. 

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -188,6 +188,10 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 == Changelog ==
 
+= [5.4.2] TBD =
+
+* Fix - No need to promoter ET+ if it's already installed. [ET-1528]
+
 = [5.4.1] 2022-06-08 =
 
 * Enhancement - Expanded list of supported currencies for Tickets Commerce, for details visit: https://evnt.is/tec-tc-currencies. [ET-1454, ET-1455, ET-1456]
--- a/src/Tickets/Commerce/Payments_Tab.php
+++ b/src/Tickets/Commerce/Payments_Tab.php
@@ -294,17 +294,25 @@ class Payments_Tab extends tad_DI52_ServiceProvider {
 	public function get_tickets_commerce_section_fields() {
 		$fields = [];
 
-		// If no gateway section is selected, show main settings.
-		$plus_link = sprintf(
-			'<a href="https://evnt.is/19zl" target="_blank" rel="noopener noreferrer">%s</a>',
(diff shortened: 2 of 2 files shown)


bordoni (dismissed):

readme.txt
```diff
@@ -188,6 +188,10 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 == Changelog ==
 
+= [5.4.2] TBD =
+
+* Fix - No need to promoter ET+ if it's already installed. [ET-1528]
```
juanfra, line 193: `promoter`. Maybe we can try to be more specific with the readme entry? For example, explaining where the change is: "Removed Event Tickets Plus upsell text on Tickets Commerce when Event Tickets Plus is installed and active".

src/Tickets/Commerce/Payments_Tab.php
```diff
@@ -294,17 +294,25 @@ public function get_section_gateway() {
...
+			esc_html_x( 'Tickets Commerce provides a simple and flexible ecommerce checkout for purchasing tickets. Just choose your payment gateway and configure checkout options and you\'re all set.', 'about Tickets Commerce', 'event-tickets' ),
+		];
 
-		$plus_message = sprintf(
-		// Translators: %1$s: The Event Tickets Plus link.
-			esc_html_x( 'Tickets Commerce provides a simple and flexible ecommerce checkout for purchasing tickets. Just choose your payment gateway and configure checkout options and you\'re all set.  If you need more advanced features like custom attendee information, QR code check in, and stock sharing between tickets, take a look at %1$s for these features and more.', 'about Tickets Commerce', 'event-tickets' ),
-			$plus_link
-		);
+		// Promote Event Tickets Plus if not installed.
+		if ( ! class_exists( 'Tribe__Tickets_Plus__Main' ) ) {
```
juanfra, line 303: I thought we agreed on hooking to the array of fields and overriding this text from ET+.

readme.txt
juanfra: We should also try to avoid using acronyms in readme entries. 

ET+ => Event Tickets Plus
ARF => Attendee Registration Fields

And so on.

src/Tickets/Commerce/Payments_Tab.php
codingmusician: Ah, you're right! I'll rework this one.

PR #2439, opened 2022-06-20
[ET-1524] Let Price Block Value Override Ticket Range String

codingmusician:
### 🎫 Ticket

[ET-1524] 

### 🗒️ Description

The price block doesn't "play nice" with ET when you enter your own value into it. It takes the value you enter, mixes it in with other ticket values, then creates a string based on all values. This fix allows you to enter in a value into the price block to override the dynamic price that is created based on ticket values alone.

### 🎥 Artifacts 

https://www.screencast.com/t/ldO2PKamtI7p

### ✔️ Checklist
- [x] My code is tested. 

[ET-1524]: https://theeventscalendar.atlassian.net/browse/ET-1524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -195,6 +195,7 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 * Enhancement - Unify CSS class names for many admin elements. [ET-1536]
 * Enhancement - Add a `Currency Position` setting for Tickets Commerce. [ET-1534]
 * Fix - Some CSS issues within the tickets block in the block editor. [ET-1530]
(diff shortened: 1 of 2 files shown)


juanfra (approved):
Thanks for the artifacts! Would you please put that on the Jira ticket so that we don't lose the explanation of the way it works?

juanfra (approved):
🚀

PR #2468, opened 2022-08-02
Check to make sure the post exists before checking if it uses Elementor

codingmusician:
### 🗒️ Description

In cases where `\Elementor\Plugin::$instance->documents->get( $post->ID )` returns, false, it causes a fatal error. We can check to make sure `\Elementor\Plugin::$instance->documents->get( $post->ID )` isn't returning false, first.

### 🎥 Artifacts 

N/A

### ✔️ Checklist
- [ ] My code is tested. 


juanfra (commented):
Looks good to me!

juanfra (approved):

PR #2475, opened 2022-08-11
[ET-1568] ticket order

ScottGarrison:
### 🎫 Ticket

[[ET-1568](https://theeventscalendar.atlassian.net/browse/ET-1568)]

### 🗒️ Description

- In order to replicate this, both ET and ET+ seem to need to be activated.

Issue: When you first create Tickets and have them set in a specific order, once you edit one (like price for example), the order of the tickets gets swapped without manually saying to do so.

Solution: Making sure to get the existing order value of the ticket, and updating it when pushed to the database through params.

### 🎥 Artifacts 
https://www.loom.com/share/feee2caaf213447dadf51a750c02fda7

### ✔️ Checklist
- [ ✔️] I've included a changelog entry in the readme.txt file. 
- [✔️ ] My code is tested. 
- [✔️ ] My code has [proper inline documentation](https://the-events-calendar.github.io/products-engineering/docs/code-standards/).


src/resources/js/tickets.js
```diff
@@ -637,16 +637,21 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
 
 		$tribe_tickets.trigger( 'pre-save-ticket.tribe', e );
 
-		var $orders = $base_panel.find( '.tribe-ticket-field-order' );
+		var $ticketID = $edit_panel.find( '#ticket_id' ).val();
+		var $editParent = $base_panel.find( `[data-ticket-type-id="${$ticketID}"]` );
```
codingmusician, line 641: Only use the `$` on jQuery objects. In this case, `ticketID` is a string.
```suggestion
		var ticketID = $edit_panel.find( '#ticket_id' ).val();
		var $editParent = $base_panel.find( `[data-ticket-type-id="${ticketID}"]` );
```

src/resources/js/tickets.js
```diff
@@ -637,16 +637,21 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
+		var $orders = $editParent.find( '.tribe-ticket-field-order' ).val();
```
codingmusician, line 642:
```suggestion
		var orders = $editParent.find( '.tribe-ticket-field-order' ).val();
```

src/resources/js/tickets.js
```diff
@@ -637,16 +637,21 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
...
 			data: $edit_panel.find( 'input,textarea,select' ).serialize().replace( /\'/g, '%27' ).replace( /\:/g, '%3A' ),
 			post_id: $post_id.val(),
 			nonce: TribeTickets.add_ticket_nonce,
-			menu_order: $orders.length,
+			menu_order: $orders,
 			is_admin: $( 'body' ).hasClass( 'wp-admin' )
 		};
 
+		// ticket_menu_order is missing from the serialized string, lets add it
+		params.data = params.data.concat( "&ticket_menu_order=" + $orders )
```
codingmusician, line 653:
```suggestion
			menu_order: orders,
			is_admin: $( 'body' ).hasClass( 'wp-admin' )
		};

		// ticket_menu_order is missing from the serialized string, lets add it
		params.data = params.data.concat( "&ticket_menu_order=" + orders )
```

src/resources/js/tickets.js
juanfra, line 641: We should be ready now to use `let` and/or `const`

juanfra (commented):
Great stuff 🥇

src/resources/js/tickets.js
ScottGarrison: yes nice catch

ScottGarrison: that makes sense but i figured it was better to keep the files themselves consistent. So use `let` and `const` in v2 files but existing v1 files, use whatever the file is currently using.

Let me know if you want me to update this anyways

bordoni (approved):

juanfra:
Taking this to the finish line. @bordoni @codingmusician @rafsuntaskin  can I please get a CR?

rafsuntaskin (dismissed):

rafsuntaskin (approved):

PR #2477, opened 2022-08-22
Fix/etp 803 attendee label validation

ScottGarrison:
### 🎫 Ticket

[[ETP-803]](https://theeventscalendar.atlassian.net/browse/ETP-803)

### 🗒️ Description

-Calls the ET+ validation to check for attendee label duplicates

### ✔️ Checklist
- [ ] My code is tested. 

### DIFF
--- a/src/resources/js/tickets.js
+++ b/src/resources/js/tickets.js
@@ -1,6 +1,11 @@
 /* global tribe_event_tickets_plus, tribe, jQuery, _, tribe_l10n_datatables,
  tribe_ticket_datepicker_format, TribeTickets, tribe_timepickers  */
 
+/**
+ * Wordpress dependencies
+ */
+const { __ } = wp.i18n;
+
 // For compatibility purposes we add this
 if ( 'undefined' === typeof tribe.tickets ) {
 	tribe.tickets = {};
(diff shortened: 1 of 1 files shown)


src/resources/js/tickets.js
```diff
@@ -635,6 +635,14 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
 			return;
 		}
 
+		// Does an attendee field validation for multiple labels, if ET+ is active.
+		// Called in event-tickets-plus/src/resources/js/meta-admin.js
+		if ( tribe_event_tickets_plus.meta.admin ) {
```
codingmusician, line 640: `tribe_event_tickets_plus` won't exist if ET+ isn't active, so this will throw an error, I think.

Probably need to check for it and the meta property, first.
```suggestion
		if ( typeof tribe_event_tickets_plus !== 'undefined' && tribe_event_tickets_plus.meta && tribe_event_tickets_plus.meta.admin ) {
```

ScottGarrison: ended up removing this for a different solution Juan recommended

ScottGarrison:
@juanfra Could you have another peak at this to see if it's the direction you were talking about?

src/resources/js/tickets.js
```diff
@@ -635,6 +636,14 @@ var ticketHeaderImage = window.ticketHeaderImage || {};
 			return;
 		}
 
+		// setting triggerHandler as a variable is needed to return a new value of attendeeLablesIsValid if needed.
+		attendeeLablesIsValid = $tribe_tickets.triggerHandler( 'attendee-label-validation.tribe', [ attendeeLablesIsValid ] );
```
juanfra, line 640: Can we possibly leave triggerHandler and the variable as something more generic so that we can also use it for other types of validation we may need?

This is referring to the ticketForm, so maybe using that could be a good idea. Something like `ticketFormIsValid` or similar. The trigger also needs to be camelCased.

Then on ET+ we can use the attendeeLabel thing, as there we're validating the labels. But here it'd make more sense to have something generic that we can hook on to make other types of validations.

ScottGarrison: I updated this. But I just used an `additionalValidation` term because there is already validation happening above what I'm doing on the form. Felt this was at least something that didn't conflict with the first validation but had a more generic name

bordoni (approved):
We need to make sure we point to the correct branch here.

juanfra (approved):
🥇

PR #2490, opened 2022-09-14
Correct the typo for the dev composer package name

sc0ttkclark:
### 🎫 Ticket

[TBD]

### 🗒️ Description

This addresses a problem with `composer.json` where the incorrect package is being referenced (package doesn't exist). This was a typo from a while back, fixing it resolves the problem where the package code isn't available.


sc0ttkclark:
Realizing now that the repo itself has the wrong package name too: https://github.com/lucatume/wp-snapshot-assertions/blob/master/composer.json#L2

cc @lucatume what do you think we should do here?

PR #2523, opened 2022-11-29
[ET-1615] - Removed the locale param for PayPal JS SDK to avoid invalid local errors.

rafsuntaskin:
### 🎫 Ticket

[ET-1615] 

### 🗒️ Description

* As per PayPal recommendation we can remove the optional `locale` param to reduce JS SDK issues with this: https://d.pr/i/zRT5LX
* We have kept the `locale` method and added a new validation method in case anyone wants to use the `locale` param with  the `tec_tickets_commerce_gateway_paypal_js_sdk_url` filter.

### ✔️ Checklist
- [x] My code is tested. 

[ET-1615]: https://theeventscalendar.atlassian.net/browse/ET-1615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -194,7 +194,7 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 * Fix - Fixed the order of tickets in an event changing when you haven't manually requested it. [ET-1568]
 * Fix - Fixed shared capacity tickets only showing the lowest capacity between the shared pool tickets. [ETP-815]
-* Tweak - Removed locale param for TicketsCommerce JS SDK as per PayPal recommendation. [ET-1615]
+* Tweak - Removed locale param for Tickets Commerce JS SDK as per PayPal recommendation. [ET-1615]
 
 = [5.5.4] 2022-11-09 =
 
--- a/readme.txt
+++ b/readme.txt
@@ -194,6 +194,7 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 * Fix - Fixed the order of tickets in an event changing when you haven't manually requested it. [ET-1568]
 * Fix - Fixed shared capacity tickets only showing the lowest capacity between the shared pool tickets. [ETP-815]
+* Tweak - Removed locale param for TicketsCommerce JS SDK as per PayPal recommendation. [ET-1615]
 
 = [5.5.4] 2022-11-09 =
 
--- a/src/Tickets/Commerce/Gateways/PayPal/Merchant.php
+++ b/src/Tickets/Commerce/Gateways/PayPal/Merchant.php
@@ -912,6 +912,27 @@ class Merchant extends Abstract_Merchant {
 		 *
 		 * @param string $locale
 		 */
-		return apply_filters( 'tec_tickets_commerce_gateway_paypal_merchant_locale', $locale );
+		$locale = apply_filters( 'tec_tickets_commerce_gateway_paypal_merchant_locale', $locale );
+
+		return $this->validate_locale( $locale );
+	}
+
+	/**
+	 * Validates the locale to make sure it has a valid length.
+	 * If it's not, it will return the default locale.
+	 *
+	 * @since TBD
+	 *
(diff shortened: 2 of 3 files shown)


src/Tickets/Commerce/Gateways/PayPal/Merchant.php
```diff
@@ -912,6 +912,27 @@ public function get_locale() {
...
+	 * @since TBD
+	 *
+	 * @param string $locale The Locale code.
+	 *
+	 * @return string
+	 */
+	public function validate_locale( string $locale ) : string {
+		// check if the length is valid.
+		if ( 5 !== strlen( $locale ) ) {
+			return 'en_US';
```
juanfra, line 933: Should we store the default locale in a class var and use it from there? It'd be nice to make this value filterable

src/Tickets/Commerce/Gateways/PayPal/Merchant.php
```diff
@@ -912,6 +912,27 @@ public function get_locale() {
 		 *
 		 * @param string $locale
 		 */
-		return apply_filters( 'tec_tickets_commerce_gateway_paypal_merchant_locale', $locale );
+		$locale = apply_filters( 'tec_tickets_commerce_gateway_paypal_merchant_locale', $locale );
```
juanfra, line 915: Should we first validate and then filter? That way the filter can override any problem we may have. I don't have a strong opinion on this, just raising the question.

readme.txt
```diff
@@ -194,6 +194,7 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 * Fix - Fixed the order of tickets in an event changing when you haven't manually requested it. [ET-1568]
 * Fix - Fixed shared capacity tickets only showing the lowest capacity between the shared pool tickets. [ETP-815]
+* Tweak - Removed locale param for TicketsCommerce JS SDK as per PayPal recommendation. [ET-1615]
```
codingmusician, line 197:
```suggestion
* Tweak - Removed locale param for Tickets Commerce JS SDK as per PayPal recommendation. [ET-1615]
```

src/Tickets/Commerce/Gateways/PayPal/Merchant.php
codingmusician: I agree. I think the default should be a variable so that we could read it from outside this class, if needed.

src/Tickets/Commerce/Gateways/PayPal/Merchant.php
rafsuntaskin: Not in this case. As we need to make sure that the final value we return is valid, filtering the value may allow users to supply invalid values.

juanfra: That will happen with most filters. For example, if you filter `wp_mail` and you input a wrong email address the function will fail, but it'll be a users problem. I was thinking more about flexibility and having the capacity to use a snippet in the case there's some problem coming up with this (so that we don't have to hotfix).

codingmusician (approved):

src/Tickets/Commerce/Gateways/PayPal/Merchant.php
```diff
@@ -912,6 +912,36 @@ public function get_locale() {
...
+	 *
+	 * @return string
+	 */
+	public function validate_locale( string $locale ) : string {
+		// check if the length is valid.
+		if ( 5 === strlen( $locale ) ) {
+			return $locale;
+		}
+
+		$fallback_locale = 'en_US';
```
juanfra, line 936: Can we set this into a class variable?

rafsuntaskin: This is only a fallback value, the default one is coming from `get_locale` of core. So, if anywhere else is needed we can use that.

PR #2526, opened 2022-12-05
Fix/et 1508 2

moraleida:
### 🎫 Ticket

[ET-1508] 

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1508]: https://theeventscalendar.atlassian.net/browse/ET-1508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -191,8 +191,6 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 = [TBD] TBD =
 * Fix - Remove need for Platform Controls to verify webhook signatures in Stripe. [ET-1508]
 
-* Fix - Fixes multiple of the same ticket form being on the same page being out of sync. [GTRIA-729]
-
 = [5.5.4] 2022-11-09 =
 
 * Fix - Fixes multiple of the same ticket form being on the same page being out of sync. [GTRIA-729]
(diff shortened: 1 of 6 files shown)


readme.txt
```diff
@@ -192,6 +192,7 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 = [TBD] TBD =
 
+* Fix - Remove need for Platform Controls to verify webhook signatures in Stripe. [ET-1508]
 * Fix - Fixed the order of tickets in an event changing when you haven't manually requested it. [ET-1568]
 * Fix - Fixed shared capacity tickets only showing the lowest capacity between the shared pool tickets. [ETP-815]
 * Tweak - Display the Webhook validation response from Stripe in Tickets Commerce Stripe settings. [ET-1614]
```
rafsuntaskin, line 198:
```suggestion

```
This can be removed as we are not getting the Stripe response to be displayed anymore. But we need to log the response in some place using the `tribe_log` action. So that, in the event of an investigation we can find the actual response from Stripe.

codingmusician (approved):

readme.txt
moraleida: @rafsuntaskin there isn't a specific response for webhook validation. What we do is we create a fake payment intent and if that works we receive an event and mark webhooks as valid. Since we're no longer depending on updating the user account, as long as the account is able to process payments and the key is correct, the validation will work.

codingmusician (approved):

PR #2529, opened 2022-12-16
Fix handling some provisional IDs.

stratease:
### 🎫 Ticket

[[ECP-1426](https://theeventscalendar.atlassian.net/browse/ECP-1426)] 

### 🗒️ Description

Will intercept some of the provisional IDs that may come through various requests to ET. 
Will attempt to sniff and normalize provisional IDs in meta queries, namely ones coming from Attendee and Ticket repositories.

### 🎥 Artifacts 
image

### ✔️ Checklist
- [x] My code is tested. 

[ECP-1426]: https://theeventscalendar.atlassian.net/browse/ECP-1426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stratease:
I think it would be better to add support for handling provisional IDs, instead of chasing various current and future URL's. Especially since I believe we will be adding more granular support in ET for occurrences in the future.

@lucatume @moraleida wanted your take on this?

src/Tickets/Custom_Tables/V1/WP_Query/WP_Meta_Query_Modifier.php
```diff
@@ -0,0 +1,155 @@
...
+	 * Ticket Meta Query Modifier constructor.
+	 *
+	 * @since TBD
+	 *
+	 * @param Tribe__Tickets__Attendee_Repository $attendee_repository
+	 * @param Tribe__Tickets__Ticket_Repository   $ticket_repository
+	 * @param Provisional_Post                    $provisional_post
+	 */
+	public function __construct(
+		Tribe__Tickets__Attendee_Repository $attendee_repository,
```
lucatume, line 82: [In the provider](https://github.com/the-events-calendar/event-tickets/pull/2529/files#diff-0ad318c81052f97ad003bd713b8295ff1ec0673e3a574c11cebcec75e509d085R49) you're setting this as a singleton object, its dependencies will be injected once, when the instance is first built.
While it's fine to require an instance of the Attendee Repository and Ticket Repository to get  the `post_types` out, as it's done in the body of the `__construct` method some lines later, the method should not keep references to the particular instances of the repository in the `attendee_repository` and `ticket_repository` properties: the repository instances are built and destroyed for each query and someone could be tempted to think those properties reference some singleton instances of the repositories.
The repositories are not singleton by design, e.g. how the Ticket Repository is bound as not singleton [here](https://github.com/the-events-calendar/event-tickets/blob/8d569126e399f71000273d752505cf506d07f807/src/Tribe/Service_Providers/ORM.php#L25).

src/Tickets/Custom_Tables/V1/WP_Query/WP_Meta_Query_Modifier.php
```diff
@@ -0,0 +1,155 @@
...
+			return;
+		}
+
+		$post_types = array_values( (array) $query->get( 'post_type' ) );
+
+		if ( ! count( array_intersect( $post_types, $this->post_types ) ) ) {
+			return;
+		}
+
+		foreach ( $this->meta_keys as $prefix ) {
```
lucatume, line 117: Instead of doing it here, I think it would make sense  to filter the `event_id` in the [repositories](https://github.com/the-events-calendar/event-tickets/blob/8d569126e399f71000273d752505cf506d07f807/src/Tribe/Ticket_Repository.php#L109-L111) (this is just an example) by filtering the `event_id` variable the moment it's added to the queries.

E.g.
```diff
        public function filter_by_event_not_in( $event_id ) {
+              $event_id = tec_event_normalize_event_id( $event_id );
		$this->by( 'meta_not_in', $this->ticket_to_event_keys(), $event_id );
	}
```

The `tec_event_normalize_event_id` would be a function that would deal with the fact that CT1 is active or not and, if active, would normalize the Event ID using the `Occurrence::normalize_id` function.

Something like this (the code might contain mistakes, I typed it directly here), in TEC:

```php
function tec_event_normalize_event_id( $event_id ) {
    if ( empty ( $event_id ) ) || ! is_numeric( $event_id ) ) { return $event_id; }

    $ct1_active = tribe()->getVar( 'ct1_fully_activated' );

    if( ! $ct1_active ) {
        return $event_id;
    }
    
    return TEC\Events\Custom_Tables\V1\Models\Occurrence::normalize_id( $event_id );
}
```

lucatume: Looking at the Ticket and Attendee repositories, I do not see a lot of places where similar logic would apply.

lucatume:
> I think it would be better to add support for handling provisional IDs, instead of chasing various current and future URL's. Especially since I believe we will be adding more granular support in ET for occurrences in the future.

Part of the strategy, thus implementation, of Tickets + Recurring Events + Provisional IDs is still in the works, so I cannot provide an answer based on fixed architecture; to keep the door open to any possible direction, I would lean toward doing something filter-based as you've done that would abstract away whether CT1 is active or not and what the support choice is.

I'm not sure what "adding support for handling provisional IDs" would mean in practical terms, could you elaborate? Do you mean table/field redirection?

I do not know exactly why, but the redirection done in the meta query, whether that stays or not, seems a replica of what is done based on [the Redirection Schema](https://github.com/the-events-calendar/the-events-calendar/blob/01c3c41567e4b3ad65de7b8a3cdafff04573f5ec/src/Events/Custom_Tables/V1/WP_Query/Redirection_Schema.php#L85); maybe we should use and extend that?

stratease:
> > I think it would be better to add support for handling provisional IDs, instead of chasing various current and future URL's. Especially since I believe we will be adding more granular support in ET for occurrences in the future.
> 
> Part of the strategy, thus implementation, of Tickets + Recurring Events + Provisional IDs is still in the works, so I cannot provide an answer based on fixed architecture; to keep the door open to any possible direction, I would lean toward doing something filter-based as you've done that would abstract away whether CT1 is active or not and what the support choice is.

That's why I set things up as I did. My goal is to simply allow "finding the post ID" when a provisional ID is supplied for the attendee queries, as they are failing right now. 

> I'm not sure what "adding support for handling provisional IDs" would mean in practical terms, could you elaborate? 

Right now all of ET has no ideal what a "provisional ID" is, and the queries will assume it is always a traditional post ID. My perspective here is that it would be better for ET to start to internally control what it does with the provisional ID's, _instead of TEC (or ECP) deciding what type of ID it should allow ET to see_.

> Do you mean table/field redirection?

Not exactly - this is just normalizing the ID. In the future ET may do that, and leaves it in ET's hands. That's really the point of this approach.

> I do not know exactly why, but the redirection done in the meta query, whether that stays or not, seems a replica of what is done based on [the Redirection Schema](https://github.com/the-events-calendar/the-events-calendar/blob/01c3c41567e4b3ad65de7b8a3cdafff04573f5ec/src/Events/Custom_Tables/V1/WP_Query/Redirection_Schema.php#L85); maybe we should use and extend that?

Not sure - wouldn't that be overkill for the scope of this? We are just normalizing provisional ID's here.

src/Tickets/Custom_Tables/V1/WP_Query/WP_Meta_Query_Modifier.php
stratease: Good catch, I actually meant to refactor those out after a cleanup iteration. Whoops.

src/Tickets/Custom_Tables/V1/WP_Query/WP_Meta_Query_Modifier.php
```diff
@@ -0,0 +1,155 @@
...
+	 * @since TBD
+	 *
+	 * @param Tribe__Tickets__Attendee_Repository $attendee_repository
+	 * @param Tribe__Tickets__Ticket_Repository   $ticket_repository
+	 * @param Provisional_Post                    $provisional_post
+	 */
+	public function __construct(
+		Tribe__Tickets__Attendee_Repository $attendee_repository,
+		Tribe__Tickets__Ticket_Repository $ticket_repository,
+		Provisional_Post $provisional_post
```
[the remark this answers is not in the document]
stratease: Good catch! Auto import gets me in trouble...

src/Tickets/Custom_Tables/V1/WP_Query/WP_Meta_Query_Modifier.php
stratease: I thought about going this route, opted not to do it because I wasn't sure where else the event ID's would be used (custom queries etc) which would still receive a provisional ID.

That said, I'll take your recommendation - there's no guarantee those queries would want the post ID instead of the provisional ID in the future.

stratease: Went a slightly different way, this change does not require a TEC version dependency - sticks to a filter CT1 hooks on.

codingmusician (dismissed):
Looks good to me.

src/Tribe/Attendees_Table.php
```diff
@@ -70,7 +70,15 @@ public function __construct( $args = [] ) {
...
 		if ( ! empty( $_GET['event_id'] ) ) {
-			$this->event = get_post( absint( $_GET['event_id'] ) );
+			/**
+			 * This filter allows retrieval of an event ID to be filtered before being accessed elsewhere.
+			 *
+			 * @since TBD
+			 *
+			 * @param int The event ID to be filtered.
+			 */
+			$event_id = apply_filters( 'tec_tickets_filter_event_id', $_GET['event_id'] );
```
bordoni, line 80: We have multiple places where we apply this filter, to avoid problems in the future with which params are being set changing for one but not the others, do you mind creating a method for this?

bordoni (changes requested):

tests/ct1_integration/Attendees_TableTest.php
```diff
@@ -0,0 +1,114 @@
...
+			return [ 'post' ];
+		} );
+
+		// Enable Tribe Commerce.
+		add_filter( 'tribe_tickets_commerce_paypal_is_active', '__return_true' );
+		add_filter( 'tribe_tickets_get_modules', function ( $modules ) {
+			$modules['Tribe__Tickets__Commerce__PayPal__Main'] = tribe( 'tickets.commerce.paypal' )->plugin_name;
+
+			return $modules;
+		} );
```
rafsuntaskin, line 38: Tribe Commerce is deprecated, we should add the tests for Tickets Commerce instead. @bordoni what do you think?

bordoni: Agree with @rafsuntaskin here.

src/Tribe/Attendees_Table.php
lucatume: This is a good chance to remove the direct access to the `GET` superglobal and use some `filter_var` to sanitize to an integer value.

lucatume: The `TEC\Tickets\Event` class is a good place to put the method wrapping the application of the `tec_tickets_filter_event_id` filter. Something like this:

```php
public static function get_event_id( $event_id ): ?int;
```

lucatume (changes requested):
The code looks fine in logic; I think collecting the logic of the filter application in a method would make it a bit easier to maintain.

bordoni (dismissed):

src/Tickets/Event.php
```diff
@@ -36,4 +36,23 @@ public static function get_post_type() {
...
+	 */
+	public static function filter_event_id( $event_id ): ?int {
+		/**
+		 * This filter allows retrieval of an event ID to be filtered before being accessed elsewhere.
+		 *
+		 * @since TBD
+		 *
+		 * @param int|null The event ID to be filtered.
+		 */
+		return apply_filters( 'tec_tickets_filter_event_id', $event_id );
```
lucatume, line 56: Since filters are the Wild West of "let's see what we can retur here", I suggest casting to `int` or `null` to avoid issues with the return value type-hint:
```suggestion
		return (int)apply_filters( 'tec_tickets_filter_event_id', $event_id ) ?: null;
```

src/Tribe/Attendees_Table.php
```diff
@@ -70,7 +72,9 @@ public function __construct( $args = [] ) {
 
 		// Fetch the event Object
 		if ( ! empty( $_GET['event_id'] ) ) {
-			$this->event = get_post( absint( $_GET['event_id'] ) );
+			$event_id    = filter_var( $_GET['event_id'], FILTER_VALIDATE_INT );
+			$event_id    = Event::filter_event_id( $event_id );
+			$this->event = get_post( absint( $event_id ) );
```
lucatume, line 77: If the `Event::filter_event_id` method returns `?int`, then there should be no need for the `absint` call.

lucatume (changes requested):
More defensive code in the filtering method.

rafsuntaskin (approved):

PR #2541, opened 2023-01-25
Remove Croatian Kuna (HRK) from Currency options and replace with Euro

chigozieorunta:
### 🎫 Ticket

[1625](https://theeventscalendar.atlassian.net/browse/ET-1625)

### 🗒️ Description

I have removed the Croatian Kuna (HRK) currency option from the default currency map and added a method to help display notices for unsupported currencies in the Notices class (Tribe__Tickets__Admin__Notices). By using an array to store unsupported currencies, I have also made it flexible for us to handle more than one unsupported currency (if the occasion arises).

### ✔️ Checklist
- [X] I've included a changelog entry in the readme.txt file. 
- [X] My code is tested. 
- [X] My code has [proper inline documentation](https://the-events-calendar.github.io/products-engineering/docs/code-standards/).


src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +238,44 @@ public function maybe_display_plus_commerce_notice() {
...
+		// Get currency code option
+		$currency = tribe_get_option( 'tickets-commerce-currency-code' );
+
+		if ( array_key_exists( $currency, $unsupported_currency ) ) {
+
+			$message = sprintf(
+				'<h3>%1$s</h3><p>%2$s</p>',
+				$unsupported_currency[$currency]['heading'],
+				$unsupported_currency[$currency]['message']
+			);
```
codingmusician, line 266: We should probably escape these.
```suggestion
			$message = sprintf(
				'<h3>%1$s</h3><p>%2$s</p>',
				esc_html( $unsupported_currency[$currency]['heading'] ),
				esc_html( $unsupported_currency[$currency]['message'] )
			);
```

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +238,44 @@ public function maybe_display_plus_commerce_notice() {
 			tribe_notice( "event-tickets-plus-missing-{$provider}-support", $message, 'dismiss=1&type=warning' );
 		}
 	}
+
+	/**
+	 * Display notices for unsupported currencies.
+	 *
```
juanfra, line 244: We need to put `@since TBD` every time we're adding a new method to our codebase. Then, when we release the version and the plugin is prepared by the release engineer, the TBD placeholder will be replaced with the version number.

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +238,44 @@ public function maybe_display_plus_commerce_notice() {
...
+		// Get currencies and notice texts.
+		$unsupported_currency = [
+			'HRK' => [
+				'heading' => __( 'Tickets Commerce is now selling with Euro', 'event-tickets' ),
+				'message' => __( 'From the 1st of January 2023, the euro became the official currency for Croatia. We have removed the Croatian kuna from our currency settings and updated your settings to start selling with Euro.', 'event-tickets' ),
+				'updated' => 'EUR'
+			],
+		];
+
+		// Get currency code option
```
juanfra, line 257: All comments need to end with a full stop.

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +238,44 @@ public function maybe_display_plus_commerce_notice() {
...
+		$currency = tribe_get_option( 'tickets-commerce-currency-code' );
```
juanfra, line 258: We have the option name living in a constant, we could probably consume it from there: 
- https://github.com/the-events-calendar/event-tickets/blob/master/src/Tickets/Commerce/Utils/Currency.php#L25
- https://github.com/the-events-calendar/event-tickets/blob/master/src/Tickets/Commerce/Settings.php#L54

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +238,44 @@ public function maybe_display_plus_commerce_notice() {
...
+
+		if ( array_key_exists( $currency, $unsupported_currency ) ) {
```
juanfra, line 260: We normally try to avoid conditionals that will cause a lot of indentation in the codebase. I'll invert this conditional and return. Then after the conditional leave the code that's inside the conditional.

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +238,44 @@ public function maybe_display_plus_commerce_notice() {
...
+
+			$message = sprintf(
+				'<h3>%1$s</h3><p>%2$s</p>',
+				esc_html( $unsupported_currency[$currency]['heading'] ),
+				esc_html( $unsupported_currency[$currency]['message'] )
```
juanfra, line 265: Whenever we're using a variable as an array index we need to add whitespace before and after it.
```suggestion
				esc_html( $unsupported_currency[ $currency ]['heading'] ),
				esc_html( $unsupported_currency[ $currency ]['message'] )
```

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +238,44 @@ public function maybe_display_plus_commerce_notice() {
...
+			);
+
+			tribe_notice(
+				__FUNCTION__,
```
juanfra, line 269: We could possibly add the slug here directly as text, because the function wouldn't be super representative of the notice (and not prefixed, which is good for identification purposes.)

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +238,44 @@ public function maybe_display_plus_commerce_notice() {
...
+				$message,
+				[
+					'dismiss' => true,
+					'type'    => 'warning',
+				]
+			);
+
+			// Update currency to new option
```
juanfra, line 277:
```suggestion
			// Update currency option to the new value.
```

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +238,44 @@ public function maybe_display_plus_commerce_notice() {
...
+			tribe_update_option( 'tickets-commerce-currency-code', $unsupported_currency[$currency]['updated'] );
```
juanfra, line 278: I'm not sure if we should include this update logic within the notification. In my opinion, the logic around unsupported currencies and the update of the option should live in the [currency class](https://github.com/the-events-calendar/event-tickets/blob/master/src/Tickets/Commerce/Utils/Currency.php), and then we consume that data from the notice function.

That way we keep organization and store all the currency-related functionality in the currency class, and then keep this notification class only with the notice functionality. Feel free to reach out if you have doubts about this.

```suggestion
			tribe_update_option( 'tickets-commerce-currency-code', $unsupported_currency[ $currency ]['updated'] );
```

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +238,44 @@ public function maybe_display_plus_commerce_notice() {
...
+	 *
+	 * @return void
+	 */
+	public function maybe_display_unsupported_currency_notice() {
+		// Get currencies and notice texts.
+		$unsupported_currency = [
+			'HRK' => [
+				'heading' => __( 'Tickets Commerce is now selling with Euro', 'event-tickets' ),
+				'message' => __( 'From the 1st of January 2023, the euro became the official currency for Croatia. We have removed the Croatian kuna from our currency settings and updated your settings to start selling with Euro.', 'event-tickets' ),
+				'updated' => 'EUR'
```
juanfra, line 253: Maybe `new_value` is more self-explanatory?

juanfra (commented):
👏 Great work Chigz! I left some comments, please feel free to reach out if you have any doubt.

The one thing that's missing is a readme entry on the `readme.txt` file. Something [like this](https://github.com/the-events-calendar/event-tickets/pull/2535/files#diff-9e6e4772050998a5c0dc3c61acf3dab0a7e594566171fa5746d6b62f9598efb6), for example.

src/Tribe/Admin/Notices.php
chigozieorunta: @juanfra So true. Thanks for this! I'll refactor this into the Currency class.

src/Tribe/Admin/Notices.php
chigozieorunta: @juanfra Thanks for this.

src/Tribe/Admin/Notices.php
chigozieorunta: @juanfra Good observation. Thanks for this!

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +240,30 @@ public function maybe_display_plus_commerce_notice() {
...
+
+	/**
+	 * Display notices for unsupported currencies.
+	 *
+	 * @since 5.5.5
+	 *
+	 * @return void
+	 */
+	public function maybe_display_unsupported_currency_notice() {
+		if ( Currency::is_supported_currency() ) return;
```
juanfra, line 252:
```suggestion
		if ( Currency::is_supported_currency() ) {
			return;
		}
```
We need to add the braces for IFs statements.

We're using the `supported_currency` logic as if we're only going to have one unsupported currency. If we're going this road we need to modify the logic to be able to iterate over the unsupported currencies and make a notice for each.

chigozieorunta: Added braces for IF statement

chigozieorunta: @juanfra I'm not sure if we will need to iterate over each unsupported currency and make a notice for each one. Since the user will only be selecting one currency setting per time, we wouldn't need to make a notice for each one, except for the one the user selects that falls into the unsupported category.

src/Tickets/Commerce/Utils/Currency.php
```diff
@@ -42,6 +42,15 @@ class Currency {
...
 	public static $currency_code_fallback = 'USD';
 
+	/**
+	 * Unsupported Currency
+	 *
+	 * @since 5.5.5
+	 *
+	 * @var array
+	 */
+	public static $unsupported_currency = [];
```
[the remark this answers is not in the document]
chigozieorunta: @juanfra We are using this parameter on the Notice side of things to retrieve the heading and message of the notice.

src/Tickets/Commerce/Utils/Currency.php
```diff
@@ -1458,4 +1459,69 @@ public function get_currency_code_options() {
...
+	 *
+	 * @return array
+	 */
+	public static function get_unsupported_currencies(): array {
+		/**
+		 * Filter all unsupported currenices before returning.
+		 *
+		 * @since TBD
+		 *
+		 * @return string
```
juanfra, line 1476:
```suggestion
		 * @return array
```

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +240,32 @@ public function maybe_display_plus_commerce_notice() {
 			tribe_notice( "event-tickets-plus-missing-{$provider}-support", $message, 'dismiss=1&type=warning' );
 		}
 	}
+
+	/**
+	 * Display notices for unsupported currencies.
+	 *
+	 * @since 5.5.5
```
juanfra, line 247:
```suggestion
	 * @since TBD
```

src/Tickets/Commerce/Utils/Currency.php
```diff
@@ -1458,4 +1459,69 @@ public function get_currency_code_options() {
...
+	}
+
+	/**
+	 * Verify if currency is supported.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	public static function is_supported_currency(): bool {
```
juanfra, line 1495: Maybe naming `is_current_currency_supported` will be more self-explanatory, as `is_supported_currency` would be a better name for something that receives a parameter.

src/Tribe/Admin/Notices.php
```diff
@@ -237,4 +240,32 @@ public function maybe_display_plus_commerce_notice() {
...
+		}
+
+		$message = sprintf(
+			'<h3>%1$s</h3><p>%2$s</p>',
+			esc_html( Currency::$unsupported_currency['heading'] ),
+			esc_html( Currency::$unsupported_currency['message'] )
+		);
+
+		tribe_notice(
+			'event-tickets-unsupported-currencies',
```
juanfra, line 263: We can maybe also add the currency symbol to the notice ID, so that if they change in the future (and the notice was dismissed) we can have the notice showing.

src/Tickets/Commerce/Utils/Currency.php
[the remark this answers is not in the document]
juanfra: That's true, fair point. Thanks for clarifying.

src/Tickets/Commerce/Utils/Currency.php
```diff
@@ -1500,4 +1501,73 @@ public function get_currency_code_options() {
...
+	/**
+	 * Get unsupported currencies and notice texts.
+	 *
+	 * @since TBD
+	 *
+	 * @return array
+	 */
+	public static function get_unsupported_currencies(): array {
+		/**
+		 * Filter all unsupported currenices before returning.
```
rafsuntaskin, line 1514:
```suggestion
		 * Filter all unsupported currencies before returning.
```

src/Tickets/Commerce/Utils/Currency.php
```diff
@@ -1500,4 +1501,73 @@ public function get_currency_code_options() {
...
+		 *
+		 * @since TBD
+		 *
+		 * @return array
+		 */
+		return apply_filters( 'tec_tickets_commerce_unsupported_currencies', [
+				'HRK' => [
+					'heading'   => __( 'Tickets Commerce is now selling with Euro', 'event-tickets' ),
+					'message'   => __( 'From the 1st of January 2023, the euro became the official currency for Croatia. We have removed the Croatian kuna from our currency settings and updated your settings to start selling with Euro.', 'event-tickets' ),
```
rafsuntaskin, line 1523:
```suggestion
					'message'   => __( 'From the 1st of January 2023, the euro became the official currency for Croatia. We have removed the Croatian Kuna from our currency settings and updated your settings to start selling with Euro.', 'event-tickets' ),
```

rafsuntaskin (commented):
Minor text updates. Otherwise, good to go.

juanfra (approved):

rafsuntaskin (approved):

codingmusician (approved):

PR #2548, opened 2023-02-08
[ET-1638] Tickets emails re structure

juanfra:
### 🎫 Ticket

[ET-1638] 

### 🗒️ Description

- Re-structuring email templates to be able to use those in different emails.

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1638]: https://theeventscalendar.atlassian.net/browse/ET-1638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/views/v2/emails/template-parts/body/tickets.php
```diff
@@ -0,0 +1,46 @@
...
+$i = 0;
+
+$this->template( 'template-parts/body/tickets-total' );
+
+?>
+<tr>
+	<td style="padding:0;">
+		<table class="tec-tickets__email-table-content-tickets" role="presentation">
+		<?php foreach ( $tickets as $ticket ) : ?>
+			<?php $i++; ?>
```
codingmusician, line 31: Should this not be at the end of the loop so we're not skipping the first ticket?

juanfra: The incremental var is just for display purposes, to use it in the "Ticket X of Y". Given that I defined it as `0`, I should do the `++` here to have the first element with `1`. It could be that or setting the var to `1` and moving the incremental to the end of the loop. It wouldn't change much though as we don't use it for anything else.

Screenshot 2023-02-09 at 10 05 31

codingmusician: Ah. My fault. I didn't look to see where it was being used. Thanks!

codingmusician (approved):

PR #2555, opened 2023-02-22
[ET-1599] Auto Generate Default Emails and Email Settings

codingmusician:
### 🎫 Ticket

[ET-1599] 

### 🗒️ Description

Automatically create default emails for Tickets Emails and create a way to edit individual email settings.

### 🎥 Artifacts 

https://youtu.be/XkK7IIusXT0

### ✔️ Checklist
- [x] My code is tested. 

[ET-1599]: https://theeventscalendar.atlassian.net/browse/ET-1599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -172,4 +197,46 @@ public function get_fields(): array {
...
+	 * 
+	 * @return boolean
+	 */
+	public function is_editing_email( $email = null ) {
+		if ( ! $this->is_on_tab() ) {
+			return false;
+		}
+
+		// Get `tc-edit` query string from URL.
+		$editing_email  = tribe_get_request_var( static::$edit_query );
```
juanfra, line 216: For consistency I'd incline to use directly the `section` parameter (so that we can have this Settings > tab > section logic), because there's plans to update settings in the future with this format. I've added  a helper function for this here: https://github.com/the-events-calendar/event-tickets/pull/2550/files#diff-b864484972892032b7a96254575015028df2c5156fd149628898c69c2b99e626R129

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -172,4 +197,46 @@ public function get_fields(): array {
...
+	 * 
+	 * @since TBD
+	 *
+	 * @param Email_Abstract $email
+	 * 
+	 * @return boolean
+	 */
+	public function is_editing_email( $email = null ) {
+		if ( ! $this->is_on_tab() ) {
+			return false;
```
juanfra, line 212: By using the section [from here](https://github.com/the-events-calendar/event-tickets/pull/2550/files#diff-b864484972892032b7a96254575015028df2c5156fd149628898c69c2b99e626R129) we shouldn't have to check if is on tab.

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -140,6 +159,12 @@ public function is_on_section( $section = '' ): bool {
 	 * @return array[]
 	 */
 	public function get_fields(): array {
+
+		// Check to see if we're editing an email, first.
+		if ( $this->is_editing_email() ) {
```
juanfra, line 164: Let's check if it's on section https://github.com/the-events-calendar/event-tickets/pull/2550/files#diff-b864484972892032b7a96254575015028df2c5156fd149628898c69c2b99e626R129

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -31,6 +32,24 @@ class Emails_Tab {
...
 	public static $slug = 'emails';
 
+	/**
+	 * Slug for editing emails.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public static $edit_query = 'te-edit';
```
juanfra, line 42: Let's rely on `section`

src/Tickets/Emails/Email/Ticket.php
```diff
@@ -142,12 +133,32 @@ public function get_title(): string {
...
+
+		/**
+		 * Allow filtering the email heading.
+		 *
+		 * @since TBD
+		 *
+		 * @param string $subject  The email heading.
+		 * @param string $id       The ticket id.
+		 */
+		$heading = apply_filters( 'tec_tickets_emails_subject_' . $this->id, $heading, $this->id, $this->template );
```
juanfra, line 146:
```suggestion
		$heading = apply_filters( 'tec_tickets_emails_heading_' . $this->id, $heading, $this->id, $this->template );
```

src/Tickets/Emails/Email/Ticket.php
```diff
@@ -176,6 +187,17 @@ public function get_subject(): string {
...
 
+	/**
+	 * Get default email subject.
+	 * 
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public function get_default_subject() {
+		return esc_html__( 'Your tickets to {event-name}', 'event-tickets' );
```
juanfra, line 198: We'll probably have to work on placeholders and this one in particular, as the logic for Events will be coming from TEC. `{event_name}` is something that will be coming from TEC, hooked to the placeholders.

src/Tickets/Emails/Email_Abstract.php
```diff
@@ -362,4 +364,22 @@ public function get_additional_content(): string {
...
+	 * 
+	 * @since TBD
+	 * 
+	 * @return string
+	 */
+	public function get_edit_url() {
+		// Force the `emails` tab.
+		$args = [
+			'tab'                   => Emails_Tab::$slug,
+			Emails_Tab::$edit_query => $this->id,
```
juanfra, line 379: Let's rely on `section`

src/Tickets/Emails/Email_Handler.php
```diff
@@ -197,4 +197,22 @@ public function add_settings_per_email( array $fields ): array {
...
+	 * Get email by ID.
+	 * 
+	 * @since TBD
+	 * 
+	 * @param string $id ID of email.
+	 * 
+	 * @return Email_Abstract|boolean Email object or false if it does not exist.
+	 */
+	public function get_email_by_id( $id ) {
+		$emails = $this->get_all();
```
juanfra, line 210: We can maybe use `array_search`?

src/Tickets/Emails/Hooks.php
```diff
@@ -73,7 +73,9 @@ protected function add_filters() {
 	 *
 	 */
 	public function action_register_post_type() {
-		$this->container->make( Email_Handler::class )->register_post_type();
+		$email_handler = $this->container->make( Email_Handler::class );
+		$email_handler->register_post_type();
```
juanfra, line 77: This should be pretty much reserved for hooks. The logic should live in `Email_Handler`. We should probably have the `$this->maybe_populate_tec_tickets_emails_post_type` [here](https://github.com/the-events-calendar/event-tickets/pull/2550/files#diff-c97bfb6de59c73509e252cc9795314ed13c6b61585f3b735b84c66ca9a8a93ffR67)

src/Tickets/Emails/Admin/Emails_Tab.php
juanfra: Also, for the per email settings I initially thought to have it something like this https://github.com/the-events-calendar/event-tickets/pull/2550/files#diff-c97bfb6de59c73509e252cc9795314ed13c6b61585f3b735b84c66ca9a8a93ffR183

codingmusician: So something like `section=edit_email` and then maybe an `id=####` in the URL?

Regarding the settings, I coded it to where it's similar to how the sections on the `Payments` tab work. It shows the general settings unless you're on a particular `section`.

src/Tickets/Emails/Email_Handler.php
codingmusician: `$emails` is an array of objects in this case. Can `array_search` be used in this case?

src/Tickets/Emails/Hooks.php
codingmusician: I tested with it there and it caused an error. It needed to be run on the `init` hook. I can separate it into a different method in Hooks if you think that would be better. Let me know your thoughts.

src/Tickets/Emails/Email/Ticket.php
codingmusician: Yeah. I figured we'd need to be doing something different with all the placeholders in the future. They also shouldn't be in any of the translation strings. I figured we'd leave that for a future ticket, though.

src/Tickets/Emails/Hooks.php
codingmusician: I created a new method within the hooks file for this.

redscar (dismissed):

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -30,6 +31,51 @@ class Emails_Tab {
...
 	public static $slug = 'emails';
+	
+	/**
+	 * Holder for template object.
+	 *
+	 * @since TBD
+	 *
+	 * @var null|Tribe_Template
+	 */
+	private $template;
```
bordoni, line 42: Avoid private go with protected

bordoni (changes requested):

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -30,6 +31,51 @@ class Emails_Tab {
...
+	protected $template;
+
+	/**
+	 * Edit section slug.
+	 * 
+	 * @since TBD
+	 * 
+	 * @var string
+	 */
+	public static $edit_section_slug = 'edit_email';
```
juanfra, line 51: Can possibly go with `section` so that we have consistency on the levels Settings > Tab > Section? So that we can replicate that wherever we are going to have sections? WooCommerce does it this way, and there are plans to re-structure the settings (in general) to have these levels (Settings > tab > section)

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -30,6 +31,51 @@ class Emails_Tab {
...
+	public static $key_edit_id = 'id';
+
+	/**
+	 * Key to determine current section.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public static $key_current_section = 'tec_tickets_emails_current_section';
```
juanfra, line 69: The current section should come from the get variable `section`

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -30,6 +31,51 @@ class Emails_Tab {
...
+
+	/**
+	 * Key to determine current email id.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public static $key_current_email_id = 'tec_tickets_emails_current_email_id';
```
juanfra, line 78: The current email id will be the `section` -> something like `admin.php?page=tec-tickets-settings&tab=event-tickets&section=THE_EMAIL_ID` will be the section for each separate email.

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -172,4 +224,126 @@ public function get_fields(): array {
...
+	 * 
+	 * @return boolean
+	 */
+	public function is_editing_email( $email = null ) {
+		if ( ! $this->is_on_section( static::$edit_section_slug ) ) {
+			return false;
+		}
+
+		// Get `tc-edit` query string from URL.
+		$editing_email  = tribe_get_request_var( static::$key_edit_id );
```
juanfra, line 243: We can possibly simplify all this by using just `section` as a parameter for the different emails we have.

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -172,4 +224,126 @@ public function get_fields(): array {
...
+	 *
+	 * @since TBD
+	 * 
+	 * @return array|null Settings array
+	 */
+	public function get_email_settings() {
+		$email_id  = tribe_get_request_var( static::$key_edit_id );
+		$email = tribe( Email_Handler::class )->get_email_by_id( $email_id );
+		
+		$back_link = [[
```
juanfra, line 265: We should start the second array in a new line for readability.

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -172,4 +224,126 @@ public function get_fields(): array {
...
+			'type' => 'html',
+			'html' => $this->get_template()->template( 'back-link', [
```
juanfra, line 267: We should start the array in a new line.

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -172,4 +224,126 @@ public function get_fields(): array {
...
+				'text' => __( 'Back to Email Settings', 'event-tickets' ),
+				'url'  => $this->get_url(),
+			], false ),
+		]];
+
+		if ( ! $email ) {
+			return array_merge( $back_link, [
+				[
+					'type' => 'html',
+					'html' => '<p>' . esc_html__( 'Invalid email id selected.', 'event-tickets') . '</p>',
```
juanfra, line 277:
```suggestion
					'html' => '<p>' . esc_html__( 'Invalid email id selected.', 'event-tickets' ) . '</p>',
```

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -172,4 +224,126 @@ public function get_fields(): array {
...
+				]
+			] );
+		}
+
+		$hidden_fields = [[
```
juanfra, line 282: We should start the second array in a new line. We can simplify the checks by just having the check on `section`, we don't want to validate if it's a valid email section. It would work the same way as we have it for Tickets Commerce and the payment gateways (in that case `tc-section`, that we'll probably move to `section` in a future)

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -172,4 +224,126 @@ public function get_fields(): array {
...
+				esc_attr( static::$key_current_email_id ),
+				esc_attr( static::$key_current_email_id ),
+				esc_attr( $email_id )
+			)
+		]];
+
+		$settings = $email->get_settings();
+
+		return array_merge( $back_link, $settings, $hidden_fields );
+	}
```
juanfra, line 300: blank line between end of function and new function.

src/Tickets/Emails/Email/Ticket.php
```diff
@@ -55,15 +55,6 @@ class Ticket extends \TEC\Tickets\Emails\Email_Abstract {
 	 */
 	public $version = '1.0.0';
```
juanfra, line 56: The version will be just for the post type, so that we can perform updates if we need to (at any given point)

src/Tickets/Emails/Email/Ticket.php
```diff
@@ -142,12 +133,32 @@ public function get_title(): string {
...
-		$heading = '';
+		$heading = tribe_get_option( static::$option_heading, $this->get_default_heading() );
+
+		/**
+		 * Allow filtering the email heading.
+		 *
+		 * @since TBD
+		 *
+		 * @param string $heading  The email heading.
+		 * @param string $id       The ticket id.
```
juanfra, line 144:
```suggestion
		 * @param string $id       The email id.
```

src/Tickets/Emails/Email_Abstract.php
```diff
@@ -362,4 +364,23 @@ public function get_additional_content(): string {
...
+	 * @since TBD
+	 * 
+	 * @return string
+	 */
+	public function get_edit_url() {
+		// Force the `emails` tab.
+		$args = [
+			'tab'                    => Emails_Tab::$slug,
+			'section'                => Emails_Tab::$edit_section_slug,
+			Emails_Tab::$key_edit_id => $this->id,
```
juanfra, line 380: I think we can abstract things and just make it work with `section` as we do with Tickets Commerce and the payment gateways; there's no need to make an extra check if an existing email is being edited separately.

src/admin-views/settings/emails/message-templates/list-recipient.php
```diff
@@ -1,22 +1,19 @@
...
-if ( empty( $template ) ) {
+// If no email, bail.
+if ( empty( $email ) ) {
 	return;
 }
 
 ?>
 <div class="tec-tickets__admin-settings-emails-template-list-item-recipient">
-	To: <?php echo esc_html( $template['recipient'] ); ?>
+	To: <?php echo esc_html( $email->recipient ); ?>
```
juanfra, line 18: We need to make this string translatable, with a placeholder for the recipient.

juanfra (commented):
Looking good! I think we need to simplify the logic around settings > tab > section. And just go with `section`, as we're doing with the Tickets Commerce payment gateways section. We don't specifically need to test if we're on an existent email, because if we're using the `section` logic, if it's not on a specific email, then the settings won't even display.

src/Tickets/Emails/Admin/Emails_Tab.php
codingmusician: In the future, are we certain that `section` will only be used for editing emails?

src/Tickets/Emails/Admin/Emails_Tab.php
codingmusician: This is a variable for a hidden input field that is posted when settings are saved. The `section` variable is not available when the settings are posted. I ran into this issue when creating the provider sections for Tickets Commerce.

src/Tickets/Emails/Admin/Emails_Tab.php
codingmusician: I'll assume that we will not be using `section` for anything else other than editing emails. I was just trying to keep things flexible for the future.

src/Tickets/Emails/Email_Handler.php
```diff
@@ -197,4 +197,22 @@ public function add_settings_per_email( array $fields ): array {
 		return $fields;
```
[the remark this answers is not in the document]
codingmusician: Settings will be handled within the `Emails_Tab` class. Removing this method.

src/Tickets/Emails/Hooks.php
```diff
@@ -64,6 +66,21 @@ protected function add_filters() {
...
+	 * Filters the redirect URL to determine whether or not section key needs to be added.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $url Redirect URL.
+	 *
+	 * @return string
+	 */
+	public function filter_redirect_url( $url ) {
+		return $this->container->make( Emails_Tab::class )->filter_redirect_url( $url );
```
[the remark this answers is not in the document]
codingmusician: Unless something has changed since we added the Tickets Commerce tab, the section doesn't automatically get appended when saving the settings. In my testing, this was the case and this filter is needed.

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -127,9 +146,13 @@ public function is_on_tab(): bool {
...
 	public function is_on_section( $section = '' ): bool {
-		$settings = tribe( Plugin_Settings::class );
+		$current_section = tribe_get_request_var( 'section' );
+
+		if ( empty( $section ) ) {
+			return ! empty( $current_section );
+		}
 
-		return $settings->is_on_tab_section( self::$slug, $section );
+		return $section === $current_section;
```
juanfra, line 155: Why are we moving away from `is_on_tab_section`? That will make the check, we avoid code duplication, and at the same time, we're also checking against the tab and confirming that it's on the settings page.

juanfra: Oh, now I see that you use this on other functions to check if the `section` is not empty, assuming they're editing an email. Maybe we can rely on `is_editing_email ` as it already has that logic of the "empty section" and we keep this one more restricted to checking if it's on a particular section.

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -172,4 +201,122 @@ public function get_fields(): array {
...
+			return false;
+		}
+
+		// Get `section` query string from URL.
+		$editing_email  = tribe_get_request_var( 'section' );
+
+		// If email wasn't passed, just return whether or not string is empty.
+		if ( empty( $email ) ) {
+			return ! empty( $editing_email );
+		}
```
juanfra, line 225: This is what I mention that we can rely on here https://github.com/the-events-calendar/event-tickets/pull/2555/files#r1119849253

src/Tickets/Emails/Admin/Emails_Tab.php
```diff
@@ -172,4 +201,122 @@ public function get_fields(): array {
...
+	/**
+	 * Filters the redirect URL to include section, if applicable.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $url URL of redirection.
+	 *
+	 * @return string
+	 */
+	public function filter_redirect_url( $url ) {
```
juanfra, line 289: Do we still need this function? What would be the case in which it'll be used?

src/Tickets/Emails/Email_Handler.php
```diff
@@ -27,6 +27,15 @@ class Email_Handler extends \tad_DI52_ServiceProvider {
...
 	const POSTTYPE = 'tec_tickets_emails';
 
+	/**
+	 * Version number for custom post types.
+	 * 
+	 * @since TBD
+	 * 
+	 * @var string
+	 */
+	const VERSION = '1.0.0';
```
juanfra, line 37: I think my comments ended up confusing you here.

The version should come from each email individually as before the most recent changes, and we could use the plugin version number (set it as TBD). That way, we'll be able to update/perform update actions on each email individually rather than in all of them. What I mentioned is that we probably could avoid having that as a class variable.

src/admin-views/settings/emails/message-templates/list-recipient.php
```diff
@@ -1,22 +1,19 @@
...
-if ( empty( $template ) ) {
+// If no email, bail.
+if ( empty( $email ) ) {
 	return;
 }
 
 ?>
 <div class="tec-tickets__admin-settings-emails-template-list-item-recipient">
-	To: <?php echo esc_html( $template['recipient'] ); ?>
+	<?php echo esc_html__( 'To:', 'event-tickets' ); ?> <?php echo esc_html( $email->recipient ); ?>
```
juanfra, line 18:
```suggestion
	<?php echo sprintf( 
			// Translators: %s: The email recipient.
			esc_html__( 'To: %s', 'tribe-common' ), 
			esc_html( $email->recipient )
		); ?>
```

Normally, this type of translations need to be managed with placeholders. We don't know if, in other languages, the `To:` goes after, and so on. It's not always the case, but if we have the chance I think it's worth it to do it.

src/Tickets/Emails/Email_Handler.php
```diff
@@ -167,34 +176,28 @@ public function create_tec_tickets_emails_post_type( $email ) {
...
 	 *
-	 * @param array $fields The fields.
+	 * @since TBD
 	 *
-	 * @return array $fields
+	 * @param string $id ID of email.
+	 *
+	 * @return Email_Abstract|boolean Email object or false if it does not exist.
 	 */
-	public function add_settings_per_email( array $fields ): array {
```
juanfra, line 183: If we remove the function, we should also remove the call from the Hooks class.

juanfra (dismissed):
🏅

PR #2562, opened 2023-03-03
Fix Tickets Emails Missing Footer

codingmusician:
### 🗒️ Description

This fixes the logic that was causing the footer not to appear in the email template.

### ✔️ Checklist
- [x] My code is tested. 

### DIFF
--- a/src/views/v2/emails/template-parts/body/footer.php
+++ b/src/views/v2/emails/template-parts/body/footer.php
@@ -20,7 +20,7 @@
  * @var string $header_text_color Header text color.
  */
 
-if ( empty( $footer_content ) || empty( $footer_credit ) ) {
+if ( empty( $footer_content ) && empty( $footer_credit ) ) {
 	return;
 }
 ?>


src/views/v2/emails/template-parts/body/footer.php
```diff
@@ -20,7 +20,7 @@
  * @var string $header_text_color Header text color.
  */
 
-if ( empty( $footer_content ) || empty( $footer_credit ) ) {
+if ( empty( $footer_content ) && empty( $footer_credit ) ) {
```
juanfra, line 23: Even though, if any is empty, it should bail. Have you tested that's fixing the issue?

rafsuntaskin (approved):

src/views/v2/emails/template-parts/body/footer.php
juanfra: I mean, if the content is empty OR the footer credit is set to OFF, it should bail.

codingmusician: And, yes. I tested this and it fixes the issue.

juanfra: Ahh, right! I thought the footer content was the content of the credit. The footer content is actually what's coming from the setting. Thanks for the clarification.

PR #2563, opened 2023-03-06
[ET-1612] Fix - Undefined $going variable on Ajax request.

chigozieorunta:
### 🎫 Ticket

[1612](https://theeventscalendar.atlassian.net/browse/ETP-1612)

### 🗒️ Description

If I have an RSVP ticket for an event, When I click the “Going“ button, Then it will log a PHP Notice:
```
PHP Notice: Undefined variable: going in
/var/www/public/impossible-eel/wp-content/plugins/event-tickets/src/views/v2/rsvp/ari/form.php
```
I have added a simple check for the $going variable so the error doesn't show up.

### 🎥 Artifacts 
[Screencast](https://drive.google.com/file/d/1uCvndufUlaET-YYlQz_iuK0oA1bIz9HL/view?usp=share_link)

### ✔️ Checklist
- [X] I've included a changelog entry in the readme.txt file. 
- [] My code is tested. 
- [X] My code has [proper inline documentation](https://the-events-calendar.github.io/products-engineering/docs/code-standards/).


src/views/v2/rsvp/ari/form.php
```diff
@@ -18,6 +18,9 @@
  * @version 5.0.0
  */
 
+// Ensure $going is not undefined in Ajax call.
+$going = isset( $going ) ? $going : null;
```
juanfra, line 22: We can probably avoid modifying the template and rather setting the `null` [here](https://github.com/the-events-calendar/event-tickets/blob/master/src/Tribe/Tickets_View.php#L1232).

Every time we update a template we'll have to put a `@since TBD Description of the changes`, and then set the version number of the template to `@version TBD`. So that can be adjusted before the release.

juanfra (changes requested):
We need to double-check the syncs for common, and utilities. And see if that's necessary, they shouldn't be part of a normal PR.

We should't be adding changes to the `package-lock.json` file, unless is precisely related.

Finally, we always need to add a readme entry and share the artifacts if the fix in the PR description.

src/views/v2/rsvp/ari/form.php
chigozieorunta: @juanfra I have attempted setting the null value for the $going variable in the Tribe__Tickets__Tickets_View class, but unfortunately, this doesn't effect the chang adequately. I have used the get method for retrieving the status as used in other template forms like so - [here](https://github.com/the-events-calendar/event-tickets/blob/master/src/views/v2/rsvp/form/form.php#L19).

src/views/v2/rsvp/ari/form.php
```diff
@@ -13,12 +13,14 @@
  * @var Tribe__Tickets__Ticket_Object $rsvp The rsvp ticket object.
  * @var WP_Post|int $post_id The post object or ID.
  *
- * @since 4.12.3
```
redscar, line 16: Instead of overwriting the `@since` we usually append a new one. Each `@since` should outline what was changed exactly like what you did.

chigozieorunta: @redscar Thanks for pointing this out to me.

bordoni (approved):

bordoni (changes requested):
Lets make sure the tests are passing here. We can take a look on that Monday.

redscar:
The code looks good. The tests are failing currently. You may need to update snapshots.

src/views/v2/rsvp/ari/form.php
```diff
@@ -14,11 +14,14 @@
...
  * @since 4.12.3
+ * @since TBD Ensured that $going is not undefined in Ajax call.
  *
- * @version 5.0.0
+ * @version TBD
  */
 
+$going = $this->get( 'going' );
 ?>
+
```
juanfra, line 24: This extra line may be the reason why snapshots are failing. I'd remove it.

juanfra (commented):
Good progress! left some notes.

src/views/v2/rsvp/ari/form.php
chigozieorunta: @juanfra Ok.

bordoni (dismissed):

juanfra (commented):
Sorry Chigz to be persistent on this one, but I don't think we need to modify the template. Because in the end, these variables are sent to the template. If the template doesn't found it, is not a problem of the template, but rather the method that's invoking it. This is in favor of consistency and to keep templates clean for people working with them (as users can override them).

Can you please try adding:

`'going'      => tribe_get_request_var( 'going', '' ),`

To [this array](https://github.com/the-events-calendar/event-tickets/blob/master/src/Tribe/RSVP.php#L317)?

That should make it.

chigozieorunta:
Thank you @juanfra you are absolutely spot on with the point you raised on the template consistency and the ability for users to override them. I have modified accordingly.

src/views/v2/rsvp/ari/form.php
```diff
@@ -14,10 +14,10 @@
  * @var WP_Post|int $post_id The post object or ID.
  *
  * @since 4.12.3
+ * @since TBD Ensured that $going is not undefined in Ajax call.
```
juanfra, line 17: We need to revert this as we're not changing the template.
```suggestion
```

src/views/v2/rsvp/ari/form.php
```diff
@@ -14,10 +14,10 @@
...
  *
- * @version 5.0.0
+ * @version TBD
```
juanfra, line 19:
```suggestion
 * @version 5.0.0
```

juanfra:
Thanks @chigozieorunta! We need to revert the changes on the template, as there are no new changes. We only update the template version number if there are changes in the template.

juanfra (approved):
🚀

src/Tribe/RSVP.php
```diff
@@ -323,6 +323,7 @@ public function render_rsvp_step( $ticket_id, $step = null ) {
 			'must_login' => ! is_user_logged_in() && $this->login_required(),
 			'login_url'  => self::get_login_url( $post_id ),
 			'threshold'  => $blocks_rsvp->get_threshold( $post_id ),
+			'going'      => tribe_get_request_var( 'going', '' ),
```
redscar, line 326: This looks good. Make sure you add a new `@since` tag within the `render_rsvp_step` docblock.

src/views/v2/rsvp/ari/form.php
```diff
@@ -17,7 +17,6 @@
  *
  * @version 5.0.0
  */
-
```
redscar, line 20: Is the removal of this line necessary?

redscar (commented):
I added some comments.  I also see a test is failing.

src/views/v2/rsvp/ari/form.php
chigozieorunta: Yes, because we are not modifying anything in that template anymore.

src/Tribe/RSVP.php
chigozieorunta: Thanks @redscar

src/views/v2/rsvp/ari/form.php
chigozieorunta: Sorry @redscar I see what you mean with the snapshots failing. I have added the new line spacing to see if this fixes it.

tests/wpunit/Tribe/Tickets/__snapshots__/RSVPTest__it_should_render_rsvp_step with data set ari__1.php
```diff
@@ -108,7 +108,7 @@ class="tribe-tickets__rsvp-ar-guest-list-item-button tribe-tickets__rsvp-ar-gues
 		<div class="tribe-tickets__rsvp-ar-form">
 
 	<input type="hidden" name="tribe_tickets[TICKET_ID][ticket_id]" value="TICKET_ID">
-	<input type="hidden" name="tribe_tickets[TICKET_ID][attendees][0][order_status]" value="not-going">
+	<input type="hidden" name="tribe_tickets[TICKET_ID][attendees][0][order_status]" value="">
```
redscar, line 111: Since the original snapshot had the value as "not-going" should the default for going be "not-going" instead of empty ""?

chigozieorunta: Yeah. That's the one I updated with Slic. I think its not supposed to have "not-going" @juanfra

redscar (approved):
Great Job :+1:

PR #2568, opened 2023-03-08
Tickets Emails: New Emails and Settings

codingmusician:
### 🎫 Ticket

[ET-1598] 
[ET-1622] 
[ET-1621] 
[ET-1620] 

### 🗒️ Description

Adds new emails and setting for Tickets Emails feature.

### 🎥 Artifacts 

image

### ✔️ Checklist
- [x] My code is tested. 

[ET-1598]: https://theeventscalendar.atlassian.net/browse/ET-1598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-1622]: https://theeventscalendar.atlassian.net/browse/ET-1622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-1621]: https://theeventscalendar.atlassian.net/browse/ET-1621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-1620]: https://theeventscalendar.atlassian.net/browse/ET-1620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Emails/Email/Ticket.php
```diff
@@ -235,22 +295,40 @@ public function get_settings(): array {
...
+				'label'               => esc_html__( 'Subject', 'event-tickets' ),
 				'default'             => $this->get_default_subject(),
-				'placeholder'         => esc_html__( 'Your tickets to {event_name}', 'event-tickets' ),
+				'placeholder'         => $this->get_default_subject(),
+				'size'                => 'large',
+				'validation_callback' => 'is_string',
+			],
+			static::$option_subject_multiple => [
+				'type'                => 'text',
+				'label'               => esc_html__( 'Subject (multiple)', 'event-tickets' ),
```
juanfra, line 312: Any reason why you chose multiple over plural like it was in the ticket description?

src/Tickets/Emails/Email_Abstract.php
```diff
@@ -376,10 +376,21 @@ public function get_edit_url() {
...
 	}
+
+	/**
+	 * Get ID.
+	 * 
+	 * @since TBD
+	 * 
+	 * @return string
+	 */
+	public function get_id() {
```
juanfra, line 393:
```suggestion
	public function get_id(): string {
```

juanfra (commented):
Looking good! I left some comments, feel free to reach out if you want to talk about any

src/Tickets/Emails/Email/Ticket.php
```diff
@@ -186,7 +224,29 @@ public function get_subject(): string {
...
+	}
+
+	/**
+	 * Get default email subject for multiple tickets.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public function get_default_subject_multiple() {
```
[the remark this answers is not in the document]
codingmusician: I meant to bring this up. I think `plural` is more of a grammatical adjective. It describes the way a word is being used. Whereas `multiple` make more sense when we're describing the actual tickets.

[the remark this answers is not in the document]
codingmusician: For example: "You purchased `multiple` tickets" make more sense than "You purchased `plural` tickets".

src/Tickets/Emails/Email/Completed_Order.php
```diff
@@ -0,0 +1,357 @@
...
+
+		/**
+		 * Allow filtering the settings for this email.
+		 *
+		 * @since TBD
+		 *
+		 * @param array          $settings  The settings array.
+		 * @param Email_Abstract $this      Email object.
+		 */
+		return apply_filters( 'tec_tickets_emails_settings_' . self::$id, $settings, $this );
```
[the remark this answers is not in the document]
codingmusician: Rather than using a variable in the filter names, I went ahead and just added each of the email names in so they made sense. For example, this one you pointed out is now `tec_tickets_emails_completed_order_settings`

src/Tickets/Emails/Email/Completed_Order.php
```diff
@@ -0,0 +1,402 @@
...
+		/**
+		 * Allow filtering the email heading globally.
+		 *
+		 * @since TBD
+		 *
+		 * @param string          $heading  The email heading.
+		 * @param string          $id       The email id.
+		 * @param Tribe__Template $template Template object.
+		 */
+		$heading = apply_filters( 'tec_tickets_emails_heading', $heading, self::$id, $this->template );
```
juanfra, line 202: The same comment about the filter naming and parameter as above.

src/Tickets/Emails/Email/Completed_Order.php
```diff
@@ -0,0 +1,402 @@
...
+		/**
+		 * Allow filtering the email subject globally.
+		 *
+		 * @since TBD
+		 *
+		 * @param string          $subject  The email subject.
+		 * @param string          $id       The email id.
+		 * @param Tribe__Template $template Template object.
+		 */
+		$subject = apply_filters( 'tec_tickets_emails_subject', $subject, self::$id, $this->template );
```
juanfra, line 253: The same comment about the filter naming and parameter as above. We're repeating these all over, and we'll be doing it in every email that we'll register. If we're doing a generic filter we should centralize it.

src/Tickets/Emails/Email/Completed_Order.php
```diff
@@ -0,0 +1,402 @@
...
+				'default'             => $this->get_default_heading(),
+				'placeholder'         => $this->get_default_heading(),
+				'size'                => 'large',
+				'validation_callback' => 'is_string',
+			],
+			static::$option_add_content => [
+				'type'                => 'wysiwyg',
+				'label'               => esc_html__( 'Additional content', 'event-tickets' ),
+				'default'             => '',
+				'tooltip'             => esc_html__( 'Additional content will be displayed below the tickets in your email.', 'event-tickets' ),
```
juanfra, line 345:
```suggestion
				'tooltip'             => esc_html__( 'Additional content will be displayed below the order details.', 'event-tickets' ),
```

src/Tickets/Emails/Email/Completed_Order.php
```diff
@@ -0,0 +1,402 @@
...
+
+		/**
+		 * Allow filtering the settings for this email.
+		 *
+		 * @since TBD
+		 *
+		 * @param array  $settings  The settings array.
+		 * @param string $id        Email ID.
+		 */
+		return apply_filters( 'tec_tickets_emails_settings', $settings, self::$id );
```
juanfra, line 382: The same comment about the filter naming and parameter as above.

src/Tickets/Emails/Email/Failed_Order.php
```diff
@@ -0,0 +1,402 @@
...
+	 * Email heading option key.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public static $option_heading = 'tec-tickets-emails-failed-order-heading';
+
+	/**
+	 * Email heading option key.
```
juanfra, line 85: Additional content

src/Tickets/Emails/Email/Failed_Order.php
```diff
@@ -0,0 +1,402 @@
...
+	 * Get email subject.
+	 *
+	 * @since TBD
+	 *
+	 * @return string The email subject.
+	 */
+	public function get_subject(): string {
+		$subject = tribe_get_option( static::$option_subject, $this->get_default_subject() );
+
+		$subject = $this->format_string( $subject );
```
juanfra, line 228: We should probably format the string later, as they will probably want to use the placeholders if they apply filters.

src/Tickets/Emails/Email/Purchase_Receipt.php
```diff
@@ -0,0 +1,340 @@
...
+	 * Email heading option key.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public static $option_heading = 'tec-tickets-emails-purchase-receipt-heading';
+
+	/**
+	 * Email heading option key.
```
juanfra, line 76: Additional content.

src/Tickets/Emails/Email/Purchase_Receipt.php
```diff
@@ -0,0 +1,340 @@
...
+	 * Get email subject.
+	 *
+	 * @since TBD
+	 *
+	 * @return string The email subject.
+	 */
+	public function get_subject(): string {
+		$subject = tribe_get_option( static::$option_subject, $this->get_default_subject() );
+
+		$subject = $this->format_string( $subject );
```
juanfra, line 173: We should do the format string later, as the last step, so that there's access to the placeholders when they filter.

juanfra (commented):
Looking good, I think we need to change how we do it with the filters, it'd make more sense to use the variable as part of the filter name (specific to the email), and then if we have one generic filter it should come from a centralized place rather than repeating the same filter throughout the different classes (probably in the abstract class). 

And we should always format the strings late, before "returning" them. As the format string method has its own filter, and the other filters should have the placeholders available.

src/Tickets/Emails/Email/Completed_Order.php
```diff
@@ -0,0 +1,402 @@
...
+		/**
+		 * Allow filtering the email recipient for Completed Order.
+		 *
+		 * @since TBD
+		 *
+		 * @param string          $heading  The email recipient.
+		 * @param string          $id       The email id.
+		 * @param Tribe__Template $template Template object.
+		 */
+		$recipient = apply_filters( 'tec_tickets_emails_completed_order_recipient', $recipient, self::$id, $this->template );
```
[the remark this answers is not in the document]
codingmusician: I was thinking about the actual name of the hook if we were to actually use the id name. It would be `tec_tickets_emails_recipient_tec_tickets_emails_completed_order` vs just using `tec_tickets_emails_completed_order_recipient`. Plus, I didn't see many examples in our codebase where we use variables within a hook name. What are your thoughts about using the `template` name instead of `id`? Example: `tec_tickets_emails_{$template}_recipient`. That would make more sense as far as readability. And each will have to be a unique template name anyway.

src/Tickets/Emails/Email/Purchase_Receipt.php
```diff
@@ -115,38 +68,14 @@ public function get_title(): string {
...
-		 *
-		 * @param string          $heading  The email heading.
-		 * @param string          $id       The email id.
-		 * @param Tribe__Template $template Template object.
-		 */
-		$heading = apply_filters( 'tec_tickets_emails_heading', $heading, self::$id, $this->template );
-
-		return $this->format_string( $heading );
+	public function get_default_recipient(): string {
+		return '{purchaser-email}';
```
juanfra, line 78: Not sure if this should come from a placeholder, as we would only use placeholders for "printable" chunks (subject, title, additional content).

In any case the format is with underscores `{purchaser_email}`. We could probably define this afterwards, when we properly populate the email with the context data.

```suggestion
		return '{purchaser_email}';
```

src/Tickets/Emails/Email/Ticket.php
```diff
@@ -134,38 +58,14 @@ public function get_title(): string {
...
-		 *
-		 * @param string          $heading  The email heading.
-		 * @param string          $id       The email id.
-		 * @param Tribe__Template $template Template object.
-		 */
-		$heading = apply_filters( 'tec_tickets_emails_heading', $heading, self::$id, $this->template );
-
-		return $this->format_string( $heading );
+	public function get_default_recipient(): string {
+		return '{attendee-email}';
```
juanfra, line 68: Again, not sure if this will work, as the ticket email is sent out to the purchaser, and attendee (if IAC). Also, we should use underscores for the placeholder format.

```suggestion
		return '{attendee_email}';
```

src/Tickets/Emails/Email/Ticket.php
```diff
@@ -184,13 +84,13 @@ public function get_default_heading() {
...
 	 *
 	 * @since TBD
 	 *
 	 * @return string
 	 */
-	public function get_default_heading_multiple() {
+	public function get_default_heading_plural(): string {
 		return sprintf(
 			// Translators: %s Lowercase plural of tickets.
 			esc_html__( 'Here are your %s, {attendee-name}!', 'event-tickets' ),
```
juanfra, line 96:
```suggestion
			esc_html__( 'Here are your %s, {attendee_name}!', 'event-tickets' ),
```

src/Tickets/Emails/Email/Ticket.php
```diff
@@ -199,43 +99,42 @@ public function get_default_heading_multiple() {
...
+		 * @param string $heading  The email heading.
+		 * @param string $id       The email id.
+		 * @param string $template Template name.
 		 */
-		$subject = apply_filters( 'tec_tickets_emails_ticket_subject', $subject, self::$id, $this->template );
+		$heading = apply_filters( 'tec_tickets_emails_heading_plural', $heading, $this->id, $this->template );
 
 		/**
-		 * Allow filtering the email subject.
+		 * Allow filtering the email heading for Completed Order.
```
juanfra, line 126: Let's fix the docblock, it's not for the completed order.

src/Tickets/Emails/Email/Ticket.php
```diff
@@ -199,43 +99,42 @@ public function get_default_heading_multiple() {
...
-		 * @param string          $subject  The email subject.
-		 * @param string          $id       The email id.
-		 * @param Tribe__Template $template Template object.
+		 * @param string $heading  The email heading.
+		 * @param string $id       The email id.
+		 * @param string $template Template name.
 		 */
-		$subject = apply_filters( 'tec_tickets_emails_subject', $subject, self::$id, $this->template );
+		$template_name = static::$template;
+		$heading = apply_filters( "tec_tickets_emails_{$template_name}_heading_plural", $heading, $this->id, $this->template );
```
juanfra, line 135: We should use the ID as part of the hook filter. The idea is that people will be able to extend this functionality and use their own templates if they want. The template is meant to store the filename that's going to be called to get the content of the email, and that _could_ change if extended. The ID will always be the same for the email.

src/Tickets/Emails/Email/Ticket.php
```diff
@@ -272,37 +171,53 @@ public function get_default_subject_multiple() {
...
+		/**
+		 * Allow filtering the email subject globally.
+		 *
+		 * @since TBD
+		 *
+		 * @param string $subject  The email heasubjectding.
+		 * @param string $id       The email id.
+		 * @param string $template Template name.
+		 */
+		$subject = apply_filters( 'tec_tickets_emails_subject_plural', $subject, $this->id, $this->template );
```
juanfra, line 195: I think we're only going to have plural subjects for Tickets and RSVPs, not sure if we need to have a global filter. We might if we repeat this in the RSVP, though.

src/Tickets/Emails/Email_Abstract.php
```diff
@@ -391,6 +381,260 @@ public function get_edit_url() {
...
+	/**
+	 * Get setting option key.
+	 * 
+	 * @since TBD
+	 * 
+	 * @return string
+	 */
+	public function get_option_key( $option ): string {
+		$template_name = $this->template;
+		return "tec-tickets-emails-{$template_name}-{$option}";
```
juanfra, line 396: As mentioned above, the template name will refer to the file that's being called to generate the HTML. I like the idea of having something for that, maybe we need to define a different variable for it? Like a "prefix" or something. But we need to reserve the `$template` to be extended.

That way, if someone (we or third parties) want to have their own template for X email, they can do it.

src/Tickets/Emails/Email_Abstract.php
```diff
@@ -391,6 +381,260 @@ public function get_edit_url() {
...
+		 * Allow filtering the email recipient for Completed Order.
+		 *
+		 * @since TBD
+		 *
+		 * @param string $recipient  The email recipient.
+		 * @param string $id         The email id.
+		 * @param string $template   Template name.
+		 */
+		$template_name = $this->template;
+		$recipient = apply_filters( "tec_tickets_emails_{$template_name}_recipient", $recipient, $this->id, $this->template );
```
juanfra, line 445: The same thing about the `$template`, we should rely on the $id or something new that stores a unique immutable value.

src/Tickets/Emails/Email_Abstract.php
```diff
@@ -391,6 +381,260 @@ public function get_edit_url() {
...
+		 * Allow filtering the email subject.
+		 *
+		 * @since TBD
+		 *
+		 * @param string $subject  The email subject.
+		 * @param string $id       The email id.
+		 * @param string $template Template name.
+		 */
+		$template_name = $this->template;
+		$subject = apply_filters( "tec_tickets_emails_{$template_name}_subject", $subject, $this->id, $this->template );
```
juanfra, line 484: The same comment about `$template_name`

src/Tickets/Emails/Email_Abstract.php
```diff
@@ -391,6 +381,260 @@ public function get_edit_url() {
...
+		 * Allow filtering the email heading for Completed Order.
+		 *
+		 * @since TBD
+		 *
+		 * @param string $heading  The email heading.
+		 * @param string $id       The email id.
+		 * @param string $template Template name.
+		 */
+		$template_name = $this->template;
+		$heading = apply_filters( "tec_tickets_emails_{$template_name}_heading", $heading, $this->id, $this->template );
```
juanfra, line 523: The same comment about `$template_name`.

src/Tickets/Emails/Email_Abstract.php
```diff
@@ -391,6 +381,260 @@ public function get_edit_url() {
...
+	/**
+	 * Get email content.
+	 *
+	 * @since TBD
+	 *
+	 * @param array $args The arguments.
+	 *
+	 * @return string The email content.
+	 */
+	public function get_content( $args = [] ): string {
```
juanfra, line 537: We will need to set the context differently per email, and send those things to the template to receive the HTML. I'd keep this method in the other classes.

src/Tickets/Emails/Email/Purchase_Receipt.php
codingmusician: Yeah. I figured we'd be changing this when we start implementing the data that goes into it. I just wanted to put in a placeholder for now.

src/Tickets/Emails/Email/Ticket.php
codingmusician: Yeah. I figured we'd be changing this when we start implementing the data that goes into it. I just wanted to put in a placeholder for now.

src/Tickets/Emails/Email/Ticket.php
codingmusician: I created a new variable specifically for filters and options called `slug`. That way filter and option names won't be so long.

src/Tickets/Emails/Email/Ticket.php
codingmusician: Right. I was thinking about RSVPs, actually.

src/Tickets/Emails/Email_Abstract.php
codingmusician: We'll use the new `slug` variable.

src/Tickets/Emails/Email_Abstract.php
codingmusician: We'll use the new `slug` variable.

src/Tickets/Emails/Email_Abstract.php
codingmusician: We'll use the new `slug` variable.

src/Tickets/Emails/Email_Abstract.php
codingmusician: We'll use the new `slug` variable.

src/Tickets/Emails/Email_Abstract.php
codingmusician: I moved the `get_content()` method back into each email class.

juanfra (approved):
🏅

PR #2575, opened 2023-03-14
Adds support for WPML

dgwatkins:
The idea is to add the ids for the translation when querying the database. It is done in a way that we can reuse the code for WooCommerce, EDD & built-in tickets.

This is more a proof of concept to get the ball going so I didn't update the changelog and I don't have a ticket id to fill in the template.


dgwatkins:
Closed in favor of #3044

PR #2578, opened 2023-03-16
[ET-1623] Add RSVP Email

codingmusician:
### 🎫 Ticket

[ET-1623] 

### 🗒️ Description

Adds the RSVP email and settings to the Tickets Emails system.

### 🎥 Artifacts 

image
image

### ✔️ Checklist
- [x] My code is tested. 

[ET-1623]: https://theeventscalendar.atlassian.net/browse/ET-1623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Emails/Email/RSVP.php
+++ b/src/Tickets/Emails/Email/RSVP.php
@@ -7,6 +7,7 @@
 
 namespace TEC\Tickets\Emails\Email;
 
+use TEC\Tickets\Commerce\Settings as Settings;
 use \TEC\Tickets\Emails\Email_Template;
 
 /**
@@ -85,9 +86,9 @@ class RSVP extends \TEC\Tickets\Emails\Email_Abstract {
 	 */
 	public function get_default_heading(): string {
 		return sprintf(
-			// Translators: %s Lowercase singular of rsvp.
+			// Translators: %s Lowercase singular of ticket.
 			esc_html__( 'Here\'s your %s, {attendee_name}!', 'event-tickets' ),
-			tribe_get_rsvp_label_singular()
+			tribe_get_ticket_label_singular_lowercase()
 		);
 	}
 
@@ -100,9 +101,9 @@ class RSVP extends \TEC\Tickets\Emails\Email_Abstract {
 	 */
 	public function get_default_heading_plural(): string {
 		return sprintf(
-			// Translators: %s Lowercase plural of rsvps.
+			// Translators: %s Lowercase plural of tickets.
 			esc_html__( 'Here are your %s, {attendee_name}!', 'event-tickets' ),
-			tribe_get_rsvp_label_plural()
+			tribe_get_ticket_label_plural_lowercase()
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


src/Tickets/Emails/Email/RSVP.php
```diff
@@ -0,0 +1,319 @@
+<?php
+/**
+ * Class RSVP
+ *
+ * @package TEC\Tickets\Emails
+ */
+
+namespace TEC\Tickets\Emails\Email;
+
+use TEC\Tickets\Commerce\Settings;
```
juanfra, line 10: Do we need this here?

src/Tickets/Emails/Email/RSVP.php
```diff
@@ -0,0 +1,319 @@
...
+			[
+				'type' => 'html',
+				'html' => '<p>' . esc_html__( 'Registrants will receive an email including their RSVP info upon registration. Customize the content of this specific email using the tools below. The brackets {event_name}, {event_date}, and {rsvp_name} can be used to pull dynamic content from the RSVP into your email. Learn more about customizing email templates in our Knowledgebase.' ) . '</p>',
+			],
+			$this->get_option_key( 'enabled' ) => [
+				'type'                => 'toggle',
+				'label'               => esc_html__( 'Enabled', 'event-tickets' ),
+				'default'             => true,
+				'validation_type'     => 'boolean',
+			],
```
juanfra, line 242: We're missing the `Use Ticket Email` described on the Jira ticket.

juanfra (commented):
Looking good. Left a couple of comments. I remember we talked about the "order" in the post type arguments, seeing the screenshots reinforces the idea that we'll need that as the first item on the list should be the "Ticket" email.

src/Tickets/Emails/Email/RSVP.php
codingmusician: Ah. No we don't. It was copied from the Ticket email class. I'll remove it.

src/Tickets/Emails/Email/RSVP.php
codingmusician: Well... I noticed that the JIRA ticket referenced a lot of "ticket" info, so I assumed it was copied over and not yet updated. I'll add it, though.

juanfra: The idea is that, as a user, you can fall back to the Ticket email template and settings in case you want to have the same thing for both.

juanfra: The other things referencing tickets (subject, heading, and so on) were also intentional. Missed that on the CR.

codingmusician: Oh...

codingmusician: Added it. I also added the logic to not show the remaining settings if they opt to use the ticket email settings instead.

src/Tickets/Emails/Email/RSVP.php
codingmusician: It was added back, now that I see that we want to use the original setting from Commerce Settings as the default.

juanfra (approved):

PR #2605, opened 2023-04-18
Tickets Emails: Dynamization of Orders and Tickets in Emails

codingmusician:
### 🎫 Ticket

[TICKET_ID] 

### 🗒️ Description

Working on dynamizing the email templates and making sure that preview data matches what live data will actually look like. This PR focuses on orders and the tickets contained within each order.

### ✔️ Checklist
- [x] My code is tested. 


src/Tickets/Commerce/Models/Order_Model.php
```diff
@@ -57,6 +57,15 @@ protected function build_properties( $filter ) {
 			$events_in_order  = (array) Arr::get( $post_meta, [ Order::$events_in_order_meta_key ] );
 			$tickets_in_order = (array) Arr::get( $post_meta, [ Order::$tickets_in_order_meta_key ] );
 
+			$tickets = [];
```
juanfra, line 60: We already have `$tickets_in_order`, wouldn't it be redundant to have the tickets in two different places? For performance purposes I'd use our `Lazy` collection or some other way to make it more performant.

@bordoni

src/views/v2/emails/template-parts/body/order/ticket-totals/ticket-title.php
```diff
@@ -23,7 +23,7 @@
  * @var \WP_Post       $order              The order object.
  */
 
-$ticket_title = empty( $ticket['title'] ) ? '' : $ticket['title'];
+$ticket_title = get_the_title( $ticket );
```
juanfra, line 26: If we're already saving the ticket title in the `$order` object, let's make use of that to make it more performant, rather than re-requesting the title with `get_the_title`

src/Tickets/Commerce/Models/Order_Model.php
codingmusician: I saw that `tickets_in_order` was being used elsewhere in the codebase, so I didn't want to break anything by changing the structure.

codingmusician: Here's where I saw the reference to `tickets_in_order`: https://github.com/the-events-calendar/event-tickets/blob/3a0958cf333841760ca42feb2f86c3530b9c2929/src/Tickets/Commerce/Repositories/Order_Repository.php#L141-L182

codingmusician: The `tickets_in_order` array is specifically an array of ticket IDs that are being stored in the meta as `_tec_tc_order_tickets_in_order`.

codingmusician:
We're going to work on dynamizing the attendee data in a separate PR. This PR handles Orders and Tickets.

src/Tickets/Commerce/Models/Order_Model.php
codingmusician: @Camwyn or @bordoni could y'all offer some feedback regarding the above convo? Do you see a problem in having a `tickets` array alongside the `tickets_in_order` array?

rafsuntaskin: I think we should avoid adding all ticket data into the Order Model like above. As this will impact the ORM performance for all other queries even without templates.

1. I would suggest that we add a context variable of the type array of `tickets` and organize the data structure for the child templates as well so that we won't need to include separate classes in the template files like this: https://github.com/the-events-calendar/event-tickets/blob/53233e484be7d86f989c7d0278dd2ea3f0af4298/src/views/v2/emails/template-parts/body/order/ticket-totals/ticket-price.php#L26-L28

2. If we change the order model then we should do it like @juanfra suggested and implement `Lazy_Collection` for the `tickets`

bordoni (approved):

PR #2613, opened 2023-04-25
[ET-1579] Spotfix for unavailable message not filtered fully.

redscar:
### 🎫 Ticket

[ET-1579] 

This ticket is a continuation of https://github.com/the-events-calendar/event-tickets/pull/2600

While testing the filter wasn't being applied to all the correct locations. This adds it to the last location to allow for testing.

The filter can be used by using the below code - 
```
function my_event_tickets_unavailable_message( $message, $ticket ) {
    // Modify the $message variable here
    $message = 'Oopsie! This ticket is currently not available at all. Sorry!';

    // Return the modified $message variable
    return $message;
}
add_filter( 'event_tickets_unvailable_message', 'my_event_tickets_unavailable_message', 10, 2 );
```

Screenshot of above code applied.
image


src/views/v2/tickets/item/inactive.php
```diff
@@ -61,6 +61,15 @@
...
 	/* translators: %s: Tickets label */
 	: $provider->get_tickets_unavailable_message( $tickets );
+
+/**
+ * Filters the unavailability message for a ticket collection
+ *
+ * @param $message string Unavailability message
+ * @param $tickets array Collection of tickets
+ */
+$message = apply_filters( 'event_tickets_unvailable_message', $message, $tickets );
```
bordoni, line 71: This is not the correct namespace for this filter. Add changelog

bordoni (changes requested):

src/Tribe/Tickets.php
```diff
@@ -3460,10 +3460,10 @@ public function maybe_show_tickets_unavailable_message( $tickets, $post_id = nul
 				 *
 				 * @since TBD
 				 *
-				 * @param string Unavailability message
-				 * @param array Collection of tickets
+				 * @param $message string Unavailability message
+				 * @param $tickets array Collection of tickets
 				 */
-				echo (string) apply_filters( 'event_tickets_unvailable_message', $message, $tickets );
+				echo (string) apply_filters( 'tec_tickets_unavailable_message', $message, $tickets );
```
rafsuntaskin, line 3466: This is an existing filter and we should not rename it. Instead let's deprecate the previous one and then add the new filter as well.

src/views/v2/tickets/item/inactive.php
```diff
@@ -61,6 +61,17 @@
...
+
+/**
+ * Filters the unavailability message for a ticket collection
+ *
+ * @since TBD
+ *
+ * @param $message string Unavailability message
+ * @param $tickets array Collection of tickets
+ */
+$message = apply_filters( 'tec_tickets_unavailable_message', $message, $tickets );
```
rafsuntaskin, line 73: If the filter is already within `get_tickets_unavailable_message` then we should not need it here again.

rafsuntaskin (changes requested):

src/views/v2/tickets/item/inactive.php
rafsuntaskin: Also, this will disregard the $is_sale_past message and always show the `tec_tickets_unavailable_message` filtered value when this filter is used.

redscar: @rafsuntaskin I don't disagree with you. However, by removing the filter from this file we default to `sprintf( __( '%s are no longer available', 'event-tickets' ),`. Shouldn't that be filterable as well?

rafsuntaskin: That is within the template, so if anyone wants to change that they can simply overrride the template. Having a filter makes more sense as well, but I just think that's outside the scope for this ticket.

codingmusician: @redscar If anything, it should be a new filter. Something other than `tec_tickets_unavailable_message`. Could cause problems filtering it twice.

codingmusician: Plus, like @rafsuntaskin mentioned, this is within the template file. It's pretty uncommon to use filters here.

redscar:
Punting until we can get acceptance/user stories for the original ticket.

PR #2615, opened 2023-04-27
Simplify the Handler with Post Type class.

bordoni:
None of this code here is actually supposed to go into `release/G23.edamame`,  here as a placeholder. Moving functionality to simplify classes.

This code changes no behavior at all, simply trying to simplify the code-base.

PR #2616, opened 2023-04-27
Remove the usage of __get and __set from Emails

bordoni:
This code depends on changes made on #2615, no behavior changes.


juanfra:
Can you expand on why they're not being used correctly from your perspective? I am fine changing how that's handled, but I [wouldn't call it](https://www.php.net/manual/en/language.oop5.overloading.php#object.set) wrong.

codingmusician (approved):

PR #2619, opened 2023-04-27
Update to avoid creating more bloat on main Module files

bordoni:
I noticed that the changes we made in Tickets Emails were adding more code to the legacy Modules, which we want to avoid as they are already impossible to read, 3000 lines each.

📹 https://i.bordoni.me/2BwGnstc

### DIFF
--- a/src/Tickets/Emails/Hooks.php
+++ b/src/Tickets/Emails/Hooks.php
@@ -71,13 +71,13 @@ class Hooks extends tad_DI52_ServiceProvider {
 
 		add_filter( 'wp_redirect', [ $this, 'filter_redirect_url' ] );
 
-		// Legacy emails.
+		// Inject the new email into the legacy codebase for emails.
 		add_filter( 'tec_tickets_send_rsvp_email_pre', [ $this, 'filter_legacy_sent_rsvp_emails' ], 20, 4 );
 		add_filter( 'tec_tickets_send_tickets_email_for_attendee_pre', [ $this, 'filter_legacy_sent_tickets_attendees_emails' ], 20, 5 );
 	}
 
 	/**
-	 * Filters the redirect URL to determine whether or not section key needs to be added.
+	 * Filters the redirect URL to determine whether section key needs to be added.
 	 *
 	 * @since 5.5.9
 	 *
@@ -136,7 +136,7 @@ class Hooks extends tad_DI52_ServiceProvider {
 	 * @since 5.5.7
 	 */
 	public function action_add_preview_modal() {
-		echo $this->container->make( Admin\Preview_Modal::class )->render_modal();
+		$this->container->make( Admin\Preview_Modal::class )->render_modal();
 	}
 
 	/**
@@ -239,7 +239,7 @@ class Hooks extends tad_DI52_ServiceProvider {
 	}
 
 	/**
-	 * Hooks to the legacy modules to send Tickets emails with the new system.
+	 * Hooks to the legacy Tickets Module to send Tickets emails with the new system.
 	 *
 	 * @since TBD
 	 *
--- a/src/Tickets/Emails/Hooks.php
+++ b/src/Tickets/Emails/Hooks.php
@@ -19,6 +19,7 @@ namespace TEC\Tickets\Emails;
 
 use \tad_DI52_ServiceProvider;
 use TEC\Tickets\Emails\Admin\Emails_Tab;
+use Tribe__Tickets__Tickets as Tickets_Module;
 
 /**
  * Class Hooks.
@@ -69,6 +70,10 @@ class Hooks extends tad_DI52_ServiceProvider {
 		add_filter( 'tribe_tickets_admin_manager_request', [ $this, 'filter_add_preview_modal_content' ], 15, 2 );
 
 		add_filter( 'wp_redirect', [ $this, 'filter_redirect_url' ] );
+
+		// Legacy emails.
+		add_filter( 'tec_tickets_send_rsvp_email_pre', [ $this, 'filter_legacy_sent_rsvp_emails' ], 20, 4 );
+		add_filter( 'tec_tickets_send_tickets_email_for_attendee_pre', [ $this, 'filter_legacy_sent_tickets_attendees_emails' ], 20, 5 );
 	}
 
 	/**
@@ -215,6 +220,42 @@ class Hooks extends tad_DI52_ServiceProvider {
 		}
 
 		return $this->container->make( Admin\Preview_Modal::class )->get_modal_content_ajax( $render_response, $vars );
+	}
+
+	/**
+	 * Hooks to the legacy modules to send RSVP emails with the new system.
+	 *
+	 * @see Legacy::send_rsvp_email
+	 *
+	 * @since TBD
+	 *
+	 * @param null|boolean   $pre      Previous value from the filter, mostly will be null.
+	 * @param int            $order_id The order ID.
+	 * @param int            $event_id The event ID.
+	 * @param Tickets_Module $module   Commerce module we are using for these emails.
+	 */
+	public function filter_legacy_sent_rsvp_emails( $pre, $order_id, $event_id = null, $module = null ) {
+		return $this->container->make( Legacy::class )->send_rsvp_email( $pre, $order_id, $event_id, $module );
+	}
+
+	/**
+	 * Hooks to the legacy modules to send Tickets emails with the new system.
+	 *
+	 * @since TBD
+	 *
+	 * @see Legacy::send_tickets_email_for_attendee
+	 *
+	 * @param null|boolean   $pre         Previous value from the filter, mostly will be null.
+	 * @param string         $to          The email to send the tickets to.
+	 * @param array          $tickets     The list of tickets to send.
+	 * @param array          $args        See the rest of the documentation for this on Legacy::send_tickets_email_for_attendee
+	 *
+	 * @param Tickets_Module $module      Commerce module we are using for these emails.
+	 *
+	 * @return bool Whether email was sent to attendees.
+	 */
+	public function filter_legacy_sent_tickets_attendees_emails( $pre, $to, $tickets, $args = [], $module = null ) {
+		return $this->container->make( Legacy::class )->send_tickets_email_for_attendee( $pre, $to, $tickets, $args, $module );
 	}
 
 	/**
--- a/src/Tickets/Emails/Legacy.php
+++ b/src/Tickets/Emails/Legacy.php
@@ -1,0 +1,228 @@
+<?php
+
+namespace TEC\Tickets\Emails;
+
+use Tribe__Tickets__Tickets as Tickets_Module;
+
+/**
+ * Class Legacy
+ *
+ * @since   TBD
+ *
+ * @package TEC\Tickets\Emails
+ */
+class Legacy {
+
+	/**
+	 * Send RSVPs/tickets email for an attendee by injecting itself into the legacy Tickets codebase.
+	 *
+	 * @since TBD
+	 *
+	 * @param null|boolean   $pre         Previous value from the filter, mostly will be null.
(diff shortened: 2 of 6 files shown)


src/Tickets/Emails/Hooks.php
```diff
@@ -69,6 +70,10 @@ protected function add_filters() {
 		add_filter( 'tribe_tickets_admin_manager_request', [ $this, 'filter_add_preview_modal_content' ], 15, 2 );
 
 		add_filter( 'wp_redirect', [ $this, 'filter_redirect_url' ] );
+
+		// Legacy emails.
```
juanfra, line 74: This is not adding any clarity at all. It's not just legacy that we're hooking on to. It's the point where emails are being delivered.

The point of having legacy methods in the modules class was to quickly deprecate the old emails whenever we wanted by deprecating that new `_legacy` function that was just holding how the legacy emails were sent.

src/Tickets/Emails/Legacy.php
```diff
@@ -0,0 +1,228 @@
...
+		}
+
+		$email_class->set( 'post_id', $event_id );
+		$email_class->set( 'tickets', $all_attendees );
+		$subject     = $email_class->get_subject();
+		$content     = $email_class->get_content( [ 'tickets' => $all_attendees ] );
+		$headers     = $email_class->get_headers();
+		$attachments = $email_class->get_attachments();
+
+		$sent = tribe( Email_Sender::class )->send( $to, $subject, $content, $headers, $attachments );
```
juanfra, line 209: This should be replaced with:

```
$email_class->set( 'post_id', $event_id );
$email_class->set( 'tickets', $all_attendees );
$email_class->recipient = $to;

$sent = $email_class->send();
```

juanfra (commented):
I think this is generating more confusion than organization. I think we're mistakenly calling things legacy when they are not, and it'll probably make things more challenging to deprecate, when we are ready. Because email deliverability for the tickets/rsvp already lives in the modules.

juanfra:
We also need artifacts showing that these changes don't affect deliverability for the RSVP and Tickets emails.

juanfra:
We're also not checking if `tec_tickets_emails_is_enabled()` is true or not, which means that it'll be delivering both emails, the old and the new.

codingmusician:
A couple of questions:
- Is the purpose of the `Legacy` class to simply check to see if we should be using Tickets Emails or not, then passing the functionality on to TE, when applicable? It seems to be doing a lot more than that.
- Shouldn't one of the first checks be to check if Tickets Emails is enabled? Juan mentioned using `tec_tickets_emails_is_enabled()`. That makes the most sense to me.

src/Tickets/Emails/Hooks.php
```diff
@@ -217,6 +222,42 @@ public function filter_add_preview_modal_content( $render_response, $vars ) {
 		return $this->container->make( Admin\Preview_Modal::class )->get_modal_content_ajax( $render_response, $vars );
 	}
 
+	/**
+	 * Hooks to the legacy modules to send RSVP emails with the new system.
```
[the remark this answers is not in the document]
bordoni: The new Module is Tickets Commerce, that is why I am calling them Legacy, which we are using this directly.

src/Tickets/Emails/Legacy.php
bordoni: Added a `@todo` so I can handle that afterwards.

src/Tribe/Tickets.php
```diff
@@ -2448,117 +2449,21 @@ public function send_tickets_email_for_attendees( $attendees, $args = [] ) {
...
+			/**
+			 * @since TBD
+			 *
+			 * @param null|mixed $pre     Determine if we should continue.
+			 * @param string     $to      The email to send the tickets to.
+			 * @param array      $tickets The list of tickets to send.
+			 * @param array      $args    The list of arguments to use for sending ticket emails.
+			 * @param static     $module  Instance of the Tickets Module.
+			 */
+			$pre = apply_filters( 'tec_tickets_send_tickets_email_for_attendee_pre', null, $to, $tickets, $args, $this );
```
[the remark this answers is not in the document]
bordoni: `pre` is a WP Core standards for when the filter is present just for short-circuiting a particular method:
Examples: `pre_update_option` or `pre_http_request`

src/Tickets/Emails/Legacy.php
```diff
@@ -0,0 +1,228 @@
...
+ * Class Legacy
+ *
+ * @since   TBD
+ *
+ * @package TEC\Tickets\Emails
+ */
+class Legacy {
+
+	/**
+	 * Send RSVPs/tickets email for an attendee by injecting itself into the legacy Tickets codebase.
```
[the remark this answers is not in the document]
bordoni: Legacy is the `Tribe__Tickets__Tickets` class, in the codebase that is Legacy, I would like to start shirking that class and eventually have us deprecate it. I know it will take a long time but that is the architecture that I would like for us to aim at.

src/Tickets/Emails/Hooks.php
bordoni: Modified the comment for a bit more clarity.

src/Tickets/Emails/Legacy.php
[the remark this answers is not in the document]
codingmusician: Is the future plan to integrate a lot of what the `Legacy` class is doing into the `Email_Sender` class or somewhere else?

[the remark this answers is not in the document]
bordoni: Possibly, a lot of the logic there should not be part of the email piece. Determining if we use the RSVP or the Tickets, or Going not Going should actually happen on the Commerce level. Like which email are we sending should not be part of the logic of the Email System. It should be part of the business logic which tends to be part of the Commerce. 

The only problem is currently on the `Tribe__Tickets__Tickets` these domains are merged, which makes it impossible for you to maintain the code without feeling like you are handling a house of cards.

src/Tickets/Emails/Hooks.php
```diff
@@ -217,6 +222,42 @@ public function filter_add_preview_modal_content( $render_response, $vars ) {
...
+	 * @see Legacy::send_rsvp_email
+	 *
+	 * @since TBD
+	 *
+	 * @param null|boolean   $pre      Previous value from the filter, mostly will be null.
+	 * @param int            $order_id The order ID.
+	 * @param int            $event_id The event ID.
+	 * @param Tickets_Module $module   Commerce module we are using for these emails.
+	 */
+	public function filter_legacy_sent_rsvp_emails( $pre, $order_id, $event_id = null, $module = null ) {
```
borkweb, line 237: Perhaps change this to:

```
public function filter_hijack_legacy_sent_rsvp_emails( $pre, $order_id, $event_id = null, $module = null ) {
```

src/Tickets/Emails/Hooks.php
```diff
@@ -217,6 +222,42 @@ public function filter_add_preview_modal_content( $render_response, $vars ) {
...
+	 * @param null|boolean   $pre         Previous value from the filter, mostly will be null.
+	 * @param string         $to          The email to send the tickets to.
+	 * @param array          $tickets     The list of tickets to send.
+	 * @param array          $args        See the rest of the documentation for this on Legacy::send_tickets_email_for_attendee
+	 *
+	 * @param Tickets_Module $module      Commerce module we are using for these emails.
+	 *
+	 * @return bool Whether email was sent to attendees.
+	 */
+	public function filter_legacy_sent_tickets_attendees_emails( $pre, $to, $tickets, $args = [], $module = null ) {
```
borkweb, line 257:
```
public function filter_hijack_legacy_sent_tickets_attendees_emails( $pre, $to, $tickets, $args = [], $module = null ) {
```

src/Tickets/Emails/Hooks.php
```diff
@@ -69,10 +70,14 @@ protected function add_filters() {
 		add_filter( 'tribe_tickets_admin_manager_request', [ $this, 'filter_add_preview_modal_content' ], 15, 2 );
 
 		add_filter( 'wp_redirect', [ $this, 'filter_redirect_url' ] );
+
+		// Inject the new email into the legacy codebase for emails.
+		add_filter( 'tec_tickets_send_rsvp_email_pre', [ $this, 'filter_legacy_sent_rsvp_emails' ], 20, 4 );
+		add_filter( 'tec_tickets_send_tickets_email_for_attendee_pre', [ $this, 'filter_legacy_sent_tickets_attendees_emails' ], 20, 5 );
```
borkweb, line 76:
```
		add_filter( 'tec_tickets_send_rsvp_email_pre', [ $this, 'filter_hijack_legacy_sent_rsvp_emails' ], 20, 4 );
		add_filter( 'tec_tickets_send_tickets_email_for_attendee_pre', [ $this, 'filter_hijack_legacy_sent_tickets_attendees_emails' ], 20, 5 );
```

src/Tickets/Emails/Legacy.php
```diff
@@ -0,0 +1,241 @@
...
+use Tribe__Tickets__Tickets as Tickets_Module;
+
+/**
+ * Class Legacy
+ *
+ * @since   TBD
+ *
+ * @package TEC\Tickets\Emails
+ */
+class Legacy {
```
borkweb, line 14: Calling this `Legacy` does feel weird. Perhaps change it to what it is actually doing: `Legacy_Hijack` or `Legacy_Hijacker`. That'd make it clear that this code is meant to override legacy logic and isn't, itself, legacy logic.

borkweb (commented):
I know this is merged already, but I felt that these comments / changes are warranted.

src/Tickets/Emails/Hooks.php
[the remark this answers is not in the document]
juanfra: > The new Module is Tickets Commerce, that is why I am calling them Legacy, which we are using this directly.

Tickets Commerce may be the new module, but the email sending is still centralized in these functions from the module. This naming doesn't ring any bell and makes deprecating things more challenging. At the same time, you're taking this approach for a couple of emails and not for the rest. This needs to be more consistent, and there needs to be clarity from a dev perspective on what's supposed to live on these files.

src/Tribe/RSVP.php
```diff
@@ -1194,106 +1194,26 @@ public function generate_tickets( $post_id = null, $redirect = true ) {
...
-		 * @param array $to_send        list of tickets to be sent out by email
-		 * @param array $all_attendees  list of all attendees/tickets, including those already sent out
-		 * @param int   $order_id
+		 * @param null|mixed $pre      Determine if we should continue.
+		 * @param int        $order_id The order ID.
+		 * @param int        $event_id The event ID.
+		 * @param static     $module   Instance of the Tickets Module.
 		 */
-		$to_send = (array) apply_filters( 'tribe_tickets_rsvp_tickets_to_send', $to_send, $all_attendees, $order_id );
+		$pre = apply_filters( 'tec_tickets_send_rsvp_email_pre', null, $order_id, $event_id, $this );
```
juanfra, line 1211: So once you deprecate the old email sending, you'll have the email sending function with only a filter named `pre_something` where's no `something`?

I think will create a lot of confusion when deprecating old emails (as well as now), and it is something that we want to do soon. That's why we had a different approach. It would be better if we talked about these things, so you can have more context instead of creating PRs and merging them immediately.

src/Tribe/Tickets.php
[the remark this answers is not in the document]
juanfra: The nomenclature of WP usually has the `pre_` as a prefix of an action/procedure, not as a suffix. 

Still, this will become super confusing once we deprecate the old email delivery. Which is something that we plan to do soon. This method will have just a `pre_` filter with nothing else, so even the naming won't make sense.

juanfra (commented):
I would have loved to continue the conversation before these breaking changes were merged.

PR #2620, opened 2023-04-28
Tickets Emails > Remove conditional, introducing the feature.

juanfra:
### 🎫 Ticket

[ET-1617] 

### 🗒️ Description

- Removing the conditional to load the feature as we'll be introducing it in escarole.

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1617]: https://theeventscalendar.atlassian.net/browse/ET-1617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Emails/Provider.php
+++ b/src/Tickets/Emails/Provider.php
@@ -25,12 +25,6 @@ class Provider extends tad_DI52_ServiceProvider {
 	 */
 	public function register() {
 
-		// If not enabled, do not load Tickets Emails system.
-		// @todo @codingmusician @rafsuntaskin @juanfra: Remove this for 5.7.0 (When we release Tickets Emails)
-		if ( ! tec_tickets_emails_is_enabled() ) {
-			return;
-		}
-
 		$this->register_assets();
 		$this->register_hooks();
 


bordoni:
Should we remove the method too?

redscar (approved):
This looks good. Do you want to implement a filter that flips the logic to disable the feature if needed? I did that with Celery for the CT logic.

bordoni:
Example of what we did with Views V2:
https://github.com/the-events-calendar/the-events-calendar/blob/master/src/functions/views/provider.php#L31-L46

codingmusician (approved):

juanfra:
We still need the method, because there will be a toggle for old installs, so that users can "upgrade" to this new functionality.

This is the method:

https://github.com/the-events-calendar/event-tickets/blob/master/src/functions/emails/provider.php#L12-L36

It still takes into account for the constant, and defaulting to ON for installs after `5.6.0`. 

**Please keep in mind that by removing this check, we're showing the `Emails` tab, but the method for `tec_tickets_emails_is_enabled` can still return false.**

And [here you can find the logic for the toggle](https://github.com/the-events-calendar/event-tickets/blob/master/src/Tickets/Emails/Admin/Settings.php#L339). So the toggle will be only available for users that installed the plugin before `5.6.0`

bordoni (approved):

bordoni:
Perfecto, that is good to know. :shipit:

PR #2621, opened 2023-04-28
Move towards the Dispatcher nomenclature.

bordoni:
📹 https://i.bordoni.me/RZX5ZMww

Depends on #2619

### DIFF
--- a/src/Tickets/Emails/Dispatcher.php
+++ b/src/Tickets/Emails/Dispatcher.php
@@ -213,7 +213,6 @@ class Dispatcher {
 		 * @since TBD
 		 *
 		 * @param array          $headers The headers.
-		 * @param string         $id      The email ID.
 		 * @param Email_Abstract $this    The email object.
 		 */
 		$headers = apply_filters( 'tec_tickets_emails_dispatcher_headers', $this->headers, $this );
@@ -226,7 +225,6 @@ class Dispatcher {
 		 * @since TBD
 		 *
 		 * @param array          $headers The headers.
-		 * @param string         $id      The email ID.
 		 * @param Email_Abstract $this    The email object.
 		 */
 		$headers = apply_filters( "tec_tickets_emails_dispatcher_{$email_slug}_headers", $headers, $this );
@@ -344,8 +342,6 @@ class Dispatcher {
 		 * @since TBD
 		 *
 		 * @param string         $subject  The email subject.
-		 * @param string         $id       The email id.
-		 * @param string         $template Template name.
 		 * @param Email_Abstract $this     The email object.
 		 */
 		$subject = apply_filters( 'tec_tickets_emails_dispatcher_subject', $this->subject, $this );
@@ -358,8 +354,6 @@ class Dispatcher {
 		 * @since TBD
 		 *
 		 * @param string         $subject  The email subject.
-		 * @param string         $id       The email id.
-		 * @param string         $template Template name.
 		 * @param Email_Abstract $this     The email object.
 		 */
 		$subject = apply_filters( "tec_tickets_emails_dispatcher_{$email_slug}_subject", $subject, $this );
--- a/src/Tickets/Emails/Dispatcher.php
+++ b/src/Tickets/Emails/Dispatcher.php
@@ -243,7 +243,7 @@ class Dispatcher {
 		 */
 		$to = apply_filters( "tec_tickets_emails_dispatcher_{$email_slug}_to", $to, $this );
 
-		return (string) $to;
+		return null === $to ? $to : (string) $to;
 	}
 
 	/**
@@ -291,7 +291,7 @@ class Dispatcher {
 		 */
 		$subject = apply_filters( "tec_tickets_emails_dispatcher_{$email_slug}_subject", $subject, $this );
 
-		return (string) $subject;
+		return null === $subject ? $subject : (string) $subject;
 	}
 
 	/**
@@ -335,7 +335,7 @@ class Dispatcher {
 		 */
 		$content = apply_filters( "tec_tickets_emails_dispatcher_{$email_slug}_content", $content, $this );
 
-		return (string) $content;
+		return null === $content ? $content :  (string) $content;
 	}
 
 	/**
--- a/src/Tickets/Emails/Dispatcher.php
+++ b/src/Tickets/Emails/Dispatcher.php
@@ -82,13 +82,14 @@ class Dispatcher {
 	protected ?string $subject = null;
 
 	/**
-	 * Sets the Email instance that will be used for this Dispatcher.
+	 * Sets the Email instance that will be used for this Dispatcher, intentionally this is protected method.
+	 * You can only generate a new Dispatcher from the Factory Method.
 	 *
 	 * @since TBD
 	 *
 	 * @param Email_Abstract $email
 	 */
-	public function set_email( Email_Abstract $email ): void {
+	protected function set_email( Email_Abstract $email ): void {
 		$this->email = $email;
 	}
 
@@ -101,6 +102,90 @@ class Dispatcher {
 	 */
 	public function get_email(): ?Email_Abstract {
 		return $this->email;
+	}
+
+	/**
+	 * Prepares the dispatcher for this Email.
+	 *
+	 * @since TBD
+	 *
+	 * @param Dispatcher $dispatcher
+	 *
+	 * @return Dispatcher
+	 */
+	protected function prepare_dispatcher( Dispatcher $dispatcher ): Dispatcher {
+		// Enforce text/html content type header.
+		$dispatcher->add_header( 'Content-Type', 'text/html; charset=utf-8' );
+
+		$email = $dispatcher->get_email();
+
+		if ( ! $email ) {
+			return $dispatcher;
+		}
+
+		$from_email = $email->get_from_email();
+		$from_name  = $email->get_from_name();
+
+		// Add From name/email to headers if no headers set yet, and we have a valid From email address.
+		if ( ! empty( $from_name ) && ! empty( $from_email ) && is_email( $from_email ) ) {
+			$from_email = sanitize_email( $from_email );
+
+			$dispatcher->add_header( 'From', sprintf(
+				'%1$s <%2$s>',
+				stripcslashes( $from_name ),
+				$from_email
+			) );
+
+			$dispatcher->add_header( 'Reply-To', $from_email );
+		}
+
+		$dispatcher->set_to( $email->get_recipient() );
+		$dispatcher->set_subject( $email->get_subject() );
+		$dispatcher->set_content( $email->get_content() );
+
+		return $dispatcher;
+	}
+
+	/**
+	 * From a given email type instance generate a new Dispatcher and prepare it for usage.
+	 *
+	 * @since TBD
+	 *
+	 * @param Email_Abstract $email
+	 *
+	 * @return Dispatcher
+	 */
+	public static function from_email( Email_Abstract $email ): Dispatcher {
+		// Generate a new dispatcher every time.
+		$dispatcher = tribe( static::class );
+
+		// Associate this email to the dispatcher.
+		$dispatcher->set_email( $email );
+
+		// Prepare the dispatcher to send an email.
+		$dispatcher = $dispatcher->prepare_dispatcher( $dispatcher );
+
+		/**
+		 * Allows modifications of the Email Dispatcher to all Email Types.
+		 *
+		 * @since TBD
+		 *
+		 * @param Dispatcher     $dispatcher Which dispatcher instance will be used for the email sent.
+		 * @param Email_Abstract $email      Which instance of the email that will be attached to this dispatcher.
+		 */
+		$dispatcher = apply_filters( 'tec_tickets_emails_dispatcher', $dispatcher, $email );
+
+		$email_slug = $email->slug;
+
+		/**
+		 * Allows modifications of the Email Dispatcher specific to this Email Type.
+		 *
+		 * @since TBD
+		 *
+		 * @param Dispatcher     $dispatcher Which dispatcher instance will be used for the email sent.
+		 * @param Email_Abstract $email      Which instance of the email that will be attached to this dispatcher.
+		 */
+		return apply_filters( "tec_tickets_emails_{$email_slug}_dispatcher", $dispatcher, $email );
 	}
 
 	/**
@@ -318,8 +403,8 @@ class Dispatcher {
 		 *
 		 * @since TBD
 		 *
-		 * @param string         $content  The email subject.
-		 * @param Email_Abstract $this     The email object.
+		 * @param string         $content The email subject.
+		 * @param Email_Abstract $this    The email object.
 		 */
 		$content = apply_filters( 'tec_tickets_emails_dispatcher_content', $this->content, $this );
 
@@ -330,12 +415,12 @@ class Dispatcher {
 		 *
 		 * @since TBD
 		 *
-		 * @param string         $content  The email subject.
-		 * @param Email_Abstract $this     The email object.
+		 * @param string         $content The email subject.
+		 * @param Email_Abstract $this    The email object.
 		 */
 		$content = apply_filters( "tec_tickets_emails_dispatcher_{$email_slug}_content", $content, $this );
 
-		return null === $content ? $content :  (string) $content;
+		return null === $content ? $content : (string) $content;
 	}
 
 	/**
@@ -404,7 +489,7 @@ class Dispatcher {
 	 * the wp_mail was actually sent at this moment.
 	 *
 	 * If you need to make another email using a similar dispatcher please make use of the Email associated with
-	 * this dispatcher by `$dispatcher->get_email()->get_dispatcher()->send()` which will generate a new dispatcher
+	 * this dispatcher by `Dispatcher::from_email( $dispatcher->get_email() )->send()` which will generate a new dispatcher
 	 * and attempt to send the email.
 	 *
 	 * @since TBD
--- a/src/Tickets/Emails/Dispatcher.php
+++ b/src/Tickets/Emails/Dispatcher.php
@@ -1,8 +1,8 @@
 <?php
 /**
- * Email Sender class.
+ * Email Dispatcher class.
  *
- * @since 5.5.9
+ * @since TBD
  *
  * @package TEC\Tickets\Emails
  */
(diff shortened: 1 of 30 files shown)


src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,76 @@
...
+	public function get_email(): ?Email_Abstract {
+		return $this->email;
+	}
+
+	/**
+	 * Send an email.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool success True if the email was sent.
```
Camwyn, line 58:
```suggestion
	 * @return bool Whether the email was sent successfully.
```
Copying the return of `wp_mail` as that covers both true and false.

src/Tickets/Emails/Email_Abstract.php
```diff
@@ -107,6 +107,15 @@ abstract class Email_Abstract {
 	 */
 	protected $data = [];
 
+	/**
+	 * Instance of the Dispatcher class associated with this
```
Camwyn, line 111:
```suggestion
	 * Instance of the Dispatcher class associated with this Email Instance.
```

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,76 @@
...
+		if ( ! $email ) {
+			return false;
+		}
+
+		$to          = $email->get_recipient();
+		$subject     = $email->get_subject();
+		$message     = $email->get_content();
+		$headers     = $email->get_headers();
+		$attachments = $email->get_attachments();
+
```
Camwyn, line 73: `if ( empty( $to ) ) { return false; }`
sanity check(s)? Could also throw exception for bad data.

Camwyn: Otherwise, why even set them as variables?

src/Tickets/Emails/Email_Abstract.php
```diff
@@ -122,6 +131,34 @@ public function hook() {
...
+
+	/**
+	 * Gets the defined instance of the dispatcher for this email.
+	 *
+	 * @since TBD
+	 *
+	 * @return Dispatcher
+	 */
+	public function get_dispatcher(): Dispatcher {
+		return $this->dispatcher;
```
Camwyn, line 159: Do we want a check here to be sure we're returning something useful? Call `set_dispatcher()` if not?

Camwyn: If we do that, the emails don't have to `set` the Dispatcher just `get` it, unless they are using a specific one.

Camwyn (commented):
Questions/suggestions

src/Tickets/Emails/Dispatcher.php
bordoni: Thank you! I will address those when I refactor that piece.

src/Tickets/Emails/Email_Abstract.php
bordoni: Not in this case, it's part of the building of the method, I would rather always  check for if dispatcher is not null. Need to change the signature of the method to be `?Dispatcher`

codingmusician:
Looing throughout the code, I see where you `->get()` the `Dispatcher`, however I do not ever see it being `->set()`. It seems like all the email classes are doing is pulling a `null` Dispatcher. Am I missing something?

src/Tickets/Emails/Email/Completed_Order.php
```diff
@@ -9,6 +9,7 @@
 
 use \TEC\Tickets\Emails\Email_Template;
 use TEC\Tickets\Emails\Admin\Preview_Data;
+use \TEC\Tickets\Emails\Email_Abstract;
```
borkweb, line 12:
```suggestion
use TEC\Tickets\Emails\Email_Abstract;
```

src/Tickets/Emails/Email/Ticket.php
```diff
@@ -9,6 +9,7 @@
 
 use TEC\Tickets\Commerce\Settings;
 use \TEC\Tickets\Emails\Email_Template;
+use \TEC\Tickets\Emails\Email_Abstract;
```
borkweb, line 12:
```suggestion
use TEC\Tickets\Emails\Email_Template;
use TEC\Tickets\Emails\Email_Abstract;
```

borkweb:
I'm liking the direction that this is headed with regards to the Dispatcher class. A bit more DRY and testable!

borkweb (approved):

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,525 @@
...
+	/**
+	 * Prepares the dispatcher for this Email.
+	 *
+	 * @since TBD
+	 *
+	 * @param Dispatcher $dispatcher
+	 *
+	 * @return Dispatcher
+	 */
+	protected function prepare_dispatcher( Dispatcher $dispatcher ): Dispatcher {
```
borkweb, line 116: This needs to be `public` so that it is actually callable on line 166.

```suggestion
	public function prepare_dispatcher( Dispatcher $dispatcher ): Dispatcher {
```

borkweb (changes requested):
After the small change and once tests pass, I'd say this is good.

src/Tickets/Emails/Dispatcher.php
bordoni: Not needed, marking as resolved.

codingmusician (dismissed):
Good to go. I do notice that we are using `to` and `recipient` in different classes when they mean the same thing. Not sure if we want to make that more unified or not.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,513 @@
...
+		 * Allow filtering the email subject globally.
+		 *
+		 * @since TBD
+		 *
+		 * @param string         $subject  The email subject.
+		 * @param string         $id       The email id.
+		 * @param string         $template Template name.
+		 * @param Email_Abstract $this     The email object.
+		 */
+		$subject = apply_filters( 'tec_tickets_emails_dispatcher_subject', $this->subject, $this );
```
redscar, line 351: Is the docblock above correct? I don't see `$id` or `$template`.

bordoni: Thank you!

src/Tickets/Emails/Email_Abstract.php
```diff
@@ -268,105 +270,6 @@ public function get_from_name(): string {
...
-		/**
-		 * Filter the attachments for the particular email.
-		 *
-		 * @since 5.5.10
-		 *
-		 * @param array          $attachments The attachments.
-		 * @param string         $id          The email ID.
-		 * @param Email_Abstract $this        The email object.
-		 */
-		$attachments = apply_filters( "tec_tickets_emails_{$this->slug}_attachments", $attachments, $this->id, $this );
```
juanfra, line 365: Removing this will basically break the TEC integration with the inclusion of the `.ics` file.

src/Tickets/Emails/Email_Abstract.php
```diff
@@ -268,105 +270,6 @@ public function get_from_name(): string {
...
-		/**
-		 * Filter the attachments.
-		 *
-		 * @since 5.5.9
-		 *
-		 * @param array          $attachments The attachments.
-		 * @param string         $id          The email ID.
-		 * @param Email_Abstract $this        The email object.
-		 */
-		$attachments = apply_filters( 'tec_tickets_emails_attachments', $attachments, $this->id, $this );
```
juanfra, line 354: We're moving away from a naming that makes sense, introducing the `dispatcher` name in the filters, adding more noise and confusion.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+	}
+
+	/**
+	 * Get dispatcher headers.
+	 *
+	 * @since TBD
+	 *
+	 * @return array
+	 */
+	public function get_headers(): array {
```
juanfra, line 209: Why don't you keep the headers method in the email class (where it belongs, because it's tied to emails) and grab the headers from there by using the `$this->get_email()->get_headers()`?

Naming this with `dispatcher` is adding more noise, and it makes it more difficult to have access to the actual email if you're trying (as a dev that's extending the functionality) to get access to the actual email.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+	public function get_attachments(): array {
+		/**
+		 * Filter the attachments.
+		 *
+		 * @since TBD
+		 *
+		 * @param array          $attachments The attachments.
+		 * @param Email_Abstract $this        The email object.
+		 */
+		$attachments = apply_filters( 'tec_tickets_emails_dispatcher_attachments', $this->attachments, $this );
```
juanfra, line 270: Why don't you keep the attachment method in the email class (where it belongs, because it's tied to emails) and grab the headers from there by using the `$this->get_email()->get_attachments()`?

Naming this with `dispatcher` adds more noise, making it more difficult to access the actual email if you're trying (as a dev extending the functionality) to access the actual email.

⚠️ **This particular change breaks the TEC implementation**. We should be more careful with these things. There's a lack of context, and we're introducing breaking changes in the week that we should mark this feature as ready.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+	}
+
+	/**
+	 * Sets the content for the dispatcher.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $value
+	 */
+	public function set_content( string $value ): void {
```
juanfra, line 415: This type of thing belongs to the dispatcher class, not the getters.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+	/**
+	 * From a given email type instance generate a new Dispatcher and prepare it for usage.
+	 *
+	 * @since TBD
+	 *
+	 * @param Email_Abstract $email
+	 *
+	 * @return Dispatcher
+	 */
+	public static function from_email( Email_Abstract $email ): Dispatcher {
```
juanfra, line 149: This makes it so confusing. The `from` word is pretty much reserved for other types of usages in the "emails" world.

I think for the sake of clarity this method should be renamed.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+
+		/**
+		 * Allow filtering the email subject.
+		 *
+		 * @since TBD
+		 *
+		 * @param string         $subject  The email subject.
+		 * @param Email_Abstract $this     The email object.
+		 */
+		$subject = apply_filters( "tec_tickets_emails_dispatcher_{$email_slug}_subject", $subject, $this );
```
juanfra, line 359: So, as far as I understand, we're having this coming from the email abstract class, where we already have two filters for the subject (particular and global). What's the point of having this here as well?

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+	public function get_content(): ?string {
+		/**
+		 * Allow filtering the email content globally.
+		 *
+		 * @since TBD
+		 *
+		 * @param string         $content The email subject.
+		 * @param Email_Abstract $this    The email object.
+		 */
+		$content = apply_filters( 'tec_tickets_emails_dispatcher_content', $this->content, $this );
```
juanfra, line 391: So, as far as I understand, we're having this coming from the email abstract class, where we already have two filters for the content (particular and global). What's the point of having this here as well?

I think these things makes it more confusing, adding the `dispatcher` as part of the naming. It should be clear that the email content is populated via the email class, and that this is just a wrapper to send (or dispatch, how you now named it).

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+	 * @return ?string
+	 */
+	public function get_content(): ?string {
+		/**
+		 * Allow filtering the email content globally.
+		 *
+		 * @since TBD
+		 *
+		 * @param string         $content The email subject.
+		 * @param Email_Abstract $this    The email object.
```
juanfra, line 389: docblocks are wrong.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+
+		$email_slug = $this->get_email()->slug;
+
+		/**
+		 * Allow filtering the email content.
+		 *
+		 * @since TBD
+		 *
+		 * @param string         $content The email subject.
+		 * @param Email_Abstract $this    The email object.
```
juanfra, line 401: Docblocks are wrong.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+	 * @return ?string
+	 */
+	public function get_subject(): ?string {
+		/**
+		 * Allow filtering the email subject globally.
+		 *
+		 * @since TBD
+		 *
+		 * @param string         $subject  The email subject.
+		 * @param Email_Abstract $this     The email object.
```
juanfra, line 345: Docblock is wrong, not Email_Abstract but dispatcher.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+
+		$email_slug = $this->get_email()->slug;
+
+		/**
+		 * Allow filtering the email subject.
+		 *
+		 * @since TBD
+		 *
+		 * @param string         $subject  The email subject.
+		 * @param Email_Abstract $this     The email object.
```
juanfra, line 357: Docblock is wrong.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+
+		$email_slug = $this->get_email()->slug;
+
+		/**
+		 * Allow filtering the email recipient for the particular email.
+		 *
+		 * @since TBD
+		 *
+		 * @param string         $to   The email recipient.
+		 * @param Email_Abstract $this The email object.
```
juanfra, line 313: `$this` is not the email abstract class.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+	 * @return ?string The email recipient.
+	 */
+	public function get_to(): ?string {
+		/**
+		 * Allow filtering the email recipient globally.
+		 *
+		 * @since TBD
+		 *
+		 * @param string         $to   The email recipient.
+		 * @param Email_Abstract $this The email object.
```
juanfra, line 301: `$this` is not the email abstract class.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+
+		$email_slug = $this->get_email()->slug;
+
+		/**
+		 * Filter the attachments for the particular email.
+		 *
+		 * @since TBD
+		 *
+		 * @param array          $attachments The attachments.
+		 * @param Email_Abstract $this        The email object.
```
juanfra, line 280: `$this` is not the email abstract class.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+	 * @return array
+	 */
+	public function get_attachments(): array {
+		/**
+		 * Filter the attachments.
+		 *
+		 * @since TBD
+		 *
+		 * @param array          $attachments The attachments.
+		 * @param Email_Abstract $this        The email object.
```
juanfra, line 268: `$this` is not the email abstract class.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+
+		$email_slug = $this->get_email()->slug;
+
+		/**
+		 * Filter the headers for the particular email using this dispatcher.
+		 *
+		 * @since TBD
+		 *
+		 * @param array          $headers The headers.
+		 * @param Email_Abstract $this    The email object.
```
juanfra, line 228: `$this` is not the email abstract class.

src/Tickets/Emails/Dispatcher.php
```diff
@@ -0,0 +1,507 @@
...
+	 * @return array
+	 */
+	public function get_headers(): array {
+		/**
+		 * Filter the headers for all dispatchers.
+		 *
+		 * @since TBD
+		 *
+		 * @param array          $headers The headers.
+		 * @param Email_Abstract $this    The email object.
```
juanfra, line 216: `$this` is not the email abstract class.

src/Tickets/Emails/Email_Abstract.php
juanfra: The attachments and headers belong to the Email and not the delivery. 

**It is wrong to move these out of the email class**.

juanfra (commented):
We're making structural changes in the last week, breaking functionality and that are rushed. This shouldn't be happening.

bordoni:
Resolving all your comments suggestion on #2627

src/Tickets/Emails/Email_Abstract.php
bordoni: Resolved in: https://github.com/the-events-calendar/the-events-calendar/pull/4236

src/Tickets/Emails/Dispatcher.php
bordoni: This is marking that we are generating a dispatcher from a Email Type instance. It's a factory method.

src/Tickets/Emails/Dispatcher.php
bordoni: Correct, I want to avoid having the Email class being the holder of the data, the Email Type classes are more like templates of what the email will be, not the actual email. But I cannot address that particular code change now, it's too late in the cycle.

src/Tickets/Emails/Dispatcher.php
bordoni: This is part of the dispatcher class, this is used to avoid having `content` be a public prop.

src/Tickets/Emails/Dispatcher.php
bordoni: We are separating concerns, Email Type class is where currently we hold data, and eventually refactored to just hold the pattern of how to generate an email, not the actual data of the email or the sending of the email.

src/Tickets/Emails/Email_Abstract.php
bordoni: I do not agree with you, separating concerns is what makes our code easier to maintain long term.

PR #2632, opened 2023-05-09
Update emails descriptions, use new link for the tickets emails docs.

juanfra:
### 🗒️ Description

Updating emails descriptions and use the link that'll contain documentation for Tickets Emails.


juanfra:
@codingmusician Initially we're going to have a central KB article that will explain how to customize, and it'll link to each article separately.

codingmusician (approved):

PR #2666, opened 2023-05-26
[TE-31DB] Fix - Apply contrast text color for live preview.

rafsuntaskin:
### 🎫 Ticket

[TE-31DB] 

### 🗒️ Description

* Live preview was not showing the adjusted contrast Text Color, but in the real email it shows up. Now the live preview matches with real email.

### ✔️ Checklist
- [ ] My code is tested. 

### DIFF
--- a/src/Tickets/Emails/Admin/Preview_Modal.php
+++ b/src/Tickets/Emails/Admin/Preview_Modal.php
@@ -189,7 +189,8 @@ class Preview_Modal {
 		$ticket_bg_color = Arr::get( $vars, 'ticketBgColor', '' );
 
 		if ( ! empty( $ticket_bg_color ) ) {
-			$preview_context['ticket_bg_color'] = wp_kses( $ticket_bg_color, [] );
+			$preview_context['ticket_bg_color']   = wp_kses( $ticket_bg_color, [] );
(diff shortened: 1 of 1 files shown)


src/Tickets/Emails/Admin/Preview_Modal.php
```diff
@@ -189,7 +189,8 @@ public function get_modal_content_ajax( $render_response, $vars ) {
 		$ticket_bg_color = Arr::get( $vars, 'ticketBgColor', '' );
 
 		if ( ! empty( $ticket_bg_color ) ) {
-			$preview_context['ticket_bg_color'] = wp_kses( $ticket_bg_color, [] );
+			$preview_context['ticket_bg_color']   = wp_kses( $ticket_bg_color, [] );
+			$preview_context['ticket_text_color'] = \Tribe__Utils__Color::get_contrast_color( $preview_context['ticket_bg_color'] );
```
codingmusician, line 193: Since we're wrapping the bg color in `wp_kses()`, we should probably do the same with the text color.

codingmusician: This one and the other one below

rafsuntaskin: @codingmusician we are re-using the var from the line above where it was already sanitized.

PR #2667, opened 2023-05-26
Tickets Emails: Move Event Titles to TEC

codingmusician:
### Issues

- TE-8ADA
- TE-DC94

### Description

The event titles were initially added into ET, but were set to be moved to TEC eventually. That never happened, so this PR handles removing the event titles from ET.

### Abstract

https://www.loom.com/share/8dc76ea87ad2419490a714ba6a2514d7


codingmusician:
We are adding the generic post titles to all emails in another PR, therefore I'm closing this one.

PR #2674, opened 2023-06-01
[ET-1767] Attendee Registration Loop issue on certain themes.

redscar:
### 🎫 Ticket

[ET-1767] 

### 🗒️ Description

When using Astra theme, the attendee registration form on the attendee registration page is being generated infinite times(on loop), and the page keeps loading. This issue seems to mostly occur with Astra version 4.x.  

Mike found the original snippet that I have used https://gist.github.com/codingmusician/da909aa95bc08e80baaab3456f586014. 

### 🎥 Artifacts 

Original Issue - 
image

Fixed -

image

### ✔️ Checklist
- [ ] My code is tested. 

[ET-1767]: https://theeventscalendar.atlassian.net/browse/ET-1767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -196,6 +196,7 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 * Fix - Updated the page heading when on the Orders Report page. [ET-1737]
 * Fix - When no providers are enabled, a warning will display above the `New Ticket` and `New RSVP` area explaining that at least one must be enabled. [ET-1696]
 * Fix - Corrected an issue with the `New ticket` button having invalid HTML. [ET-1631]
+* Fix - Resolved an issue that caused compatibility problems between specific themes and the Attendee Registration page. [ET-1767]
 
 = [5.5.11.1] 2023-05-09 =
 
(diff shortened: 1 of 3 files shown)


rafsuntaskin:
@redscar Attendee registration is part of ETP. Can we add this filter into ETP?

juanfra (changes requested):
I think we can avoid re-filtering this and setting the `is_singular` to true [here](https://github.com/the-events-calendar/event-tickets/blob/afaf3197d94b8022b150c1d43b4672b2e7db33fa/src/Tribe/Attendee_Registration/Template.php#L84).

All the conditionals that you're applying on `tec_tickets_mock_ar_as_singular ` made sense on a snippet (where maybe we used them as safe checks), but many of those are not really useful there (checking if ET is active in ET for example).

redscar:
> I think we can avoid re-filtering this and setting the `is_singular` to true [here](https://github.com/the-events-calendar/event-tickets/blob/afaf3197d94b8022b150c1d43b4672b2e7db33fa/src/Tribe/Attendee_Registration/Template.php#L84).
> 
> All the conditionals that you're applying on `tec_tickets_mock_ar_as_singular ` made sense on a snippet (where maybe we used them as safe checks), but many of those are not really useful there (checking if ET is active in ET for example).

@juanfra Good call. By making that change it seems the issue is still fixed and working properly.

codingmusician (approved):

juanfra (approved):

PR #2686, opened 2023-06-06
Inlcude Telemetry from the Tickets perspective

bordoni:


src/Tickets/Telemetry/Telemetry.php
```diff
@@ -0,0 +1,269 @@
...
+		if (
+			TEC::POSTTYPE === $screen->id
+			|| TEC::ORGANIZER_POST_TYPE === $screen->id
+			|| TEC::VENUE_POST_TYPE === $screen->id
+		) {
+			return false;
+		}
+
+		return true;
+	}
```
bordoni, line 221: @Camwyn you should make sure this is changed to only show on ET pages.

PR #2714, opened 2023-07-07
Ensure the Attendees page displays correctly when accessed through th…

EugeneKyale:
Ensure the Attendees page displays correctly when accessed through the Events Manager.

### 🎫 Ticket

[TICKET_ID] 

### ✔️ Checklist
- [ ] My code is tested. 

### DIFF
--- a/src/Tribe/Attendees.php
+++ b/src/Tribe/Attendees.php
@@ -151,16 +151,25 @@ class Tribe__Tickets__Attendees {
 	 * Returns the full URL to the attendees report page.
 	 *
 	 * @since 4.6.2
+	 * @since TBD - Added $post_id parameter to allow for conversion of provisional ID's into regular Post ID's.
 	 *
 	 * @param WP_Post $post
+	 * @param int $post_id The Post ID of the event.
 	 *
 	 * @return string
 	 */
 	public function get_report_link( $post ) {
+		// Convert provisional ID's into regular Post ID's.
+		if ( class_exists( '\TEC\Events\Custom_Tables\V1\Models\Occurrence', false ) ) {
+			$post_id = \TEC\Events\Custom_Tables\V1\Models\Occurrence::normalize_id( $post->ID );
+		} else {
+			$post_id = $post->ID;
+		}
+
 		$args = [
 			'post_type' => $post->post_type,
 			'page'      => $this->slug(),
-			'event_id'  => $post->ID,
+			'event_id'  => $post_id,
 		];
 
 		$url = add_query_arg( $args, admin_url( 'edit.php' ) );
--- a/readme.txt
+++ b/readme.txt
@@ -199,6 +199,7 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 * Fix - Ticketed Commerce events will now be accurately categorized and counted under the ticketed tab in the Dashboard Event List. [ET-1774]
 * Fix - The attendee export functionality for old converted recurring events has been improved to accurately export attendees. [ET-1739]
 * Fix - The Attendee List will now be correctly displayed when the 'Show attendees list on event page' option is enabled within the classic editor.  [ETP-623]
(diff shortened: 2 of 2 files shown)


src/Tribe/Attendees.php
```diff
@@ -151,16 +151,25 @@ public function print_checkedin_totals() {
...
+	 * @since TBD - Added $post_id parameter to allow for conversion of provisional ID's into regular Post ID's.
 	 *
 	 * @param WP_Post $post
+	 * @param int $post_id The Post ID of the event.
 	 *
 	 * @return string
 	 */
 	public function get_report_link( $post ) {
+		// Convert provisional ID's into regular Post ID's.
+		if ( class_exists( '\TEC\Events\Custom_Tables\V1\Models\Occurrence', false ) ) {
```
juanfra, line 163: @EugeneKyale do you think this case could occur in other places? I see at least [one other place](https://github.com/the-events-calendar/event-tickets/blob/8661b22a06dbaff45d61b61e4e2e710dd17bc787/src/Tribe/Tickets.php#L424).

Would it make sense to make a wrapper function for the id normalization we could call in different places?

EugeneKyale: @juanfra yeah it makes sense to have a helper function so can ensure consistency in handling IDs throughout our code. Which class do you think is best suited to host the helper function?

src/Tribe/Attendees.php
```diff
@@ -151,16 +151,25 @@ public function print_checkedin_totals() {
...
+			$post_id = \TEC\Events\Custom_Tables\V1\Models\Occurrence::normalize_id( $post->ID );
+		} else {
+			$post_id = $post->ID;
+		}
+
```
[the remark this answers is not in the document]
EugeneKyale: @redscar the filter doesn't normalize the ids for me. It's still returns the provisional ID when applied.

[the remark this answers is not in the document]
redscar: Definitely odd, I had the hook working in another method in that class. https://github.com/the-events-calendar/event-tickets/blob/release/G23.greenpeas/src/Tribe/Attendees.php#L619

I saw your work with the helper function, I think that will work and if we need to add the hook into it in the future it would allow us to.

redscar (dismissed):

src/functions/views/provider.php
```diff
@@ -17,3 +17,20 @@ function tribe_events_tickets_views_v2_is_enabled() {
...
+ *
+ * @return int The normalized post ID.
+ */
+function tribe_events_tickets_normalize_post_id( $post_id ) {
+	if ( class_exists( '\TEC\Events\Custom_Tables\V1\Models\Occurrence', false ) ) {
+		return \TEC\Events\Custom_Tables\V1\Models\Occurrence::normalize_id( $post_id );
+	}
+
+	return $post_id;
+}
```
rafsuntaskin, line 36: We already have a filter for this here: https://github.com/the-events-calendar/event-tickets/blob/2198626e84d17071a1fe9d6e17f2e2f0b540c9e0/src/Tickets/Custom_Tables/V1/Provider.php#L51

So, the wrapper method can just contain the filter and won't need to check the class anymore. Also, we are moving away from `tribe` namespace so, the method can be named the same as the filter.


```suggestion
function tec_tickets_filter_event_id( $post_id ) {
	return apply_filters( 'tec_tickets_filter_event_id', $post_id );
}
```

rafsuntaskin (changes requested):

src/functions/views/provider.php
EugeneKyale: @rafsuntaskin I've updated the function name to use the `tec` prefix. Using the filter in this scenario doesn't normalize the `$post_id`. I think there might be an issue with the filter execution order.

codingmusician (approved):
🚀

codingmusician (approved):

PR #2727, opened 2023-07-26
[Tests] - ETP `restv1` fixes.

rafsuntaskin:
### 🎫 Ticket

[TICKET_ID] 

### 🗒️ Description

* Depends on: https://github.com/the-events-calendar/tribe-common/pull/1965
* Relates to: https://github.com/the-events-calendar/event-tickets-plus/pull/1442

### ✔️ Checklist
- [ ] My code is tested. 


src/Tribe/Attendee_Repository.php
```diff
@@ -672,7 +673,7 @@ public function filter_by_checkedin( $checkedin ) {
 			return Tribe__Repository__Query_Filters::meta_in( $meta_keys, '1', 'is-checked-in' );
 		}
 
-		return Tribe__Repository__Query_Filters::meta_not_in_or_not_exists( $meta_keys, '1', 'is-not-checked-in' );
+		$this->filter_query->meta_not( $meta_keys, '1', 'is-not-checked-in' );
```
codingmusician, line 676: Should we be returning this?
```suggestion
		return $this->filter_query->meta_not( $meta_keys, '1', 'is-not-checked-in' );
```

tests/_support/Testcases/REST/V1/BaseTicketEditorCest.php
```diff
@@ -175,7 +175,7 @@ protected function create_rsvp_using_rest( Restv1Tester $I, array $variation, ar
 				'_tribe_ticket_capacity'         => '',
 				'_ticket_start_date'             => '2020-01-02 08:00:00',
 				'_ticket_end_date'               => '2050-03-02 20:00:00',
-				'_tribe_ticket_show_not_going'   => 'false',
+				'_tribe_ticket_show_not_going'   => false,
```
codingmusician, line 178: Should this also be zero, like you did below?
```suggestion
				'_tribe_ticket_show_not_going'   => 0,
```

src/Tribe/Attendee_Repository.php
rafsuntaskin: Nope, this is handled within the `meta_not` method which directly modifies the query and doesn't need to be returned.

tests/_support/Testcases/REST/V1/BaseTicketEditorCest.php
rafsuntaskin: The other ones are actually the odd ones, there is some issue with that particular test where sending `false` is going as ` `. That is why I used 0 to get rid of that issue. But `false` is the correct one to use.

rafsuntaskin: Ideally, it should have been fixed within this same scope but as many other PRs for ETP are blocked by this and it's a minor issue impacting this test only, therefore, we have made the decision to go with this workaround for now. We will definitely revisit this later to see why this particular test method is having issues.

I have checked with POSTman directly and the original API works as expected.

PR #2729, opened 2023-07-31
[ET-1773] Tickets Commerce PayPal - Show Notice if Not SSL

codingmusician:
### 🎫 Ticket

[ET-1773] 

### 🗒️ Description

Show notice in TC PayPal settings if site is running `http` only that keeps user from trying to connect to PayPal. This will not affect users that are already connected to PayPal.

### 🎥 Artifacts 

https://www.loom.com/share/d8cb5287870a4fe78b408c4b139f6c64?sid=fc556254-bfa5-40eb-9d98-7a0df03160b1

### ✔️ Checklist
- [x] My code is tested. 

[ET-1773]: https://theeventscalendar.atlassian.net/browse/ET-1773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/admin-views/settings/tickets-commerce/paypal/connect/inactive.php
```diff
@@ -13,12 +13,16 @@
  * @var TEC\Tickets\Commerce\Gateways\PayPal\Signup   $signup                [Global] The Signup class.
  * @var bool                                          $is_merchant_active    [Global] Whether the merchant is active or not.
  * @var bool                                          $is_merchant_connected [Global] Whether the merchant is connected or not.
+ * @var bool                                          $is_ssl                [Global] Whether the site is SSL or not.
```
redscar, line 16: Although I might be mistaken, wouldn't it be more appropriate to pass in these variables instead of setting $is_ssl below?

redscar: Also the `@since` needs to be updated.

redscar (commented):
Since I see that the snapshot updated to the non-SSL warning message, should we make another snapshot test to test when SSL is set?

src/admin-views/settings/tickets-commerce/paypal/connect/inactive.php
redscar: Would it make sense to check if the variable is empty that way is_ssl() isn't overriding the global variable?

codingmusician: Regarding adding a test, you're right. I'll need to write this in a test-friendly way, too. Back to the drawing board. Thanks, @redscar!

src/admin-views/settings/tickets-commerce/paypal/connect/non-ssl-notice.php
```diff
@@ -0,0 +1,47 @@
+<?php
+/**
+ * The Template for displaying the Tickets Commerce PayPal Settings notice when `https` is not being used.
+ *
+ * @version TBD
+ *
+ * @since   TBD
```
rafsuntaskin, line 7:
```suggestion
 * @since TBD
```

src/admin-views/settings/tickets-commerce/paypal/connect/signup-link.php
```diff
@@ -0,0 +1,24 @@
+<?php
+/**
+ * The Template for displaying the Tickets Commerce PayPal Settings signup link.
+ *
+ * @version TBD
+ *
+ * @since   TBD
```
rafsuntaskin, line 7:
```suggestion
 * @since TBD
```

redscar:
I see that the `commerce_integration` test failed.

redscar (approved):

PR #2732, opened 2023-08-02
[ET-1776] Tickets Emails: Capitalize Provider Names

codingmusician:
### 🎫 Ticket

[ET-1776] 

### 🗒️ Description

It was found that the "Completed Order" email was showing 'stripe' and 'paypal' as lower case. This PR formats them more formally. I also notices that the gateway info was not being shown within the preview, so I corrected that within the preview data.

### 🎥 Artifacts 

Before:
image

After:
image

### ✔️ Checklist
- [x] My code is tested. 

### DIFF
--- a/src/views/emails/template-parts/body/order/payment-info.php
+++ b/src/views/emails/template-parts/body/order/payment-info.php
@@ -12,6 +12,7 @@
  * @version 5.5.11
  *
  * @since 5.5.11
+ * @since TBD    Capitalize payment gateway name.
  *
  * @var Tribe__Template                    $this               Current template object.
  * @var \TEC\Tickets\Emails\Email_Abstract $email              The email object.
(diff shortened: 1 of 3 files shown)


redscar (commented):
This look's good. Instead of relying on `Ucase` would it be better to have the Provider name in each Provider? That way in the future if you have an instance of `PayPal` you don't have to add to the case?

Also I see the tests only test Stripe at the moment, should we test for each provider?

src/views/emails/template-parts/body/order/payment-info.php
```diff
@@ -27,15 +28,21 @@
 	return;
 }
 
+// Make sure payment gateway name is uppercase. If gateway is PayPal, ensure proper capitalization.
+$gateway_name = ucwords( $order->gateway );
+if ( 'Paypal' === $gateway_name ) {
+	$gateway_name = esc_html__( 'PayPal', 'event-tickets' );
+}
+
```
rafsuntaskin, line 36: There is a `get_label` method on gateways. Using that should be easier: https://github.com/the-events-calendar/event-tickets/blob/1a857b5135c22927fa9b3b5f187d7aef95e97028/src/Tickets/Commerce/Gateways/PayPal/Gateway.php#L73

rafsuntaskin (changes requested):

src/views/emails/template-parts/body/order/payment-info.php
codingmusician: Yep. I was actually working on that. Got a new solution for it, now.

redscar (dismissed):

rafsuntaskin (changes requested):
While you are at it, I found another place where the static method is called as instant method here: https://github.com/the-events-calendar/event-tickets/blob/1cb3283d998fb5c9429547537bccb8cd87e90b02/src/admin-views/settings/tickets-commerce/gateways/button.php#L40

rafsuntaskin (approved):

PR #2735, opened 2023-08-04
[ET-1776] Tickets Emails: Capitalize Provider Names (enhanced)

codingmusician:
### 🎫 Ticket

[ET-1776] 

### 🗒️ Description

In the previous PR https://github.com/the-events-calendar/event-tickets/pull/2732, it was pointed out that adding dynamic data to the Order Model was not the most efficient way to go. The PR reverts that change and utilized the `Order::get_gateway_label()` method that I was unaware of.

I also needed to update the preview data for the order to reflect actual data from live orders so that the `Order` methods would function properly and not throw any errors. The `gateway` and `post_status` were update and I added a `gateway_payload` param to the Preview Order data.

### 🎥 Artifacts 

Before:
image

After:
image

### ✔️ Checklist
- [x] My code is tested. 

[ET-1776]: https://theeventscalendar.atlassian.net/browse/ET-1776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Models/Order_Model.php
+++ b/src/Tickets/Commerce/Models/Order_Model.php
@@ -10,7 +10,6 @@
 namespace TEC\Tickets\Commerce\Models;
 
 use TEC\Tickets\Commerce;
-use TEC\Tickets\Commerce\Gateways\Manager;
 use TEC\Tickets\Commerce\Module;
 use TEC\Tickets\Commerce\Order;
 use TEC\Tickets\Commerce\Status\Status_Handler;
@@ -45,8 +44,6 @@ class Order_Model extends Base {
 			$gateway_slug        = Arr::get( $post_meta, [ Order::$gateway_meta_key, 0 ] );
 			$gateway_order_id    = Arr::get( $post_meta, [ Order::$gateway_order_id_meta_key, 0 ] );
 			$gateway_payload     = $this->get_gateway_payloads( $post_meta );
-			$gateway             = tribe( Manager::class )->get_gateway_by_key( $gateway_slug );
-			$gateway_name        = $gateway ? $gateway::get_label() : ucwords( str_replace( '_', ' ', $gateway_slug ) );
 			$status_log          = $this->get_status_log( $post_meta );
 			$status              = tribe( Status_Handler::class )->get_by_wp_slug( $this->post->post_status );
 			$flag_action_markers = $this->get_flag_action_markers( $post_meta );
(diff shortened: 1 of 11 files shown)


redscar (dismissed):

bordoni (approved):

src/Tickets/Commerce/Models/Order_Model.php
```diff
@@ -45,8 +44,6 @@ protected function build_properties( $filter ) {
 			$gateway_slug        = Arr::get( $post_meta, [ Order::$gateway_meta_key, 0 ] );
 			$gateway_order_id    = Arr::get( $post_meta, [ Order::$gateway_order_id_meta_key, 0 ] );
 			$gateway_payload     = $this->get_gateway_payloads( $post_meta );
-			$gateway             = tribe( Manager::class )->get_gateway_by_key( $gateway_slug );
-			$gateway_name        = $gateway ? $gateway::get_label() : ucwords( str_replace( '_', ' ', $gateway_slug ) );
```
rafsuntaskin, line 49: I will suggest to have the gateway name included from the Order Model as we can re-use this in other places if needed: https://github.com/the-events-calendar/event-tickets/pull/2735/files#diff-2a52eb3e077e150d63a4b0cc3e1ba88c713f66a1499cdf6e4d4b3b8e848a57e2R31

bordoni: This is what I was saying that we should avoid, adding complexity to the Order Model.
We either pull the gateway object by default or we pull just the slug and let the specific code pull the object if they need.

codingmusician: Getting the Gateway name is a simple call, too:
```$gateway_name = tribe( TEC\Tickets\Commerce\Order::class )->get_gateway_label( $order );```

codingmusician: I just wasn't aware it existed

rafsuntaskin: @bordoni for me, it makes more sense to have the core info bundled into Order Model data as much as we can. Unless we are making expensive queries.

rafsuntaskin: I agree that having the whole $gateway object makes more sense, instead of having the properties seprately.

bordoni: My only concern in general is that this model is not using JIT and Caching yet, which makes each operation that we bind to the model more expensive.

bordoni (approved):
Perfecto.

PR #2753, opened 2023-08-31
Remove cache, go with fast queries

lucatume:
This PR is born in the context of the Flexible Tickets project.

I found myself needing to modify the query to report the Ticketed/Unticketed numbers to take Series Passes into account.
While working on the code to find an entry point I've found pretty old code in charge of populating the values and filtering the posts.
The code was not performant or scalable:
* all queries are reported as "slow" by Query Monitor
* all queries are unbound, broken into multiple queries (all of them unbound) ...
* ... leading to a lot of arrays moving around in PHP (i.e. RAM) and TCP (the Db) that are, then, discarded to run just a `count()` on them

So, pretty bad. I'm proud to say that stupid code is mostly mine. Go me!

Also: no tests.

This PR:
* adds the entry points I require for FT
* adds testing coverage for the changed code
* removes the Cache and related abstractions (must have been a dogmatic moment for me ...)
* updates all the queries to be fast queries

Artifacts:
* [Ticketed before](https://share.cleanshot.com/BLLLtm7p) and [Unticketed before](https://share.cleanshot.com/bLdTkwwG)
* [Ticketed after](https://share.cleanshot.com/DHn7GQvM) and [Unticketed after](https://share.cleanshot.com/b7YYBpqc)

No slow queries (no object-cache installed) and ~30% bettter performance.


lucatume:
Note: tests are failing due to out-of-sync branches, I will fix them on the main branch for the FT project.

bordoni (commented):
Move the removed files to deprecated tho. Just in case. Leave them in place

PR #2757, opened 2023-09-06
Restore possibility to move Tickets/Attendees to all Single Events

lucatume:
### 🎫 Ticket

[ET-1862](https://theeventscalendar.atlassian.net/browse/ET-1862) 

### 🗒️ Description

This PR restores the functionality to allow moving Attendees between any Single Event on the site, whether the Single Event belongs, in the context of CT1, to a Series or not.

A note on the fix: the exclusion of recurring events was previously done incorrectly by leveraging the support in ECP for the `post__not_in_series` query var.
That would exclude not just Recurring Events (part of a Series by default) but Single Events part of a Series as well. Herein lies the issue.
In the PR I'm setting the `post__not_recurring` query var, hinting at its future support; it's currently not part of ECP code.
Since there is not `post__not_recurring` support to leverage, and to deal with an ECP support that might come later than the shipping of ET, I've created a method that will handle the exclusion part of the `Tribe__Tickets__Admin__Move_Tickets` class, the `exclude_recurring_events` one. That method is meant to be a "bridge" to provide the Recurring Event  exclusion functionality while the support for `post__not_recurring` makes its way into ECP code, [see the ECP PR](https://github.com/the-events-calendar/events-pro/pull/2315) 

Lastly: the PR reintroduces the run of the `ct1_integration` suite in CI.

### 🎥 Artifacts 
* [Screencast](https://share.cleanshot.com/T73qWKK8) 
* tests

### ✔️ Checklist
- [x] My code is tested. 

[ET-1862]: https://theeventscalendar.atlassian.net/browse/ET-1862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7.2",
     "fzaninotto/faker": "^1.8",
     "lucatume/wp-browser": "^3.1.4",
-    "lucatume/wp-snapshot-assertions": "^1.1.0",
+    "lucatume/wp-snaphot-assertions": "^1.1.0",
     "phpunit/phpunit": "^6.5.14",
     "spatie/phpunit-snapshot-assertions": "^1.4.2",
     "the-events-calendar/coding-standards": "dev-master",
--- a/.github/workflows/tests-php-ct1.yml
+++ b/.github/workflows/tests-php-ct1.yml
@@ -1,4 +1,4 @@
-name: 'Codeception Tests'
+name: 'CT1 Codeception Tests'
 on: [ pull_request ]
 jobs:
   test:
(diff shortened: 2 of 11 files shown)


composer.json
```diff
@@ -25,7 +25,7 @@
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7.2",
     "fzaninotto/faker": "^1.8",
     "lucatume/wp-browser": "^3.1.4",
-    "lucatume/wp-snapshot-assertions": "^1.1.0",
+    "lucatume/wp-snaphot-assertions": "^1.1.0",
```
rafsuntaskin, line 28:
```suggestion
    "lucatume/wp-snapshot-assertions": "^1.1.0",
```

https://github.com/lucatume/wp-snapshot-assertions this one, right?

lucatume: [The name is historically incorrect](https://github.com/lucatume/wp-snapshot-assertions/blob/master/composer.json#L2); the correct Composer name to use `snapHot-assertions`.

rafsuntaskin (approved):

PR #2766, opened 2023-09-19
[ETP-882] AR modal not showing after install despite of being default selection

redscar:
[ETP-882] 

### 🗒️ Description

While this ticket is for ET+ all of the logic is within ET.  By switching to `tribe_get_option` from `$settings_manager::get_option` we can set the default to true. Originally the option was coming back null which was causing it to return false.

### 🎥 Artifacts 

This video is from a fresh sandbox install. I enabled Paypal, created a ticket and took the video.
[artifact-ETP-882.webm](https://github.com/the-events-calendar/event-tickets/assets/12241059/aeea560f-3deb-4d3d-94f0-1cbd32d95371)

### ✔️ Checklist
- [x] My code is tested. 

[ETP-882]: https://theeventscalendar.atlassian.net/browse/ETP-882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Attendee_Registration/Main.php
```diff
@@ -211,18 +211,17 @@ public function get_attendee_registration_page() {
...
 	 */
 	public function is_modal_enabled( $post = null ) {
-		/** @var $settings_manager Tribe__Settings_Manager */
-		$settings_manager = tribe( 'settings.manager' );
 
-		$event_tickets_plus = class_exists( 'Tribe__Tickets_Plus__Main' );
+		$event_tickets_plus = class_exists( 'Tribe__Tickets_Plus__Main', false );
 
-		$option = $settings_manager::get_option( 'ticket-attendee-modal' );
+		$option = tribe_get_option( 'ticket-attendee-modal', true );
```
codingmusician, line 224: `Tribe__Settings_Manager::get_option()` has a parameter for a default value. Why did you choose to go with `tribe_get_option()`?

redscar: I was originally looking at how some of our other options are grabbed and they were using `tribe_get_option`. It seems like `tribe_get_option` actually calls `Tribe__Settings_Manager` with a filter as well.

readme.txt
```diff
@@ -193,6 +193,7 @@ Check out our extensive [knowledgebase](https://evnt.is/18wm) for articles on us
 
 = [TBD] TBD =
 
+* Fix - On a new installation of Event Tickets Plus, the Attendee Registration feature will now use the modal by default.  [ETP-882]
```
codingmusician, line 196: I think this could be simplified
```suggestion
* Fix - The Attendee Registration feature will now use the modal by default.  [ETP-882]
```

rafsuntaskin:
@redscar we also have the `default` param available from the existing method: https://github.com/the-events-calendar/tribe-common/blob/abb3e84806fb7d5b02a25430cebf3796b7ef0ad3/src/Tribe/Settings_Manager.php#L101

Was that not working? Or any other particular reason to use `tribe_get_option` instead?

redscar:
> @redscar we also have the `default` param available from the existing method: [the-events-calendar/tribe-common@`abb3e84`/src/Tribe/Settings_Manager.php#L101](https://github.com/the-events-calendar/tribe-common/blob/abb3e84806fb7d5b02a25430cebf3796b7ef0ad3/src/Tribe/Settings_Manager.php#L101)
> 
> Was that not working? Or any other particular reason to use `tribe_get_option` instead?

I decided to switch back to the original code with the default value. @rafsuntaskin I honestly switched to `tribe_get_options` so that we didn't have to load the settings manager in the method. For my knowledge, does the original method offer any better functionality than tribe_get_option?

rafsuntaskin:
> > @redscar we also have the `default` param available from the existing method: [the-events-calendar/tribe-common@`abb3e84`/src/Tribe/Settings_Manager.php#L101](https://github.com/the-events-calendar/tribe-common/blob/abb3e84806fb7d5b02a25430cebf3796b7ef0ad3/src/Tribe/Settings_Manager.php#L101)
> > Was that not working? Or any other particular reason to use `tribe_get_option` instead?
> 
> I decided to switch back to the original code with the default value. @rafsuntaskin I honestly switched to `tribe_get_options` so that we didn't have to load the settings manager in the method. For my knowledge, does the original method offer any better functionality than tribe_get_option?

It has a different filter than the one that is being used within `tribe_get_option` and ideally we should totally move these methods to ETP. For me any of them were fine, I was just curious whether you had any other scenario in mind that I might be missing. 

Nice work overall. 👍

rafsuntaskin (approved):

PR #2768, opened 2023-09-19
Address some dynamic property issues with PHP 8.1

Camwyn:
This has no ticket (yet) - it just addresses a few reported warnings about dynamic properties in php 8.1+


bordoni (changes requested):
Add changelog

bordoni (approved):

PR #2781, opened 2023-09-26
Include a new Facade that handles QR code properly for WordPress

bordoni:
### 🎫 Ticket

[ETWP-29] 

### 🗒️ Description

Create a initial version of the QR Code facade, I've intentionally only created the methods we will need to do the operations we are handling currently on our current codebase.

This is very much a draft that I am looking to get some input from you all. Tomorrow my goal is to add tests. but doing QA on the code it covers exactly what we were doing before but with a layer to allow us to switch to a better library in the future.

### 🎥 Artifacts 
- **QR Code for Setting up ET+ App** 
   https://i.bordoni.me/sjyGpvb6

[ETWP-29]: https://theeventscalendar.atlassian.net/browse/ETWP-29?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/QR/Controller.php
```diff
@@ -0,0 +1,172 @@
...
+
+	/**
+	 * Boot the Controller.
+	 *
+	 * This function is used to instantiate the singleton classes and register any other providers.
+	 *
+	 * @since   TBD
+	 */
+	public function boot() {
+		$this->container->bind( Facade::class, [ $this, 'bind_facade_or_error' ] );
```
lucatume, line 35: Do not use the `boot` method to register a binding, use the `do_register` method of the Controller.
The `boot` method is meant to contain methods that would **build** things that are required at the application boot time, in contrast with the application initialization phase that will rely on the Providers/Controllers `do_register` methods.

src/Tickets/QR/Controller.php
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	/**
+	 * Register the controller.
+	 *
+	 * @since   TBD
+	 */
+	public function do_register(): void {
+		$this->boot();
```
lucatume, line 44: Avoid overriding the default `boot` method if there is nothing to really boot. What this controller should do, is just register a binding, see my previous comment.

src/Tickets/QR/Controller.php
```diff
@@ -0,0 +1,172 @@
...
+
+	/**
+	 * Determines if the QR code library is loaded.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	protected function has_library_loaded(): bool {
+		return class_exists( 'QRencode', false ) || class_exists( 'QRcode', false );
```
lucatume, line 135: If I understand this correctly, the idea is to avoid auto-loading the library from **another** source that might provide it (e.g. another plugin); but this will not avoid a conflict with that library.
If it's already loaded, it might not be the version required by ET.
If it's not loaded, and is not loaded at this stage since autoloading is omitted, then loading the ET version of the library would not avoid a possible version conflict.
Since the library is shipped with the plugin anyway, why not vendor-prefix the QR library and just always load the vendor-prefixed version?

lucatume (changes requested):
Some preliminary observations, a question in line.
Another question: I understand the `cache` directory is shipped as the library cannot be bent into use WordPress filesystem layer, but is this really required? I see a `QR_CACHEABLE` constant that leads me to believe that could be avoided. What is the performance price of having cache off? From what I understand, the QR is used in admin UI and in mail dispatching.

src/Tickets/QR/Controller.php
borkweb: Agreed. We definitely need to do vendor prefixing

src/Tickets/QR/Controller.php
```diff
@@ -0,0 +1,172 @@
...
+
+	/**
+	 * Register the controller.
+	 *
+	 * @since   TBD
+	 */
+	public function do_register(): void {
+		$this->boot();
+
+		$this->add_actions();
```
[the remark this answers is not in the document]
bordoni: Sounds good. I will go that route.

src/Tickets/QR/Facade.php
```diff
@@ -0,0 +1,177 @@
...
+namespace TEC\Tickets\QR;
+
+/**
+ * Class Facade
+ *
+ * @since   TBD
+ *
+ * @package TEC\Tickets\QR
+ */
+class Facade {
```
borkweb, line 12: I'm not a huge fan of using the word Facade for the actual class name. It can be a facade, but be a name that is more recognizable. Perhaps: `TEC\Tickets\QR\QR`

That way, any operation/interaction is immediately recognizable without the need for a `use` alias.

```php
tribe( QR::class )->get_png_as_string( $data );
```

bordoni:
@lucatume I will persue the route of not including the cache and adding it to our namespace.
I will do both manually since this is not included via Composer, I wanted to stick as close to the one in ET+ as possible.

src/Tickets/QR/QR.php
```diff
@@ -0,0 +1,179 @@
...
+		$folder = '/' . ltrim( $folder, '/' );
+		$png_as_string = $this->get_png_as_string( $data );
+
+		// Filters the upload directory but still use `wp_upload_bits` to create the file.
+		$upload_bits_filter = static function( $arr ) use ( $folder ) {
+	        $arr['url'] = str_replace( $arr['subdir'], $folder, $arr['url'] );
+	        $arr['path'] = str_replace( $arr['subdir'], $folder, $arr['path'] );
+			$arr['subdir'] = $folder;
+		    return $arr;
+		};
```
codingmusician, line 168: Indention here:
```suggestion
		$upload_bits_filter = static function( $arr ) use ( $folder ) {
			$arr['url'] = str_replace( $arr['subdir'], $folder, $arr['url'] );
			$arr['path'] = str_replace( $arr['subdir'], $folder, $arr['path'] );
			$arr['subdir'] = $folder;
			return $arr;
		};
```

src/Tickets/QR/QR.php
```diff
@@ -0,0 +1,179 @@
+<?php
+
+namespace TEC\Tickets\QR;
+
+use TEC\Tickets\phpqrcode\QRcode;
+
+/**
+ * Class Facade
```
redscar, line 8: Should this be `Class QR` now?

codingmusician (approved):

src/Tickets/QR/QR.php
```diff
@@ -0,0 +1,179 @@
...
+	/**
+	 * Get the QR code as a file uploaded to WordPress.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $data
+	 * @param string $name
+	 * @param string $folder
+	 *
+	 * @return array
```
lucatume, line 156:
```suggestion
	 * @return array{file: string, url: string, type: string, error: string|false} The QR uploaded file information.
```

tests/integration/TEC/Tickets/QR/QR_Test.php
```diff
@@ -0,0 +1,123 @@
...
+class QR_Test extends \Codeception\TestCase\WPTestCase {
+	use MatchesSnapshots;
+
+	protected $driver;
+
+	public function setUp(): void {
+		parent::setUp();
+
+		// Setup a new HTML output driver to make sure our stuff is tolerable.
+		$this->driver = new WPHtmlOutputDriver( home_url(), 'http://views.dev' );
```
lucatume, line 24: This builds a test dependency on the snapshot framework. 
I would remove this one and go with the general solution of filtering the `home_url` function to return `https://views.dev` to stabilize the snapshots.

Also: do not extend the `setUp` method, use a method with a `@before` annotation for better cross-compatibility with Codeception and PHPUnit versions and better readability.

tests/integration/TEC/Tickets/QR/QR_Test.php
```diff
@@ -0,0 +1,123 @@
...
+
+		$this->assertMatchesSnapshot( json_encode( $upload ), $this->driver );
+		$this->assertFileExists( $upload['file'] );
+		$this->assertEmpty( $upload['error'] );
+
+		if ( ! empty( $folder ) ) {
+			$this->assertContains( $folder, $upload['file'] );
+		}
+
+		unlink( $upload['file'] );
```
lucatume, line 119: You cannot do this here: if the test does not pass, this line will never execute and the file will never be removed leaving a dirty state.
Move this into a method annotated with `@after` that will always execute, no matter the test's success or failure.

tests/integration/TEC/Tickets/QR/QR_Test.php
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Class QR_Test.
+ *
+ * @since   TBD
+ *
+ * @package TEC\Tickets\QR
+ */
+class QR_Test extends \Codeception\TestCase\WPTestCase {
+	use MatchesSnapshots;
```
lucatume, line 16: I would use the `codeception-snapshot-asssertion` library: it's made for Codeception, uses Codeception-format snapshots, and produces readable snapshots.
This snapshot library was made for PHPUnit and jerry-rigged into Codeception when there was no alternative.

lucatume (changes requested):
Doc-block cleanup, test iteration.

src/Tickets/QR/Controller.php
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	/**
+	 * Unregister the controller.
+	 *
+	 * @since   TBD
+	 *
+	 * @return void
+	 */
+	public function unregister(): void {
```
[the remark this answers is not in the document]
bordoni: Cannot remove it, it's a required method.

src/Tickets/QR/QR.php
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * @package TEC\Tickets\QR
+ */
+class QR {
+	/**
+	 * The level of the QR code.
+	 *
+	 * @since TBD
+	 *
+	 * @var int
```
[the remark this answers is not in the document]
bordoni: What do you mean here? The description?

src/Tickets/QR/Controller.php
```diff
@@ -0,0 +1,109 @@
...
+ * @package TEC\Tickets\QR
+ */
+class Controller extends Controller_Contract {
+	/**
+	 * Determines if this controller will register.
+	 * This is present due to how UOPZ works, it will fail if method belongs to the parent/abstract class.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
```
lucatume, line 21:
```suggestion
	 * @return bool Whether the controller is active or not.
```

lucatume (approved):

borkweb (dismissed):
```
🌤️ 
      🎈 
                    🏃‍♂️ 
📗📗📗📗📗📗📗📗📗📗📗📗📗📗📗
```

PR #2788, opened 2023-09-29
Move the location of service provider to avoid problems with ECP

bordoni:
### 🎫 Ticket

[ET-1886] 

### 🎥 Artifacts 
No artifacts in this case, it just works. Nothing changes.

[ET-1886]: https://stellarwp.atlassian.net/browse/ET-1886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Main.php
+++ b/src/Tribe/Main.php
@@ -368,9 +368,6 @@ class Tribe__Tickets__Main {
 	 * @since 4.10
 	 */
 	public function bootstrap() {
-		// Initialize the Service Provider for Tickets.
-		tribe_register_provider( 'Tribe__Tickets__Service_Provider' );
-
 		$this->hooks();
 
 		$this->register_active_plugin();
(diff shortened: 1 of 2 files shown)


src/Tribe/Main.php
```diff
@@ -401,6 +398,9 @@ public function bootstrap() {
 	public function bind_implementations() {
 		tribe_singleton( 'tickets.main', $this );
 
+		// Initialize the Service Provider for Tickets.
+		tribe_register_provider( Tribe__Tickets__Service_Provider::class );
+
```
redscar, line 403: Should the `@since` be updated for the `bind_implementations` and `bootstrap` method?

bordoni: Not in this case, no major changes to the existence of the method signature of the method/class.

redscar (approved):

PR #2800, opened 2023-10-09
Move all pieces from QR code

bordoni:
### 🎫 Ticket

[ETWP-39] 

### 🗒️ Description

Please watch the video, I know it's long, but it explains what was done here.

### 🎥 Artifacts 
https://i.bordoni.me/PN1xq2Jt

[ETWP-39]: https://stellarwp.atlassian.net/browse/ETWP-39?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/QR/Connector.php
```diff
@@ -0,0 +1,237 @@
...
+
+		if ( empty( $link ) ) {
+			return null;
+		}
+
+		if ( is_wp_error( $qr_code ) ) {
+			return null;
+		}
+
+		// @ToDo @rafsuntaskin @juanfra change the default directory to `tec/tickets-plus/qr-codes` for storing images.
```
codingmusician, line 163: Is this @todo relevant anymore since QRs will be accessible from different plugins?

src/Tickets/QR/Controller.php
```diff
@@ -27,21 +27,108 @@ public function is_active(): bool {
...
+	 * @uses  Notices::register_admin_notices()
+	 *
+	 * @return void
 	 */
 	public function do_register(): void {
 		$this->container->bind( QR::class, [ $this, 'bind_facade_or_error' ] );
+		$this->container->singleton( Settings::class, Settings::class );
+		$this->container->singleton( Notices::class, Notices::class );
+		$this->container->singleton( Connector::class, Connector::class );
+		$this->container->singleton( Observer::class, Observer::class );
```
codingmusician, line 41: Since the implementation is same as class name, the 2nd arg isn't necessary.
```suggestion
		$this->container->singleton( Settings::class );
		$this->container->singleton( Notices::class );
		$this->container->singleton( Connector::class );
		$this->container->singleton( Observer::class );
```

src/Tickets/QR/Notices.php
```diff
@@ -0,0 +1,87 @@
...
+	 * Gets the notice for the QR code dependency.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public function get_dependency_notice_contents(): string {
+		$html = '<h2>' . esc_html__( 'QR codes for tickets not available.', 'event-tickets' ) . '</h2>';
+		$html .= esc_html__( 'In order to have QR codes for your tickets you will need to have both the `php_gd2` and `gzuncompress` PHP extensions installed on your server. Please contact your hosting provider.', 'event-tickets' );
+		$html .= ' <a target="_blank" href="https://evnt.is/XXXX">' . esc_html__( 'Learn more.', 'event-tickets' ) . '</a>';
```
juanfra, line 83: We need a todo to change this URL and to have an article.

src/Tickets/QR/Settings.php
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	/**
+	 * Get the API hash key, if none exists, generate one.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public function get_api_hash(): string {
```
juanfra, line 144: I find this confusing, it's the api key you're getting by using the method. And the term we use for it is Api Key. Naming this api hash makes it more confusing than anything else.

src/Tickets/QR/Settings.php
```diff
@@ -0,0 +1,165 @@
...
+	/**
+	 * Get the option key that will be used to store the value.
+	 *
+	 * Option keys that are user controlled, need to be kept as public.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public static function get_enabled_key(): string {
```
juanfra, line 47: I believe we should move all the settings to one place. Partially having some of them here, and then the rest duplicated in two plugins will make this piece really difficult to maintain. We could find a way to elegantly enable things from the "satellite" plugins.

bordoni: No, the goal is there is one QR code enabled setting. Not one per plugin.

src/Tickets/QR/Settings.php
bordoni: So in the previous code there was two different terminologies, I picked hash because that is what we do with the method, we create a md5 hash of a random password. Now if we want to go with `get_api_key` we can go that route too.

src/Tickets/QR/Connector.php
bordoni: Partially so I will modify.

src/Tickets/QR/Observer.php
```diff
@@ -0,0 +1,282 @@
...
+		 * @since TBD
+		 *
+		 * @param string $url             URL to redirect to, gets escaped upstream.
+		 * @param int    $event_id        Event Post ID.
+		 * @param int    $ticket_id       Ticket Post ID.
+		 * @param bool   $user_had_access Whether the logged-in user has permission to perform check ins.
+		 */
+		$url = apply_filters( 'tec_tickets_qr_observer_handle_checkin_redirect', $url, $event_id, $ticket_id, $user_had_access );
+
+		wp_redirect( esc_url_raw( $url ) );
```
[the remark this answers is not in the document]
bordoni: We can, but not the goal of this PR. I just ported code.

src/Tickets/QR/Observer.php
```diff
@@ -0,0 +1,282 @@
...
+		$event_id  = tribe_get_request_var( 'event_id', false );
+
+		// Check all the data we need is there.
+		if ( empty( $ticket_id ) || empty( $event_id ) ) {
+			return;
+		}
+
+		$event_id      = (int) $event_id;
+		$ticket_id     = (int) $ticket_id;
+		$security_code = (string) esc_attr( tribe_get_request_var( 'security_code' ) );
```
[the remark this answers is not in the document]
bordoni: We could but this way we avoid checking this until we have ticket and event ID.

codingmusician (approved):

PR #2818, opened 2023-10-25
[ET-1723] - Feat - Updated Event order pages from series with new order report design.

rafsuntaskin:
### 🎫 Ticket

[ET-1723] 

### ✔️ Checklist
- [x] My code is tested. 

[ET-1723]: https://stellarwp.atlassian.net/browse/ET-1723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/integration/_bootstrap.php
```diff
@@ -10,11 +10,6 @@
 DB::query( "ALTER TABLE $wpdb->posts AUTO_INCREMENT = 5096" );
 
 // Disconnect Promoter to avoid license-related notices.
-
-// Start the posts auto-increment from a high number to make it easier to replace the post IDs in HTML snapshots.
-global $wpdb;
-DB::query( "ALTER TABLE $wpdb->posts AUTO_INCREMENT = 5096" );
```
lucatume, line 16: Why remove this?

tests/integration/Tribe/Admin/OrderReportTest.php
```diff
@@ -77,8 +77,8 @@ function (): array {
 				$ticket_id = $this->create_tc_ticket( $event_id, 10 );
 
 				$this->set_fn_return( 'current_time', '2020-02-22 22:22:22' );
-
 				$order_a = $this->create_order( [ $ticket_id => 2 ], [ 'purchaser_email' => 'purchaser@test.com' ] );
+				$this->set_fn_return( 'current_time', '2020-02-23 22:22:22' );
```
lucatume, line 81: I understand why this might be required in concept, either:
1. Remove this and use another method to update the Order date (e.g. targeted `update_post_meta` or `wp_update_post`) after you've inserted the order.
2. Add a comment to explain why you're doing this.

Of the two, I would prefer the first option: altering time can very easily have unforeseen consequences and implications that are not immediate. If what you're trying to do is simulate an order placed in the past, just alter the order data.

src/Tickets/Flexible_Tickets/Base.php
```diff
@@ -508,4 +512,93 @@ public function render_series_details_for_attached_event( int $post_id ): void {
...
+	}
+
+	/**
+	 * Filters the order report to remove the series passes from the event sales data.
+	 *
+	 * @since TBD
+	 *
+	 * @param bool          $include Whether to include the event sales data.
+	 * @param Ticket_Object $ticket  The ticket object.
+	 * @param array         $quantity_by_status The quantity by status.
```
lucatume, line 586: Document this with the array key and value types.
e.g. `array<string,int>`.

src/Tickets/Flexible_Tickets/Base.php
```diff
@@ -508,4 +512,93 @@ public function render_series_details_for_attached_event( int $post_id ): void {
...
+	/**
+	 * Renders the series details on order report page for an event attached to a series.
+	 *
+	 * @since TBD
+	 *
+	 * @param int $post_id The ID of the post being displayed.
+	 *
+	 * @return void
+	 */
+	public function render_series_details_for_attached_event_order_report( int $post_id ): void {
```
lucatume, line 525: It makes sense to do this in the `Base` controller, but it's growing too large.
Please extract this method, and others that might be about altering the reports, into a dedicated, proxied class (e.g. [like this](https://github.com/the-events-calendar/event-tickets/blob/7285785ea2a74359e84afdc42b351e6b5894e9a5/src/Tickets/Flexible_Tickets/Series_Passes.php#L100) and [this](https://github.com/the-events-calendar/event-tickets/blob/7285785ea2a74359e84afdc42b351e6b5894e9a5/src/Tickets/Flexible_Tickets/Series_Passes.php#L341)).

src/Tickets/Commerce/Reports/Data/Order_Summary.php
```diff
@@ -291,4 +299,28 @@ public function get_event_sales_data(): array {
...
+		 * Filters if the ticket sales data should be included into event sales data.
+		 *
+		 * @since TBD
+		 *
+		 * @param bool $should_include Whether to include the sales data into event sales data.
+		 * @param Ticket_Object $ticket The ticket object.
+		 * @param array<string,int> $quantity_by_status The quantity by status.
+		 * @param Order_Summary $this The order summary object.
+		 */
+		return apply_filters( 'tec_tickets_commerce_order_report_summary_should_include_event_sales_data', true, $ticket, $quantity_by_status, $this );
```
lucatume, line 324: Break down the arguments into separate lines to avoid long lines wrapping.

src/Tickets/Commerce/Reports/Data/Order_Summary.php
```diff
@@ -21,7 +21,7 @@ class Order_Summary {
 	 *
 	 * @var int The post ID.
 	 */
-	protected int $post_id;
+	public int $post_id;
```
lucatume, line 24: Agh, no. 
If you need to have this `public` to read its value, add a getter method and leave this `protected`.
Whenever you make a property `public`, you should add all the code that, on each method is making sure that property value, which can then be changed from any piece of code at any moment, still makes sense.

lucatume (changes requested):

tests/integration/_bootstrap.php
rafsuntaskin: It's not removed. It's just Github file is not refreshed or something.Check the whole file link and it exists.

tests/integration/Tribe/Admin/OrderReportTest.php
rafsuntaskin: I actually tried the 1st method that you proposed. But the thing is once a post is updated the created/updated date is also updated. that's a loop. Doesn't matter if you are updating the created/updated date or not.

rafsuntaskin: `update_post_meta` won't help as the date is coming from the post table data.

src/Tickets/Commerce/Reports/Data/Order_Summary.php
```diff
@@ -235,6 +251,9 @@ public function get_tickets(): array {
 	 * @return array<string, array{ ticket: Ticket_Object, label: string, type: string, qty_data: array, qty_by_status: string }> A map from ticket types to the Tickets of that type.
 	 */
 	public function get_tickets_by_type(): array {
+		// Sort the tickets by keys to display default types first.
+		ksort( $this->tickets_by_type );
```
lucatume, line 255: Meh, hee you're relying on `default` coming before `series_pass` or whatever type; it's not a solid solution as it will have the side-effect of changing if a ticket type is defined that comes before, alpha-numerically speaking, `default`.

Take the robust and verbose route here with an `array_merge`:
```php
$ordered = array_merge(['default' => []], $this->tickets_by_type);
```

src/Tickets/Flexible_Tickets/Series_Passes/Reports.php
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	/**
+	 * Registers the hooks.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function register_hooks(): void {
```
lucatume, line 48: I meant something else.
Any `add_action` or `add_filter` call should stay in the Controller, the `Base` one in this case.
The link I had provided in my comment showed the Controller delegating the actual method to a specialized class.

src/Tickets/Flexible_Tickets/Series_Passes.php
```diff
@@ -132,6 +144,8 @@ public function is_active(): bool {
 	protected function do_register(): void {
 		$this->container->singleton( Series_Passes\Repository::class, Series_Passes\Repository::class );
 		$this->container->singleton( Series_Passes\Metadata::class, Series_Passes\Metadata::class );
+		// Register reporting functions.
+		$this->reports->register_hooks();
```
lucatume, line 148: See my other note: **only controllers** should hook actions and filters.

lucatume (changes requested):

src/Tickets/Flexible_Tickets/Series_Passes/Reports.php
rafsuntaskin: As this was handling reporting related to Series Passes only (for now) so, I thought it belonged under Series pass namespace.

src/Tickets/Flexible_Tickets/Series_Passes.php
rafsuntaskin: Looks like this is set as tolerable and is working as expected to avoid conflict: https://github.com/the-events-calendar/event-tickets-plus/blob/c3d06592c99c9767cb5f12346262f310cbc8ea09/tests/wpunit/Tribe/Tickets_Plus/Partials/V2/Modal/Attendee_Registration_Test.php#L48

lucatume (approved):

PR #2837, opened 2023-11-09
Fix PHP8.1 deprecation notices.

juanfra:
### 🎫 Ticket

N/A

### 🗒️ Description

- Fix PHP8.1 deprecation notices.


bordoni (approved):
I think that is good unless @lucatume knows a better way around this.

PR #2838, opened 2023-11-09
DQA 11,13,14 My Tickets & Your Tickets button stacking

redscar:
### 🎫 Ticket

DQA 11 & DQA 13 & DQA 14

### 🗒️ Description

This is an extension to my main PR https://github.com/the-events-calendar/event-tickets-wallet-plus/pull/77
This adds the setup to be able to use a container query.

Video of the "My Tickets" page.
[artifact_dqa_13.webm](https://github.com/the-events-calendar/event-tickets/assets/12241059/e6cf9f07-44c2-4992-94d0-1db6e0d1cbde)

image

image

image

image

image

### DIFF
--- a/src/resources/postcss/my-tickets/_orders.pcss
+++ b/src/resources/postcss/my-tickets/_orders.pcss
@@ -28,6 +28,7 @@
 		padding: 0;
 
 		> .tribe-item {
+			container-type: inline-size;
 			margin: 0;
 			padding: 0;
 		}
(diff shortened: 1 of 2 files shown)


src/resources/postcss/my-tickets/_orders.pcss
```diff
@@ -86,22 +87,23 @@
 
 				& > div {
 					flex: 1 0 100%;
+					flex-basis: var(--tec-grid-width-1-of-4);
```
redscar, line 90: I added this piece because the content wasn't breaking properly when resizing the page. However, I'm a bit on the fence about locking the "My Tickets" page into always having four columns. As it stands, the page shifts between two, three, or four columns, but setting it to four by default might throw things off, especially if ET+ or ETWP aren't being used, or if down the line we decide to switch up the column count. Does anyone have any ideas?

Camwyn: You can set it to whatever here and then override it in ET+/W+ although that might make it harder to modify later on (more places to change)

Camwyn (approved):

PR #2843, opened 2023-11-14
Forgotten TBDs from older version.

bordoni:


redscar (approved):

juanfra (dismissed):

src/Tickets/QR/Controller.php
```diff
@@ -51,7 +51,11 @@ public function do_register(): void {
 	/**
 	 * Unregister the controller.
 	 *
+<<<<<<< HEAD
```
estebanhdzc, line 54: Looks like a conflic here 😅

PR #2908, opened 2024-01-24
[ET-1925] Fix fatal found during testing

redscar:
### 🎫 Ticket

[ET-1925] 

### 🗒️ Description
When testing on a fresh install a fatal occurred due to an issue with the list of previous installed versions. This fixes the error and another I found during testing when TEC is disabled.

image

[ET-1925]: https://stellarwp.atlassian.net/browse/ET-1925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rafsuntaskin (dismissed):
returning inventory here will not return proper value.
apply like this:
https://github.com/the-events-calendar/event-tickets/blob/26f698e8d26b81150df2b66e1bc17cb002786f96/src/Tribe/Ticket_Object.php#L678

src/Tickets/Commerce/Provider.php
```diff
@@ -28,7 +28,7 @@ class Provider extends Service_Provider {
 	 */
 	public function register() {
 		// Specifically prevents anything else from loading.
-		if ( ! tec_tickets_commerce_is_enabled() ) {
+		if ( ! \tec_tickets_commerce_is_enabled() ) {
```
rafsuntaskin, line 31: do we still need the `\`. I don't think the fails earlier were related to this one.

src/Tickets/Site_Health/Subsections/Plugins/Plugin_Data_Subsection.php
```diff
@@ -404,7 +409,7 @@ private function get_number_of_attendees(): int {
 	 * @return array Associative array with formatted average, max, and min ticket prices.
 	 */
 	private function get_formatted_prices(): array {
-		$ticket_prices = tribe( 'tickets.event-repository' )->per_page( -1 )->where( 'has_tickets' )->pluck( 'cost' );
+		$ticket_prices = tribe( 'tickets.ticket-repository' )->per_page( -1 )->where( 'has_tickets' )->pluck( 'cost' );
```
rafsuntaskin, line 412: I think it should be event, as there are no `has_tickets` filter in the tickets repo.

tests/integration/_bootstrap.php
```diff
@@ -12,8 +12,9 @@
 // Disconnect Promoter to avoid license-related notices.
 remove_action( 'tribe_tickets_promoter_trigger', [ tribe( Dispatcher::class ), 'trigger' ] );
 
+require_once trailingslashit( EVENT_TICKETS_DIR ) . 'src/template-tags/commerce.php';
 // Ensure Ticket Commerce is enabled.
-if ( ! tec_tickets_commerce_is_enabled() ) {
+if ( ! \tec_tickets_commerce_is_enabled() ) {
```
codingmusician, line 17:
```suggestion
if ( ! tec_tickets_commerce_is_enabled() ) {
```

src/Tickets/Site_Health/Subsections/Plugins/Plugin_Data_Subsection.php
```diff
@@ -229,15 +229,20 @@ protected function generate_subsection(): array {
...
-					[]
-				)
-			),
+		$previous_versions = array_filter(
+			(array) tribe_get_option(
+				'previous_event_tickets_versions',
+				[]
+			)
+		);
+
```
rafsuntaskin, line 238: Bail out here with '' if `$previous_versions` is empty. Then just return the list on the next. Saves running the `Arr::tolist` on empty values.

tests/integration/_bootstrap.php
```diff
@@ -12,6 +12,7 @@
 // Disconnect Promoter to avoid license-related notices.
 remove_action( 'tribe_tickets_promoter_trigger', [ tribe( Dispatcher::class ), 'trigger' ] );
 
+require_once trailingslashit( EVENT_TICKETS_DIR ) . 'src/template-tags/commerce.php';
```
rafsuntaskin, line 15: This should also be not required.

src/Tickets/Site_Health/Subsections/Plugins/Plugin_Data_Subsection.php
codingmusician:
```suggestion
		if ( empty( $previous_versions ) ) {
			return '';
		}

```

codingmusician (approved):

PR #2937, opened 2024-02-23
[ET-2007] Decode html entites in email subject string

codingmusician:
### 🎫 Ticket

[ET-2007]

### 🗒️ Description
It was reported the Ampersands were being encoded when an event (or post) title included one in its title and when using the `{event_name}` placeholder within the subject line of an email. We found that `wp_kses()` was causing this to occur. We are using `wp_specialchars_decode()` just before sending the email to prevent these characters from being sent out.
### 🎥 Artifacts 

https://www.loom.com/share/e0918416a8a242fd893200d26713d04e

### ✔️ Checklist

[ET-2007]: https://stellarwp.atlassian.net/browse/ET-2007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Emails/Dispatcher.php
+++ b/src/Tickets/Emails/Dispatcher.php
@@ -552,9 +552,12 @@ class Dispatcher {
 			return false;
 		}
 
+		// Handle any encoded characters in the subject.
+		$subject = html_entity_decode( $this->get_subject() );
+
 		$sent = (bool) wp_mail(
 			$this->get_to(),
-			$this->get_subject(),
+			$subject,
 			$this->get_content(),
 			$this->get_headers_formatted(),
 			$this->get_attachments()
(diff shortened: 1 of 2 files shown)


redscar (approved):

src/Tickets/Emails/Dispatcher.php
```diff
@@ -552,9 +553,12 @@ public function send(): bool {
 			return false;
 		}
 
+		// Handle any encoded characters in the subject.
+		$subject = wp_specialchars_decode( $this->get_subject() );
```
bordoni, line 557: I wonder if we should move the escaping to the method itself.

bordoni (changes requested):

src/Tickets/Emails/Dispatcher.php
codingmusician: I had thought about that, but because of the filters in that method, I felt we wanted to decode the string really late. That's when I figured we'd just decode the string right before it's sent. That guarantees that it will be decoded.

bordoni (approved):

PR #2950, opened 2024-03-05
[ET-2018] Classic Editor > ECP upsell showing when it should not

codingmusician:
### 🎫 Ticket
[ET-2018]

### 🗒️ Description
The ECP upsell (for multiple ticket types) should not be showing when ECP is installed and activated. It should also not be showing when editing a non-event. In both cases, the upsell was showing. This PR adds the logic to fix that.

I also added a check to make sure we're in the admin side when editing/adding a ticket. This will prevent the upsell from showing within the Community Tickets interface on the front-end.

### 🎥 Artifacts 
[UPDATED]
https://www.loom.com/share/f105503368504f5592d391deb3a985ac?sid=62acd840-178c-4d09-b9b1-da979f777466

### ✔️ Checklist

[ET-2018]: https://stellarwp.atlassian.net/browse/ET-2018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Admin/Upsell.php
+++ b/src/Tickets/Admin/Upsell.php
@@ -228,11 +228,6 @@ class Upsell {
 	 * @param Template $template Current instance of the Tribe__Template.
 	 */
 	public function render_ticket_type_upsell_notice( string $file, array $name, Template $template ): void {
-		// If they already have ECP activated or are not within the admin area, then bail.
-		if ( class_exists( 'Tribe__Events__Pro__Main', false ) ) {
-			return;
-		}
-
 		// Check if post type is an event.
 		if ( ! function_exists( 'tribe_is_event' ) || ! tribe_is_event() ) {
 			return;
--- a/src/Tickets/Admin/Upsell.php
+++ b/src/Tickets/Admin/Upsell.php
@@ -227,6 +227,16 @@ class Upsell {
 	 * @param Template $template Current instance of the Tribe__Template.
 	 */
 	public function render_ticket_type_upsell_notice( string $file, array $name, Template $template ): void {
+		// If they already have ECP activated or are not within the admin area, then bail.
+		if ( class_exists( 'Tribe__Events__Pro__Main', false ) ) {
+			return;
+		}
+
+		// Check if post type is an event.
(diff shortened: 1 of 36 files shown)


src/Tickets/Admin/Upsell.php
```diff
@@ -227,6 +227,21 @@ public function maybe_show_paystack_promo( $settings ) {
 	 * @param Template $template Current instance of the Tribe__Template.
 	 */
 	public function render_ticket_type_upsell_notice( string $file, array $name, Template $template ): void {
+		// If they already have ECP activated or are not within the admin area, then bail.
```
redscar, line 230: We should probably update the `@since`.

rafsuntaskin (changes requested):
This hook should be unhooked from ECP instead of doing Class_exist checks inside this one.

Or we can use the Flexible Tickets base provider as well as that is only registered when ECP is enabled: https://github.com/the-events-calendar/event-tickets/blob/7927ad86111bc17e3c3fb2c83db53f9abbbd516e/src/Tickets/Flexible_Tickets/Base.php#L95

src/Tickets/Admin/Upsell.php
```diff
@@ -221,12 +221,23 @@ public function maybe_show_paystack_promo( $settings ) {
...
 	public function render_ticket_type_upsell_notice( string $file, array $name, Template $template ): void {
+		// Check if post type is an event.
+		if ( ! function_exists( 'tribe_is_event' ) || ! tribe_is_event() ) {
+			return;
+		}
+
+		// If not within the admin area, then bail.
+		if ( ! is_admin() ) {
+			return;
+		}
```
bordoni, line 239: I would use an action like `did_action( 'action_that_only_happens_in_pro' )`, that tends to be a safer bet.

bordoni (dismissed):

codingmusician:
Added this check, which is an action that runs once within ECP upon `plugins_loaded`:
```did_action( 'tribe_events_pro_init_apm_filters' )```

redscar (approved):

rafsuntaskin (approved):

PR #2964, opened 2024-03-14
[ET-2015] Adjusted the logic for calculating fees when using Stripe.

redscar:
### 🎫 Ticket

[ET-2015]

### 🗒️ Description

In a previous PR (https://github.com/the-events-calendar/event-tickets/pull/2403) we removed certain parts of the `is_licensed_plugin` logic due to its resource-intensive nature, especially noticeable on the checkout page when Stripe is enabled.  This PR reintroduces the logic with the idea of caching the revalidation of `id_current_license_valid`. 

Scenario - 
When $revalidate is true we need to check the most recent license. This happens [here](https://github.com/the-events-calendar/event-tickets/blob/master/src/Tickets/Commerce/Gateways/Stripe/Application_Fee.php#L34). This method is always called when you checkout for Stripe to see if the fee must be added.

I added in additional caching when `$revalidate` is true. The cache will only last an hour and will help alleviate multiple calls to `is_current_license_valid` which is resource intensive.

I originally tackled this using Tribe Cache. However, I found that the value wasn't being kept in the cache and it was blank every page load. Therefore I moved forward using transients instead.

### ✔️ Checklist

[ET-2015]: https://stellarwp.atlassian.net/browse/ET-2015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Settings.php
+++ b/src/Tickets/Commerce/Settings.php
@@ -594,12 +594,18 @@ class Settings {
 		$cache_key = __METHOD__;
 		$cached    = get_transient( $cache_key );
 
-		if ( ! empty( $cached ) ) {
-			return $cached;
+		// Decode the JSON string. If $cached is false (transient not set), json_decode returns null.
+		$cached_value = false !== $cached ? json_decode( $cached, true ) : null;
+
+		// Check explicitly for null to determine if the transient was not set.
+		if ( null !== $cached_value ) {
+			return $cached_value;
 		}
-		// Forcing a revalidate of is_current_license_valid can be expensive. Cache it for 1 hour.
+
 		$is_license_valid = $pue->is_current_license_valid( $revalidate );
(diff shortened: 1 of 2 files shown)


src/Tickets/Commerce/Settings.php
```diff
@@ -567,9 +567,10 @@ public function skip_sanitization( $meta_value ) {
 	/**
 	 * Is a valid license of Event Tickets Plus available?
 	 *
+	 * @since TBD Added caching.
 	 * @since 5.3.0
```
codingmusician, line 571: Switch these lines:
```suggestion
	 * @since 5.3.0
	 * @since TBD Added caching.
```

codingmusician (approved):

PR #2983, opened 2024-04-01
[ET-2062] QR Code creating deprecation notices

redscar:
### 🎫 Ticket

[ET-2062]

### 🗒️ Description

When using PHP 8.1+ deprecation notices are generated when a QR code is generated. The exact issue has to do with the index of the `rsblocks` index being a float instead of an integer. By casting it as an int we fix the deprecation notice and the QR codes continue to work properly.

Before PHP 8.1 it was automatically converted to an int.

### ✔️ Checklist

### DIFF
--- a/vendor/phpqrcode/qrencode.php
+++ b/vendor/phpqrcode/qrencode.php
@@ -139,7 +139,7 @@ namespace TEC\Tickets\phpqrcode;
                 if($col >= $this->rsblocks[0]->dataLength) {
                     $row += $this->b1;
                 }
-				$ret = $this->rsblocks[$row]->data[(int)$col];
+				$ret = $this->rsblocks[(int)$row]->data[(int)$col];
             } else if($this->count < $this->dataLength + $this->eccLength) {
                 $row = ($this->count - $this->dataLength) % $this->blocks;
                 $col = ($this->count - $this->dataLength) / $this->blocks;
--- a/vendor/phpqrcode/qrencode.php
+++ b/vendor/phpqrcode/qrencode.php
@@ -139,7 +139,7 @@ namespace TEC\Tickets\phpqrcode;
                 if($col >= $this->rsblocks[0]->dataLength) {
                     $row += $this->b1;
                 }
-                $ret = $this->rsblocks[$row]->data[$col];
+				$ret = $this->rsblocks[$row]->data[(int)$col];
             } else if($this->count < $this->dataLength + $this->eccLength) {
                 $row = ($this->count - $this->dataLength) % $this->blocks;
                 $col = ($this->count - $this->dataLength) / $this->blocks;


codingmusician (approved):

vendor/phpqrcode/qrencode.php
```diff
@@ -133,20 +133,22 @@ public function getCode()
...
+				$col = $this->count / $this->blocks;
+				if ( $col >= $this->rsblocks[0]->dataLength ) {
+					$row += $this->b1;
+				}
+				$ret = $this->rsblocks[ (int) $row ]->data[ (int) $col ];
+			} else {
+				if ( $this->count < $this->dataLength + $this->eccLength ) {
+					$row = ( $this->count - $this->dataLength ) % $this->blocks;
+					$col = ( $this->count - $this->dataLength ) / $this->blocks;
+					$ret = $this->rsblocks[ $row ]->ecc[ $col ];
```
rafsuntaskin, line 147: should we also cast this one?

vendor/phpqrcode/qrencode.php
```diff
@@ -133,20 +133,22 @@ public function getCode()
...
-            } else if($this->count < $this->dataLength + $this->eccLength) {
-                $row = ($this->count - $this->dataLength) % $this->blocks;
-                $col = ($this->count - $this->dataLength) / $this->blocks;
-                $ret = $this->rsblocks[$row]->ecc[$col];
-            } else {
-                return 0;
-            }
+			if ( $this->count < $this->dataLength ) {
+				$row = $this->count % $this->blocks;
+				$col = $this->count / $this->blocks;
```
rafsuntaskin, line 138:
```suggestion
				$row = (int) $this->count % $this->blocks;
				$col = (int) $this->count / $this->blocks;
```
Why not cast here?

rafsuntaskin (changes requested):

vendor/phpqrcode/qrencode.php
redscar: I do agree we should cast on line 147.

vendor/phpqrcode/qrencode.php
redscar: Is this cast necessary if we are casting them in the indexes? Especially since there is additional logic below?

rafsuntaskin: I just didn't like the casting inside the indexes, so wanted to have them earlier. Although, it feels same as row and columns should be int always but let's proceed with this one if only you can verify them with your own tests.

Otherwise, just cast them in a separate line above before using them inside the index. That feels more clear.

vendor/phpqrcode/qrencode.php
rafsuntaskin: Maybe cast them before the indexes.

PR #3027, opened 2024-04-26
[ET-2089] Free Tickets: Update emails to handle free orders

codingmusician:
### 🎫 Ticket
[ET-2089]

### 🗒️ Description
Making some tweaks to a couple of emails.
- Completed Order - Remove the Gateway Order #, since it's the same as the main order #. Remove the "Payment completed with..." text when the Gateway label is blank.
- Purchase Receipt - Remove the "Payment completed with..." text when the Gateway label is blank.

### 🎥 Artifacts 
Before: Note the redundancy of order ID and Gateway order ID, which are the same:
image

After:
image

### ✔️ Checklist

[ET-2089]: https://stellarwp.atlassian.net/browse/ET-2089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/views/emails/template-parts/body/order/order-gateway-data.php
+++ b/src/views/emails/template-parts/body/order/order-gateway-data.php
@@ -12,6 +12,7 @@
  * @version 5.6.0
  *
  * @since 5.6.0
+ * @since TBD Don't show if gateway order number is same as regular order number.
  *
  * @var Tribe__Template                    $this               Current template object.
  * @var \TEC\Tickets\Emails\Email_Abstract $email              The email object.
--- a/src/views/emails/template-parts/body/order/order-gateway-data.php
+++ b/src/views/emails/template-parts/body/order/order-gateway-data.php
@@ -28,6 +28,11 @@ if ( empty( $order->gateway_order_id )  ) {
 	return;
 }
 
+// No need to show gateway ID if it's the same as the order ID.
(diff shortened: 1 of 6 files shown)


src/views/emails/template-parts/body/order/order-gateway-data.php
```diff
@@ -28,6 +29,11 @@
 	return;
 }
 
+// No need to show gateway ID if it's the same as the order ID.
+if ( is_numeric( $order->gateway_order_id ) && intval( $order->ID ) === intval( $order->gateway_order_id ) ) {
+	return;
+}
```
rafsuntaskin, line 35: This logic is a bit more complex than necessary, there could be other gateways where the `gateway_order_Id` could be same as `Order_id`, let's just skip this for free orders depending on the `total_value` as well.

rafsuntaskin (changes requested):

src/views/emails/template-parts/body/order/order-gateway-data.php
codingmusician: In my opinion, if the gateway order is the same as the order id (whether it's free or not), then this information is redundant and shouldn't be shown. What is the reason we're not showing the Gateway ID for free orders? An ID exists, right? If the reasoning is that we don't need to show it due to the order ID being shown elsewhere in the email template, then this logic should suffice.

rafsuntaskin (approved):

PR #3029, opened 2024-05-01
Initial SLR move

lucatume:
WiP

### DIFF
--- a/src/Tickets/Seating/Controller.php
+++ b/src/Tickets/Seating/Controller.php
@@ -1,0 +1,190 @@
+<?php
+/**
+ * The main Event Assigned Controller plugin controllers, it bootstraps the ancillary controllers and binds the main
+ * definitions.
+ *
+ * @since   1.0.0
+ *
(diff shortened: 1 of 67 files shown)


src/Tickets/Seating/Controller.php
```diff
@@ -0,0 +1,190 @@
...
+	 * Registers some common assets that will be used in Admin, Frontend and Blocks context.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	private function register_common_assets(): void {
+		Asset::add(
+			'tec-events-assigned-seating-vendor',
+			'vendor.js',
```
borkweb, line 126: I think this should be `app/vendor.js`. I just downloaded ET and the vendor file lives in `common/src/resources/js/app/vendor.min.js`

lucatume: That part of the code is not connected yet, so it's still very much a copy and paste.
I will wire it correctly in next iterations.

PR #3032, opened 2024-05-02
Add variables to docblock

andrasguseo:
Variables were missing from the docblock. This likely influenced that our documentation on this filter was incompolete. https://www.dropbox.com/scl/fi/qqgf8qvjetg8inw1bs18b/shot_240502_160124.jpg?rlkey=ctce8mev4d8bp56f75kgxxiu6&dl=0

### DIFF
--- a/src/Tribe/Attendees.php
+++ b/src/Tribe/Attendees.php
@@ -647,7 +647,9 @@ class Tribe__Tickets__Attendees {
 
 		/**
 		 * Used to modify what columns should be shown on the CSV export.
-		 * The column name should be the Array Index and the Header is the array Value.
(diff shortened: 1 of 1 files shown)


bordoni (changes requested):
Add `@since` here if we are making changes.

andrasguseo:
Added `@since TBD`.
I'm not sure when this filter was introduced though, and adding `@since {upcoming version}` doesn't feel the best, but surely better than nothing.
I'd love to learn how to find when a piece of code was added.

bordoni (approved):

Camwyn:
Y'all merged a TBD into master?

PR #3044, opened 2024-05-14
Adds support for WPML

dgwatkins:
The idea is to add the ids for the translation when querying the database. It is done in a way that we can reuse the code for WooCommerce, EDD & built-in tickets.

This is more a proof of concept to get the ball going so I didn't update the changelog and I don't have a ticket id to fill in the template.


dgwatkins:
I updated the pull request but its no longer working.
The reasons are 2:
- The integrations manager class is deprecated, I don't think its loading the integration code
- The event_id that we intercept used to be an integer value, now its an array. The code needs some minor adjustments.

dgwatkins:
This was a proof of concept that no longer works, closing.

PR #3051, opened 2024-05-28
[SL-89] Feat - Layout cards display

rafsuntaskin:
### 🎫 Ticket

[SL-89]

### 🗒️ Description

* Display the Layout cards inside the Seating Layout Tab.

### 🎥 Artifacts 
📷 screenshot(s): https://share.cleanshot.com/XNyk5xW7

### ✔️ Checklist

[SL-89]: https://stellarwp.atlassian.net/browse/SL-89?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Seating/Admin/Tabs/Layout_Card.php
+++ b/src/Tickets/Seating/Admin/Tabs/Layout_Card.php
@@ -1,0 +1,138 @@
+<?php
+/**
+ * The Layout card object class.
+ *
+ * @since TBD
+ *
+ * @package TEC\Tickets\Seating\Admin\Tabs;
+ */
+
+namespace TEC\Tickets\Seating\Admin\Tabs;
+
+/**
+ * The Layout_Card Class.
+ *
+ * @since TBD
+ *
+ * @package TEC\Tickets\Seating\Admin\Tabs;
+ */
+class Layout_Card {
+	/**
+	 * The Layout ID.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	protected string $id;
+		
+	/**
+	 * The Layout name.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	protected string $name;
+	
+	/**
+	 * The Layout map ID.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	protected string $map;
+		
+	/**
+	 * The number of seats in the Layout.
+	 *
+	 * @since TBD
+	 *
+	 * @var int
+	 */
+	protected int $seats;
+		
+	/**
+	 * The URL to the Layout's screenshot.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	protected string $screenshot_url;
+		
+	/**
+	 * Layout_Card constructor.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $id The Layout ID.
+	 * @param string $name The Layout name.
+	 * @param string $map The Layout map ID.
+	 * @param int    $seats The number of seats in the Layout.
+	 * @param string $screen_shot_url The URL to the Layout's screenshot.
+	 */
+	public function __construct( string $id, string $name, string $map, int $seats, string $screen_shot_url ) {
+		$this->id             = $id;
+		$this->name           = $name;
+		$this->map            = $map;
+		$this->seats          = $seats;
+		$this->screenshot_url = $screen_shot_url;
+	}
+		
+	/**
+	 * Returns the Layout ID.
+	 *
+	 * @since TBD
+	 *
+	 * @return string The Layout ID.
+	 */
+	public function get_id(): string {
+		return $this->id;
+	}
+		
+	/**
+	 * Returns the Layout name.
+	 *
+	 * @since TBD
+	 *
+	 * @return string The Layout name.
+	 */
+	public function get_name(): string {
+		return $this->name;
+	}
+	
+	/**
+	 * Returns the Layout map ID.
+	 *
+	 * @since TBD
+	 *
+	 * @return string The Layout map.
+	 */
+	public function get_map(): string {
+		return $this->map;
+	}
+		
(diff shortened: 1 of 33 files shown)


tests/slr_integration/Admin/Maps_Layout_Homepage_Test.php
```diff
@@ -67,4 +68,43 @@ public function test_maps_tab_card_listing() {
 
 		$this->assertMatchesHtmlSnapshot( $html );
 	}
+	
+	public function test_layouts_tab_card_listing() {
```
lucatume, line 72: Add a test for empty (0 cards), 1 card.

src/Tickets/Seating/app/admin/layouts/style.pcss
```diff
@@ -1 +1,61 @@
-@import "./../../postcss/_tabs.pcss";
\ No newline at end of file
+@import "./../../postcss/_tabs.pcss";
+
+.tec-tickets__tab-content__empty {
```
lucatume, line 3: Here and elsewhere: if the classes do not come from existing code, try to avoid using BEM with more than one block separator (`__`).

E.g. rename this class to `.tec-tickets-seating__tab--empty`.
Avoid reproducing the `Very__Long__Class__Names__From__Previous__Versions__Of__The__Plugins`.

src/Tickets/Seating/Tables/Layouts.php
```diff
@@ -87,6 +87,7 @@ protected function get_definition() {
 				`created_date` datetime NOT NULL,
 				`map` varchar(36) NOT NULL,
 				`seats` int(11) NOT NULL DEFAULT '0',
+				`screenshot_url` varchar(255) NOT NULL,
```
lucatume, line 90: Set a default of empty string?
The value might legitimately be an empty string, let's account for that.

src/Tickets/Seating/Service/Layouts.php
```diff
@@ -125,7 +128,45 @@ public function get_in_option_format() {
...
+				$layout_cards[] = new Layout_Card(
+					$row->id,
+					$row->name,
+					$row->map,
+					$row->seats,
+					$row->screenshot_url
+				);
+			}
+			
+			wp_cache_set(
```
lucatume, line 159: Let's avoid this caching.

You're just moving data from one database, to another.
The query is simple enough and fast enough to have equivalent times, the data handling is minor and not involving complicate computations that you would save with a caching.

Instead of caching this value, memoize it.

lucatume (changes requested):

lucatume (changes requested):
Good work.

The only thing that remains is testing both Maps and Layouts tabs with only 1 card.

src/Tickets/Seating/Service/Layouts.php
```diff
@@ -125,7 +128,40 @@ public function get_in_option_format() {
...
+				$layout_cards[] = new Layout_Card(
+					$row->id,
+					$row->name,
+					$row->map,
+					$row->seats,
+					$row->screenshot_url
+				);
+			}
+			
+			tribe_set_var( $mem_key, $layout_cards );
```
lucatume, line 159: To memoize use `tribe_cahe()[$mem_key]`. That will go into non-persistent caching in WP cache layer and will be naturally flushed with a `wp_cache_flush` call.
The same is not true for `tribe_set_var()`; the var will be set for the whole request (or a whole test process run)

lucatume (changes requested):

src/Tickets/Seating/Admin/Tabs/Layout_Card.php
```diff
@@ -0,0 +1,138 @@
...
+	 * @param int    $seats The number of seats in the Layout.
+	 * @param string $screen_shot_url The URL to the Layout's screenshot.
+	 */
+	public function __construct( string $id, string $name, string $map, int $seats, string $screen_shot_url ) {
+		$this->id             = $id;
+		$this->name           = $name;
+		$this->map            = $map;
+		$this->seats          = $seats;
+		$this->screenshot_url = $screen_shot_url;
+	}
```
bordoni, line 82: I know is a bit silly, but would love to see a factory method, with the screenshot url being a setter method since it shouldn't be required, I wonder is the screenshot URL an external link or an image that we upload?

src/Tickets/Seating/Service/Layouts.php
```diff
@@ -125,7 +128,41 @@ public function get_in_option_format() {
...
+			$layout_cards = [];
+			foreach ( Layouts_Table::fetch_all() as $row ) {
+				$layout_cards[] = new Layout_Card(
+					$row->id,
+					$row->name,
+					$row->map,
+					$row->seats,
+					$row->screenshot_url
+				);
+			}
```
bordoni, line 158: If we have it as a factory style method we could do something like:
```
$cache[ $mem_key ] = array_map( [ Layout_Card::class, 'from_row' ], Layouts_Table::fetch_all() );
```

bordoni (changes requested):

src/Tickets/Seating/Service/Layouts.php
rafsuntaskin: The fetch method from table is a generator so we can't feed that directly to [`array_map`](https://share.cleanshot.com/69kW7XhR).

For now, we should settle with this structure and later I will try to use the generator method in future PRs and try to simplify this one as possible.

lucatume: We're using a Generator to limit the number of stuff pulled from the db and the number of items to keep in memory.
The factory method would require, under the hood, an unroll of the Generator with an `iterator_to_array` call that would still pull items in batches (good), but would store all of them in memory (bad).

lucatume (approved):

PR #3058, opened 2024-06-11
Order Management: Add Orders Submenu

codingmusician:
### 🎫 Ticket
[ET-2124]

### 🗒️ Description
The minimum amount needed to add "Orders" to the ticket menu. Note: Still a work in progress, because I think we want to hide this if TC isn't enabled.

### 🎥 Artifacts 
image

### ✔️ Checklist

[ET-2124]: https://stellarwp.atlassian.net/browse/ET-2124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Admin/Orders_Page.php
+++ b/src/Tickets/Commerce/Admin/Orders_Page.php
@@ -19,10 +19,6 @@ use Tribe\Admin\Pages;
  *
  * @package TEC\Tickets\Commerce\Admin
  */
-
-/**
- * Manages the admin settings UI in relation to ticket configuration.
- */
 class Orders_Page {
 
 	/**
--- a/src/Tickets/Commerce/Admin/Orders_Page.php
+++ b/src/Tickets/Commerce/Admin/Orders_Page.php
@@ -44,16 +44,14 @@ class Orders_Page {
 	 * @since TBD
 	 */
 	public function add_orders_page() {
-		$admin_pages = tribe( 'admin.pages' );
-
-		$admin_pages->register_page(
-			[
-				'id'       => static::$slug,
-				'path'     => 'edit.php?post_type=' . Order::POSTTYPE,
-				'parent'   => static::$parent_slug,
-				'title'    => esc_html__( 'Orders', 'event-tickets' ),
-				'position' => 1.7,
-			]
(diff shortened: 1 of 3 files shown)


src/Tickets/Commerce/Admin/Orders_Page.php
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Class Orders_Page.
+ *
+ * @since   TBD
+ *
+ * @package TEC\Tickets\Commerce\Admin
+ */
+
+/**
+ * Manages the admin settings UI in relation to ticket configuration.
```
redscar, line 24: Does this need to be updated?

codingmusician: Good call. That doesn't even need to be there.

redscar (approved):

src/Tickets/Commerce/Admin/Orders_Page.php
```diff
@@ -0,0 +1,47 @@
...
+			static::$parent_slug,
+			esc_html__( 'Orders', 'event-tickets' ),
+			esc_html__( 'Orders', 'event-tickets' ),
+			Pages::get_capability(),
+			'edit.php?post_type=' . Order::POSTTYPE,
+			'',
+			1.7
+		);
+	}
+}
```
bordoni, line 47: This needs to be registered as a singleton, otherwise it will spawn a new instance every time, I would like to avoid the public static prop, would love to go all non-static methods and protected props to start.

A couple of other things:
1. Create a method for all 7 items used by `add_orders_page()`.
1. Lets make sure we have tests as we start here.
1. Handle the `edit.php` url using `add_query_arg` not string concat.
1. Make sure you save the return of the `add_submenu_page` to a prop on this class, so we can re-use, this particular class will need to have a bunch o bell an whistles before we are over here, since quite a bite o filtering will be needed.

codingmusician: I'll update this. Thank you!

PR #3079, opened 2024-06-24
Feat/data range filter

dpanta94:
### 🎫 Ticket

[ET-2122]

### 🗒️ Description

Adds Date Range filters in Orders Table admin screen.

### 🎥 Artifacts 
image

### ✔️ Checklist

[ET-2122]: https://stellarwp.atlassian.net/browse/ET-2122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Commerce/Admin_Tables/Orders_Table.php
```diff
@@ -438,6 +442,35 @@ public function column_total( $item ) {
 		return $item->total_value->get_currency();
 	}
 
+	/**
+	 * Handler for the post parent column
```
redscar, line 446:
```suggestion
	 * Handler for the post parent column.
```

src/Tickets/Commerce/Admin_Tables/Orders_Table.php
```diff
@@ -615,6 +650,65 @@ protected function extra_tablenav( $which ) {
...
+		 *
+		 * @param bool   $disable   Whether to disable the drop-down. Default false.
+		 * @param string $post_type The post type.
+		 */
+		if ( apply_filters( 'tec_tc_orders_disable_date_range_dropdown', false, $post_type ) ) {
+			return;
+		}
+
+		$date_from = sanitize_text_field( $_GET['tec_tc_date_range_from'] ?? '' ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$date_to   = sanitize_text_field( $_GET['tec_tc_date_range_to'] ?? '' ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
bordoni, line 676: Use `tribe_get_request_var()` for both.

src/Tickets/Commerce/Admin_Tables/Orders_Table.php
```diff
@@ -698,4 +794,68 @@ protected function gateways_dropdown( $post_type ) {
...
+		 * @param string $post_type The post type.
+		 */
+		if ( apply_filters( 'tec_tc_orders_disable_post_parent_dropdown', false, $post_type ) ) {
+			return;
+		}
+
+		// Event options are being filtered in the Frontend after the user starts typing in the search box.
+		// Except for when the user has already filtered by an event. We take the event ID from the URL and add it to the dropdown.
+
+		$e = absint( $_GET['tec_tc_events'] ?? 0 ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
```
bordoni, line 823: `tribe_get_request_var()`

src/Tickets/Commerce/Admin_Tables/Orders_Table.php
```diff
@@ -698,4 +794,68 @@ protected function gateways_dropdown( $post_type ) {
...
+		add_filter( 'tribe_field_div_start', '__return_empty_string' );
+		add_filter( 'tribe_field_div_end', '__return_empty_string' );
+		?>
+		<label for="tec_tc_events-select" class="screen-reader-text"><?php esc_html_e( 'Filter By Event', 'event-tickets' ); ?></label>
+		<?php
+		new Tribe__Field( 'tec_tc_events', $field, $e );
+		remove_filter( 'tribe_field_start', '__return_empty_string', 10 );
+		remove_filter( 'tribe_field_end', '__return_empty_string', 10 );
+		remove_filter( 'tribe_field_div_start', '__return_empty_string', 10 );
+		remove_filter( 'tribe_field_div_end', '__return_empty_string', 10 );
```
bordoni, line 859: I would just print the field manually, using `Tribe__Field` creates more problems than it solves honestly here.

redscar (commented):
Look's good. One small change.

src/Tickets/Commerce/Hooks.php
```diff
@@ -143,6 +145,54 @@ protected function add_filters() {
...
+
+		if ( empty( $query->posts ) ) {
+			return $results;
+		}
+
+		$results = array_map(
+			function ( $r ) {
+				return [
+					'id'   => $r->ID,
+					'text' => apply_filters( 'the_title', $r->post_title ),
```
bordoni, line 189: Why not use `get_the_title()`? It handles all the things we need.

src/Tickets/Commerce/Admin_Tables/Orders_Table.php
```diff
@@ -698,4 +794,68 @@ protected function gateways_dropdown( $post_type ) {
...
+
+		$event = $e ? get_post( $e ) : null;
+
+		$event = $event instanceof WP_Post ? $event : null;
+
+		$events_formatted = [
+			'' => esc_html__( 'All Events', 'event-tickets' ),
+		];
+
+		$events_formatted += $event ? [ (string) $event->ID => apply_filters( 'the_title', $event->post_title ) ] : [];
```
bordoni, line 833: Why not use `get_the_title()`? It handles all the things we need.

src/Tickets/Commerce/Hooks.php
```diff
@@ -202,6 +293,16 @@ public function pre_filter_admin_order_table( $query ) {
 			];
 		}
 
+		$event_filter = absint( $_GET['tec_tc_events'] ?? 0 ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
bordoni, line 296: `tribe_get_request_var()`

bordoni (changes requested):

src/Tickets/Commerce/Hooks.php
dpanta94: I am not a 100% if get_the_title would utilize already queried and retrieved object. To avoid that (and a potential double query) i applied the filters in the existing value.

JPry: If you pass the post ID to `get_the_title()`, then you don't have to worry about it using the correct object.

> To avoid that (and a potential double query)

Within a single request, all queries to post objects are cached. You can see this in `\WP_Post::get_instance()`.

src/Tickets/Commerce/Hooks.php
```diff
@@ -143,6 +145,54 @@ protected function add_filters() {
...
+		];
+
+		$query = new WP_Query( $args );
+
+		if ( empty( $query->posts ) ) {
+			return $results;
+		}
+
+		$results = array_map(
+			function ( $r ) {
```
JPry, line 186: This is a throwaway anonymous function, but it would still be better to use a variable name that makes it clear what the data is. Even more ideal would be to type-hint `WP_Post` as the parameter type.

src/Tickets/Commerce/Hooks.php
```diff
@@ -211,6 +312,31 @@ public function pre_filter_admin_order_table( $query ) {
...
+		}
+
+		$date_from_ts = strtotime( $date_from );
+		$date_to_ts   = strtotime( $date_to );
+
+		if ( $date_to_ts >= $date_from_ts ) {
+			return true;
+		}
+
+		return false;
```
JPry, line 337: At this point, we could simplify this to a single line:

```suggestion
		return $date_to_ts >= $date_from_ts;
```

src/Tickets/Commerce/Hooks.php
dpanta94: Thanks @JPry ! I definitely can pass the post id. In this case i had some doubts since im specifally telling wp to avoid updating meta and term cache since i wasn't going to use them. I guess it has no reason not to cache what has been retrieved so it shouldn't perform a double query for what is already retrieved

src/Tickets/Commerce/Hooks.php
```diff
@@ -143,6 +146,54 @@ protected function add_filters() {
...
+			function ( WP_Post $result ) {
+				return [
+					'id'   => $result->ID,
+					'text' => get_the_title( $result->ID ),
+				];
+			},
+			$query->posts
+		);
+
+		return [ 'results' => $results ];
```
JPry, line 196: I'm noticing now that the structure of the array is different here compared to line 183 above. Should they both be the same structure, even if the results are empty?

dpanta94: No it doesn't matter, when there are results though the Tribe__Ajax__Dropdown::route method expects them in this format.

src/Tickets/Commerce/Admin_Tables/Orders_Table.php
```diff
@@ -676,26 +769,64 @@ protected function gateways_dropdown( $post_type ) {
...
 
-		add_filter( 'tribe_field_start', '__return_empty_string' );
-		add_filter( 'tribe_field_end', '__return_empty_string' );
-		add_filter( 'tribe_field_div_start', '__return_empty_string' );
-		add_filter( 'tribe_field_div_end', '__return_empty_string' );
+		$events_formatted += $event ? [ (string) $event->ID => get_the_title( $event->ID ) ] : [];
 		?>
-		<label for="tec_tc_gateway-select" class="screen-reader-text"><?php esc_html_e( 'Filter By Gateway', 'event-tickets' ); ?></label>
+		<label for="tec_tc_events-select" class="screen-reader-text"><?php esc_html_e( 'Filter By Event', 'event-tickets' ); ?></label>
+		<select name="tec_tc_events" id='tec_tc_events-select' class='tribe-dropdown' data-freeform="1" data-force-search="1" data-searching-placeholder="<?php esc_attr_e( 'Searching...', 'event-tickets' ); ?>" data-source="tec_tc_order_table_events" data-source-nonce="<?php echo esc_attr( wp_create_nonce( 'tribe_dropdown' ) ); ?>">
```
bordoni, line 825: I was going to ask, give each attr a new line for readability.

bordoni (changes requested):

src/Tickets/Commerce/Admin_Tables/Orders_Table.php
```diff
@@ -676,26 +769,64 @@ protected function gateways_dropdown( $post_type ) {
...
+
+		$events_formatted = [
+			'' => esc_html__( 'All Events', 'event-tickets' ),
 		];
 
-		add_filter( 'tribe_field_start', '__return_empty_string' );
-		add_filter( 'tribe_field_end', '__return_empty_string' );
-		add_filter( 'tribe_field_div_start', '__return_empty_string' );
-		add_filter( 'tribe_field_div_end', '__return_empty_string' );
+		$events_formatted += $event ? [ (string) $event->ID => get_the_title( $event->ID ) ] : [];
```
rafsuntaskin, line 822: Let's skip the one line for a if($event) check to add the data to the array Otherwise, it will run the [loop](https://github.com/the-events-calendar/event-tickets/pull/3079/files#diff-cd5f0a73f32dd0d696f1cf3cb5b89c236eae6675e25a064f77150f1c6582895dR826) for the `[]` if that ever gets added,

src/Tickets/Commerce/Hooks.php
```diff
@@ -143,6 +146,54 @@ protected function add_filters() {
...
+		add_filter( 'tribe_dropdown_tec_tc_order_table_events', [ $this, 'provide_events_results_to_ajax' ], 10, 2 );
+	}
+
+	/**
+	 * Provides the results for the events dropdown in the Orders table.
+	 *
+	 * @since TBD
+	 *
+	 * @param array $results The results.
+	 * @param array $search The search.
```
rafsuntaskin, line 159:
```suggestion
	 * @param array<string,mixed> $results The results.
	 * @param array<string,string> $search The search.
```

src/Tickets/Commerce/Hooks.php
```diff
@@ -143,6 +146,54 @@ protected function add_filters() {
...
+	 *
+	 * @return array
```
rafsuntaskin, line 161:
```suggestion
	 * @return array<string,mixed>
```

src/Tickets/Commerce/Hooks.php
```diff
@@ -185,14 +236,55 @@ public function pre_filter_admin_order_table( $query ) {
 			}
 		}
 
+		$date_from = sanitize_text_field( tribe_get_request_var( 'tec_tc_date_range_from', '' ) );
+		$date_to   = sanitize_text_field( tribe_get_request_var( 'tec_tc_date_range_to', '' ) );
```
rafsuntaskin, line 240: `tribe_get_request_var` should already sanitize the values.

src/Tickets/Commerce/Hooks.php
```diff
@@ -185,14 +236,55 @@ public function pre_filter_admin_order_table( $query ) {
...
+
+		$date_from = Tribe__Date_Utils::is_valid_date( $date_from ) ? $date_from : '';
+		$date_to   = Tribe__Date_Utils::is_valid_date( $date_to ) ? $date_to : '';
+
+		$date_range_valid = $this->is_valid_date_range( $date_from, $date_to );
+
+		if ( ! $date_range_valid ) {
+			// If invalid, adjust the to date to be the same as the from date.
+			// @todo show a message.
```
rafsuntaskin, line 249: Is this in progress?

src/Tickets/Commerce/Admin_Tables/Orders_Table.php
dpanta94: No it doesn't. Adding an empty array to an already existing array results to the original array.

```php
<?php

$test = [ '1' => 2 ];

$test += [];

var_dump( $test );
```

```php
array(1) {
  [1]=>
  int(2)
}
```

rafsuntaskin (changes requested):

src/Tickets/Commerce/Hooks.php
dpanta94: It does a sanitization of removing html special chars and removing tags. `sanitize_text_field` covers more stuff so id like to leave it there.

rafsuntaskin (approved):

PR #3095, opened 2024-07-08
Fix stripe refunds

dpanta94:
### 🎫 Ticket

[ET-2142]

### 🗒️ Description

We were facing 2 issues with Stripe refunds.

1. Refunds would never reflect in WP. Reason is that in part of fixing ET-1792 we introduced a new bug. In ET-1792 before processing an incoming webhook we would check if there was already a successful payment intent present in Stripe. Overcoming that way Race conditions. BUT webhooks for refunds always did have a successful payment intent and we were wrongly ignoring them. We are overcoming this one by also checking against the type of incoming webhook.

2. Multiple refunds from stripe would fail to be registered in WP. Reason was that the code was failing early because it could not update the order status from `refunded` to `refunded`. Fixing this by allowing explicitly this status update.

Pending/possible second PR
- Check the same bug for PayPal and if it's present fix it.

### ✔️ Checklist

[ET-2142]: https://stellarwp.atlassian.net/browse/ET-2142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "event-tickets",
-  "version": "5.9.1",
+  "version": "5.11.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "event-tickets",
-      "version": "5.9.1",
+      "version": "5.11.0",
       "hasInstallScript": true,
       "dependencies": {
         "@babel/runtime": "^7.15.3",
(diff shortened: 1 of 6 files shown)


Camwyn:
I'm taking this ticket over for Dimi, so ping me if there is an issue/question - not Dimi 😉

rafsuntaskin:
@Camwyn tests are pending on this one.

Camwyn:
Added some new tests by duplicating and modifying the Order tests at `tests/commerce_integration/TEC/Gateways/Stripe/OrderTest.php`

bordoni (approved):

PR #3101, opened 2024-07-11
SLR - fix session persistence

lucatume:
In this PR:
- **fix(Seating) print the new session token in all the timer instances on initial render**
- **fix(Seating) cs fixes**
- **refactor(Seating) on interrupt, remove token session from DB**

Follow-up on the work done for SL-92 to fix some issues with session persistence and window/tab re-use: going back to the same browser window to start a new seat-selection session would create inconsistent session data, this fixes it.

Also: on interrupt (i.e. when the timer runs out) remove the token session from the db instead of just emptying it.

[Demo](https://share.cleanshot.com/g5ZTY13C)

### DIFF
--- a/src/Tickets/Seating/Admin/Ajax.php
+++ b/src/Tickets/Seating/Admin/Ajax.php
@@ -396,8 +396,9 @@ class Ajax extends Controller_Contract {
 		}
 
 		// Remove the `tec-tickets-commerce-cart` cookie.
+		$cookie_name = Cart::$cart_hash_cookie_name;
 		setcookie(
-			Cart::$cart_hash_cookie_name,
+			$cookie_name,
 			'',
 			time() - DAY_IN_SECONDS,
 			COOKIEPATH,
@@ -405,5 +406,6 @@ class Ajax extends Controller_Contract {
 			true,
 			true
 		);
+		unset( $_COOKIE[ $cookie_name ] );
 	}
 }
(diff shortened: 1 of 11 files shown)


src/Tickets/Seating/Admin/Ajax.php
```diff
@@ -396,14 +396,16 @@ public function clear_commerce_cart_cookie( int $post_id ): void {
 		}
 
 		// Remove the `tec-tickets-commerce-cart` cookie.
+		$cookie_name = Cart::$cart_hash_cookie_name;
 		setcookie(
```
rafsuntaskin, line 400:
```suggestion
                 // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.cookies_setcookie
		setcookie(
```

src/Tickets/Seating/Admin/Ajax.php
```diff
@@ -396,14 +396,16 @@ public function clear_commerce_cart_cookie( int $post_id ): void {
...
-			Cart::$cart_hash_cookie_name,
+			$cookie_name,
 			'',
 			time() - DAY_IN_SECONDS,
 			COOKIEPATH,
 			COOKIE_DOMAIN,
 			true,
 			true
 		);
+		unset( $_COOKIE[ $cookie_name ] );
```
rafsuntaskin, line 409:
```suggestion
		// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE
		unset( $_COOKIE[ $cookie_name ] );
```

rafsuntaskin (approved):

PR #3117, opened 2024-07-19
Fix Attendees CSV Export

pattihis:
### 🎫 Ticket

[ET-2094]

### 🗒️ Description

In our previous CSV export functions we expected an associated `event_id`, that would exist as a query parameter. In our main attendees table we are now exporting all of them instead ( `event_id=all` ).

### ✔️ Checklist


src/Tribe/Attendees.php
```diff
@@ -820,6 +825,47 @@ public function maybe_generate_csv() {
...
+
+		// Sanitize items for CSV usage.
+		$items = $this->sanitize_csv_rows( $items );
+
+		if ( ! empty( $items ) ) {
+			$charset  = get_option( 'blog_charset' );
+			$filename = sanitize_file_name( __( 'All event attendees', 'event-tickets' ) );
+
+			// Output headers so that the file is downloaded rather than displayed.
+			header( "Content-Type: text/csv; charset=$charset" );
```
JPry, line 851: While not strictly necessary, I find that including brackets around interpolated variables helps with readability.

```suggestion
			header( "Content-Type: text/csv; charset={$charset}" );
```

src/Tribe/Attendees.php
```diff
@@ -820,6 +825,47 @@ public function maybe_generate_csv() {
...
+			header( "Content-Disposition: attachment; filename=$filename.csv" );
```
JPry, line 852:
```suggestion
			header( "Content-Disposition: attachment; filename={$filename}.csv" );
```

src/Tribe/Attendees.php
```diff
@@ -820,6 +825,47 @@ public function maybe_generate_csv() {
...
+			return;
+		}
+
+		// Generate filtered list of attendees.
+		$items = $this->generate_filtered_list( 'all' );
+
+		// Sanitize items for CSV usage.
+		$items = $this->sanitize_csv_rows( $items );
+
+		if ( ! empty( $items ) ) {
```
JPry, line 846: It would be more readable to invert this conditional and then return early, instead of keeping all the subsequent logic within this if statement.

src/Tribe/Attendees.php
```diff
@@ -820,6 +825,47 @@ public function maybe_generate_csv() {
 		}
 	}
 
+	/**
+	 * Admin requested a CSV export from the attendees table.
+	 * Generates the download and finishes the execution.
+	 *
+	 * @since TBD
+	 *
+	 */
```
JPry, line 834:
```suggestion
	 */
```

src/Tribe/Attendees.php
```diff
@@ -820,6 +825,47 @@ public function maybe_generate_csv() {
 		}
 	}
 
+	/**
+	 * Admin requested a CSV export from the attendees table.
+	 * Generates the download and finishes the execution.
```
JPry, line 830: I think this could be tweaked to better describe the function itself.

```suggestion
	 * Export a CSV containing all attendees.
	 *
	 * Called when an Admin requested a CSV export from the attendees table. Generates
	 * the download file and finishes the execution.
```

src/Tribe/Attendees.php
```diff
@@ -575,7 +579,7 @@ private function generate_filtered_list( $event_id ) {
 		$filter_name = "manage_{$this->page_id}_columns";
 		add_filter( $filter_name, [ $this->attendees_table, 'get_columns' ], 15 );
 
-		$items = Tribe__Tickets__Tickets::get_event_attendees( $event_id );
+		$items = 'all' === $event_id ? Tribe__Tickets__Tickets::get_attendees_by_args()['attendees'] : Tribe__Tickets__Tickets::get_event_attendees( $event_id );
```
JPry, line 582: Even though the use of `Tribe__Tickets__Tickets` isn't new here, it might be a good time to use an alias with the class import so that it is easier to read.

JPry (changes requested):

src/Tribe/Attendees.php
```diff
@@ -820,6 +825,47 @@ public function maybe_generate_csv() {
...
+			$filename = sanitize_file_name( __( 'All event attendees', 'event-tickets' ) );
+
+			// Output headers so that the file is downloaded rather than displayed.
+			header( "Content-Type: text/csv; charset=$charset" );
+			header( "Content-Disposition: attachment; filename=$filename.csv" );
+
+			// Create the file pointer connected to the output stream.
+			$output = fopen( 'php://output', 'w' );
+
+			$delimiter = apply_filters( 'tribe_tickets_attendees_csv_export_delimiter', ',' );
```
bordoni, line 857: Docblock here.

src/Tribe/Attendees.php
```diff
@@ -820,6 +825,47 @@ public function maybe_generate_csv() {
...
+	 * Admin requested a CSV export from the attendees table.
+	 * Generates the download and finishes the execution.
+	 *
+	 * @since TBD
+	 *
+	 */
+	public function export_csv_for_all() {
+		if ( empty( $_GET['attendees_csv'] ) || empty( $_GET['attendees_csv_nonce'] ) || ! current_user_can( 'manage_options' ) ) {
+			return;
+		}
```
bordoni, line 838: I would love to see this being separated in to three separate conditionals with a bail for each. Helps readability.

bordoni (changes requested):

JPry (approved):
Looks good!

src/Tickets/Flexible_Tickets/Repositories/Event_Repository.php
```diff
@@ -93,14 +93,12 @@ private function include_series_in_meta_value_compare( string $alias ): string {
 
 		// Include Events that are either directly related to a Ticket (Single and Series Tickets) or
 		// are related to a Series that is related to a Ticket (Series Passes).
-		return $wpdb->prepare( "
+		return "
```
JPry, line 96: This is a bit of a nit-pick, but for a multi-line string I think it would be better to use the [heredoc](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc) syntax. Interpolated variables are still allowed in a heredoc string.

src/Tribe/Attendees.php
```diff
@@ -790,34 +804,95 @@ public function maybe_generate_csv() {
...
-				fputcsv( $output, $item, $delimiter );
-			}
+		if ( empty( $items ) ) {
+			return;
+		}
 
-			fclose( $output );
-			exit;
+		$charset  = get_option( 'blog_charset' );
+		$filename = sanitize_file_name( $event->post_title . '-' . __( 'attendees', 'event-tickets' ) );
```
JPry, line 812: Minor: this would be a good place to use a context translation string (`_x()`) to specify how "attendees" is used. It's possible that this could change how the word is translated, since it is appended to the event's title.

src/Tribe/Attendees.php
```diff
@@ -790,34 +804,95 @@ public function maybe_generate_csv() {
...
+	/**
+	 * Export a CSV containing all attendees.
+	 *
+	 * Called when an Admin requested a CSV export from the attendees table. Generates
+	 * the download file and finishes the execution.
+	 * 
+	 * @since TBD
+	 */
+	public function export_csv_for_all() {
+		if ( empty( $_GET['attendees_csv'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
JPry, line 848: Since we are doing this same check for both of the export functions, I think we should have a separate method that does these checks and determines whether we should continue or `return`. Then both methods call that one method.

src/Tribe/Attendees.php
```diff
@@ -790,34 +804,95 @@ public function maybe_generate_csv() {
...
+		$items = $this->generate_filtered_list( 'all' );
+
+		// Sanitize items for CSV usage.
+		$items = $this->sanitize_csv_rows( $items );
+
+		if ( empty( $items ) ) {
+			return;
+		}
+
+		$charset  = get_option( 'blog_charset' );
```
JPry, line 870: Similar to the above  comment, we have two methods that are doing almost the same thing: building a list of attendees and creating a file download.

It would be nicer to see a single method that accepts `$items` as a parameter, and generates the CSV with those items. Then each of the methods could simply handle how `$items` is created and pass that to the new method.

JPry (changes requested):

pattihis:
Creating a new clean branch based on `release/T24.centaur`

PR #3139, opened 2024-08-01
[ET-2171] Replaces DOMNodeInserted event with a custom built animation

dpanta94:
### 🎫 Ticket

[ET-2171]

### 🗒️ Description

Replaces DOMNodeInserted event with a custom built one using css.

How does it work ?

We are creating an animation `nodeInserted` which we apply to an element we want to listen for when its injected. We listen for the animation start event.

Why like this ?

This is a widely used technique for replacing `DOMNodeInserted` and its guaranteed that it will continue to work for a long time.

### 🎥 Artifacts 
https://jumpshare.com/embed/uINU2b1hMBjXmwivzsjg

### ✔️ Checklist
- [x] Ran `npm run changelog` to add changelog file(s).

[ET-2171]: https://stellarwp.atlassian.net/browse/ET-2171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/resources/js/commerce/gateway/paypal/checkout.js
+++ b/src/resources/js/commerce/gateway/paypal/checkout.js
@@ -544,7 +544,7 @@ tribe.tickets.commerce.gateway.paypal.checkout = {};
 	 * @since TBD Replaced DOMNodeInserted with animationstart event.
 	 */
 	obj.buttonsLoaded = function ( event ) {
-		if ( event.originalEvent.animationName !== 'nodeInserted' ) {
+		if ( event.originalEvent.animationName !== 'node_inserted' ) {
 			return;
 		}
 
--- a/src/resources/js/commerce/gateway/paypal/checkout.js
+++ b/src/resources/js/commerce/gateway/paypal/checkout.js
@@ -541,6 +541,7 @@ tribe.tickets.commerce.gateway.paypal.checkout = {};
 	 * Handle actions when checkout buttons are loaded.
 	 *
 	 * @since 5.1.10
+	 * @since TBD Replaced DOMNodeInserted with animationstart event.
 	 */
 	obj.buttonsLoaded = function ( event ) {
 		if ( event.originalEvent.animationName !== 'nodeInserted' ) {
@@ -593,6 +594,7 @@ tribe.tickets.commerce.gateway.paypal.checkout = {};
 	 * Setup the triggers for Ticket Commerce loader view.
 	 *
 	 * @since 5.1.10
+	 * @since TBD Replaced DOMNodeInserted with animationstart event.
 	 *
 	 * @return {void}
 	 */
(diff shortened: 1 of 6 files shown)


src/resources/js/commerce/gateway/paypal/checkout.js
```diff
@@ -541,9 +542,15 @@ tribe.tickets.commerce.gateway.paypal.checkout = {};
 	 *
 	 * @since 5.1.10
```
redscar, line 543: We should update the `@since`.

src/resources/js/commerce/gateway/paypal/checkout.js
```diff
@@ -593,7 +600,9 @@ tribe.tickets.commerce.gateway.paypal.checkout = {};
 		$document.trigger( tribe.tickets.commerce.customEvents.showLoader );
```
redscar, line 600: We should update the `@since`

redscar:
I'm seeing some issues with the lint test in regards to ordering. I think once that's fixed it should be good to go.

redscar (approved):

src/resources/js/commerce/gateway/paypal/checkout.js
```diff
@@ -540,10 +541,17 @@ tribe.tickets.commerce.gateway.paypal.checkout = {};
...
+	obj.buttonsLoaded = function ( event ) {
+		if ( event.originalEvent.animationName !== 'node_inserted' ) {
+			return;
+		}
+
 		$document.trigger( tribe.tickets.commerce.customEvents.hideLoader );
-		$( tribe.tickets.commerce.selectors.checkoutContainer ).off( 'DOMNodeInserted', obj.selectors.buttons, obj.buttonsLoaded );
+		$( tribe.tickets.commerce.selectors.checkoutContainer ).off( 'animationstart', obj.selectors.animatedElement, obj.buttonsLoaded );
+		$( tribe.tickets.commerce.selectors.checkoutContainer ).off( 'MSAnimationStart', obj.selectors.animatedElement, obj.buttonsLoaded );
+		$( tribe.tickets.commerce.selectors.checkoutContainer ).off( 'webkitAnimationStart', obj.selectors.animatedElement, obj.buttonsLoaded );
```
Camwyn, line 554: I wonder if the on/off is large enough to warrant it's own toggle function 🤔 (not a request, just a thing to think about)

Camwyn: Also wonder if the browser-prefixes are necessary? The [_element API_](https://caniuse.com/?search=animationstart) for  animationstart has nearly 97% and all browsers since 2020 are good with it.

Camwyn: Ditto for prefixed @keyframes and the animation-duration prop (below)

Camwyn (approved):

src/resources/js/commerce/gateway/paypal/checkout.js
dpanta94: They might not be, but i preferred to leave them there just in case. I wasnt able to test this in all modern browsers. Happy to follow your lead on this one!

Regarding creating methods for the on/off for now i think its an overkill. If we would need to do it one more or if there were more events to on/off i think it would be worth it.

PR #3149, opened 2024-08-01
Fix Attendees CSV Export

pattihis:
### 🎫 Ticket

[ET-2094]

### 🗒️ Description

In our previous CSV export functions we expected an associated `event_id`, that would exist as a query parameter. In our main attendees table we are now exporting all of them instead ( `event_id=all` ).

### ✔️ Checklist

### DIFF
--- a/src/Tribe/Attendees.php
+++ b/src/Tribe/Attendees.php
@@ -758,7 +758,7 @@ class Tribe__Tickets__Attendees {
 	 * @return void
 	 */
 	public function maybe_generate_csv( $all = false ) {
-		if ( 
+		if (
 				! isset( $_GET['attendees_csv_nonce'] )
 				|| ! wp_verify_nonce( sanitize_key( $_GET['attendees_csv_nonce'] ), 'attendees_csv_nonce' )
 				|| empty( $_GET['attendees_csv'] )
--- a/src/Tribe/Attendees.php
+++ b/src/Tribe/Attendees.php
@@ -751,9 +751,9 @@ class Tribe__Tickets__Attendees {
 	 * Checks if the user requested a CSV export from the attendees list.
 	 * If so, generates the download and finishes the execution.
 	 *
-	 * @param bool $all Whether to generate a CSV for attendees of all events or just the current one.
-	 *
 	 * @since 4.6.2
+	 *
+	 * @param bool $all Whether to generate a CSV for attendees of all events or just the current one.
 	 */
 	public function maybe_generate_csv( $all = false ) {
 
@@ -762,7 +762,7 @@ class Tribe__Tickets__Attendees {
 		}
 
 		if ( ! $all ) {
-			$event_id = absint( sanitize_key( $_GET['event_id'] ) );
+			$event_id = isset( $_GET['event_id'] ) ? absint( sanitize_key( $_GET['event_id'] ) ) : 0;
 			$event_id = Event::filter_event_id( $event_id, 'attendee-csv-report' );
 			$event    = get_post( $event_id );
 
--- a/src/Tribe/Attendees.php
+++ b/src/Tribe/Attendees.php
@@ -2,6 +2,7 @@
 
 use TEC\Tickets\Event;
 use Tribe__Utils__Array as Arr;
+use Tribe__Tickets__Tickets as Tickets;
 use Tribe__Tickets__Ticket_Object as Ticket;
 use Tribe__Tickets__Global_Stock as Global_Stock;
 use TEC\Tickets\Admin\Attendees\Page as Attendees_Page;
@@ -174,7 +175,7 @@ class Tribe__Tickets__Attendees {
 
 		$total_checked_in = $this->get_checkedin_total();
 		$check_in_percent = $this->get_checkedin_percentage( $post_id );
-		$total_attendees  = Tribe__Tickets__Tickets::get_event_attendees_count( $post_id );
+		$total_attendees  = Tickets::get_event_attendees_count( $post_id );
 
 		/** @var Tribe__Tickets__Admin__Views $admin_views */
 		$admin_views = tribe( 'tickets.admin.views' );
@@ -208,7 +209,7 @@ class Tribe__Tickets__Attendees {
 	 */
 	public function get_checkedin_percentage( $post_id ): string {
 		$total_checked_in = $this->get_checkedin_total();
-		$total            = Tribe__Tickets__Tickets::get_event_attendees_count( $post_id );
+		$total            = Tickets::get_event_attendees_count( $post_id );
 
 		// Remove the "Not Going" RSVPs.
 		$not_going = tribe( 'tickets.rsvp' )->get_attendees_count_not_going( $post_id );
@@ -278,7 +279,7 @@ class Tribe__Tickets__Attendees {
 			return $actions;
 		}
 
-		$tickets = Tribe__Tickets__Tickets::get_event_tickets( $post->ID );
+		$tickets = Tickets::get_event_tickets( $post->ID );
 
 		// Only proceed if there are tickets.
 		if ( empty( $tickets ) ) {
@@ -486,7 +487,7 @@ class Tribe__Tickets__Attendees {
 		} else {
 			$this->attendees_table = new Tribe__Tickets__Attendees_Table();
 
-			$this->maybe_generate_csv();
+			$this->maybe_generate_csv( 'all' === ( $_GET['event_id'] ?? '' ) ); // phpcs:ignore WordPress.Security
 
 			add_filter( 'admin_title', [ $this, 'filter_admin_title' ], 10, 2 );
 			add_filter( 'admin_body_class', [ $this, 'filter_admin_body_class' ] );
@@ -575,7 +576,7 @@ class Tribe__Tickets__Attendees {
 		$filter_name = "manage_{$this->page_id}_columns";
 		add_filter( $filter_name, [ $this->attendees_table, 'get_columns' ], 15 );
 
-		$items = Tribe__Tickets__Tickets::get_event_attendees( $event_id );
+		$items = 'all' === $event_id ? Tickets::get_attendees_by_args()['attendees'] : Tickets::get_event_attendees( $event_id );
 
 		// Add Handler for Community Tickets to Prevent Notices in Exports.
 		if ( ! is_admin() ) {
@@ -596,6 +597,7 @@ class Tribe__Tickets__Attendees {
 				'purchaser',
 				'status',
 				'attendee_actions',
+				'attendee_event',
 			]
 		);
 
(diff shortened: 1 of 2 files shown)


src/Tribe/Attendees.php
```diff
@@ -752,72 +754,97 @@ public function sanitize_csv_value( $value ) {
...
-		 * Allow for filtering and modifying the list of attendees that will be exported via CSV for a given event.
-		 *
-		 * @param array $items    The array of attendees that will be exported in this CSV file.
-		 * @param int   $event_id The ID of the event these attendees are associated with.
-		 */
-		$items = apply_filters( 'tribe_events_tickets_attendees_csv_items', $items, $event_id );
+			$event_id = Event::filter_event_id( $event_id, 'attendee-csv-report' );
+
+			// Verify event ID is a valid integer and the nonce is accepted.
+			if ( empty( $event_id ) || ! wp_verify_nonce( $_GET['attendees_csv_nonce'], 'attendees_csv_nonce' ) ) {
```
JPry, line 780: Why only do the nonce verification for a specific event and not `all`?

src/Tribe/Attendees.php
```diff
@@ -752,72 +754,97 @@ public function sanitize_csv_value( $value ) {
 	 * @since 4.6.2
 	 *
 	 */
-	public function maybe_generate_csv() {
-		if ( empty( $_GET['attendees_csv'] ) || empty( $_GET['attendees_csv_nonce'] ) || empty( $_GET['event_id'] ) ) {
+	public function maybe_generate_csv( $all = false ) {
```
JPry, line 757: Two things:

1. We'll need to update the docblock to reflect the new parameter.
2. I'm not so sure that `$all` as a boolean value is the best option here. Maybe something like `$event_id` with possible values as a string (for `'all'`) or an integer (for the event ID). Bonus would be using `is_numeric()` to ensure a string version of an ID also works in that case.

src/Tribe/Attendees.php
Camwyn, line 780: If you move the `wp_verify_nonce` check to the top when we check if it exists, and move that check to the start, you _should_ be able to remove all the `phpcs:ignore`

src/Tribe/Attendees.php
```diff
@@ -752,72 +754,97 @@ public function sanitize_csv_value( $value ) {
...
-			foreach ( $items as $item ) {
-				fputcsv( $output, $item, $delimiter );
-			}
+		if ( empty( $items ) ) {
+			return;
+		}
 
-			fclose( $output );
-			exit;
+		$charset  = get_option( 'blog_charset' );
```
Camwyn, line 817:
```suggestion
		$charset = get_option( 'blog_charset' );
```

src/Tribe/Attendees.php
```diff
@@ -752,72 +754,97 @@ public function sanitize_csv_value( $value ) {
...
 			return;
 		}
 
-		$event_id = absint( $_GET['event_id'] );
-
-		$event_id = Event::filter_event_id( $event_id, 'attendee-csv-report' );
-
-		// Verify event ID is a valid integer and the nonce is accepted.
-		if ( empty( $event_id ) || ! wp_verify_nonce( $_GET['attendees_csv_nonce'], 'attendees_csv_nonce' ) ) {
+		if ( empty( $_GET['attendees_csv_nonce'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
JPry, line 762: Looking at these checks, I feel like there are 2 things that would make this a bit better:

1. Have a dedicated function to check the various `$_GET` values, and trigger an early return if we determine we don't have what we need. This would help make the CSV logic easier to read and understand.
2. The nonce check should also happen as part of the above, rather than farther down the function.

src/Tribe/Attendees.php
```diff
@@ -752,72 +754,97 @@ public function sanitize_csv_value( $value ) {
...
+			$event = get_post( $event_id );
 
-		if ( ! empty( $items ) ) {
-			$charset  = get_option( 'blog_charset' );
-			$filename = sanitize_file_name( $event->post_title . '-' . __( 'attendees', 'event-tickets' ) );
+			// Verify event exists and current user has access to it.
+			if (
+				! $event instanceof WP_Post
+				|| ! $this->user_can( 'edit_posts', $event_id )
+			) {
```
JPry, line 790: For readability, I think it might be better to rearrange this conditional.

I don't know that it's long enough to necessitate two lines, but it's also fine to leave it split.

```suggestion
			if ( ! ( $event instanceof WP_Post && $this->user_can( 'edit_posts', $event_id ) ) ) {
```

With that said... this could also be subjective, so if you feel more strongly about leaving it as you've written it, that's fine by me too.

src/Tribe/Attendees.php
```diff
@@ -752,72 +754,97 @@ public function sanitize_csv_value( $value ) {
...
+				return;
+			}
 
-			// Output headers so that the file is downloaded rather than displayed.
-			header( "Content-Type: text/csv; charset=$charset" );
-			header( "Content-Disposition: attachment; filename=$filename.csv" );
+			$items = $this->generate_filtered_list( $event_id );
+		} else {
+
+			$items = $this->generate_filtered_list( 'all' );
```
JPry, line 797: No need for an extra line.

```suggestion
			$items = $this->generate_filtered_list( 'all' );
```

src/Tribe/Attendees.php
```diff
@@ -486,7 +487,7 @@ public function screen_setup() {
 		} else {
 			$this->attendees_table = new Tribe__Tickets__Attendees_Table();
 
-			$this->maybe_generate_csv();
+			$this->maybe_generate_csv( 'all' === ( $_GET['event_id'] ?? '' ) ); // phpcs:ignore WordPress.Security
```
Camwyn, line 490: This is too broad - it should be for something specific (like`WordPress.Security.NonceVerification.Recommended` verification, which should happen earlier)

Camwyn (changes requested):
Nonce verification needs to happen and happen first.

src/Tribe/Attendees.php
Camwyn: TL/DR: if the nonce is invalid we stop - immediately. We should do that before we do anything else (even checking permissions).

src/Tribe/Attendees.php
```diff
@@ -575,7 +576,7 @@ private function generate_filtered_list( $event_id ) {
 		$filter_name = "manage_{$this->page_id}_columns";
 		add_filter( $filter_name, [ $this->attendees_table, 'get_columns' ], 15 );
 
-		$items = Tribe__Tickets__Tickets::get_event_attendees( $event_id );
+		$items = 'all' === $event_id ? Tickets::get_attendees_by_args()['attendees'] : Tickets::get_event_attendees( $event_id );
```
JPry, line 579: The docblock for the filter above (and implicitly the whole method) indicates that `$event_id` should be an integer, but now we're passing a string. We should update the two docblocks to reflect the changed use.

src/Tribe/Attendees.php
```diff
@@ -752,72 +754,97 @@ public function sanitize_csv_value( $value ) {
...
-			 *
-			 * @since 5.1.3
+			 * Allow for filtering and modifying the list of attendees that will be exported via CSV for a given event.
 			 *
-			 * @param string $delimiter The field delimiter used in the CSV export file.
+			 * @param array $items    The array of attendees that will be exported in this CSV file.
+			 * @param int   $event_id The ID of the event these attendees are associated with.
 			 */
-			$delimiter = apply_filters( 'tribe_tickets_attendees_csv_export_delimiter', ',' );
+			$items = apply_filters( 'tribe_events_tickets_attendees_csv_items', $items, $event_id );
```
JPry, line 810: It feels odd to me to have this filter here instead of where the event items are generated. Or perhaps it feels more odd to have the filter only apply to single event items instead of all.

I understand that this reflects the usage before the change to allow `all` as an option, but perhaps we should consider making the filter apply regardless of whether one event or `all` has been set.

src/Tribe/Attendees.php
```diff
@@ -752,72 +754,97 @@ public function sanitize_csv_value( $value ) {
...
+		if ( empty( $items ) ) {
+			return;
+		}
 
-			fclose( $output );
-			exit;
+		$charset  = get_option( 'blog_charset' );
+
+		if ( ! $all ) {
+			$filename = sanitize_file_name( $event->post_title . '-' . _x( 'attendees', 'CSV export file name', 'event-tickets' ) );
```
JPry, line 820: `_x()` -> ❤️

src/Tribe/Attendees.php
```diff
@@ -752,72 +754,97 @@ public function sanitize_csv_value( $value ) {
 	 * @since 4.6.2
 	 *
 	 */
-	public function maybe_generate_csv() {
-		if ( empty( $_GET['attendees_csv'] ) || empty( $_GET['attendees_csv_nonce'] ) || empty( $_GET['event_id'] ) ) {
+	public function maybe_generate_csv( $all = false ) {
+		if ( empty( $_GET['attendees_csv'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
Camwyn, line 758:
```suggestion
		if ( ! isset( $_GET['attendees_csv_nonce'] ) || ! wp_verify_nonce( sanitize_key( $_GET['attendees_csv_nonce'] ), 'attendees_csv_nonce' ) || empty( $_GET['attendees_csv'] )  ) { 
```

Camwyn: Then remove all the other stuff it covers (and the ignores)

JPry (changes requested):

src/Tribe/Attendees.php
Camwyn: Filters need docblcoks

src/Tribe/Attendees.php
Camwyn: Also recall you can use tribe_get_request_var to bypass the nonce warning - where it's not appropriate (I don't think there is a nonce available here)

Camwyn (approved):

src/Tribe/Attendees.php
```diff
@@ -555,15 +556,15 @@ public function render() {
 	 *
 	 * @since 4.6.2
 	 *
-	 * @param $event_id
+	 * @param mixed $event_id The ID of the event to export the list for or 'all' for all events.
```
JPry, line 559: Minor: `mixed` is accurate, but since we're only expecting 2 types we could do it this way:

```suggestion
	 * @param int|string $event_id The ID of the event to export the list for or 'all' for all events.
```

Same for the filter below.

src/Tribe/Attendees.php
```diff
@@ -751,73 +753,79 @@ public function sanitize_csv_value( $value ) {
...
 	 * @since 4.6.2
 	 *
+	 * @param bool $all Whether to generate a CSV for attendees of all events or just the current one.
+	 *
+	 * @return void
 	 */
-	public function maybe_generate_csv() {
-		if ( empty( $_GET['attendees_csv'] ) || empty( $_GET['attendees_csv_nonce'] ) || empty( $_GET['event_id'] ) ) {
+	public function maybe_generate_csv( $all = false ) {
+		if ( ! isset( $_GET['attendees_csv_nonce'] )
```
JPry, line 761: Minor: For multiline if statements, we should put the first comparison on a new line and leave `if (` on its own line

```suggestion
		if (
			! isset( $_GET['attendees_csv_nonce'] )
```

JPry (dismissed):
Looks good! I flagged two minor things but I'd say once those are tweaked this is good to go.

src/Tribe/Attendees.php
JPry: I should also say the same applies for the other places where we're expecting `all` or an event ID in the file.

src/Tribe/Attendees.php
```diff
@@ -751,73 +753,80 @@ public function sanitize_csv_value( $value ) {
...
 	 * @since 4.6.2
 	 *
+	 * @param bool $all Whether to generate a CSV for attendees of all events or just the current one.
+	 *
+	 * @return void
 	 */
-	public function maybe_generate_csv() {
-		if ( empty( $_GET['attendees_csv'] ) || empty( $_GET['attendees_csv_nonce'] ) || empty( $_GET['event_id'] ) ) {
+	public function maybe_generate_csv( $all = false ) {
+		if (
```
Camwyn, line 761:
```suggestion
		if (
```

Camwyn (approved):

PR #3151, opened 2024-08-01
Refactor logic for `Ticket::increase_ticket_sales_by` method

codingmusician:
### 🎫 Ticket
[ETP-920]

### 🗒️ Description
This [this PR](https://github.com/the-events-calendar/event-tickets/pull/3145), we added some logic to fix a bug. We then realized that the `$shared_capacity` param should have been a boolean value all along. To make things right, We're moving that logic into the `Increase_Sales::handle()` method so that `$shared_capacity` is a boolean value, as it should be.

### 🎥 Artifacts 
[Screencast](https://share.zight.com/2Nu8ng1P)

### ✔️ Checklist
- [x] Ran `npm run changelog` to add changelog file(s).

[ETP-920]: https://stellarwp.atlassian.net/browse/ETP-920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Flag_Actions/Increase_Sales.php
+++ b/src/Tickets/Commerce/Flag_Actions/Increase_Sales.php
@@ -65,7 +65,11 @@ class Increase_Sales extends Flag_Action_Abstract {
 
 			$global_stock = new \Tribe__Tickets__Global_Stock( $ticket->get_event_id() );
 
-			tribe( Ticket::class )->increase_ticket_sales_by( $ticket->ID, $quantity, $ticket->global_stock_mode(), $global_stock );
+			// Is ticket shared capacity?
+			$global_stock_mode  = $ticket->global_stock_mode();
+			$is_shared_capacity = ! empty( $global_stock_mode ) && 'own' !== $global_stock_mode;
+
+			tribe( Ticket::class )->increase_ticket_sales_by( $ticket->ID, $quantity, $is_shared_capacity, $global_stock );
 		}
 	}
 }
--- a/src/Tickets/Commerce/Flag_Actions/Increase_Sales.php
+++ b/src/Tickets/Commerce/Flag_Actions/Increase_Sales.php
@@ -40,6 +40,7 @@ class Increase_Sales extends Flag_Action_Abstract {
 	 * {@inheritDoc}
 	 *
 	 * @since 5.2.0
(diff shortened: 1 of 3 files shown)


src/Tickets/Commerce/Flag_Actions/Increase_Sales.php
```diff
@@ -65,7 +66,11 @@ public function handle( Status_Interface $new_status, $old_status, \WP_Post $pos
 
 			$global_stock = new \Tribe__Tickets__Global_Stock( $ticket->get_event_id() );
 
-			tribe( Ticket::class )->increase_ticket_sales_by( $ticket->ID, $quantity, $ticket->global_stock_mode(), $global_stock );
+			// Is ticket shared capacity?
+			$global_stock_mode  = $ticket->global_stock_mode();
```
Camwyn, line 70: Any reason we didn't use https://github.com/the-events-calendar/event-tickets/blob/master/src/Tribe/Tickets_Handler.php#L1000

Camwyn: I guess since we have the ticket object, this might make more sense. I'm on the fence.

Camwyn (commented):
Thought/question

src/Tickets/Commerce/Flag_Actions/Increase_Sales.php
JPry: @Camwyn I didn't know about that (TIL), but it looks like `Tickets_Handler::gas_shared_capacity()` is wanting a `WP_Post` object to work with, whereas we have a `Tribe__Tickets__Ticket_Object` as `$ticket` here.

To me it feels like it's easier to use the method here rather than the `Tickets_Handler` class, but with that said I think it would be all around better to have this kind of functionality as part of the ticket object itself in some way.

I don't know enough about the existing code structure to say what that should look like at this point, but maybe the current approach (this PR) is fine for now.

What do you think since you have more experience here?

Camwyn: See, that's why I am on the fence - we have the object. I agree it seems like that functionality should move, but that's OOS for this PR/ticket.

Camwyn (approved):

PR #3158, opened 2024-08-06
Fix CSV export for all attendees

pattihis:
### 🎫 Ticket

[ET-2094]

### 🗒️ Description

This is a follow up to https://github.com/the-events-calendar/event-tickets/pull/3117

### ✔️ Checklist


src/Tribe/Attendees.php
```diff
@@ -751,73 +753,80 @@ public function sanitize_csv_value( $value ) {
 	 *
 	 * @since 4.6.2
```
codingmusician, line 754: Need an updated `@since` here, too.

src/Tribe/Attendees.php
```diff
@@ -208,7 +209,7 @@ public function get_checkedin_total(): int {
 	 */
 	public function get_checkedin_percentage( $post_id ): string {
```
codingmusician, line 210: Should add a new `@since` for this method regarding the `all` logic.

src/Tribe/Attendees.php
```diff
@@ -555,15 +556,15 @@ public function render() {
 	 *
 	 * @since 4.6.2
```
codingmusician, line 557: New `@since` needs to be added here, too.

src/Tribe/Attendees.php
```diff
@@ -832,7 +841,6 @@ public function maybe_generate_csv() {
 	 * @param ?string|?int $send_to  The recipient's ID or email.
 	 *                               If $type is 'user', this should be the user ID.
 	 *
-	 *
 	 * @return true|WP_Error
 	 */
 	public function has_attendees_list_access( $event_id = null, ?string $nonce = null, ?string $type = 'user', $send_to = null ) {
```
codingmusician, line 846: New `@since`?

src/Tribe/Attendees.php
```diff
@@ -1175,7 +1184,7 @@ public function include_export_button_title( $event_id, Tribe__Tickets__Attendee
 	 * @return array<string,mixed> The context used to render the Attendees page.
 	 */
 	public function get_render_context( int $post_id ): array {
```
codingmusician, line 1186: New `@since`?

stratease (changes requested):
We need tests?

Camwyn (changes requested):
Tests please

tests/wpunit/Tribe/Tickets/Attendees_Test.php
```diff
@@ -248,6 +254,55 @@ public function test_event_details_top( Closure $fixture ): void {
...
+
+		$attendees = new Attendees();
+		$attendees->attendees_table = new Tribe__Tickets__Attendees_Table();
+
+		// Generate filtered list of attendees for the latter event.
+		$items = $attendees->generate_filtered_list( $post_id );
+
+		// Get the 'Ticket Holder Name' column from the arrays.
+		$this->assertEquals( $items[0][5], 'Ticket Holder Name' );
+		$this->assertEquals( $items[1][5], 'Jane Doe' );
```
Camwyn, line 295: 1) can we add a fail message to indicate what part of this test is failing?(here and elsewhere) - or split the tests up into logically named tests.
2) can we include a check that there is only one ticket in this iteration? i.e. `$this->assertCount( 2, $items, 'There should be only two tickets' )` (assuming that the count is tickets + header here)

tests/wpunit/Tribe/Tickets/Attendees_Test.php
```diff
@@ -248,6 +254,55 @@ public function test_event_details_top( Closure $fixture ): void {
...
+
+		// Generate filtered list of attendees for all events.
+		$items = $attendees->generate_filtered_list( 'all' );
+
+		// Get the 'Ticket Holder Name' column from the arrays.
+		$this->assertEquals( $items[0][5], 'Ticket Holder Name' );
```
Camwyn, line 301: Can we include a check that there are two and only two tickets in this iteration? i.e `$this->assertCount( 3, $items, 'There should be only two tickets' )`(assuming that the count is tickets + header here)

Camwyn (changes requested):
test tweaks

tests/wpunit/Tribe/Tickets/Attendees_Test.php
pattihis: @Camwyn 

We have two events. Both have attendees. I am showing that the `generate_filtered_list()` function, the one that populates the CSV, returns a proper lists if we pass an `event_id` or `all`.

What's the point to check on the number of the tickets? I think that it's irrelevant. Our function should return an attendee list either for a specific event or for all existing events. That's what this test does.

stratease (changes requested):

tests/ft_integration/Tribe/Tickets/__snapshots__/MetaboxTest__test_get_panels_without_providers__event__0.snapshot.html
```diff
@@ -247,7 +247,7 @@ <h4>Ticket Settings	</h4>
 			class="tribe-datepicker tribe-field-start_date ticket_field"
 			name="ticket_start_date"
 			id="ticket_start_date"
-			value="8/5/2024"
+			value="8/8/2024"
```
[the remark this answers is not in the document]
pattihis: @stratease I have no idea. I ran the failing test with `--debug` and it changed the snapshot. 

Brian Krane worked on that test [two days ago](https://github.com/the-events-calendar/event-tickets/commit/ecfd3edc21759f1d33636881c31c56e66310cfb8).

tests/wpunit/Tribe/Tickets/Attendees_Test.php
```diff
@@ -248,6 +254,55 @@ public function test_event_details_top( Closure $fixture ): void {
 		}
 	}
 
+	/**
+	 * @test
+	 */
+	public function test_generate_filtered_list() {
```
redscar, line 260: We may want to rename the test to remove the `test`prefix.For example, `it_should_generate_a_filtered_list`.

I also suggest possibly adding in a dataprovider that way you can test the scenario of one (or many) ticket with `generate_filtered_list()` and then with multiple tickets using `generate_filtered_list('all')`

It could help clean the code up a bit. You could also keep the expected data in an array and test that it matches what `generate_filtered_list()` returns for consistency.

tests/ft_integration/Tribe/Tickets/__snapshots__/MetaboxTest__test_get_panels_without_providers__event__0.snapshot.html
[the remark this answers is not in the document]
redscar: @pattihis For that failing test we may need to add the below into `test_get_panels_without_providers` similar to how I did it for `test_get_panels`

```
// Set up a fake "now".
$date = new \DateTime( '2019-09-11 22:00:00', new \DateTimeZone( 'America/New_York' ) );
$now  = $date->getTimestamp();
// Alter the concept of the `now` timestamp to return the timestamp for `2019-09-11 22:00:00` in NY timezone.
uopz_set_return(
	'strtotime', static function ( $str ) use ( $now ) {
	return $str === 'now' ? $now : strtotime( $str );
},  true
);
// Make sure that `now` (string) will be resolved to the fake date object.
uopz_set_return( Date_Utils::class, 'build_date_object', $date );```

tests/wpunit/Tribe/Tickets/Attendees_Test.php
pattihis: Thank you @redscar . About the name, isn't the `test_` prefix the proper one to use before the actual name of the function being tested? 

Cause that's the function we are testing here
`Tribe__Tickets__Attendees::generate_filtered_list()`

Of course we can use a `it_should_whatever` naming approach but I find the actual function names prepended with `test_` a more consistent way.

<hr>

We don't want multiple scenarios, just one scenario is enough: `more than 1 events`. And we check the function with one event and with all of them in the same context.

redscar: I understand, the scenario may play a part into it. Usually if I add `@test` to the docblock I try to stay away from the `test_` prefix.

---
That makes sense, I was thinking more in regards to making sure that passing `all` or passing an Event ID to the function returns the expected output for both scenarios since we don't want the scenario where passing the Event ID to return all attendees.

pattihis: Added your suggestions, thanks!

tests/ft_integration/Tribe/Tickets/__snapshots__/MetaboxTest__test_get_panels_without_providers__event__0.snapshot.html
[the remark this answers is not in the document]
pattihis: Added your fix, thanks!

Camwyn (approved):

stratease (approved):
Groovy.

PR #3177, opened 2024-08-15
[ET-2179] All Tickets Page > sorting and design updates

codingmusician:
### 🎫 Ticket
[ET-2179]
[ET-2181]

### 🗒️ Description
Implements the query, sorting and some design. It also implements the links and quicklinks.

### 🎥 Artifacts 
Screencast: https://share.zight.com/BluPgjP9

### ✔️ Checklist

[ET-2179]: https://stellarwp.atlassian.net/browse/ET-2179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ET-2181]: https://stellarwp.atlassian.net/browse/ET-2181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,324 @@ public function get_sortable_columns() {
...
+
+		$edit_post_url  = get_edit_post_link( $event );
+		$edit_post_link = sprintf(
+			'<a href="%s" class="tec-tickets-all-tickets-table-event-link" target="_blank" rel="nofollow noopener">%s</a>',
+			esc_url( $edit_post_url ),
+			get_the_title( $event )
+		);
+
+		$orders_report_url  = sprintf(
+			'https://wp.lndo.site/wp-admin/edit.php?post_type=%s&page=tickets-orders&event_id=%d',
```
rafsuntaskin, line 298:
```suggestion
			'https://wp.lndo.site/wp-admin/edit.php?post_type=%s&page=tickets-orders&event_id=%d',
```

Needs to be dynamic

rafsuntaskin (dismissed):

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,324 @@ public function get_sortable_columns() {
...
+			$event->ID
+		);
+		$orders_report_link = sprintf(
+			'<a href="%s" class="tec-tickets-all-tickets-table-event-link" target="_blank" rel="nofollow noopener">%s</a>',
+			esc_url( $orders_report_url ),
+			esc_html__( 'Orders', 'event-tickets' )
+		);
+
+		$attendees_report_url  = sprintf(
+			'https://wp.lndo.site/wp-admin/edit.php?post_type=%s&page=tickets-attendees&event_id=%d',
```
rafsuntaskin, line 309: Needs to be dynamic

codingmusician: Doh!! 🤦‍♂️

codingmusician: I'll fix this next week.

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,324 @@ public function get_sortable_columns() {
...
+		 *
+		 * @param array $actions The actions for the event in the All Tickets Table.
+		 */
+		$actions = apply_filters( 'tec_tickets_all_tickets_table_event_actions', $actions, $event, $item );
+
+		return sprintf( '%1$s %2$s', $edit_post_link, $this->row_actions( $actions ) );
+	}
+
+	/**
+	 * Get the column start value.
```
stratease, line 337:
```suggestion
	 * Get the column start date value.
```
Start is a bit vague, can we be more specific?

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,324 @@ public function get_sortable_columns() {
...
+	 */
+	public function column_start( $item ): string {
+		$date_format = tribe_get_date_format( true );
+		$datetime    = $item->start_date( false );
+
+		return $datetime->format( $date_format );
+	}
+
+	/**
+	 * Get the column end value.
```
stratease, line 353:
```suggestion
	 * Get the column end date value.
```
Same.

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,324 @@ public function get_sortable_columns() {
...
+
+	/**
+	 * Get the query args for the list table.
+	 *
+	 * @since TBD
+	 *
+	 * @return array
+	 */
+	public function get_query_args() {
+
```
stratease, line 499:
```suggestion
```

stratease (changes requested):

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,329 @@ public function get_sortable_columns() {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param Tribe__Tickets__Ticket_Object $item The current item.
+	 *
+	 * @return string
+	 */
+	public function column_id( $item ): string {
+		return $item->ID;
+	}
```
dpanta94, line 273: The return here is string or int ? I would update to `return absint( $item->ID );` and the docs to reflect that.

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,329 @@ public function get_sortable_columns() {
...
+	 *
+	 * @param Tribe__Tickets__Ticket_Object $item The current item.
+	 *
+	 * @return string
+	 */
+	public function column_start( $item ): string {
+		$date_format = tribe_get_date_format( true );
+		$datetime    = $item->start_date( false );
+
+		return $datetime->format( $date_format );
```
dpanta94, line 355:
```suggestion
		return esc_html( $datetime->format( $date_format ) );
```

Also maybe we could use the `<time>` element here ? see an example [here](https://github.com/the-events-calendar/event-tickets/blob/master/src/Tickets/Commerce/Admin_Tables/Orders_Table.php#L406)

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,329 @@ public function get_sortable_columns() {
...
+	 *
+	 * @param Tribe__Tickets__Ticket_Object $item The current item.
+	 *
+	 * @return string
+	 */
+	public function column_end( $item ): string {
+		$date_format = tribe_get_date_format( true );
+		$datetime    = $item->end_date( false );
+
+		return $datetime->format( $date_format );
```
dpanta94, line 371: Same as above!

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,329 @@ public function get_sortable_columns() {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param Tribe__Tickets__Ticket_Object $item The current item.
+	 *
+	 * @return string
+	 */
+	public function column_remaining( $item ): string {
+		$available = $item->available();
+		return $available < 0 ? '-' : $available;
```
dpanta94, line 432: We define that we return string, i guess because of the `'-'` option which is fine. But i would suggest to typecast $available to string then.
```suggestion
		return $available < 0 ? '-' : (string) $available;
```

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,329 @@ public function get_sortable_columns() {
...
+	/**
+	 * Modify the sort arguments.
+	 *
+	 * @since TBD
+	 *
+	 * @param array $args The arguments used to query the tickets for the All Tickets Table.
+	 *
+	 * @return array
+	 */
+	public function modify_sort_args( $args ) {
```
dpanta94, line 457:
```suggestion
	public function modify_sort_args( $args ): array {
```

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,329 @@ public function get_sortable_columns() {
...
+	}
+
+	/**
+	 * Get the query args for the list table.
+	 *
+	 * @since TBD
+	 *
+	 * @return array
+	 */
+	public function get_query_args() {
```
dpanta94, line 504:
```suggestion
	public function get_query_args(): array {
```

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -241,18 +522,35 @@ public function prepare_items() {
...
+			'total_items' => $total_items,
+			'per_page'    => $this->get_items_per_page( $this->per_page_option ),
+		];
 
 		$this->_column_headers = [
 			$this->get_table_columns(),
 			$this->get_hidden_columns(),
 			$this->get_sortable_columns(),
 		];
-		$this->items           = $items;
```
dpanta94, line 255: Dont we need $this->items defined ?

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,329 @@ public function get_sortable_columns() {
...
+	 * Get the column sales value.
+	 *
+	 * @since TBD
+	 *
+	 * @param Tribe__Tickets__Ticket_Object $item The current item.
+	 *
+	 * @return string
+	 */
+	public function column_sales( $item ): string {
+		return tribe_format_currency( number_format( $item->qty_sold() * $item->price, 2 ), $item->ID );
```
dpanta94, line 445: This one has a logic issue. The sales can't always be quantity sold times price since the price could have changed or we could have sales. This needs to be calculated differently.

src/Tickets/Admin/All_Tickets/Provider.php
```diff
@@ -51,5 +51,9 @@ protected function register_hooks() {
 	 * @since TBD
 	 */
 	protected function register_assets() {
+		$assets = new Assets( $this->container );
+		$assets->register();
+
+		$this->container->singleton( Assets::class, $assets );
```
dpanta94, line 57:
```suggestion
		$this->container->register( Assets::class );
```
Same result since the container's register method will create a sisngleton and jut call the register method of the registered interface immediately.

src/admin-views/all-tickets/table-column-name.php
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * Table column template.
+ *
+ * @since TBD
```
dpanta94, line 5: Probably we also need a `@version` tag

dpanta94 (changes requested):

src/Tickets/Admin/All_Tickets/Assets.php
```diff
@@ -0,0 +1,65 @@
...
+		tribe_asset(
+			$plugin,
+			'tec-tickets-all-tickets-styles',
+			'all-tickets.css',
+			[],
+			'admin_enqueue_scripts',
+			[
+				'conditionals' => [ $this, 'should_enqueue_assets' ],
+			]
+		);
```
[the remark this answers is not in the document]
codingmusician: Ah!! I had forgotten about the Stellar lib. Thanks Dimi!

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,329 @@ public function get_sortable_columns() {
...
 	 */
 	public function column_default( $item, $column_name ): string {
-		return $item[ $column_name ];
+		// If the column name is empty or the item does not have the column name, return an empty string.
+		if ( empty( $column_name ) || ! isset( $item->$column_name ) ) {
+			return '';
+		}
+
+		// If value is not a string or a number, return an empty string.
+		if ( ! is_string( $item->$column_name ) && ! is_numeric( $item->$column_name ) ) {
```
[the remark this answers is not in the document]
codingmusician: Well... that would allow boolean values through, right? We only want values that could be printed.

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,329 @@ public function get_sortable_columns() {
...
-	 * @since  TBD
+	 * @since TBD
 	 *
-	 * @param object $item        The current item.
-	 * @param string $column_name The current column name.
+	 * @param Tribe__Tickets__Ticket_Object $item The current item.
+	 *
+	 * @return string HTML for the ticket type icon.
+	 */
+	public function get_ticket_type_icon( $item ) {
```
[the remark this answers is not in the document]
codingmusician: Not currently. We can use `protected`.

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -189,49 +190,329 @@ public function get_sortable_columns() {
...
+		}
 
-		$pagination_args = [
-			'total_items' => 1,
-			'per_page'    => $per_page,
+		$edit_post_url  = get_edit_post_link( $event );
+		$edit_post_link = sprintf(
+			'<a href="%s" class="tec-tickets-all-tickets-table-event-link" target="_blank" rel="nofollow noopener">%s</a>',
+			esc_url( $edit_post_url ),
+			esc_html( $item->name )
```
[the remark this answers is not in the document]
codingmusician: The `$item` is the ticket itself and this is for the `Name` column, which is the Ticket Name. Thanks for double-checking.

src/Tickets/Admin/All_Tickets/List_Table.php
codingmusician: This is to print out the ID for the column header, so a string would be ideal. No need to force to an int.

src/Tickets/Admin/All_Tickets/List_Table.php
codingmusician: Yes. We will need to discuss this with @bordoni

src/Tickets/Admin/All_Tickets/List_Table.php
codingmusician: It's defined in the [parent class](https://developer.wordpress.org/reference/classes/wp_list_table/#properties)

src/Tickets/Admin/All_Tickets/List_Table.php
dpanta94: Maybe force to a string then its a good idea ? Im just a bit worried if that could ever be an int where we would have an issue with declared return type of the method.

dpanta94 (approved):

PR #3181, opened 2024-08-19
All Tickets Page > Search and Filter

codingmusician:
### 🎫 Tickets
[ET-2176]
[ET-2177]

### 🗒️ Description
Implement the search and select filter form and logic for the All Tickets table.

### 🎥 Artifacts 
Screencast (updated): https://share.zight.com/nOu2wJKb

### ✔️ Checklist

[ET-2176]: https://stellarwp.atlassian.net/browse/ET-2176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-2177]: https://stellarwp.atlassian.net/browse/ET-2177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -29,6 +30,71 @@ class List_Table extends WP_List_Table {
 	 */
 	protected $per_page_option;
 
+
+	/**
+	 * The template object.
+	 *
+	 * @var Tribe__Template
+	 */
+	protected $template;
```
dpanta94, line 39: One thing i am not a fan about this class is that it has grown to do a lot more stuff than the List Table now :( Maybe some logic needs to be separated into different files.

dpanta94 (changes requested):

src/admin-views/all-tickets/filters/status-select.php
```diff
@@ -0,0 +1,19 @@
...
+ * @since TBD
+ *
+ * @var \Tribe__Template  $this            Current template object.
+ * @var string[]          $status_options  The list table for the All Tickets screen.
+ * @var string            $current_status  Currently selected filter.
+ */
+
+?>
+<select name="status-filter" id="tec-tickets-all-tickets-status-filter">
+	<?php foreach ( $status_options as $value => $label ) : ?>
```
dpanta94, line 14: Is $status_options reflecting statuses that are present in the DB ? Or are they simply all the values ? If it's the first case, probably we should check that its `count() > 1` before we print the `<select/>`

src/admin-views/all-tickets/filters/status-select.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/**
+ * All Tickets list table status filter template.
+ *
+ * @since TBD
```
dpanta94, line 5: a couple of template files are missing `@version` tag :D

tests/integration/TEC/Tickets/Admin/All_Tickets/__snapshots__/List_TableTest__test_filters__filter by discounted__0.snapshot.json
```diff
@@ -0,0 +1 @@
+null
```
dpanta94, line 1: I would double check the tests producing these `null` snapshots.

tests/integration/TEC/Tickets/Admin/All_Tickets/__snapshots__/PageTest__test_render_tec_tickets_all_tickets_page__0.snapshot.html
```diff
@@ -2,11 +2,33 @@
...
+			<option value="active"  selected='selected'>
+			Active Tickets		</option>
+			<option value="past" >
+			Past Tickets		</option>
+			<option value="upcoming" >
+			Upcoming Tickets		</option>
+			<option value="discounted" >
+			Discounted Tickets		</option>
+			<option value="all" >
+			All Tickets		</option>
```
dpanta94, line 21: I guess every status option is being printed. Not sure if it would be a better experience to print only what is present in the DB

src/admin-views/all-tickets/filters/status-select.php
dpanta94: Do we also need an `All` option ?

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -498,31 +567,111 @@ public function modify_sort_args( $args ): array {
...
 	 *
+	 * @param array $args The arguments used to query the tickets for the All Tickets Table.
+	 *
 	 * @return array
 	 */
-	public function get_query_args(): array {
+	public function modify_filter_args( $args ) {
+		$filter = tribe_get_request_var( self::$status_key, self::$default_status );
+
+		if ( 'all' === $filter ) {
```
dpanta94, line 587:
```suggestion
		if ( ! $filter || 'all' === $filter ) {
```

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -498,31 +567,111 @@ public function modify_sort_args( $args ): array {
...
+	 *
+	 * @return array|bool
+	 */
+	public function get_query_args() {
 		$current_page = $this->get_pagenum();
 		$per_page     = $this->get_items_per_page( $this->per_page_option );
 
+		$provider_options = $this->get_provider_options();
+		$default_provider = empty( $provider_options ) ? '' : key( $provider_options );
+		$current_provider = tribe_get_request_var( self::$provider_key, $default_provider );
```
dpanta94, line 658: could `tribe_get_request_var` return `''` instead of the ~$default_provider` if the var is set in the `$_REQUEST` as an empty string ? If yes, we should make sure we are not continuing with empty string as the `$current_provider`

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -498,31 +567,111 @@ public function modify_sort_args( $args ): array {
...
+			'return_total_found'     => true,
+			'post_type'              => $current_provider,
 		];
 
+		$args = $this->modify_filter_args( $args );
 		$args = $this->modify_sort_args( $args );
 
+		if ( isset( $args['meta_query'] ) && count( $args['meta_query'] ) > 1 ) {
+			$args['meta_query']['relation'] = 'AND';
+		}
```
dpanta94, line 673: I would do this inside the method i am modifying the `meta_query`. Just so i am sure, i didnt modify the `relation` if i didn't modify the `meta_query`

src/Tickets/Admin/All_Tickets/List_Table.php
```diff
@@ -541,14 +690,22 @@ public function get_query_args(): array {
...
-		$args               = $this->get_query_args();
-		$tickets_repository = tribe_tickets()->by_args( $args );
-		$total_items        = $tickets_repository->found();
-		$items              = $tickets_repository->all();
+		add_filter( 'posts_clauses', [ $this, 'filter_query_clauses' ], 10, 2 );
+
+		$args        = $this->get_query_args();
+		$this->query = new WP_Query( $args );
+		$total_items = $this->query->found_posts;
+		$items       = $this->query->posts;
```
dpanta94, line 699: I don't want to cause any big changes here, but this is starting to look a lot more like the `WP_Posts_List_Table` class. Could you please have a look into extending this one instead of `WP_List_Table` ? Maybe worth it to have a chat about it.

If the change is not worth it, just let it as is. I believe though, that we could take advantage of functionalities being already offered by the Post List Table class.

dpanta94 (changes requested):
Looks good! Left a few comments

tests/integration/TEC/Tickets/Admin/All_Tickets/__snapshots__/List_TableTest__test_filters__filter by discounted__0.snapshot.json
codingmusician: This is actually correct. I purposely made no discounted or past tickets so the tests would be blank.

src/admin-views/all-tickets/filters/status-select.php
codingmusician: `status` is a little confusing in this case. They are not WP statuses. We have a set list of them. They are listed [here](https://github.com/the-events-calendar/event-tickets/blob/d11100b18c16d90a33bf1cce49362f9d88947ef5/src/Tickets/Admin/All_Tickets/List_Table.php#L901-L907). `All` is included as an option.

tests/integration/TEC/Tickets/Admin/All_Tickets/__snapshots__/PageTest__test_render_tec_tickets_all_tickets_page__0.snapshot.html
codingmusician: See my comment about the statuses above. They are not in the DB. Most of them are based on the start/end sales dats (past, upcoming, active, etc.).

src/Tickets/Admin/All_Tickets/List_Table.php
codingmusician: Good idea. We should probably check the value against the array of providers to make sure it's valid, too. Thanks!

src/Tickets/Admin/All_Tickets/List_Table.php
codingmusician: I thought about this, but because we're displaying multiple `post_types`, I ultimately went against it. `WP_Posts_List_Table` seemed to be specifically designed for a single `post_type`.

bordoni (changes requested):
I would love to not call this All Tickets, and move more towards just Tickets. Since for all intents and purposes all Admin Ticket stuff should live in that folder.

src/Tickets/Admin/All_Tickets/List_Table.php
dpanta94: That's a good point. Yea if we need to have different post types in the same list then `WP_List_Table` is a better option

dpanta94 (commented):
Looks good! Would have approved if not for Gustavo's comment!

dpanta94 (approved):

src/admin-views/all-tickets/filters.php
```diff
@@ -0,0 +1,22 @@
...
+ * @var string            $current_filter  Currently selected filter.
+ * @var string            $search_id       ID for the search input element.
+ * @var string            $search_value    Current search value.
+ */
+
+?>
+<div class="alignleft actions">
+	<?php $this->template( 'all-tickets/filters/status-select' ); ?>
+	<?php $this->template( 'all-tickets/filters/provider-select' ); ?>
+	<?php $this->template( 'all-tickets/filters/search-input' ); ?>
```
bordoni, line 20: Same here, lets move away from `all-tickets` directory.

bordoni (changes requested):

bordoni (approved):

PR #3183, opened 2024-08-20
[SL-102] Calculate event availability accounting for Seating

dpanta94:
### 🎫 Ticket

[SL-102]

### 🗒️ Description

An event's availability was being calculated by checking the event's related meta key. When seating is enabled and used for that event, this functionality needs to be adjusted in order to take into consideration:

- Different Tickets being sold for the same ticket type
- Seats being blocked even though the map has the seats available in other layouts.

So for seating enabled events we calculate the event's availability as the total of each seat's type availability.

#### Additionally

Fixes issue with capacity when purchasing different tickets of the same seating type during one order.

@lucatume after you see the below artifact let me know please why i saw the session expired during the end (and the order still went through) if you can think of something. I didn't take more than 15 mins to complete the checkout session for sure.

### 🎥 Artifacts 
https://drive.google.com/file/d/1QL4YqlynE3NdTyok25RnB9Jdy4yJR8Wz/view?usp=drive_link

### ✔️ Checklist

[SL-102]: https://stellarwp.atlassian.net/browse/SL-102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Seating/Frontend.php
```diff
@@ -116,31 +116,33 @@ public function print_tickets_block( $html, $file, $name, $template, $context ):
...
 		$prices   = [];
-		$provider = Tickets::get_event_ticket_provider_object( $post_id );
-		foreach ( tribe_tickets()->where( 'event', $post_id )->get_ids( true ) as $ticket_id ) {
-			$ticket = $provider->get_ticket( $post_id, $ticket_id );
-			if ( ! $ticket ) {
-				continue;
-			}
-			$prices[] = $ticket->price;
-		}
+		$provider = Tickets::get_event_ticket_provider_object( $post_id ); // @todo - i see the below method can return false but we are never checking against it.
```
lucatume, line 119: If `$provider` is `false`, then it should be handled like not having tickets.

src/Tickets/Seating/Frontend.php
```diff
@@ -116,31 +116,33 @@ public function print_tickets_block( $html, $file, $name, $template, $context ):
...
+			array_map(
+				static function ( $ticket_id ) use ( $provider, $post_id ) {
+					$ticket = $provider->get_ticket( $post_id, $ticket_id );
+					if ( ! $ticket ) {
+						return false;
+					}
+
+					return $ticket;
+				},
+				tribe_tickets()->where( 'event', $post_id )->get_ids()
```
lucatume, line 131: `->get_ids()` will run an unbounded query. 
Whenever you're using the repository (not only for tickets), use a Generator that will run batches of queries.
That's the reason why the loop was using `foreach` originally; restore the `foreach` loop.

src/Tickets/Seating/Frontend.php
```diff
@@ -116,31 +116,33 @@ public function print_tickets_block( $html, $file, $name, $template, $context ):
...
+			)
+		);
 
-		if ( ! count( $prices ) ) {
+		if ( empty( $tickets ) ) {
 			// Why are we here at all?
 			return $html;
 		}
 
+		$prices     = wp_list_pluck( $tickets, 'price' );
```
lucatume, line 140: Restoring the `foreach` loop this will not be required. It's just more efficient to do that in the loop.

src/Tickets/Seating/Frontend.php
```diff
@@ -167,6 +169,50 @@ public function print_tickets_block( $html, $file, $name, $template, $context ):
...
+			array_map(
+				static function ( $ticket_id ) use ( $provider, $event_id ) {
+					$ticket = $provider->get_ticket( $event_id, $ticket_id );
+					if ( ! $ticket ) {
+						return false;
+					}
+
+					return $ticket;
+				},
+				tribe_tickets()->where( 'event', $event_id )->get_ids()
```
lucatume, line 198: Here as well, use a `foreach` loop and `->get_ids(true)` to use the generator.

src/Tickets/Seating/Frontend.php
```diff
@@ -167,6 +169,50 @@ public function print_tickets_block( $html, $file, $name, $template, $context ):
...
+
+		if ( empty( $tickets ) ) {
+			return 0;
+		}
+
+		$available = [];
+
+		foreach ( $tickets as $ticket ) {
+			// The array's keys are the seating types. In order for us to calculate the stock per type and NOT per ticket.
+			$available[ get_post_meta( $ticket->ID, Meta::META_KEY_SEAT_TYPE, true ) ] = $ticket->stock();
```
lucatume, line 210: This operation can be done in the `foreach` loop.

tests/_data/seating.sql
lucatume, line 1: What is the purpose of this file?

lucatume (changes requested):

lucatume (commented):
In your screencast, the count should go from 94 to 91, but goes to 92. Did you track that down? If that is solved, could you provide an updated screencast?

tests/_data/seating.sql
dpanta94: I deleted this since we are not using it currently.

src/Tickets/Seating/Commerce/Controller.php
```diff
@@ -48,6 +56,68 @@ public function unregister(): void {
...
+			return $stock;
+		}
+
+		$provider = Tickets::get_event_ticket_provider_object( $event->ID );
+
+		if ( ! $provider ) {
+			return $stock;
+		}
+
+		foreach ( tribe_tickets()->where( 'event', $event->ID )->get_ids( true ) as $ticket_id ) {
```
rafsuntaskin, line 101:
```suggestion
		foreach ( tribe_tickets()->where( 'event', $event->ID )->get_ids( true ) as $ticket_id ) {
```

Let's use `tec_tc_tickets` instead as this only impacts TicketsCommerce..

src/Tickets/Seating/Commerce/Controller.php
```diff
@@ -48,6 +56,68 @@ public function unregister(): void {
...
+		if ( ! $event instanceof WP_Post || ! $event->ID ) {
+			return $stock;
+		}
+
+		$ticket_seat_key = get_post_meta( $ticket->ID, Meta::META_KEY_SEAT_TYPE, true );
+
+		// Not a seating ticket. We should not modify the stock.
+		if ( ! $ticket_seat_key ) {
+			return $stock;
+		}
```
rafsuntaskin, line 93: Moving this to the top will reduce unwanted operations.

src/Tickets/Seating/Commerce/Controller.php
rafsuntaskin: We may also use `meta_not_equals` clause to filter the same seated key tickets.

src/Tribe/Ticket_Object.php
```diff
@@ -706,6 +709,10 @@ public function inventory() {
 			}
 
 			$inventory = min( $inventory );
+
+			// Prevent inventory higher than the stock.
+			$inventory = min( $inventory, $this->stock() );
```
rafsuntaskin, line 714: Can't we just update the previous line? We are already doing `min` before this line.

rafsuntaskin (changes requested):

src/Tickets/Commerce/Flag_Actions/Decrease_Stock.php
```diff
@@ -105,7 +105,18 @@ public function handle( Status_Interface $new_status, $old_status, \WP_Post $pos
...
+			 *
+			 * @since TBD
+			 *
+			 * @param int      $update_stock The updated stock value.
+			 * @param Ticket   $ticket       The ticket object.
+			 * @param \WP_Post $post         The order post object.
+			 */
+			$update_stock = apply_filters( 'tec_tickets_commerce_decrease_ticket_stock', $ticket->stock() - $quantity, $ticket, $post );
+
+			update_post_meta( $ticket->ID, Ticket::$stock_meta_key, $update_stock );
```
lucatume, line 119: This method, as well as the other one in the `Increase_Stock` action, are very specific to Tickets Commerce.
I would suggest a more general-purpose method, still based around the sync of the `_stock` meta key (which is the same for all providers) and "sitting" on the updates of the `_stock` meta key.

The logic would not change: when an update to the `_stock` meta key comes in for an ASC ticket, set the stock of all other ASC tickets sharing the seat type (the one dictating the initial `_stock` value) to the new value.
Remove the filtering method temporarily to avoid recursions while running `update_post_meta( $id, '_stock', $value )` on the other tickets, and keep invalidating the cache like you're doing now.

src/Tickets/Seating/Commerce/Controller.php
```diff
@@ -34,6 +39,9 @@ protected function do_register(): void {
 			'tec_tickets_seating_timer_token_object_id_entries',
 			[ $this, 'filter_timer_token_object_id_entries' ],
 		);
+
+		add_filter( 'tec_tickets_commerce_increase_ticket_stock', [ $this, 'sync_seated_tickets_stock' ], 10, 3 );
```
lucatume, line 43: As suggested in my other comment, I would generalize this by watching for updates to the `_stock` meta key.

lucatume (changes requested):
Good idea, and good catch!
I would just make the solution more general purpose.

src/Tickets/Seating/Commerce/Controller.php
dpanta94: Will modify to impact other providers as well, so no longer revelant :D

tests/_data/seating.sql
dpanta94: The file was for playwright tests, has been restored

PR #3200, opened 2024-08-27
Fixing capacity issues

dpanta94:
### 🎫 Ticket

Issue 81 from spreadsheet.

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Seating/Commerce/Controller.php
+++ b/src/Tickets/Seating/Commerce/Controller.php
@@ -137,6 +137,13 @@ class Controller extends Controller_Contract {
 			return;
 		}
 
+		$stock = (int) $meta_value;
+
+		if ( 0 > $stock ) {
+			// We are not syncing bugs. Seats can be infinite.
+			return;
+		}
+
 		$seat_type = get_post_meta( $object_id, Meta::META_KEY_SEAT_TYPE, true );
 
 		// Not a seating ticket. We should not modify the stock.
@@ -155,8 +162,6 @@ class Controller extends Controller_Contract {
 		if ( ! $event instanceof WP_Post || ! $event->ID ) {
 			return;
 		}
-
-		$stock = (int) $meta_value;
 
 		// Remove the action to avoid infinite loops.
 		remove_action( 'tec_tickets_commerce_increase_ticket_stock', [ $this, 'sync_seated_tickets_stock' ] );
--- a/src/Tickets/Seating/Commerce/Controller.php
+++ b/src/Tickets/Seating/Commerce/Controller.php
@@ -164,7 +164,7 @@ class Controller extends Controller_Contract {
 		}
 
 		// Remove the action to avoid infinite loops.
-		remove_action( 'tec_tickets_commerce_increase_ticket_stock', [ $this, 'sync_seated_tickets_stock' ] );
+		remove_action( 'updated_postmeta', [ $this, 'sync_seated_tickets_stock' ] );
 
 		foreach (
 			tribe_tickets()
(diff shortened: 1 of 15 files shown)


src/Tribe/Editor/REST/V1/Endpoints/Single_Ticket.php
```diff
@@ -328,7 +328,7 @@ public function add_ticket( WP_REST_Request $request, $nonce_action ) {
 		 * @param object              $ticket      Ticket that was just added.
 		 * @param array<string,mixed> $ticket_data The body of the request.
 		 */
-		do_action( 'tribe_tickets_ticket_added', $post_id, $ticket_id, $ticket_data );
+		do_action( 'tribe_tickets_ticket_added', $post_id, $ticket, $ticket_data );
```
bordoni, line 331: I am not sure about this change here, since someone could be using this hook already, expecting an ID, and it suddenly is an object now?

bordoni (changes requested):

src/Tickets/Seating/Commerce/Controller.php
```diff
@@ -137,6 +137,13 @@ public function sync_seated_tickets_stock( $meta_id, $object_id, $meta_key, $met
 			return;
 		}
 
+		$stock = (int) $meta_value;
+
+		if ( 0 > $stock ) {
+			// We are not syncing bugs. Seats can be infinite.
```
rafsuntaskin, line 143: I think you meant Seats can NOT be infinite?

rafsuntaskin: Also, we should move this after the seat type check below to avoid interference with regular tickets.

src/Tribe/Editor/REST/V1/Endpoints/Single_Ticket.php
rafsuntaskin: @dpanta94 the doc block needs to be updated here, the `$ticket_id` should be fine.

src/Tickets/Seating/Commerce/Controller.php
dpanta94: Yes i meant not infinite

dpanta94: For the other comment, i think being above the seat check is better. It's a faster check so it should come before slower checks. It can quickly tell us that its not about seating if its `-1`

src/Tribe/Editor/REST/V1/Endpoints/Single_Ticket.php
dpanta94: @bordoni @rafsuntaskin `$ticket` is still the ticket id not an object. The difference is, that prior `$ticekt_id` would only get a value when the operation was about an update and not about create. `$ticket` stores the ticket id in both cases

rafsuntaskin: @dpanta94 https://github.com/the-events-calendar/event-tickets/pull/3200/files#r1734907208

src/Tickets/Seating/Commerce/Controller.php
rafsuntaskin: @dpanta94 you are typecasting the value so it may not be the original one that is now returned. There are tests failing for this.

tests/wpunit/Tribe/Tickets/Template_TagsTest.php
```diff
@@ -1124,6 +1124,10 @@ public function it_should_delete_capacity_from_a_ticket() {
 	 */
 	public function it_should_delete_capacity_from_an_event() {
 		$event_id  = $this->factory()->event->create();
+		$tickets_handler = tribe( 'tickets.handler' );
+
+		update_post_meta( $event_id, $tickets_handler->key_capacity, 20 );
```
lucatume, line 1129: Will existing posts/events have this key correctly set?

dpanta94:
Closing in favot of #3206

PR #3208, opened 2024-08-29
Fixing capacity totals in the block editor

dpanta94:
### 🎫 Ticket

Issue 90 from gsheet


dpanta94:
Tests being added in #3211

PR #3214, opened 2024-09-02
Hide progress bar quantity when ticket is ASC

dpanta94:
### 🎫 Ticket

Issue 89 from gsheet

### DIFF
--- a/src/Tickets/Blocks/Ticket/app/editor/template.js
+++ b/src/Tickets/Blocks/Ticket/app/editor/template.js
@@ -17,6 +17,7 @@ import './style.pcss';
 import TicketContainer from './container/container';
 import TicketDashboard from './dashboard/container';
 import MoveModal from '@moderntribe/tickets/elements/move-modal';
+import { applyFilters } from '@wordpress/hooks';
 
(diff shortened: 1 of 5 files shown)


src/Tickets/Blocks/Ticket/app/editor/template.js
```diff
@@ -52,26 +53,39 @@ class Ticket extends PureComponent {
...
+							'tribe-editor__ticket--has-tickets-plus':
+								hasTicketsPlus,
+						},
+						{
+							'tribe-editor__ticket--is-asc': applyFilters(
+								'tribe.editor.ticket.isAsc',
+								false,
+								clientId
+							),
+						}
```
rafsuntaskin, line 73: @dpanta94 we try to avoid putting integration/feature codes directly into the ET code as much as possible.

So, I would suggest to have the classNames filtered and then add the asc class using the filter.

dpanta94: @rafsuntaskin can do that not a problem. WIll change that ;)

PR #3218, opened 2024-09-03
Tickets Home > Optimize DB Queries

codingmusician:
### 🎫 Ticket
[ET-2212]

### 🗒️ Description
Optimizing the queries for enhancing performance.

### 🎥 Artifacts 
Before Optimization:
image

After Optimization:
image

### ✔️ Checklist

[ET-2212]: https://stellarwp.atlassian.net/browse/ET-2212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Commerce/Ticket.php
```diff
@@ -181,6 +190,18 @@ class Ticket {
...
+	 * Sets the pending attendee count for a ticket id.
+	 *
+	 * @since TBD
+	 *
+	 * @param int $ticket_id      The ticket ID.
+	 * @param int $attendee_count The number of attendees.
+	 */
+	public static function set_attendees_by_ticket( $ticket_id, $attendee_count ) {
+		static::$pending_attendees_by_ticket[ $ticket_id ] = $attendee_count;
+	}
```
bordoni, line 203: We need to improve the name of these variables and methods, as of right now this will cause confusion.

We need to allow attendees by ticket to be stored for each one of the possible statuses, which this will not allow.

So potentially store attendees by ticket and each status as a new index, so we can get all the counts per ticket ID.

src/Tickets/Commerce/Ticket.php
```diff
@@ -396,7 +417,7 @@ public function get_ticket( $ticket_id ) {
 			$return->global_stock_cap( $capped );
 		}
 
-		$qty_cancelled = $this->get_cancelled( $ticket_id );
+		$qty_cancelled = 0;
```
bordoni, line 420: Dont stop using the method we still need to fetch how many were cancelled.

src/Tickets/Commerce/Ticket.php
```diff
@@ -431,26 +452,17 @@ public function get_ticket( $ticket_id ) {
...
-			'post_status'    => Denied::SLUG,
-			'posts_per_page' => - 1,
-		), [
-			'items',
-		] );
-
-		$denied = 0;
-		foreach ( $denied_orders as $denied_order ) {
-			$denied += $denied_order->get_item_quantity( $ticket_id );
-		}
```
bordoni, line 451: We still need to fetch this data, even if the existing one was fetched incorrectly. The method is not deprecated, we just need to fix it.

src/Tickets/Commerce/Module.php
```diff
@@ -214,6 +214,37 @@ public function __construct() {
...
+		}
+
+		// If no attendee, bail.
+		if ( ! $attendee ) {
+			return;
+		}
+
+		$this->attendees_by_ticket_id[ $ticket_id ][] = $attendee;
+	}
+
```
bordoni, line 247: Do not add new methods to the Module. This needs to live in the Commerce Attendee class.

src/Tickets/Commerce/Module.php
```diff
@@ -350,6 +381,30 @@ public function get_attendees_by_id( $post_id, $post_type = null ) {
...
+		} else {
+			/** @var Tribe__Tickets__Attendee_Repository $repository */
+			$repository = tec_tc_attendees( $this->orm_provider );
+			$attendees  = $repository->by( 'ticket_id', $ticket_id )->all();
+
+			$this->attendees_by_ticket_id[ $ticket_id ] = $attendees;
+		}
+
+		return $this->get_attendees_from_module( $attendees );
+	}
```
bordoni, line 406: This method should live in the Attendees class.

bordoni (changes requested):

src/Tickets/Commerce/Module.php
codingmusician: @bordoni The `get_attendees_by_ticket_id()` method is actually a protected method that exists in the parent class we're extending here. In `Tribe__Tickets__Tickets`, the `get_attendees_by_ticket_id()` method tries to use `tribe_attendees()` to find the attendee by querying all of the different repositories. I wanted to use `tec_tc_attendees()` and only search the Tickets Commerce repo.

src/Tickets/Commerce/Module.php
codingmusician: I actually added this to the main `Tribe__Tickets__Tickets` class so it could be used within any of the providers.

src/Tickets/Commerce/Module.php
bordoni: Where is that being used? I mean the method: `get_attendees_by_ticket_id`.

I think the code is good, but I want to reduce reliance on the Module class, look at the method `get_price_html` from this class, it's just a bypass to a separate class.

codingmusician: Here's where it is defined within `Tribe__Tickets__Tickets`:
https://github.com/the-events-calendar/event-tickets/blob/54b05bd623d88688b9ed8c086d3406c46a5cb22b/src/Tribe/Tickets.php#L874-L879

Here's where it is being used within TC `Module`:
https://github.com/the-events-calendar/event-tickets/blob/54b05bd623d88688b9ed8c086d3406c46a5cb22b/src/Tickets/Commerce/Module.php#L342

The method within `Tribe__Tickets__Tickets` uses `tribe_attendees()`, however there is not a way to only reference TC attendees. It queries all attendee post types. By overriding and using `tec_tc_attendees()` we can narrow it down to only TC attendees.

bordoni: Oh yeah for sure, that is correct. What I am suggesting is the following;

Classes `Tribe__Tickets__Tickets` and any child of it needs to be reduced in complexity at all costs, it is currently why it's so hard to maintain the codebase in Event Tickets.

`\TEC\Tickets\Commerce\Module` currently extends `Tribe__Tickets__Tickets` which is why your code added changes to both. 

What I am suggesting is the following; Move both of the methods you added to a Commerce specific class, like the `\TEC\Tickets\Commerce\Ticket` than make the changes you suggested of routing the Tickets Commerce method to make sure we are not using the wrong provider.

These methods "aliases" need to be kept as simple as possible so we can potentially migrate the codebase out of using them.

If you feel like the Ticket class in the Commerce folder might not be the correct place, feel free to create a new class for Ticket Memoization if we need it.

Camwyn: My 2¢
> ...tries to use tribe_attendees() to find the attendee by querying all of the different repositories

`tribe_attendees` [uses the ticket's ORM provider already](https://github.com/the-events-calendar/event-tickets/blob/aedc8a68be32a4706092b62c3558554cdaffa8c2/src/template-tags/tickets.php#L1294)

```
function tribe_attendees( $repository = 'default' ) {...}
```

So I'm not seeing the benefit here. Unless you are addressing a bug? Maybe the bug is that we have a separate `tec_tc_attendees` function instead of adding it to `tribe_attendees` via the filter as we do for all other providers?

That being said, the memoization should be at the "base" (parent/abstract/trait) layer unless there is good reason to do otherwise. We want to minimize the amount of things we need to override in children as much as possible.

codingmusician: @Camwyn `tribe_attendees()` attempts to find attendees from all repositories (there isn't an option to only query Tickets Commerce), which creates unnecessary queries. By using `tec_tc_attendees()`, it only queries Tickets Commerce queries. The only time the `get_attendees_by_ticket_id()` method is being used is from within the `TEC\Tickets\Commerce\Module` class, which would only be querying Tickets Commerce attendees.

src/Tickets/Admin/Tickets/List_Table.php
```diff
@@ -389,7 +419,7 @@ public function column_name( $item ): string {
 			return esc_html( $item->name );
 		}
 
-		$edit_post_url  = get_edit_post_link( $event );
+		$edit_post_url  = $this->get_event_edit_url( $event->ID );
 		$edit_post_link = sprintf(
 			'<a href="%s" class="tec-tickets-admin-tickets-table-event-link" target="_blank" rel="nofollow noopener">%s</a>',
 			esc_url( $edit_post_url ),
```
Camwyn, line 425: Any reason to not jsut do:
```suggestion
		$edit_post_link = sprintf(
			'<a href="%s" class="tec-tickets-admin-tickets-table-event-link" target="_blank" rel="nofollow noopener">%s</a>',
			esc_url( $this->get_event_edit_url( $event->ID ) ),
```

src/Tickets/Commerce/Memoize_Attendees.php
```diff
@@ -0,0 +1,131 @@
...
+namespace TEC\Tickets\Commerce;
+
+/**
+ * Class Memoize_Attendees
+ *
+ * @since TBD
+ *
+ * @package TEC\Tickets\Admin\Commerce
+ */
+class Memoize_Attendees {
```
bordoni, line 19: I am wondering if we need an extra class for this, why no use `tribe_cache()` and store as two separate arrays there, it might make it easier for tests because we can tell it to ignore cache.

bordoni (changes requested):

src/Tickets/Admin/Tickets/List_Table.php
```diff
@@ -949,6 +981,72 @@ public function get_query_args() {
...
+				'update_post_meta_cache' => true,
+			]
+		);
+
+		// Only continue if we're looking at Tickets Commerce tickets.
+		if ( Page::get_current_provider() !== TicketsCommerce\Module::class ) {
+			return;
+		}
+
+		// @todo @codingmusician - Add query priming solutions for other providers.
```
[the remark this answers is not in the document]
codingmusician: Right. That's what I'm thinking, too. For now, Gustavo said to just worry about Tickets Commerce to get this out the door. We will optimize other providers in phase 2.

src/Tickets/Admin/Tickets/Page.php
```diff
@@ -74,7 +74,75 @@ public static function get_provider_options() {
...
+	/**
+	 * Get the currently selected provider.
+	 *
+	 * @since TBD
+	 *
+	 * @return string;
+	 */
+	public static function get_current_provider() {
+		$provider_info    = static::get_provider_info();
+		$default_provider = empty( $provider_info ) ? '' : addslashes( key( $provider_info ) );
```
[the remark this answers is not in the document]
codingmusician: I don't think we care what the default is, but we want to use one that is enabled (Tickets Commerce may not be).

src/Tickets/Commerce/Memoize_Attendees.php
codingmusician: The idea came out of a discussion we were having about it. We can use `tribe_cache()` though. No problem.

Camwyn (approved):

bordoni (approved):

PR #3219, opened 2024-09-03
Moved attendees stock issue

pattihis:
### 🎫 Ticket

[ET-2098]

### 🗒️ Description
When attendees were moved from one ticket to another the stock/sales meta was not updated properly.

### ✔️ Checklist

[ET-2098]: https://stellarwp.atlassian.net/browse/ET-2098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tribe/Tickets_Handler.php
```diff
@@ -880,9 +880,13 @@ public function get_ticket_totals( $ticket ) {
 
 		$totals = array_map( 'intval', $totals );
 
-		// Remove Pending from total
+		// Remove Pending from total.
 		$totals['sold'] -= $totals['pending'];
 
+		// Update Stock.
+		$totals['stock'] = tribe_tickets_get_capacity( $ticket->ID ) - $totals['sold'];
```
dpanta94, line 887: `tribe_tickets_get_capacity` should calculate RSVPs as well. So i believe the logic is not right.

Additionally shouldn't we subtract `$totals['pending']` as well since they are no longer in the stock ?

src/Tribe/Tickets_Handler.php
```diff
@@ -880,9 +880,13 @@ public function get_ticket_totals( $ticket ) {
...
+		update_post_meta( $ticket->ID, '_stock', $totals['stock'] );
```
dpanta94, line 888: Can we find a more appropriate place for such an operation ? Like when the attendee is moved. I would not expect any write operations to be part of a method `get_ticket_totals`

dpanta94 (changes requested):
Left some comments. Also please, we need extensive tests when we edit capacity related fields. This could affect the work we are doing in the SLR as well, so let's cover a lot of possibilities with integration tests

src/Tribe/Tickets_Handler.php
pattihis: This is the `get_ticket_totals()` functions, it works for RSVPs as well.

The 'pending' have been subtracted in the line above, or am I missing something?

src/Tribe/Tickets_Handler.php
pattihis: During the move there are are a couple of update posts, that are wrong because this function returns wrong totals. Have [a look here](https://github.com/the-events-calendar/event-tickets/blob/master/src/Tribe/Admin/Move_Tickets.php#L637)

pattihis:
@dpanta94 tests will of course be added, after the solution has been accepted.

src/Tribe/Tickets_Handler.php
dpanta94: Yes because they have been substracted from the `$totals['sold']` they need to be substracted from the `$totals['stock']` again or move line 884 after 887

src/Tribe/Tickets_Handler.php
dpanta94: This placement is definitely problematic @pattihis. It can very easily cause infinite loops. see method `migrate_object_capacity` which can be called by `tribe_tickets_get_capacity` which in turn calls `get_ticket_totals`. I would look for another place for such an operation during a different hook

dpanta94: If you check CI tests are failing because of infinite loop as well currently

src/Tribe/Admin/Move_Tickets.php
```diff
@@ -672,8 +672,8 @@ public function move_tickets( array $ticket_ids, $tgt_ticket_type_id, $src_event
 			if ( 'Tribe__Tickets__RSVP' === $ticket['provider'] ) {
 				$src_stock ++;
 				$tgt_stock --;
-				update_post_meta( $src_ticket_type_id, '_stock', $src_stock );
-				update_post_meta( $tgt_ticket_type_id, '_stock', $tgt_stock );
```
bordoni, line 676: So we dont change the ticket stock too? Just the Event global Stock?

bordoni (changes requested):

src/Tribe/Admin/Move_Tickets.php
```diff
@@ -672,8 +672,8 @@ public function move_tickets( array $ticket_ids, $tgt_ticket_type_id, $src_event
...
+				update_post_meta( $src_event_id, '_stock', $src_stock );
+				update_post_meta( $tgt_event_id, '_stock', $tgt_stock );
```
dpanta94, line 676: @pattihis i think that the meta `_stock` is relevant only at a ticket. Not the event itself. The event has different meta to evaluate its total capacity (see global stock consts, and tickets handler key_capacity property.

dpanta94 (changes requested):
I read a bit on the issue, and i think (just by reading though i didnt dig into it) that we have issues on other fields rather than the stock. Maybe on total_sales ?

I would watch the db and the relevant fields (of tickets and event) to try and determine what is not goign as expected.

src/Tribe/Admin/Move_Tickets.php
dpanta94: @bordoni The '_stock' meta, i think doesn;t have any value in an ticket-able post type. They use diffenent meta for that. I think its `tribe( 'tickets.handler' )->key_capacity` and `Global_Stock::GLOBAL_STOCK_LEVEL`.

pattihis: @bordoni maybe we need to update both meta

src/Tribe/Admin/Move_Tickets.php
```diff
@@ -662,18 +657,15 @@ public function move_tickets( array $ticket_ids, $tgt_ticket_type_id, $src_event
...
 
-			// adjust sales numbers - don't allow negatives
-			$src_qty_sold --;
-			$tgt_qty_sold ++;
-			update_post_meta( $src_ticket_type_id, 'total_sales', $src_qty_sold );
-			update_post_meta( $tgt_ticket_type_id, 'total_sales', $tgt_qty_sold );
-
-			//adjust stock numbers for RSVP Tickets
+			// adjust stock numbers for RSVP and Commerce Tickets.
 			if ( 'Tribe__Tickets__RSVP' === $ticket['provider'] ) {
```
stratease, line 661:
```suggestion
			if ( 'Tribe__Tickets__RSVP' === $ticket['provider'] ) {
```
Looking at this function I'm wondering if it is time to refactor so it isn't so big and logic heavy. We are about 200 lines long so far.

Are there points of diverging logic like this check, that we can separate into `do_rsvp_stuff()`. Don't force it but if you see some quick wins here that would be good toward improving maintainability in the future.

pattihis: Definitely agree @stratease . 

This was a bit confusing for me. Even variable naming was off, imo. `$ticket_id` was ref to the Attendee instead of the Ticket, which was named `$product_id`... took me some time to break down all the logic.

My fix now is towards that direction: to use the methods from RSVP/Ticket classes, instead of updating a bunch of meta fields directly.

Also, [this conditional here](https://github.com/the-events-calendar/event-tickets/pull/3219/files#diff-98b75fdcf668fd80207d85d830c6fcb599c8bed942f16edcefed124f460ac54aR632), and all updates after it, should be centralized under `increase_rsvp()` or `increase_ticket()`

What do you think?

stratease (dismissed):
Thanks for cleanup.

stratease (approved):

stratease (approved):

PR #3221, opened 2024-09-04
Fix state's seat type by post id

dpanta94:
### 🎫 Ticket

Fixing an issue we had with the state's seatTypesByPostId property.

When that property was being set, the ticket id was not existent yet. So we are udpating it during ticket's save/update hooks.


src/Tickets/Seating/app/blockEditor/store/index.js
```diff
@@ -104,7 +111,26 @@ const store = createReduxStore(storeName, {
...
-						[ticketPostId]: action.seatTypeId,
+						[ticketPostId || action.clientId]: action.seatTypeId,
+					},
+				};
+			case 'SET_TICKET_SEAT_TYPE_BY_POST_ID':
+				const ticketId = getTicketIdFromCommonStore(action.clientId);
+
+				const { seatTypesByPostId, seatTypesByClientId } = state;
+
+				const seatTypeId =
```
lucatume, line 122: I'm not sure about the lookup order here: shouldn't the ticket ID be looked up first, then the post, then the client ID? I.e. I _think_ the order should be the reverse of what it is now.

src/modules/data/blocks/ticket/sagas.js
```diff
@@ -675,6 +675,8 @@ export function* createNewTicket( action ) {
 				put( actions.setTicketHasChanges( clientId, false ) ),
 			] );
 
+			doAction( 'tec.tickets.blocks.ticketCreated', clientId );
```
lucatume, line 678: Document this action.

src/modules/data/blocks/ticket/sagas.js
```diff
@@ -817,6 +819,8 @@ export function* updateTicket( action ) {
 				put( actions.setTicketTempSaleEndDateInput( clientId, saleEndDateInput ) ),
 				put( actions.setTicketTempSaleEndDateMoment( clientId, saleEndDateMoment ) ),
 			] );
+
+			doAction( 'tec.tickets.blocks.ticketUpdated', clientId );
```
lucatume, line 823: Document this action.

src/Tickets/Seating/app/blockEditor/store/index.js
```diff
@@ -104,7 +111,26 @@ const store = createReduxStore(storeName, {
...
+					seatTypesByClientId[action.clientId] ||
+					seatTypesByPostId[action.clientId] ||
+					seatTypesByPostId[ticketId];
+
+				delete seatTypesByPostId[action.clientId];
```
lucatume, line 127: Why remove this and not **update** it?

src/Tickets/Seating/app/blockEditor/update-states-seat-type-by-ticket-id.js
```diff
@@ -0,0 +1,6 @@
+import { storeName } from './store';
+import { dispatch } from '@wordpress/data';
+
+export const useUpdateStatesSeatTypeByTicketId = (clientId) => {
```
lucatume, line 4: The `use` part of the name is suggestive of a React hook, since this code is running in the context of a React application this  might be confusing.

I suggest `setSeatTypeForTicket` as an alternative.

But then: one file, for one function with one line of code; does it make sense? I would just call `dispatch(...` directly in the  code. If the code needs to become more complex than a dispatch, then that can be extracted into a function in a separate file and so on.

src/Tickets/Seating/app/blockEditor/filters.js
```diff
@@ -120,3 +121,15 @@ addFilter(
 	'tec.tickets.seating',
 	filterTicketIsAsc
 );
+
+addAction(
+	'tec.tickets.blocks.ticketUpdated',
+	'tec.tickets.seating',
+	useUpdateStatesSeatTypeByTicketId
```
lucatume, line 128: See my note in the file defining this function to inline the `dispatch` call directly.

lucatume (changes requested):

src/Tickets/Seating/app/blockEditor/update-states-seat-type-by-ticket-id.js
dpanta94: @lucatume i agree it would be better having it in the filter.js since this is a one liner. Regarding tests, i've had issues trying to import anything from the filters.js file in a test file.

src/Tickets/Seating/app/blockEditor/store/index.js
dpanta94: I think the current approach is correct. Let me explain my reasoning.

1. If the seat type id is present in the `seatTypesByClientId` property this is the more recent since this gets updated in every change. 
2. Otherwise, we check `seatTypesByPostId` in the clientId offset since this is being set up along with the previous value. - making this unnecessary as a result
3. We check in `seatTypesByPostId` for when the value is not present `seatTypesByClientId` - e.g. during initial load of the block.

src/Tickets/Seating/app/blockEditor/store/index.js
dpanta94: We could also skip setting a value inside `seatTypesByPostId` until we have a ticket id.

tests/slr_jest/blockEditor/hook-callbacks.spec.js
```diff
@@ -0,0 +1,508 @@
...
+
+jest.mock('@wordpress/data', () => ({
+	select: jest.fn(),
+	createReduxStore: jest.fn(),
+	register: jest.fn(),
+	dispatch: jest.fn(),
+}));
+
+describe('setSeatTypeForTicket', () => {
+	beforeEach(() => {
```
lucatume, line 21: You can do this only once, in a wrapping `describe` block for the whole file. E.g. `describe('hook-callbacks', ...`

tests/slr_jest/blockEditor/hook-callbacks.spec.js
```diff
@@ -0,0 +1,508 @@
...
+		jest.resetModules();
+		jest.resetAllMocks();
+	});
+
+	afterEach(() => {
```
lucatume, line 26: Like the `beforeEach`, I would run the `afterEach` only once in a wrapping `describe`.

lucatume (approved):
A really small cleanup change, in the tests, then good to merge.

tests/slr_jest/blockEditor/hook-callbacks.spec.js
dpanta94: So nice to know that! Thank you!

tests/slr_jest/blockEditor/hook-callbacks.spec.js
dpanta94: Cleaned up and mergin!

PR #3226, opened 2024-09-06
Issues from gsheet rel to SL-97

dpanta94:
### 🎫 Ticket

[SL-97]

### Description
Issues from the gsheet related to SL-97

Fixing issues 72, 118, 119, 121.

While working on the above found and fixed 2 additional issues:
- Ticket quantity in AR modal even though the +/- icons were not present, user could change it by clicking on it and entering a number. (like he would for any `input[type="number"]`)
- The close modal icon was not visible in some themes (e.g. twentytwenty)

[SL-97]: https://stellarwp.atlassian.net/browse/SL-97?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Seating/Frontend/Session.php
```diff
@@ -340,4 +341,50 @@ public function confirm_all_reservations(): bool {
...
+	 * Returns the seat label for a ticket in an event's registration modal/page.
+	 *
+	 * @since TBD
+	 *
+	 * @param ?int $event_id  The event ID.
+	 * @param ?int $ticket_id The ticket ID.
+	 *
+	 * @return ?array The seat reservations for the ticket.
+	 */
+	public function get_events_registrations_ticket_seat_label( int $event_id = null, int $ticket_id = null ) {
```
lucatume, line 355: Add return type-hinting here.

src/Tickets/Seating/Frontend/Session.php
```diff
@@ -340,4 +341,50 @@ public function confirm_all_reservations(): bool {
 
 		return $confirmed;
 	}
+
+	/**
+	 * Returns the seat label for a ticket in an event's registration modal/page.
```
lucatume, line 346: This returns a list, update the doc-block to reflect that.
Also: this is the doc-block of the method, not how/where it's used; I would remove the "in an event..." part.
Finally, why not just return an array of (string) seat labels from this method (as the name suggest) in place of shuttling around the whole reservation object?

src/views/v2/tickets/item.php
```diff
@@ -117,6 +119,15 @@
 	$attributes['data-available-count'] = (string) $available_count;
 }
 
+$seat_labels = tribe( Session::class )->get_events_registrations_ticket_seat_label( $post_id, $ticket->ID );
```
lucatume, line 122: This cannot be done here. The Seating feature might or might be not registered and the ticket template should know nothing about it.

Filter the `$attributes` and hook on that from a Seating controller.

src/views/v2/tickets/item.php
```diff
@@ -117,6 +119,15 @@
...
+
+$seat_labels = implode(
```
lucatume, line 124: As mentioned above, this template should not contain any Seating feature related knowledge.

tests/ft_integration/TEC/Tickets/Flexible_Tickets/Series_Passes/__snapshots__/BaseTest__should_not_replace_tickets_block_on_events_not_in_series__0.snapshot.html
```diff
@@ -30,6 +30,7 @@ <h3 class="tribe-common-h7 tribe-tickets-notice__title">Checkout Unavailable!</h
 
 		<div
 	id="tribe-block-tickets-item-{{ticket_2}}"
+	data-seat-labels=""
```
lucatume, line 33: Per my comment above, this snapshot should not change when the Seating feature is not active.

tests/slr_integration/__snapshots__/Frontend_Test__should_display_ticket_block_when_seating_is_not_enabled__0.snapshot.html
```diff
@@ -29,6 +29,7 @@ <h2 class="tribe-common-h4 tribe-common-h--alt tribe-tickets__tickets-title">
 
 		<div
 	id="tribe-block-tickets-item-{{ticket_1}}"
+	data-seat-labels=""
```
lucatume, line 32: This should not be here when seating is not enabled.

lucatume (changes requested):
Some changes required following two notion:
1. as logic-less templates as possible
2. Seating should modify the code via filters, not by direct insertion

src/views/seating/seat-selection-timer.php
```diff
@@ -42,4 +42,4 @@
 	</div>
 	<div class="tec-tickets-seating__dialog-append-target"></div>
 </div>
-
+<?php
```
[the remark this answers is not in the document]
dpanta94: It' suggested to close php files being included with an open php tag.

src/views/v2/tickets/item.php
dpanta94: Completely forgot about that -_- Changing it :D

src/Tickets/Seating/Frontend/Session.php
lucatume: > I did that initially - then i though i was i need of the reservation id as well. So i did a small refactor. Turns out i didn't need it though, but i kept the logic because it could prove more valuble in the future to have the method return a list of all the reservation details for a specific event and ticket and the logic utilizing that to decide what's needed.

It might, and when it will be needed, it will be created. Either you need it now, or there is no point in indulging speculative engineering: add something as it _might_ be needed.

lucatume (approved):

lucatume (approved):

PR #3233, opened 2024-09-12
Fixed bleeding state - added more tests

dpanta94:
### 🎫 Ticket

Issue 110 from gsheet

### DIFF
--- a/src/Tickets/Seating/app/blockEditor/store/selectors.js
+++ b/src/Tickets/Seating/app/blockEditor/store/selectors.js
@@ -37,15 +37,27 @@ export const selectors = {
 		});
 	},
 	getCurrentLayoutId(state) {
+		if (!!!state.isUsingAssignedSeating) {
+			return null;
+		}
+
 		return state?.currentLayoutId || null;
 	},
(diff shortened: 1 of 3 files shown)


src/Tickets/Seating/app/blockEditor/store/selectors.js
```diff
@@ -37,15 +37,27 @@ export const selectors = {
 		});
 	},
 	getCurrentLayoutId(state) {
+		if (!!!state.isUsingAssignedSeating) {
```
lucatume, line 40: Here and elsewhere why not use just one `!`? `!!!` is `negation of negation of negation of state.isUsingAssignedSeating` that results in `negation of state.isUsingAssignedSeating`. Am I not getting something?

lucatume (changes requested):

src/Tickets/Seating/app/blockEditor/store/selectors.js
dpanta94: No you are right. I was not thinking it seems. Fixed

lucatume (approved):

PR #3236, opened 2024-09-16
[Issue-123] Fixing sold out view on event listing for Seated tickets.

rafsuntaskin:
### 🎫 Ticket

[issue-123]

### 🗒️ Description

* On the event listing view the seated ticket events were not showing as sold out when it was.

### 🎥 Artifacts 
📷 screenshot(s): https://share.cleanshot.com/hFTgnzFG

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Seating/Commerce/Controller.php
+++ b/src/Tickets/Seating/Commerce/Controller.php
@@ -87,6 +87,9 @@ class Controller extends Controller_Contract {
 		$tickets = tribe_tickets()
 			->where( 'event', $post_id )
 			->get_ids( true );
+	
+		$capacity_by_type   = [];
+		$total_sold_by_type = [];
 		
 		foreach ( $tickets as $ticket_id ) {
 			$ticket = Tickets::load_ticket_object( $ticket_id );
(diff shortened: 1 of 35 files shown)


src/Tickets/Seating/Commerce/Controller.php
```diff
@@ -56,8 +57,59 @@ public function unregister(): void {
...
+	 * @param int                             $post_id The post ID.
+	 *
+	 * @return array<string,array<string|int>> The types of tickets.
+	 */
+	public function set_event_stock_counts( $types, $post_id ): array {
+		$has_seating = get_post_meta( $post_id, Meta::META_KEY_ENABLED, true );
+		
+		if ( ! $has_seating ) {
+			return $types;
+		}
```
dpanta94, line 79: Why not use `tec_tickets_seating_enabled`. Important note that mentioned function is checking against `Meta::META_KEY_LAYOUT_ID` instead. If we are incosistent accross the database as to whether a post has seating enabled this can create issues. Rather use one single function for that, and if we need to change it in the future to only have to change that.

src/Tickets/Seating/Commerce/Controller.php
```diff
@@ -56,8 +57,59 @@ public function unregister(): void {
...
+			
+			if ( ! tribe_events_ticket_is_on_sale( $ticket ) ) {
+				continue;
+			}
+			
+			++$types['tickets']['count'];
+			
+			$stock_level                    = max( 0, $ticket->available() );
+			$types['tickets']['stock']     += $stock_level;
+			$types['tickets']['available'] += $stock_level;
```
dpanta94, line 108: I have doubts here. I believe you are not considering stock of tickets for the same seat type. Can you make sure that's accounted for ? Also check the method `get_events_ticket_capacity_for_seating` it's a good example of correctly calculating tickets stock of seated events.

tests/slr_integration/Commerce/Controller_Test.php
```diff
@@ -313,4 +313,85 @@ public function test_filter_timer_token_object_id_entries( Closure $fixture ): v
 			$filtered_entries,
 		);
 	}
+	
+	public function test_stock_count_for_seated_tickets() {
+		$controller = $this->make_controller();
+		$controller->register();
+		$event_id = tribe_events()->set_args(
+			[
```
dpanta94, line 321: Here, please include a scenario where we have 2 different tickets sharing the same seat type.

dpanta94 (changes requested):
Left some comments to be amended !

dpanta94 (approved):
Looks good, let's see though why the slr tests are failing!

PR #3237, opened 2024-09-16
Load the Payment Element if we have Wallets

pattihis:
### 🎫 Ticket

[ETP-942]

### 🗒️ Description
Load the full payment element if we have wallets enabled, even though only `card` is selected as payment method.

Related with https://github.com/the-events-calendar/event-tickets-plus/pull/1642

### 🎥 Artifacts 
Screenshot_1
This combination of settings should load the full Payment Element. Disabling the wallets above, would load the simple Card Element.

### ✔️ Checklist

[ETP-942]: https://stellarwp.atlassian.net/browse/ETP-942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Commerce/Gateways/Stripe/Stripe_Elements.php
```diff
@@ -69,4 +75,22 @@ public function card_element_type() {
...
+	 *
+	 * @return bool
+	 */
+	public function wallets_enabled() {
+		if ( ! defined( 'EVENT_TICKETS_PLUS_FILE' ) ) {
+			return false;
+		}
+
+		$settings = tribe( \TEC\Tickets_Plus\Commerce\Gateways\Stripe\Settings::class );
+		$wallets  = tribe_get_option( $settings::$option_checkout_element_wallets, [] );
```
dpanta94, line 92: Let's add a `use \TEC\Tickets_Plus\Commerce\Gateways\Stripe\Settings;`

No need to potentially init an object to access it's static property.
```suggestion
		$wallets  = tribe_get_option( Settings::$option_checkout_element_wallets, [] );
```

src/Tickets/Commerce/Gateways/Stripe/Stripe_Elements.php
```diff
@@ -69,4 +75,22 @@ public function card_element_type() {
...
+	 * Are Apple/Google Pay wallets enabled?
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	public function wallets_enabled() {
+		if ( ! defined( 'EVENT_TICKETS_PLUS_FILE' ) ) {
+			return false;
+		}
```
dpanta94, line 89: Stephen had shared a map of actions to check per plugin instead of defined. Maybe speak with him.

My opinion though is checking constants its the fastest solution we have in our hands, but if i am not mistaken the constants could be defined but the plugin still not fully loaded. (Min requirements failed for example)

dpanta94 (changes requested):
Requested some changes and we need some tests for this.

src/Tickets/Commerce/Gateways/Stripe/Stripe_Elements.php
```diff
@@ -54,8 +53,15 @@ public function include_payment_element() {
 
 		$payment_methods = ( new Merchant() )->get_payment_method_types();
 
-		// Don't load the Payment Element if just the Credit Card method is selected.
-		return ! ( 1 === count( $payment_methods ) && 'card' === $payment_methods[0] );
+		if ( 1 > count( $payment_methods ) ) {
```
dpanta94, line 56:
```suggestion
		if ( 1 < count( $payment_methods ) ) {
```

Shouldn't this be like that instead ?

src/Tickets/Commerce/Gateways/Stripe/Stripe_Elements.php
pattihis: Yes, @dpanta94 I thought of that but I opted not to, because we won't know if ETP will be available. 

Is it ok to use a namespaced class that might not exist? No warnings or errors will be trown?

src/Tickets/Commerce/Gateways/Stripe/Stripe_Elements.php
pattihis: Yes, this is a good catch, thank you!

src/Tickets/Commerce/Gateways/Stripe/Stripe_Elements.php
pattihis: @Camwyn Would be very interested to see that list! Thanks

src/Tickets/Commerce/Gateways/Stripe/Stripe_Elements.php
dpanta94: Yes it is! When you reach that part of the code, you have already checked that ETP exists.

Just `use` a non existent class triggers no errors. It will when you actually use it and it doesn't exist.

So when you would reach that line :D

pattihis: Changed... thanks!

src/Tickets/Commerce/Gateways/Stripe/Stripe_Elements.php
bordoni: I would love to make sur ewe dont use constants, because they can be active when the plugin is soft deactivated, use a `did_action()`

pattihis: [The list for reference](https://lw.slack.com/archives/C01H7Q57P1C/p1725247304646939?thread_ts=1725247070.454189&cid=C01H7Q57P1C)

src/Tickets/Commerce/Gateways/Stripe/Stripe_Elements.php
```diff
@@ -54,8 +54,15 @@ public function include_payment_element() {
...
-		return ! ( 1 === count( $payment_methods ) && 'card' === $payment_methods[0] );
+		if ( 1 < count( $payment_methods ) ) {
+			return true;
+		} elseif ( 1 === count( $payment_methods ) && 'card' !== $payment_methods[0] ) {
+			return true;
+		} elseif ( 1 === count( $payment_methods ) && 'card' === $payment_methods[0] && $this->wallets_enabled() ) {
+			return true;
+		}
+
+		return false;
```
Camwyn, line 65: Instead of this - which creates an ETP dependency, let's add a filter and have ETP hook into it.

Something like:
```suggestion
		return (bool) apply_filters( 'tec_ticekts_include_payment_method', false );
```

Then kill off the `wallets_enabled` function - let ETP handle the logic (will require a linked ETP PR)

pattihis: Yes, makes more sense, will do, thanks!

src/resources/js/v2/rsvp-block.js
```diff
@@ -61,7 +61,7 @@ tribe.tickets.rsvp.block = {};
 					action: 'tribe_tickets_rsvp_handle',
 					ticket_id: rsvpId,
 					step: 'going',
-					nonce: TribeRsvp.nonces.rsvpHandle
+					nonce: TribeRsvp.nonces.rsvpHandle,
```
Camwyn, line 64: not an array - is this trailing comma legit/needed?

Camwyn (changes requested):
Quick question

src/Tickets/Commerce/Gateways/Stripe/Stripe_Elements.php
```diff
@@ -54,8 +53,20 @@ public function include_payment_element() {
...
+		}
+
+		/**
+		 * Filter to allow loading the Payment Element component
+		 *
+		 * @since TBD
+		 *
+		 * @param bool $include_payment_element Whether to include the Payment Element.
+		 */
+		return (bool) apply_filters( 'tec_tickets_commerce_stripe_include_payment_element', false );
```
Camwyn, line 69: Might be worth passing either `$payment_methods` methods or the class object (`$this`) here

pattihis: Added, thanks!

Camwyn (approved):

PR #3259, opened 2024-09-25
Allow Attendees CSV export per role

pattihis:
### 🎫 Ticket

[ET-2226]

### 🗒️ Description
Allow the Attendees CSV export for Admin and Editor roles and add a filter if customization is needed.

### ✔️ Checklist

[ET-2226]: https://stellarwp.atlassian.net/browse/ET-2226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Attendees.php
+++ b/src/Tribe/Attendees.php
@@ -832,7 +832,7 @@ class Tribe__Tickets__Attendees {
 		$can_export = current_user_can( 'publish_pages' );
 
 		/**
-		 * Filter roles that are allowed to export the attendees list as a CSV.
+		 * Filter if the current user can export the Attendees list as a CSV.
 		 *
 		 * This allows developers to customize the function to grant permission
 		 * to additional roles or specific users.
--- a/src/Tribe/Attendees.php
+++ b/src/Tribe/Attendees.php
@@ -753,7 +753,7 @@ class Tribe__Tickets__Attendees {
 				! isset( $_GET['attendees_csv_nonce'] )
 				|| ! wp_verify_nonce( sanitize_key( $_GET['attendees_csv_nonce'] ), 'attendees_csv_nonce' )
 				|| empty( $_GET['attendees_csv'] )
-				|| ! current_user_can( 'manage_options' ) ) {
+				|| ! $this->user_can_export_attendees_csv() ) {
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


src/Tribe/Attendees.php
```diff
@@ -820,6 +820,32 @@ public function maybe_generate_csv( $all = false ) {
...
+	 * @since TBD
+	 *
+	 * @return boolean
+	 */
+	public function user_can_export_attendees_csv() {
+		$user = wp_get_current_user();
+
+		$allowed_roles = [ 'administrator', 'editor' ];
+
+		$can_export = ! empty( array_intersect( $allowed_roles, $user->roles ) );
```
dpanta94, line 835: I don't like checking roles instead of capabilities to be honest. Can we check an admin's cap and an editor's cap instead ?

Checking for roles, we are forcing the role to be given to a user that needs to perform that action (along with all the other caps that come with the role). Just checking a cap, we allow for users to create custom roles with a set of caps, or specifically grant users (or roles) a cap instead.

dpanta94: https://wp-kama.com/note/wp-capabilities-list we could use a cap admin and editors share only (and of course super admins)

pattihis: @dpanta94 I agree. I don't like checking for roles either and that's why I had `current_user_can( 'manage_options' )` previously there. But,[ Jack's request](https://lw.slack.com/archives/C0714HQSMEW/p1727293533685739?thread_ts=1727206613.079549&cid=C0714HQSMEW) was to add a filter to allow for custom roles.

codingmusician: Jack mention adding a filter for the roles, not just the `$can_export` value. If we're going to change to checking roles, then add a filter for that, instead.

codingmusician: Otherwise, I would do `$can_export = current_user_can( 'manage_options' )` and then use the filter you have at the bottom.

pattihis: Yes that makes @codingmusician . I made the change, thank you both.

dpanta94 (approved):

dpanta94 (approved):

PR #3268, opened 2024-09-30
Disabled create button when ASC but not seat type selected

dpanta94:
### 🎫 Ticket

Issue 111 from gsheet.

### 🗒️ Description

In the PR the create button of a ticket is enabled only when a layout AND a seat type are selected.

Refactors non asc ticket but without introducing any changes. Only makes reading it easier and introduces filters.

Makes sure we access the current layout property if its defined and adds a method in order to enable us to notify the common store that the ticket has changes.

### 🎥 Artifacts 

https://www.loom.com/share/6c6832c1dd2344efa8f3f5d140a4432d

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Blocks/Ticket/app/editor/dashboard/container.js
+++ b/src/Tickets/Blocks/Ticket/app/editor/dashboard/container.js
@@ -8,6 +8,7 @@ import { compose } from 'redux';
  * WordPress dependencies
  */
 import { dispatch as wpDispatch } from '@wordpress/data';
+import { applyFilters } from '@wordpress/hooks';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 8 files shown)


src/Tickets/Blocks/Ticket/app/editor/dashboard/container.js
```diff
@@ -16,12 +17,74 @@ import Template from './template';
...
+	/**
+	 * Filters whether the ticket is valid.
+	 *
+	 * @since TBD
+	 *
+	 * @param {boolean} isValid  Whether the ticket is valid.
+	 * @param {Object}  state    The state of the store.
+	 * @param {Object}  ownProps The own props of the component.
+	 */
+	const isTicketValid = applyFilters(
```
lucatume, line 40: This filter will apply before the default logic, which works for the SLR case where the determination is based on Layout  and Seat Type only, but might not work for other filtering functions.

I would apply the default logic first, then apply the filter on the result of the default logic:

```js
const getIsConfirmDisabled = (state) => {
	let isConfirmDisabled = (
		!selectors.getRSVPTempTitle(state) ||
		!selectors.getRSVPHasChanges(state) ||
		selectors.getRSVPIsLoading(state) ||
		selectors.getRSVPHasDurationError(state)
	);
	
	isConfirmDisabled = applyFilters( ... );
	
	return isConfirmDisabled;
};
```

src/Tickets/Seating/app/blockEditor/hook-callbacks.js
```diff
@@ -153,11 +153,46 @@ export const filterSettingsFields = (fields) => {
...
+ *
+ * @since TBD
+ *
+ * @param {boolean} isValid  Whether the ticket is valid.
+ * @param {Object}  state    The state of the store.
+ * @param {Object}  ownProps The own props of the component.
+ *
+ * @return {boolean} Whether the ticket is valid.
+ */
+export const filterTicketIsValid = (isValid, state, ownProps) => {
```
lucatume, line 173: This function might likely require modification following my comment on the later application of the filter.

src/Tickets/Seating/app/blockEditor/store/common-store-bridge.js
```diff
@@ -31,4 +37,15 @@ export function setCappedTicketCapacityInCommonStore(clientId, capacity) {
...
 	dispatchToCommonStore(setTicketTempCapacityType(clientId, CAPPED));
+	setTicketHasChangesInCommonStore(clientId);
+}
+
+export function setTicketHasChangesInCommonStore(clientId) {
+	const ticketTempTitle = selectFromCommonStore(getTicketTempTitle, {
+		clientId,
+	});
+
+	// "Changing" the tempTitle will trigger a re-evaluation on the confirm button's disabled status.
```
lucatume, line 48: This comments gives me pause: it makes the method look like a "hack" of how the existing block works.

If I understand this correctly, the confirm button disabled status will **not** be re-evalutated on the call to `setTicketHasChanges`, right?
This behaviour, while it might be pre-existing, does not seem correct. If a filter has been introduced to filter the ticket valid/invalid status, then a change to that status should be all that is required to trigger a re-evaluation of the confirm button disabled status.
The fact that `setTicketHasChanges`, as an API, will not work as intended and will require a call to `setTicketTempTitle` to work is not documented and will eventually break sending the next dev down a rabbit hole of debugging.

If `setTicketHasChanges` is not enough to trigger re-evaluation, this is the moment where that should be changed in the original reducers.

Am I missing something?

src/modules/data/blocks/ticket/selectors.js
```diff
@@ -268,7 +268,7 @@ export const getTicketHasBeenCreated = createSelector(
 
 export const getTicketHasChanges = createSelector(
 	[ getTicket ],
-	( ticket ) => ticket.hasChanges,
+	( ticket ) => ticket.hasChanges || ! ticket.ticketId,
```
lucatume, line 271: Adding this `! ticket.ticketId` will trigger change detection on each change of the ticket during its creation, is this intentional? How does this work when the SLR feature is disabled?

lucatume (changes requested):

src/Tickets/Seating/app/blockEditor/store/common-store-bridge.js
dpanta94: I have retested this and it seems it's not needed to trigger fake changes to tempTitle to re-evaluate the button's disabled status. Triggering that it has changes is enough. I am not sure as to why initially i though it was needed, which confuses me a bit.

src/modules/data/blocks/ticket/selectors.js
dpanta94: Couldn't see any change ... reverted since it doesn't cause any issues.

src/Tickets/Blocks/Ticket/app/editor/dashboard/container.js
dpanta94: I have added a filter for that `tec.tickets.blocks.confirmButton.isDisabled` but i think for this specific case the one being utilized is better.

The reason is that it's filtering whether or not the Ticket on its current state is valid or not, rather than if the confirm button should be disabled or not which can't be decided only by checking the ticket's validity.

Now, the `getTicketHasChanges` method in combination with the filtered `isTicketValid` are going to produce the deisred results :D

dpanta94: Have ignored that in your example you are using RSVP methods instead of tickets. I have assumed that its simply a typo.

lucatume (approved):

PR #3270, opened 2024-09-30
[SL-47] Seating UP-Sell

dpanta94:
### 🎫 Ticket

[SL-47]

### 🗒️ Description

Adds an upsell banner where the seating data for an event is present. The upsell will not appear if a license is active, expired or if a succesful activation of the seating service has ever been recorded.

### 🎥 Artifacts 
image

### ✔️ Checklist

[SL-47]: https://stellarwp.atlassian.net/browse/SL-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dpanta94:
Fixes some issues that were also fixed in #3247

src/Tickets/Seating/Orders/Seats_Report.php
```diff
@@ -212,4 +220,33 @@ public function add_seats_row_action( $actions, $post ): array {
 
 		return $actions;
 	}
+
+	/**
+	 * Checks if the upsell should be shown in the Seats report tab instead.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
```
lucatume, line 229:
```suggestion
	 * @return bool Whether the upsell should show or not.
```

src/Tickets/Seating/Orders/Seats_Report.php
```diff
@@ -212,4 +220,33 @@ public function add_seats_row_action( $actions, $post ): array {
 
 		return $actions;
 	}
+
+	/**
+	 * Checks if the upsell should be shown in the Seats report tab instead.
```
lucatume, line 225:
```suggestion
	 * Returns whether the upsell should show or not.
```

src/Tickets/Seating/Orders/Seats_Report.php
```diff
@@ -212,4 +220,33 @@ public function add_seats_row_action( $actions, $post ): array {
...
+		 * Filters whether the upsell should be shown in the Seats report tab.
+		 *
+		 * @since TBD
+		 *
+		 * @param bool                                           $show_upsell Whether the upsell should be shown.
+		 * @param TEC\Common\StellarWP\Uplink\Resources\Resource $seating     The seating service.
+		 */
+		return apply_filters(
+			'tec_tickets_seating_should_show_upsell',
+			! (
```
lucatume, line 244: Why not use AND here?

```
$seating->get_license_object()->is_valid() &&
$seating->get_license_object()->is_expired() &&
tribe_get_option( 'tec_tickets_seating_connected_on', false )
```

Furthermore, check if the value of the option returned is actually a timestamp, else any value will do.

src/Tickets/Seating/Libraries/Uplink.php
```diff
@@ -59,25 +64,19 @@ public function do_register(): void {
...
 	/**
-	 * Remove the action hooks.
+	 * Stores a successful connection through uplink.
 	 *
 	 * @since TBD
 	 */
-	public function remove_actions(): void {
-		remove_action( 'init', [ $this, 'register_plugin' ] );
+	public function store_successful_connection(): void {
+		tribe_update_option( 'tec_tickets_seating_connected_on', time() );
```
lucatume, line 79: Since the option is being stored inside an already namespace TEC array of options, I would drop the `tec_` prefix and call the option `tickets_seating_connected_on`.

Then: does the value need to be a timestamp? It might, but if it needs to be a timestamp, then it should be verified when read ("is this a valid timestamp?"). If it's sufficient for the value to be there or not, then it's fine to store a boolean-ish value.

src/Tickets/Seating/Orders/Seats_Report.php
```diff
@@ -127,11 +134,12 @@ public function setup_template_vars(): array {
...
-			'post_id'    => $post_id,
-			'iframe_url' => tribe( Service::class )->get_seat_report_url( $token, $post_id ),
-			'token'      => $token,
-			'error'      => $ephemeral_token instanceof WP_Error ? $ephemeral_token->get_error_message() : '',
+			'post'          => $post,
+			'post_id'       => $post_id,
+			'iframe_url'    => tribe( Service::class )->get_seat_report_url( $token, $post_id ),
+			'token'         => $token,
+			'error'         => $ephemeral_token instanceof WP_Error ? $ephemeral_token->get_error_message() : '',
+			'should_upsell' => $this->should_show_upsell(),
```
lucatume, line 142: I would keep the same var name:

```suggestion
			'should_show_upsell' => $this->should_show_upsell(),
```

lucatume (changes requested):
Some smaller changes.

src/Tickets/Seating/Orders/Seats_Report.php
dpanta94: It's a negation of the ORs. So i want to stop evaluation on the first true statement. I think it's faster from trying to make the expression correct but with using AND operators.

Also a license could never be valid and expired in the same time :D Unless we have a bug in uplink at least

dpanta94: Now the option is a simple bool being stored, so any value should do.

lucatume (approved):

PR #3282, opened 2024-10-04
Fix/ticket stock on seat type deletion

dpanta94:
### 🎫 Ticket

Issue 199 from gsheet.

### 🗒️ Description

Fixes the required updates when a seat type is being deleted. Explaining more details of the logic in the demo.

### 🎥 Artifacts 
https://www.loom.com/share/33fd0d7b1aea4455a35bc7cf9a38d05a

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Seating/Admin/Ajax.php
+++ b/src/Tickets/Seating/Admin/Ajax.php
@@ -1017,6 +1017,18 @@ class Ajax extends Controller_Contract {
 		global $wpdb;
 
 		try {
+			$original_seat_types_tickets = array_map(
+				'intval',
+				DB::get_col(
+					DB::prepare(
+						'SELECT DISTINCT(post_id) FROM %i WHERE meta_key = %s AND meta_value = %s',
+						$wpdb->postmeta,
+						Meta::META_KEY_SEAT_TYPE,
+						$new_seat_type['id']
+					)
+				)
+			);
+
 			$updated_seat_types_meta = DB::query(
 				DB::prepare(
 					'UPDATE %i SET meta_value = %s WHERE meta_key = %s AND meta_value = %s',
@@ -1037,7 +1049,7 @@ class Ajax extends Controller_Contract {
 		}
 
 		$updated_seat_types = $this->seat_types->update_from_service( [ $new_seat_type ] );
-		$updated_tickets    = $this->seat_types->update_tickets_capacity( [ $new_seat_type['id'] => $new_seat_type['seatsCount'] ] );
+		$updated_tickets    = $this->seat_types->update_tickets_with_calculated_stock_and_capacity( $new_seat_type['id'], $new_seat_type['seatsCount'], $original_seat_types_tickets );
 
 		wp_send_json_success(
 			[
--- a/src/Tickets/Seating/Admin/Ajax.php
+++ b/src/Tickets/Seating/Admin/Ajax.php
@@ -1016,13 +1016,16 @@ class Ajax extends Controller_Contract {
 
 		global $wpdb;
 
+		$ticket_post_types = implode( ', ', array_map( static fn( $v ) => "'" . esc_sql( $v ) . "'", array_values( tribe_tickets()->ticket_types() ) ) );
+
 		try {
 			$original_seat_types_tickets = array_map(
 				'intval',
 				DB::get_col(
 					DB::prepare(
-						'SELECT DISTINCT(post_id) FROM %i WHERE meta_key = %s AND meta_value = %s',
+						'SELECT DISTINCT(pm.post_id) FROM %i pm JOIN %i p ON p.ID=pm.post_id WHERE pm.meta_key = %s AND pm.meta_value = %s AND p.post_type IN (' . $ticket_post_types . ')',
 						$wpdb->postmeta,
+						$wpdb->posts,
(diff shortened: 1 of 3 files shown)


tests/slr_integration/Service/Seat_Types_Test.php
```diff
@@ -269,4 +269,477 @@ public function test_update_tickets_capacity(): void {
...
+		$post_1_ticket_1 = $this->create_tc_ticket( $post_id_1, 10 );
+		$post_1_ticket_2 = $this->create_tc_ticket( $post_id_1, 20 );
+
+		update_post_meta( $post_id_1, Meta::META_KEY_LAYOUT_ID, 'layout-uuid-1' );
+		update_post_meta( $post_id_1, $capacity_meta_key, 30 );
+		update_post_meta( $post_id_1, Global_Stock::GLOBAL_STOCK_LEVEL, 30 );
+
+		update_post_meta( $post_1_ticket_1, Meta::META_KEY_SEAT_TYPE, 'seat-type-uuid-1' );
+		update_post_meta( $post_1_ticket_1, $capacity_meta_key, 10 );
+		update_post_meta( $post_1_ticket_1, '_stock', 10 );
```
rafsuntaskin, line 316:
```suggestion
		update_post_meta( $post_1_ticket_1, '_stock', 10 );
```

Instead of updating the `_stock` like this we can also use args inside `create_ticket` like [this](https://github.com/the-events-calendar/event-tickets/blob/66a92ccc37ab2d0b1f450e6b1de001f5b20bace2/tests/commerce_integration/TEC/Tickets/Commerce/Stock/SharedCapacity/SharedCapacityTest.php#L60)

It uses the original API that is used for adding ticket and should emulate the real scenario better.

rafsuntaskin (changes requested):

src/Tickets/Seating/Admin/Ajax.php
```diff
@@ -1017,6 +1017,18 @@ public function handle_seat_type_deleted() {
 		global $wpdb;
 
 		try {
+			$original_seat_types_tickets = array_map(
+				'intval',
+				DB::get_col(
+					DB::prepare(
+						'SELECT DISTINCT(post_id) FROM %i WHERE meta_key = %s AND meta_value = %s',
```
lucatume, line 1024: I would modify this to `JOIN` on the `posts` table and include the post type in the query to make sure you're getting only the post IDs of tickets.

You can pull the ticket post types using the `tribe_tickets()->ticket_types()` function to build a comma-separated list.

Right now, the only entity that can be assigned post types are tickets, but that could change due to future changes or user-land modifications.

In general: always query by post type when querying posts.

src/Tickets/Seating/Service/Seat_Types.php
```diff
@@ -263,6 +266,152 @@ public function update_tickets_capacity( array $updates ): int {
...
+		$capacity_meta_key = $tickets_handler->key_capacity;
+
+		$events_primary = [];
+
+		remove_action( 'updated_postmeta', [ tribe( Commerce_Controller::class ), 'sync_seated_tickets_stock' ] );
+
+		foreach (
+			tribe_tickets()
+				->where( 'meta_equals', Meta::META_KEY_SEAT_TYPE, $new_seat_type_id )
+				->not_in( $original_ticket_ids )
```
lucatume, line 298: This method should bail on empty `$original_ticket_ids` since it will build a query with an invalid `IN` clause in that case.

lucatume (changes requested):
A bit more defensive coding, but fine otherwise.

tests/slr_integration/Service/Seat_Types_Test.php
dpanta94: I did try out and even though i think its good practice. At the end, i believe it makes the tests more difficult to read (we have to play around with the variable passed into `create_tc_ticket` since the capacity is changing) and we would have to continue manually updating `_stock` in order to fake ticket with sales.

src/Tickets/Seating/Service/Seat_Types.php
dpanta94: Im confident it should not bail, but rather account for that. Something like `! empty( $original_ticket_ids ) ? $original_ticket_ids : 0` but ill write a test case about this scenario which is not being covered yet to make sure.

src/Tickets/Seating/Admin/Ajax.php
dpanta94: Agreed. makes sense. Fixing the update query as well based on your comment!

dpanta94: Correction: Seems like i shouldn't change the update query just below because attendee's meta is expected to be uddated as well.

src/Tickets/Seating/Service/Seat_Types.php
dpanta94: The added test case confirms that we shouldn't bail but rather account for that case. Amazing catch by the way!

tests/slr_integration/Service/Seat_Types_Test.php
rafsuntaskin: For making sales we usually do make use of the `create_order` that way it emulates the sales. I understand you are taking a more unit test approach for this one which should be fine.

PR #3301, opened 2024-10-11
Coupons & Fees Frontend into Bucket Branch

redscar:
### 🎫 Ticket

[TICKET_ID]

### 🎥 Artifacts 

Screenshot 2024-10-30 at 15 03 37
Screenshot 2024-10-30 at 15 03 55
Screenshot 2024-10-30 at 15 04 14

### ✔️ Checklist


src/resources/js/v2/tickets-commerce.js
```diff
@@ -82,6 +88,174 @@ tribe.tickets.commerce = {};
...
+
+	obj.updateCouponDiscount = function (discount) {
+		const $couponValueElement = $( obj.selectors.couponAppliedValue );
+
+		// Use DOMParser to unescape the discount value
+		const parser = new DOMParser();
+		const unescapedDiscount = parser.parseFromString(`<!doctype html><body>${discount}`, 'text/html').body.textContent;
+
+		$couponValueElement.text(unescapedDiscount);
+	};
```
bordoni, line 114: Add docblock.

That seems rather complex for the unescaping, what is the goal there?

src/resources/js/v2/tickets-commerce.js
```diff
@@ -82,6 +88,174 @@ tribe.tickets.commerce = {};
...
+	obj.bindCouponApply = function() {
+		let ajaxInProgress = false;
+
+		$document.on( 'click', obj.selectors.couponApplyButton, applyCoupon );
+		$document.on( 'keydown', obj.selectors.couponInput, function( e ) {
+			if ( e.key === 'Enter' ) {
+				e.preventDefault();
+				applyCoupon();
+			}
+		});
```
bordoni, line 126: Invert the logic, switch the variable name to `event` and use `keyCode` to avoid having OS differences around `Enter`.

src/resources/js/v2/tickets-commerce.js
```diff
@@ -82,6 +88,174 @@ tribe.tickets.commerce = {};
...
+		 */
+		function applyCoupon() {
+			// Prevent multiple AJAX requests at once.
+			if ( ajaxInProgress ) {
+				return;
+			}
+
+			const couponValue = $( obj.selectors.couponInput ).val().trim();
+			const nonce = $( obj.selectors.nonce ).val();
+			const $errorMessage = $( '.tribe-tickets__commerce-checkout-cart-coupons__error' );
```
bordoni, line 141: This needs to be in a selector var.

src/resources/js/v2/tickets-commerce.js
```diff
@@ -82,6 +88,174 @@ tribe.tickets.commerce = {};
...
+		}
+	};
+
+	/**
+	 * Bind the remove coupon button.
+	 *
+	 * @since TBD
+	 */
+	obj.bindCouponRemove = function () {
+		$document.on( 'click', obj.selectors.couponRemoveButton, function () {
```
bordoni, line 204: Move the actual click method to a separate `obj.` method.

src/resources/js/v2/tickets-commerce.js
```diff
@@ -82,6 +88,174 @@ tribe.tickets.commerce = {};
...
+						// Show input and apply button again.
+						$( obj.selectors.couponInput ).val( '' ).show();
+						$( obj.selectors.couponInputLabel ).show();
+						$( obj.selectors.couponApplyButton ).show();
+
+						// Hide the applied coupon section.
+						$( obj.selectors.couponAppliedSection ).hide();
+						obj.updateTotalPrice( response.data.amount );
+					} else {
+						$errorMessage.text( response.data.message || 'Failed to remove coupon.' ).show();
```
bordoni, line 245: Needs translations.

src/resources/js/v2/tickets-commerce.js
```diff
@@ -82,6 +88,174 @@ tribe.tickets.commerce = {};
...
+					}
+				},
+				error() {
+					$errorMessage.text( 'Error removing coupon. Please try again.' ).show();
```
bordoni, line 249: These need to be translated.

src/resources/js/v2/tickets-commerce.js
```diff
@@ -82,6 +88,174 @@ tribe.tickets.commerce = {};
...
+				applyCoupon();
+			}
+		});
+
+		/**
+		 * Function to apply the coupon and handle AJAX request.
+		 *
+		 * @since TBD
+		 */
+		function applyCoupon() {
```
bordoni, line 133: Move the method to `obj.`

src/Tickets/Order_Modifiers/Modifiers/Modifier_Manager.php
```diff
@@ -149,4 +150,61 @@ public function delete_relationships_by_post( int $post_id ): void {
...
+	 */
+	public function calculate_total_fees( float $base_price, array $items ): float {
+
+		$total_fees = 0.0;
+
+		foreach ( $items as $item ) {
+			// Apply the fees directly without wrapping in an array.
+			$total_fees += $this->apply_fees_to_item( $base_price, $item );
+		}
+
```
bordoni, line 174: Do not handle money as float.

src/Tickets/Order_Modifiers/Checkout/Gateway/Paypal/Fees.php
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * This class manages fees, calculates them, and appends them to the cart
+ * during the PayPal checkout process. It integrates with various filters
+ * and hooks specific to PayPal's order and payment flow.
+ *
+ * @since TBD
+ * @package TEC\Tickets\Order_Modifiers\Checkout\Gateway\Paypal
+ */
+
+namespace TEC\Tickets\Order_Modifiers\Checkout\Gateway\Paypal;
```
bordoni, line 13: I would like to avoid adding PayPal code outside of the PayPal Gateway folder. Move this to a `Tickets\Commerce\Gateways\Paypal\Order_Modifiers`

src/Tickets/Order_Modifiers/Checkout/Abstract_Fees.php
```diff
@@ -0,0 +1,319 @@
...
+
+	/**
+	 * Subtotal value used in fee calculations.
+	 *
+	 * This represents the total amount used as a basis for calculating applicable fees.
+	 *
+	 * @since TBD
+	 * @var float
+	 */
+	protected float $subtotal = 0.0;
```
bordoni, line 79: Never store money as float.

src/Tickets/Order_Modifiers/Checkout/Abstract_Fees.php
```diff
@@ -0,0 +1,319 @@
...
+	 *
+	 * This static property ensures that the fees are only appended once during the
+	 * checkout process across multiple instances of the class. If set to `true`, the
+	 * method `append_fees_to_cart` will not add the fees again. The default is `false`,
+	 * indicating the fees have not yet been appended.
+	 *
+	 * @since TBD
+	 * @var bool
+	 */
+	protected static bool $fees_appended = false;
```
bordoni, line 104: I would rather see this stored in the cart structure, and not mix responsibilities, feels like it would be a pain to find this here after a couple of iterations of code.

bordoni (changes requested):
A couple of general things:
- [ ] Go with `() => {}` for the `function() {}`
- [ ] All of your selectors are global to the page, which is not very good for overall WordPress compatibility, create a container so this gets executed for every container. It avoids global selection on the page.
- [ ] Any `$.ajax()` should potentially be on it's own method to ensure making it simpler and separate in concerns.
- [ ] This can be addressed in a separate PR but, we need to move all this Order Modifiers code to `Tickets\Commerce` folder.
- [x] Avoid direct usage of `_GET` or `_POST` use `tribe_get_request_var()`
- [ ] Every single time we use `__construct()` paired up with instances of other classes we need to pair with a the proper dependency injection system using `tribe()->bind()` methods for registration and instance generation. Otherwise I would like to have setter and getter methods.
- [ ] There is a lot of functionality on this PR, let's make sure we have proper artifacts attached.
- [ ] We need a lot more tests for this functionality.

src/Tickets/Order_Modifiers/Checkout/Abstract_Fees.php
```diff
@@ -0,0 +1,339 @@
...
+
+		// Loop through each fee and append it to the $items array if it's not already added.
+		foreach ( $fees as $fee ) {
+			// Skip if this fee has already been added to the cart.
+			if ( array_key_exists( $fee['id'], $existing_fee_ids ) ) {
+				continue;
+			}
+
+			// Append the fee to the cart.
+			// @todo - Review what needs to be sent. Some of these are used so wp_pluck doesn't cause a warning.
```
redscar, line 319: I wanted to point this comment out. 

I know the biggest issue we had originally was the "Orders" page in the dashboard. I think the data below is good enough. However, will `ticket_id` and `event_id` 0 cause any issues in the future?

From the testing we did originally, we shouldn't have a problem anywhere else. If we do have a problem we need to incorporate the `type` check.

src/Tickets/Order_Modifiers/Checkout/Gateway/Stripe/Fees.php
```diff
@@ -0,0 +1,157 @@
...
+
+		// Fetch the combined fees associated with the items in the cart.
+		$combined_fees = $this->get_combined_fees_for_items( $items );
+
+		// If no combined fees exist, return the original cart value.
+		if ( empty( $combined_fees ) ) {
+			return $value;
+		}
+
+		// Convert each fee_amount to an integer using get_integer() and filter out negative values.
```
redscar, line 90:
```suggestion
		// Convert each fee_amount to an integer using get_decimal() and filter out negative values.
```

src/Tickets/Order_Modifiers/Checkout/Gateway/Stripe/Fees.php
```diff
@@ -0,0 +1,157 @@
...
+
+		// Loop through the fee items and format each one as "FeeName: Price".
+		foreach ( $fee_items as $fee_item ) {
+			// Skip the fee if it lacks required data or has an invalid price.
+			if ( ! isset( $fee_item['display_name'], $fee_item['price'] ) || $fee_item['price'] < 0 ) {
+				continue;
+			}
+
+			// Format the fee metadata as "FeeName: Price".
+			$fee_metadata[] = sprintf( '%s: %.2f', $fee_item['display_name'], $fee_item['price'] );
```
redscar, line 148: I'm not positive about the Stripe API. However, we may want to incorporate a new line at the end/beginning of the Fee text so it doesn't appear bunched up.

src/Tickets/Order_Modifiers/Table_Views/Order_Modifier_Table.php
```diff
@@ -256,10 +261,15 @@ public function search_box( $text, $input_id, $placeholder = '' ) {
...
 	 */
 	public function render_tabs(): void {
-		$modifiers = tribe( Controller::class )->get_modifiers();
+		$modifiers = $this->get_modifiers();
 
-		$current_modifier = tribe_get_request_var( 'modifier', 'coupon' );
+		// If we don't have multiple modifiers, don't render tabs.
+		if ( count( $modifiers ) < 2 ) {
+			return;
+		}
```
redscar, line 269: Based off of the comment, shouldn't this be `1` and not `2`?
```suggestion
		// If we don't have multiple modifiers, don't render tabs.
		if ( count( $modifiers ) < 1 ) {
			return;
		}
```

src/Tickets/Commerce/Gateways/PayPal/REST/Order_Endpoint.php
```diff
@@ -158,6 +151,95 @@ public function handle_create_order( WP_REST_Request $request ) {
...
+				 * Example: If `$type` is 'fee', the filter will be `tec_commerce_get_unit_data_fee`.
+				 *
+				 * @since TBD
+				 *
+				 * @param array   $item   The cart item for which the unit data is being generated.
+				 * @param WP_Post $order  The current order object.
+				 *
+				 * @return array The unit data for the item.
+				 */
+				$unit_data = apply_filters( "tec_commerce_paypal_order_get_unit_data_{$type}", $item, $order );
```
bordoni, line 201: `tec_tickets_commerce_` for prefix.

src/Tickets/Commerce/Gateways/PayPal/REST/Order_Endpoint.php
```diff
@@ -158,6 +151,95 @@ public function handle_create_order( WP_REST_Request $request ) {
...
+				break;
+		}
+
+		// Apply overarching filter for any further customization of the returned data.
+		return apply_filters( 'tec_commerce_paypal_order_get_unit_data', $unit_data, $item, $order );
```
bordoni, line 206: `tec_tickets_commerce_` for prefix.

src/Tickets/Commerce/Gateways/PayPal/REST/Order_Endpoint.php
```diff
@@ -158,6 +151,95 @@ public function handle_create_order( WP_REST_Request $request ) {
...
+				/**
+				 * Filters the unit data for custom item types in the cart.
+				 *
+				 * This filter allows external developers to generate and customize the unit data
+				 * for items in the cart based on the item type (other than 'ticket').
+				 *
+				 * The filter name is dynamic and uses the item type (`$type`) to provide flexibility for
+				 * different item categories.
+				 *
+				 * Example: If `$type` is 'fee', the filter will be `tec_commerce_get_unit_data_fee`.
```
bordoni, line 192: `tec_tickets_commerce_` for prefix.

src/Tickets/Order_Modifiers/Checkout/Coupons.php
```diff
@@ -0,0 +1,93 @@
...
+			[
+				$this,
+				'display_coupon_section',
+			],
+			40,
+			3
+		);
+
+		// Add asset localization to ensure the script has the necessary data.
+		add_action( 'init', fn() => $this->localize_assets() );
```
bordoni, line 56: Avoid this on production code, this hook cannot be removed.

src/Tickets/Order_Modifiers/Checkout/Fees.php
```diff
@@ -0,0 +1,53 @@
...
+	 * and modifies the total values accordingly, ensuring that the fee logic is applied agnostically
+	 * across different payment gateways.
+	 *
+	 * @since TBD
+	 */
+	public function register(): void {
+		// Hook for calculating total values, setting subtotal, and modifying the total value.
+		add_filter(
+			'tec_tickets_commerce_get_cart_total_value',
+			fn( $values, $items, $subtotal ) => $this->calculate_fees( $values, $items, $subtotal ),
```
bordoni, line 41: Avoid anonymous functions to add filters.

src/Tickets/Order_Modifiers/Checkout/Fees.php
```diff
@@ -0,0 +1,53 @@
...
+			...$this->hook_args['ten_three']
+		);
+
+		// Hook for displaying fees in the checkout.
+		add_action(
+			'tec_tickets_commerce_checkout_cart_before_footer_quantity',
+			fn( $post, $items, $template ) => $this->display_fee_section( $items, $template ),
```
bordoni, line 48: Avoid anonymous functions to add filters.

src/Tickets/Order_Modifiers/Checkout/Gateway/Paypal/Fees.php
```diff
@@ -0,0 +1,84 @@
...
+	 * This includes calculating total fees, modifying cart values, and displaying
+	 * fee sections during the PayPal checkout process.
+	 *
+	 * @since TBD
+	 */
+	public function register(): void {
+		// Hook for appending fees to the cart for PayPal processing.
+		add_filter(
+			'tec_tickets_commerce_create_order_from_cart_items',
+			fn( $items, $subtotal ) => $this->append_fees_to_cart( $items, $subtotal ),
```
bordoni, line 42: Avoid anonymous functions to add filters.

src/Tickets/Order_Modifiers/Checkout/Gateway/Paypal/Fees.php
```diff
@@ -0,0 +1,84 @@
...
+			...$this->hook_args['ten_two']
+		);
+
+		// Hook for adding fee unit data to PayPal order.
+		add_action(
+			'tec_commerce_paypal_order_get_unit_data_fee',
+			fn( $item, $order ) => $this->add_fee_unit_data_to_paypal( $item, $order ),
```
bordoni, line 49: Avoid anonymous functions to add filters.

src/Tickets/Order_Modifiers/Checkout/Gateway/Stripe/Fees.php
```diff
@@ -0,0 +1,157 @@
...
+
+		add_filter(
+			'tec_tickets_commerce_stripe_create_from_cart',
+			fn( $value, $items ) => $this->append_fees_to_cart_stripe( $value, $items ),
+			...$this->hook_args['ten_two']
+		);
+
+		add_filter(
+			'tec_tickets_commerce_stripe_update_payment_intent_metadata',
+			fn( $metadata, $order ) => $this->add_meta_data_to_stripe( $metadata, $order ),
```
bordoni, line 55: Avoid anonymous functions to add filters.

src/Tickets/Order_Modifiers/Checkout/Gateway/Stripe/Fees.php
```diff
@@ -0,0 +1,157 @@
...
+		// Hook for appending fees to the cart for Stripe processing.
+		add_filter(
+			'tec_tickets_commerce_create_order_from_cart_items',
+			fn( $items, $subtotal ) => $this->append_fees_to_cart( $items, $subtotal ),
+			...$this->hook_args['ten_two']
+		);
+
+		add_filter(
+			'tec_tickets_commerce_stripe_create_from_cart',
+			fn( $value, $items ) => $this->append_fees_to_cart_stripe( $value, $items ),
```
bordoni, line 49: Avoid anonymous functions to add filters.

src/Tickets/Order_Modifiers/Checkout/Gateway/Stripe/Fees.php
```diff
@@ -0,0 +1,157 @@
...
+	 * This includes calculating total fees, modifying cart values, and displaying
+	 * fee sections during the Stripe checkout process.
+	 *
+	 * @since TBD
+	 */
+	public function register(): void {
+		// Hook for appending fees to the cart for Stripe processing.
+		add_filter(
+			'tec_tickets_commerce_create_order_from_cart_items',
+			fn( $items, $subtotal ) => $this->append_fees_to_cart( $items, $subtotal ),
```
bordoni, line 43: Avoid anonymous functions to add filters.

src/Tickets/Order_Modifiers/Custom_Tables/Order_Modifier_Relationships.php
```diff
@@ -102,12 +94,16 @@ protected function after_update( array $results ): array {
...
 
-		// Check if the table exists first.
-		if ( ! $this->exists() ) {
-			return $results;
-		}
+		// Add foreign key for `modifier_id`.
+		$this->add_foreign_key( $table_name, 'fk_modifier_id', 'modifier_id', $parent_table_name, $parent_table_uid );
+
+		// Add foreign key for `post_id`.
+		$this->add_foreign_key( $table_name, 'fk_post_id', 'post_id', $wp_posts_table, 'ID' );
```
bordoni, line 106: Avoid Foreign keys, keep it as soft references. Avoid Delete Cascading too, there is a reason why WordPress doesn't use those, it breaks in tons of servers.

src/Tickets/Order_Modifiers/Custom_Tables/Order_Modifier_Relationships.php
```diff
@@ -64,21 +61,16 @@ class Order_Modifier_Relationships extends Table {
...
 
 		return "
 			CREATE TABLE `$table_name` (
 				`object_id` BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,
 				`modifier_id` BIGINT UNSIGNED NOT NULL,
 				`post_id`  BIGINT UNSIGNED NOT NULL,
 				`post_type` VARCHAR(20) NOT NULL,
-				PRIMARY KEY (`object_id`),
-				FOREIGN KEY (`modifier_id`) REFERENCES $parent_table_name($parent_table_uid)ON DELETE CASCADE,
-    			CONSTRAINT `fk_post_id` FOREIGN KEY (`post_id`) REFERENCES `$wp_posts_table` (`ID`) ON DELETE CASCADE
```
bordoni, line 81: Its ok to create keys for indexing so things are faster, but dont create actual MySQL references to entities, it causes lots of issues.

src/Tickets/Order_Modifiers/Custom_Tables/Abstract_Custom_Table.php
```diff
@@ -0,0 +1,132 @@
...
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery,WordPress.DB.PreparedSQL.NotPrepared
+		$updated = $wpdb->query( $sql );
+
+		$message = $updated ?
+			sprintf( 'Added index to the %s table on %s.', $table_name, $columns ) :
+			sprintf( 'Failed to add an index on the %s table for %s.', $table_name, $columns );
+
+		$results[ "{$table_name}.{$columns}" ] = $message;
+		return $results;
+	}
```
bordoni, line 59: I really like this to be in `Schema` not here. Huge side note, I believe this can have uses, but not on Plugins with wide use, it fails in a lot of servers.

src/Tickets/Order_Modifiers/Custom_Tables/Abstract_Custom_Table.php
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * Abstract class that provides utility methods for managing custom table schemas,
+ * including adding indexes and foreign key constraints.
+ *
+ * @since TBD
+ *
+ * @package TEC\Tickets\Order_Modifiers\Custom_Tables
+ */
+abstract class Abstract_Custom_Table extends Table {
```
bordoni, line 24: Move this class to Schema, as a contribution to their code.

src/Tickets/Order_Modifiers/Custom_Tables/Order_Modifiers_Meta.php
```diff
@@ -109,7 +100,12 @@ protected function after_update( array $results ): array {
 		}
 
 		global $wpdb;
-		$table_name = self::table_name( true );
+		$table_name        = self::table_name();
+		$parent_table_name = Order_Modifiers::table_name();
+		$parent_table_uid  = Order_Modifiers::uid_column();
+
+		// Add the foreign key constraint using the method from the abstract class.
+		$this->add_foreign_key( $table_name, 'fk_order_modifier', 'order_modifier_id', $parent_table_name, $parent_table_uid );
```
bordoni, line 108: Just like with the other table, avoid actual foreign keys.

tests/_bootstrap.php
```diff
@@ -40,3 +40,23 @@
...
+}
+
+$GLOBALS['wp_filter']['tec_tickets_order_modifiers_coupons_enabled'] = [
+	10 => [
+		[
+			'function'      => '__return_true',
+			'accepted_args' => 1,
+		],
+	],
+];
```
bordoni, line 62: Why not use a constant or ENV variable like the other feature flags?

src/views/v2/commerce/checkout/order-modifiers/coupons.php
```diff
@@ -0,0 +1,26 @@
...
+	<input type="text" class="tribe-tickets__commerce-checkout-cart-coupons__input" id="coupon_input" name="coupons" aria-describedby="coupon_error">
+	<button id="coupon_apply" class="tribe-tickets__commerce-checkout-cart-coupons__apply-button">Apply</button>
+	<p id="coupon_error" class="tribe-tickets__commerce-checkout-cart-coupons__error" style="display: none; color: red;" aria-live="polite" role="alert">Invalid Coupon Code</p>
+</div>
+<div class="tribe-tickets__commerce-checkout-cart-coupons__applied" style="display: none;">
+	<span class="tribe-tickets__commerce-checkout-cart-coupons__applied-text">
+		Coupon: <span class="tribe-tickets__commerce-checkout-cart-coupons__applied-value"></span> - Discount: <span class="tribe-tickets__commerce-checkout-cart-coupons__applied-discount"></span>
+	</span>
+	<button class="tribe-tickets__commerce-checkout-cart-coupons__remove-button" type="button">X</button>
+</div>
```
bordoni, line 26: Translations.

bordoni (changes requested):
I've mentioned that on another PR. Order Modifiers is part of Tickets Commerce, please move it to the `Commerce` folder and Namespace.

src/Tickets/Order_Modifiers/Checkout/Coupons.php
JPry: > this hook cannot be removed.

Making the hook difficult to remove is intentional. Many of our hooks aren't meant to be removed. If there's a reason we might want to allow certain hooks to be easily removed or modified, then it makes sense to change how they are hooked so that they can be removed more easily.

Technically speaking, these *can* still be removed, but it is difficult to do so with the current implementation of the WP Hooks API.

src/Tickets/Order_Modifiers/Checkout/Fees.php
JPry: @bordoni Are there any specific use cases where we'd want to make these hooks easily removable?

tests/_bootstrap.php
JPry: Because it was the easiest way to keep all of the coupon functionality turned off until it is ready. This filter will be removed once the coupons functionality is ready.

src/Tickets/Order_Modifiers/Custom_Tables/Abstract_Custom_Table.php
JPry: @bordoni Does that need to be done for this PR, or can we have a follow-up to move it?

tests/commerce_integration/TEC/Tickets/Commerce/Shortcodes/__snapshots__/Checkout_ShortcodeTest__ticket_on_sale_should_match_html__1.php
```diff
@@ -1,12 +1,12 @@
...
+<link rel=\'stylesheet\' id=\'tribe-tickets-commerce-paypal-style-css\' href=\'http://wordpress.test/wp-content/plugins/event-tickets/src/resources/css/tickets-commerce/gateway/paypal.min.css?ver={{ET_VERSION}}\' type=\'text/css\' media=\'all\' />
+<link rel=\'stylesheet\' id=\'tribe-common-responsive-css\' href=\'http://wordpress.test/wp-content/plugins/event-tickets/src/resources/css/common-responsive.min.css?ver={{ET_VERSION}}\' type=\'text/css\' media=\'all\' />
+<link rel=\'stylesheet\' id=\'tribe-tickets-commerce-style-css\' href=\'http://wordpress.test/wp-content/plugins/event-tickets/src/resources/css/tickets-commerce.min.css?ver={{ET_VERSION}}\' type=\'text/css\' media=\'all\' />
+<link rel=\'stylesheet\' id=\'tribe-tickets-commerce-stripe-style-css\' href=\'http://wordpress.test/wp-content/plugins/event-tickets/src/resources/css/tickets-commerce/gateway/stripe.min.css?ver={{ET_VERSION}}\' type=\'text/css\' media=\'all\' />
+<link rel=\'stylesheet\' id=\'tribe-tickets-commerce-free-style-css\' href=\'http://wordpress.test/wp-content/plugins/event-tickets/src/resources/css/tickets-commerce/gateway/free.min.css?ver={{ET_VERSION}}\' type=\'text/css\' media=\'all\' />
```
dpanta94, line 9: Before updating snapshots in your local ensure minified versions are not present. CI does not build the plugins before running tests as a result minified versions are not present in CI.

Restore the snapshot to its original form without the `.min`

dpanta94 (changes requested):

tests/integration/Tribe/Admin/__snapshots__/OrderReportTest__test_tc_order_report_display__event with multiple tickets in same order__0.snapshot.html
```diff
@@ -131,11 +131,11 @@ <h3>Sales Totals</h3>
 			<option value="id">
 			Search by Order ID		</option>
 	</select><input type="submit" id="search-submit" class="button" value="Search Orders"  /></p>
-		<input type="hidden" id="_wpnonce" name="_wpnonce" value="1234567890" /><input type="hidden" name="_wp_http_referer" value="" />	<div class="tablenav top">
+		<input type="hidden" id="_wpnonce" name="_wpnonce" value="1234{{ID}}90" /><input type="hidden" name="_wp_http_referer" value="" />	<div class="tablenav top">
 
 				<div class="alignleft actions bulkactions">
 					</div>
-					<div class="alignleft actions attendees-actions"><input type="button" name="print" class="print button action" value="Print"><a target="_blank" href="?orders_csv=1&#038;orders_csv_nonce=1234567890" class="export action button" rel="noopener noreferrer">Export</a></div>
+					<div class="alignleft actions attendees-actions"><input type="button" name="print" class="print button action" value="Print"><a target="_blank" href="?orders_csv=1&#038;orders_csv_nonce=1234{{ID}}90" class="export action button" rel="noopener noreferrer">Export</a></div>
```
dpanta94, line 138: These snapshot updates also dont look correct. Why is the {{ID}} placeholder in the center of where the nonce is supposed to be ?

tests/integration/Tribe/Admin/__snapshots__/OrderReportTest__test_tc_order_report_display__event with multiple tickets in same order__0.snapshot.html
```diff
@@ -166,7 +166,7 @@ <h3>Sales Totals</h3>
...
```
dpanta94, line 169: Same as above

dpanta94 (changes requested):

tests/commerce_integration/TEC/Tickets/Commerce/Partials/Checkout/Cart/__snapshots__/CartFooterTest__test_should_render_cart__0.snapshot.html
```diff
@@ -1,9 +1,12 @@
-<footer  class="tribe-tickets__commerce-checkout-cart-footer tribe-common-b1" >
 
-	<div class="tribe-tickets__commerce-checkout-cart-footer-quantity">
+<footer  class="tribe-tickets__commerce-checkout-cart-footer tribe-common-b1" >
+	
+<div class="tribe-tickets__commerce-checkout-cart-footer-order-modifier-fees">
+	<ul>
+			</ul>
+</div>
```
dpanta94, line 7: Maybe that this is here is an indication that something isn't going as expected ? Since there no fees should it be printed at all ?

tests/commerce_integration/TEC/Tickets/Commerce/Partials/Checkout/Cart/__snapshots__/CartTest__test_should_render_cart__0.snapshot.html
```diff
@@ -11,14 +11,17 @@ <h4 class="tribe-common-h4 tribe-common-h--alt tribe-tickets__commerce-checkout-
...
-	<footer  class="tribe-tickets__commerce-checkout-cart-footer tribe-common-b1" >
-
-	<div class="tribe-tickets__commerce-checkout-cart-footer-quantity">
+	
+<footer  class="tribe-tickets__commerce-checkout-cart-footer tribe-common-b1" >
+	
+<div class="tribe-tickets__commerce-checkout-cart-footer-order-modifier-fees">
+	<ul>
+			</ul>
+</div>
```
dpanta94, line 20: Same as above. I would expcet these snapshots to pass without a need of an update. To ensure we remain backwards compatible

tests/commerce_integration/TEC/Tickets/Commerce/Shortcodes/__snapshots__/Checkout_ShortcodeTest__ticket_not_on_sale_should_match_html__1.php
```diff
@@ -76,14 +76,17 @@ class="tribe-common-b2 tribe-common-b3--min-medium tribe-tickets__commerce-check
...
-	<footer  class="tribe-tickets__commerce-checkout-cart-footer tribe-common-b1" >
-
-	<div class="tribe-tickets__commerce-checkout-cart-footer-quantity">
+	
+<footer  class="tribe-tickets__commerce-checkout-cart-footer tribe-common-b1" >
+	
+<div class="tribe-tickets__commerce-checkout-cart-footer-order-modifier-fees">
+	<ul>
+			</ul>
+</div>
```
dpanta94, line 85: Same comment as previously

tests/commerce_integration/TEC/Tickets/Commerce/Shortcodes/__snapshots__/Checkout_ShortcodeTest__ticket_on_sale_should_match_html__1.php
```diff
@@ -96,14 +96,17 @@ class="tribe-common-b2 tribe-common-b3--min-medium tribe-tickets__commerce-check
...
-	<footer  class="tribe-tickets__commerce-checkout-cart-footer tribe-common-b1" >
-
-	<div class="tribe-tickets__commerce-checkout-cart-footer-quantity">
+	
+<footer  class="tribe-tickets__commerce-checkout-cart-footer tribe-common-b1" >
+	
+<div class="tribe-tickets__commerce-checkout-cart-footer-order-modifier-fees">
+	<ul>
+			</ul>
+</div>
```
dpanta94, line 105: Here as well

dpanta94 (changes requested):

src/Tickets/Commerce/Gateways/Contracts/Traits/Paid_Gateway.php
```diff
@@ -12,15 +12,25 @@
...
-	 * @inheritDoc
+	 * Determine whether the gateway should be shown as an available gateway.
+	 *
+	 * @since 5.1.6
+	 * @since TBD Prevented the cart totals from being recalculated multiple times.
+	 *
+	 * @return bool Whether the gateway should be shown as an available gateway.
 	 */
 	public static function should_show() {
+		static $cart_total = null;
```
dpanta94, line 25: Avoid any type of cache we can't clear. I know this only for the request's run cycle but prefer using tribe_cache instead. If for some reason, someone calls wp_cache_flush its expected that every cache will clear. The same should happen for run time cache :D

src/Tickets/Commerce/Admin_Tables/Orders_Table.php
```diff
@@ -449,6 +449,10 @@ public function column_items( $item ) {
 		}
 
 		foreach ( $item->items as $cart_item ) {
+			// Check if 'type' exists and proceed only if it's empty or equals 'ticket'.
+			if ( ! empty( $cart_item['type'] ) && 'ticket' !== $cart_item['type'] ) {
+				continue;
+			}
```
dpanta94, line 455:
```suggestion
			if ( empty( $cart_item['type'] ) || 'ticket' !== $cart_item['type'] ) {
				continue;
			}
```

Lets assure this is backwards compatbile. The typ;e is being added now, so for already existing entries its not expected to be in place. For those it should be assume the item is a ticket

src/Tickets/Commerce/Gateways/Free/Gateway.php
```diff
@@ -64,14 +64,23 @@ public static function is_active(): bool {
...
-	 * @inheritDoc
+	 * Determine whether the gateway should be shown as an available gateway.
+	 *
+	 * @since 5.1.6
+	 * @since TBD Prevented the cart totals from being recalculated multiple times.
+	 *
+	 * @return bool Whether the gateway should be shown as an available gateway.
 	 */
 	public static function should_show(): bool {
+		static $cart_total = null;
```
dpanta94, line 75: Same comment about caching. Prefer tribe_cache()

src/Tickets/Commerce/Gateways/PayPal/REST/Order_Endpoint.php
```diff
@@ -158,6 +151,95 @@ public function handle_create_order( WP_REST_Request $request ) {
...
+	 *
+	 * @return array The structured data for the item, including 'name', 'unit_amount', 'quantity', 'item_total', and
+	 *     'sku'.
+	 */
+	public function get_unit_data( array $item, $order ) {
+		if ( empty( $order ) ) {
+			return [];
+		}
+
+		$type = $item['type'] ?? 'ticket';
```
dpanta94, line 175: This one here, probably makes one of my previous comments redundant :D

src/Tickets/Commerce/Order_Modifiers/Checkout/Abstract_Fees.php
```diff
@@ -0,0 +1,339 @@
...
+
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		$this->modifier_strategy                       = tribe( Controller::class )->get_modifier( $this->modifier_type );
+		$this->manager                                 = new Modifier_Manager( $this->modifier_strategy );
+		$this->order_modifiers_repository              = new Fee_Repository();
+		$this->order_modifiers_relationship_repository = new Order_Modifier_Relationship();
+	}
```
dpanta94, line 114: Why dont you pass those in the constructor and let auto-wiring do its thing ?

src/Tickets/Commerce/Order_Modifiers/Checkout/Abstract_Fees.php
```diff
@@ -0,0 +1,339 @@
...
+		'ten_two'   => [ 10, 2 ],
+	];
+
+	/**
+	 * The modifier type used for fees.
+	 *
+	 * @since TBD
+	 * @var string
+	 */
+	protected string $modifier_type = 'fee';
```
dpanta94, line 49: Why are we becoming so specific in an abstract ? I would prefer here only defining the property and setting it in a class extending the abstract.

src/Tickets/Commerce/Order_Modifiers/Checkout/Abstract_Fees.php
```diff
@@ -0,0 +1,339 @@
...
+
+	/**
+	 * The default priority and number of arguments for hooks.
+	 *
+	 * @var array
+	 */
+	protected array $hook_args = [
+		'ten_three' => [ 10, 3 ],
+		'ten_two'   => [ 10, 2 ],
+	];
```
dpanta94, line 41: Do we need something like that ? isn't it easier of just writting `10, 3` or `10, 2` everytime ?

src/Tickets/Commerce/Order_Modifiers/Checkout/Abstract_Fees.php
```diff
@@ -0,0 +1,339 @@
...
+
+		// Fetch related ticket fees and automatic fees.
+		$related_ticket_fees = $this->order_modifiers_repository->find_relationship_by_post_ids( $ticket_ids, $this->modifier_type );
+		$automatic_fees      = $this->order_modifiers_repository->get_all_automatic_fees();
+
+		// Combine the fees and remove duplicates.
+		$combined_fees = $this->extract_and_combine_fees( $related_ticket_fees, $automatic_fees );
+
+		// Cache the combined fees for future use.
+		wp_cache_set( $cache_key, $combined_fees, 'combined_fees', 120 );
```
dpanta94, line 240: prefer tribe_cache :D

src/Tickets/Commerce/Order_Modifiers/Checkout/Abstract_Fees.php
```diff
@@ -0,0 +1,339 @@
...
+	protected function get_combined_fees_for_items( array $items ): array {
+		if ( empty( $items ) ) {
+			return [];
+		}
+
+		// Generate a cache key based on the items.
+		$cache_key = 'combined_fees_' . md5( wp_json_encode( $items ) );
+
+		// Check if the combined fees are already cached.
+		$cached_fees = wp_cache_get( $cache_key, 'combined_fees' );
```
dpanta94, line 219: tribe_cache here as well

tests/integration/Tribe/Admin/__snapshots__/OrderReportTest__test_tc_order_report_display__event with multiple tickets in same order__0.snapshot.html
JPry: It would seem that it's due to a find/replace in the test itself that looks for post IDs and replaces them with `{{ID}}`. It would see that `5678` is one of the post IDs.

dpanta94: Can you change the return value of the wp_create_nonce for the specific test ?

src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Abstract_Custom_Table.php
```diff
@@ -0,0 +1,132 @@
...
+	 * @param string $table_name        The name of the table to add the foreign key to.
+	 * @param string $foreign_key_name  The name of the foreign key constraint.
+	 * @param string $column_name       The column that references the foreign key.
+	 * @param string $referenced_table  The referenced table name.
+	 * @param string $referenced_column The referenced column in the foreign table.
+	 * @param string $on_delete_action  The action on delete (e.g., CASCADE).
+	 *
+	 * @return void
+	 */
+	protected function add_foreign_key(
```
dpanta94, line 75: I wouldn't use forgeign keys at all. Its not our DBs so its hard to support customers already. WP doesn't have a system on its own using foreign keys so i would strongly advice against using them as well.

src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Order_Modifier_Relationships.php
```diff
@@ -102,12 +94,16 @@ protected function after_update( array $results ): array {
...
 
-		// Check if the table exists first.
-		if ( ! $this->exists() ) {
-			return $results;
-		}
+		// Add foreign key for `modifier_id`.
+		$this->add_foreign_key( $table_name, 'fk_modifier_id', 'modifier_id', $parent_table_name, $parent_table_uid );
+
+		// Add foreign key for `post_id`.
+		$this->add_foreign_key( $table_name, 'fk_post_id', 'post_id', $wp_posts_table, 'ID' );
```
dpanta94, line 106: As mentioned above, i would avoid foreign keys

src/Tickets/Commerce/Order_Modifiers/Modifier_Admin_Handler.php
```diff
@@ -69,8 +73,8 @@ public static function get_page_slug(): string {
 	 * @return void
 	 */
 	public function register(): void {
-		add_action( 'admin_menu', [ $this, 'add_tec_tickets_order_modifiers_page' ], 15 );
-		add_action( 'admin_init', [ $this, 'handle_delete_modifier' ] );
+		add_action( 'admin_menu', fn() => $this->add_tec_tickets_order_modifiers_page(), 15 );
+		add_action( 'admin_init', fn() => $this->handle_delete_modifier() );
```
dpanta94, line 77: Be aware that in arrow functions you cant return void. You should always return something, even null :D

tests/_bootstrap.php
```diff
@@ -40,3 +40,23 @@
...
+}
+
+$GLOBALS['wp_filter']['tec_tickets_commerce_order_modifiers_coupons_enabled'] = [
+	10 => [
+		[
+			'function'      => '__return_true',
+			'accepted_args' => 1,
+		],
+	],
+];
```
dpanta94, line 62: Do we need to do this in all the tests ? Shouldn't existing functionality be tested with the false return value as well ?

I would prefer this to be specifically added on the _bootstrap file of its own test suite. There wp has loaded and you can use add_fitler as well

tests/_support/Traits/Order_Modifiers.php
```diff
@@ -0,0 +1,75 @@
...
+	 * @return Order_Modifier The retrieved order modifier.
+	 */
+	protected function get_order_modifier_for_test( int $modifier_id, $type = 'coupon' ) {
+		$order_modifier_repository = new Order_Modifiers_Repository( $type );
+
+		// Retrieve the modifier by ID.
+		return $order_modifier_repository->find_by_id( $modifier_id );
+	}
+
+	protected function clear_all_modifiers( $type = 'coupon' ) {
```
dpanta94, line 66: Why do we need this method ? DB's state is being reset before and after every test case on its own already.

Every test case, has a START TRANSACTION before and a ROLLBACK right after.

Unless you are creating entries outside of a testcase, you shouldn't need to manually clear your entries :D

tests/integration/TEC/Tickets/Order_Modifiers/Create_Order_Modifiers_Abstract.php
```diff
@@ -0,0 +1,400 @@
...
+
+use Codeception\TestCase\WPTestCase;
+use tad\Codeception\SnapshotAssertions\SnapshotAssertions;
+use TEC\Tickets\Commerce\Order_Modifiers\Modifier_Admin_Handler;
+use TEC\Tickets\Commerce\Order_Modifiers\Modifiers\Modifier_Abstract;
+use TEC\Tickets\Commerce\Order_Modifiers\Table_Views\Order_Modifier_Table;
+use Tribe\Tests\Traits\With_Uopz;
+use Tribe\Tickets\Test\Traits\Order_Modifiers;
+
+abstract class Create_Order_Modifiers_Abstract extends WPTestCase {
```
dpanta94, line 13: Can we make the name suggest this is a class for tests ?

tests/integration/TEC/Tickets/Order_Modifiers/Fees/Checkout/Checkout_Test.php
```diff
@@ -0,0 +1,158 @@
...
+
+	/** @var ?Repository */
+	protected ?Repository $repository;
+
+	/**
+	 * @before
+	 */
+	public function set_up() {
+		$post_id         = static::factory()->post->create( [ 'post_type' => 'page' ] );
+		$this->ticket_id = $this->create_ticket( Commerce::class, $post_id, $this->base_price );
```
dpanta94, line 38: Dont modify the DB state outside of a testcase.

You can modify the DB state inside data providers, which they do so inside a closure which is being called inside a testcase.

So please create post and ticket either directly inside a tescase or in a provider's closure.

tests/integration/TEC/Tickets/Order_Modifiers/Fees/Checkout/Order_Test.php
```diff
@@ -0,0 +1,293 @@
...
+	/**
+	 * Set up common elements for all tests.
+	 *
+	 * @before
+	 */
+	public function set_up(): void {
+		// Step 1: Create a ticket to use in all tests.
+		$post_id         = static::factory()->post->create();
+		$ticket_price    = 23;
+		$this->ticket_id = $this->create_ticket( Commerce::class, $post_id, $ticket_price );
```
dpanta94, line 62: Similar as above

tests/integration/TEC/Tickets/Order_Modifiers/Fees/Checkout/Order_Test.php
```diff
@@ -0,0 +1,293 @@
...
+
+		// Step 4: Create the order.
+		$purchaser = [
+			'purchaser_user_id'    => 0,
+			'purchaser_full_name'  => 'Test Purchaser',
+			'purchaser_first_name' => 'Test',
+			'purchaser_last_name'  => 'Purchaser',
+			'purchaser_email'      => 'test-' . uniqid() . '@test.com',
+		];
+		$order     = tribe( Order::class )->create_from_cart( tribe( Gateway::class ), $purchaser );
```
dpanta94, line 258: We have Traits to help you create orders as well.

uninstall.php
```diff
@@ -42,3 +42,4 @@
 // Drop our custom tables.
 $wpdb->query( "DROP TABLE IF EXISTS {$wpdb->prefix}tec_order_modifiers" );
 $wpdb->query( "DROP TABLE IF EXISTS {$wpdb->prefix}tec_order_modifiers_meta" );
+$wpdb->query( "DROP TABLE IF EXISTS {$wpdb->prefix}tec_order_modifier_relationships" );
```
dpanta94, line 45: Hooray ! I think all of us, we have to keep in mind clearing all the custom tables we are creating during uninstall.

Nice!

dpanta94 (changes requested):

src/Tickets/Commerce/Gateways/Contracts/Traits/Paid_Gateway.php
rafsuntaskin: I am a bit confused why are we caching values of `get_cart_total`? We already have this similarly cached in the [method](https://github.com/the-events-calendar/event-tickets/blob/6ef2fe23d8b80f86707b410ee97e194278061fb7/src/Tickets/Commerce/Cart/Abstract_Cart.php#L81) itself. We may improve that variable cache as Dimi suggested but if that cache is in place we don't need caches in other places for this call.

src/Tickets/Commerce/Gateways/Free/Gateway.php
rafsuntaskin: Same concern as the above cache comment. We had this [cached](https://github.com/the-events-calendar/event-tickets/blob/6ef2fe23d8b80f86707b410ee97e194278061fb7/src/Tickets/Commerce/Cart/Abstract_Cart.php#L81).

src/Tickets/Commerce/Gateways/PayPal/REST/Order_Endpoint.php
```diff
@@ -158,6 +151,95 @@ public function handle_create_order( WP_REST_Request $request ) {
...
+	 * @param array         $item The cart item for which to retrieve unit data.
+	 * @param WP_Post|false $order The order from the items in the cart.
+	 *
+	 * @return array The structured data for the item, including 'name', 'unit_amount', 'quantity', 'item_total', and
+	 *     'sku'.
+	 */
+	public function get_unit_data( array $item, $order ) {
+		if ( empty( $order ) ) {
+			return [];
+		}
```
rafsuntaskin, line 173: We should remove the `false` option from the arg type for `$order` as we don't use it when it is false. We expect it to be a valid WP_Post to continue the process.

src/views/v2/commerce/checkout/cart/footer.php
```diff
@@ -30,8 +30,11 @@
 	'tribe-common-b1',
 ];
 ?>
-<footer <?php tribe_classes( $classes ); ?>>
 
+<footer <?php tribe_classes( $classes ); ?>>
+	<?php
+	do_action( 'tec_tickets_commerce_checkout_cart_before_footer_quantity', $post, $items, $this );
```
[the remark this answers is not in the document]
dpanta94: I actually think in this case a simple do action is a better approach. We don't want to do all that the do_entry_point method is doing.

dpanta94 (approved):

PR #3319, opened 2024-10-17
fix(Seating) block stock updates, sync pre meta update

lucatume:
This udpates the logic used to cross-sync the stock meta (`_stock`)
across Tickes with the same seat type to do it **before** the default
meta update logic takes place and handle it using custom logic.

Since Stock = Capacity - Attendees, this update relies on a query that
will count Attendees per-post, per-tickets by seat-type to make sure the
value will be consistent.

Since the default logic will take over other meta update systems, I've
modified other code handling similar updates to remove and add again the
meta handling logic.

Finally, I've updated the `slr_integration` tests to make sure that ASC
tickets will map to a Series Type and Attendees in the database.

### DIFF
--- a/src/Tickets/Seating/Commerce/Controller.php
+++ b/src/Tickets/Seating/Commerce/Controller.php
@@ -70,20 +70,20 @@ class Controller extends Controller_Contract {
 	/**
 	 * Controller constructor.
 	 *
-	 * since TBD
+	 * @since TBD
 	 *
-	 * @param Container $container A reference to the DI container instance.
-	 * @param Service   $service   A reference to the Seating Service facade.
-	 * @param Seat_Types_Table  $seat_types_table A reference to the Seat Types Table handler.
-	 * @param Attendees $attendees A reference to the Attendees data handler.
+	 * @param Container        $container        A reference to the DI container instance.
+	 * @param Service          $service          A reference to the Seating Service facade.
+	 * @param Seat_Types_Table $seat_types_table A reference to the Seat Types Table handler.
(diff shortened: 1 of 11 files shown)


src/Tickets/Seating/Commerce/Controller.php
```diff
@@ -328,4 +310,127 @@ static function ( $item ) use ( $session_entries ) {
...
+			Ticket::$stock_meta_key,
+			$this->tickets_handler->key_capacity,
+		], true ) ) {
+			// Not a ticket meta key we care about.
+			return $check;
+		}
+
+		$ticket_post_types = tribe_tickets()->ticket_types();
+
+		if ( ! in_array( get_post_type( $object_id ), $ticket_post_types, true ) ) {
```
dpanta94, line 400:
```suggestion
		if ( ! in_array( get_post_type( $object_id ), array_values( $ticket_post_types ), true ) ) {
```

src/Tickets/Seating/Commerce/Controller.php
```diff
@@ -328,4 +310,127 @@ static function ( $item ) use ( $session_entries ) {
...
+			// Not a ticket post type.
+			return $check;
+		}
+
+		$seat_type = get_post_meta( $object_id, Meta::META_KEY_SEAT_TYPE, true );
+
+		if ( ! ( $seat_type ) ) {
```
dpanta94, line 407:
```suggestion
		if ( ! $seat_type ) {
```

dpanta94 (approved):

PR #3335, opened 2024-10-24
SLR - do not clear session for incomplete orders

lucatume:
This updates the logic used to confirm attendee reservations and clear
the current seat selection session to make sure the session will not be
cleared when the order is not complete (e.g. a pending one).

If the session is incorrectly cleared, then the frontend timer will not
find the session information and will expire sending the user a time-out
message and a redirection dialog. This will happen if the user cannot
complete the payment correctly after a failure (that would generate
attendees and a pending order) in the grace time.

Avoiding the session clear in those instances of failed payment will
allow the user to try and complete the payment beyond the grace period,
provided they still have time to do it.

TODO:
- [ ] Change timer grace time (maybe?)
- [ ] On timeout, along with the session the pending attendees should be removed
- [ ] Automated tests

### DIFF
--- a/src/Tickets/Seating/Frontend/Session.php
+++ b/src/Tickets/Seating/Frontend/Session.php
@@ -322,9 +322,11 @@ class Session {
 	 *
 	 * @since TBD
 	 *
+	 * @param bool $delete_token_session Whether to delete the token session after confirming the reservations.
+	 *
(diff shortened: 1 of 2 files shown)


lucatume:
Closing this one as replaced by #3368

PR #3342, opened 2024-10-25
[ET-2261] FSE Theme Notice

codingmusician:
### 🎫 Ticket
[ET-2261]

### 🗒️ Description
Adds a notice if the site is using an FSE theme and a custom AR Page with Shortcode hasn't been set up.

### 🎥 Artifacts 
https://share.zight.com/4guGGBpL

image

### ✔️ Checklist

[ET-2261]: https://stellarwp.atlassian.net/browse/ET-2261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Admin/Notices.php
+++ b/src/Tribe/Admin/Notices.php
@@ -28,6 +28,7 @@ class Tribe__Tickets__Admin__Notices {
 		add_action( 'admin_init', [ $this, 'maybe_display_plus_commerce_notice' ] );
 		add_action( 'admin_init', [ $this, 'maybe_display_unsupported_currency_notice' ] );
 		add_action( 'admin_init', [ $this, 'maybe_display_paystack_notice' ] );
+		add_action( 'admin_init', [ $this, 'maybe_display_fse_ar_page_notice' ] );
 	}
 
 	/**
@@ -328,5 +329,65 @@ class Tribe__Tickets__Admin__Notices {
 			]
 		);
 
+	}
+
+	/**
+	 * Display AR Page notice for FSE theme detected.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function maybe_display_fse_ar_page_notice() {
+		// Bail if we aren't in Tickets > Settings.
+		if ( \Tribe\Tickets\Admin\Settings::$settings_page_id !== tribe_get_request_var( 'page' ) ) {
+			return;
+		}
+
+		// Bail if Tickets Plus is not active.
+		if ( ! class_exists( 'Tribe__Tickets_Plus__Main' ) ) {
+			return;
+		}
+
+		// Bail if not using FSE theme.
+		if ( ! is_readable( get_stylesheet_directory() . '/templates/index.html' ) ) {
+			return;
+		}
+
+		// Bail if the attendee registration page is already set.
+		$id = Tribe__Settings_Manager::get_option( 'ticket-attendee-page-id', false );
+		if ( ! empty( $id ) ) {
+			return;
+		}
+
+		$settings_link = sprintf(
+			'<a href="%s">%s</a>',
+			esc_url( admin_url( 'admin.php?page=tec-tickets-settings&tab=attendee-registration' ) ),
+			esc_html__( 'Attendee Registration settings', 'event-tickets' )
+		);
+
+		$message_text = __(
+			sprintf(
+				// Translators: %s: The "Attendee Registration settings" link.
+				esc_html__( 'We detected that you are using a Full Site Editing theme. In order for the Attendee Registration Page to function properly, you will need to set up a page, using the [tribe_attendee_registration] shortcode in the %s.', 'event-tickets' ),
+				$settings_link
+			),
+			'text for FSE AR Modal notice',
+			'event-tickets'
+		);
+
+		$message = sprintf(
+			'<p>%s</p>',
+			$message_text
+		);
+
+		tribe_notice(
+			'tec-tickets-ar-page-with-fse-theme',
+			$message,
+			[
+				'dismiss' => false,
+				'type'    => 'error',
+			]
+		);
 	}
 }
--- a/src/Tribe/Admin/Notices.php
+++ b/src/Tribe/Admin/Notices.php
@@ -366,14 +366,10 @@ class Tribe__Tickets__Admin__Notices {
 			esc_html__( 'Attendee Registration settings', 'event-tickets' )
 		);
 
-		$message_text = __(
-			sprintf(
-				// Translators: %s: The "Attendee Registration settings" link.
-				esc_html__( 'We detected that you are using a Full Site Editing theme. In order for the Attendee Registration Page to function properly, you will need to set up a page, using the [tribe_attendee_registration] shortcode in the %s.', 'event-tickets' ),
-				$settings_link
-			),
-			'text for FSE AR Modal notice',
-			'event-tickets'
+		$message_text = sprintf(
+			// Translators: %s: The "Attendee Registration settings" link.
+			esc_html__( 'We detected that you are using a Full Site Editing theme. In order for the Attendee Registration Page to function properly, you will need to set up a page, using the [tribe_attendee_registration] shortcode in the %s.', 'event-tickets' ),
+			$settings_link
 		);
 
 		$message = sprintf(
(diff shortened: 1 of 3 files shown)


src/Tribe/Admin/Notices.php
```diff
@@ -329,4 +330,60 @@ function maybe_display_paystack_notice() {
...
+	public function maybe_display_fse_ar_page_notice() {
+		// Bail if we aren't in Tickets > Settings.
+		if ( \Tribe\Tickets\Admin\Settings::$settings_page_id !== tribe_get_request_var( 'page' ) ) {
+			return;
+		}
+
+		// Bail if Tickets Plus is not active.
+		if ( ! class_exists( 'Tribe__Tickets_Plus__Main' ) ) {
+			return;
+		}
```
dpanta94, line 350: Let's use did action instead of class_exists please

src/Tribe/Admin/Notices.php
```diff
@@ -329,4 +330,60 @@ function maybe_display_paystack_notice() {
...
+
+		// Bail if not using FSE theme.
+		if ( ! is_readable( get_stylesheet_directory() . '/templates/index.html' ) ) {
+			return;
+		}
```
dpanta94, line 355: Let's use WP API instead https://developer.wordpress.org/reference/classes/wp_theme/is_block_theme/

src/Tribe/Admin/Notices.php
```diff
@@ -329,4 +330,60 @@ function maybe_display_paystack_notice() {
...
+
+		// Bail if the attendee registration page is already set.
+		$id = Tribe__Settings_Manager::get_option( 'ticket-attendee-page-id', false );
+		if ( ! empty( $id ) ) {
+			return;
+		}
```
dpanta94, line 361: Are we sure we don't need additional validation here ?

Like that the saved id still exists ? Or if it is still published ? etc

src/Tribe/Admin/Notices.php
```diff
@@ -329,4 +330,60 @@ function maybe_display_paystack_notice() {
...
+
+		$settings_link = sprintf(
+			'<a href="%s">%s</a>',
+			esc_url( admin_url( 'admin.php?page=tec-tickets-settings&tab=attendee-registration' ) ),
```
dpanta94, line 365: Can we use methods to build the URL params instead of hard coding ? For example it should be an Admin_Page instance with slug `tec-tickets-settings`. So if we changed the slug, this URL generation would still function.

src/Tribe/Tickets.php
```diff
@@ -4368,14 +4369,20 @@ public function maybe_redirect_to_attendees_registration_screen( $redirect = nul
 			$meta = tribe( 'tickets-plus.meta' );
 
 			$cart_has_meta = true;
+			$cart_has_iac  = true;
 
-			// If the method exists (latest ET+ version), run it.
+			// If the method exists in ET+, run it.
 			if ( method_exists( $meta, 'cart_has_meta' ) ) {
 				$cart_has_meta = $meta->cart_has_meta( $tickets_in_cart );
 			}
```
rafsuntaskin, line 4377: This should already cover the case for IAC, we should not add a new logic for that. IAC is just another form of meta.

rafsuntaskin (changes requested):

src/Tribe/Tickets.php
```diff
@@ -4368,14 +4369,20 @@ public function maybe_redirect_to_attendees_registration_screen( $redirect = nul
...
 
-			// There are no meta fields on the cart tickets.
-			if ( ! $cart_has_meta ) {
+			// If the method exists in ET+, run it.
+			if ( method_exists( $meta, 'cart_has_iac' ) ) {
+				$cart_has_iac = $meta->cart_has_iac( $tickets_in_cart );
+			}
```
dpanta94, line 4382: Can we do this through a filter ? I would not introduce the $cart_has_iac variable here at all. I would apply_filters here whether should we bail or not and hook into ETP.

Honestly this whole method should be in ETP instead, if you could move would be awesome. Othewise let's just not continue adding ETP logic inside ET :D

dpanta94 (changes requested):
Left some comments - Lets make sure its also covered by tests

src/Tribe/Admin/Notices.php
```diff
@@ -329,4 +331,60 @@ function maybe_display_paystack_notice() {
...
+	 * @return void
+	 */
+	public function maybe_display_fse_ar_page_notice() {
+		// Bail if we aren't in Tickets > Settings.
+		if ( Settings::$settings_page_id !== tec_get_request_var( 'page' ) ) {
+			return;
+		}
+
+		// Bail if Tickets Plus is not active.
+		if ( ! did_action( 'tec_container_registered_provider_Tribe__Tickets_Plus__Service_Provider' ) ) {
```
dpanta94, line 349: I really like Stephen's approach regarding that and would love to be consistent accross these checks. [Here](https://lw.slack.com/archives/C01H7Q57P1C/p1725247304646939?thread_ts=1725247070.454189&cid=C01H7Q57P1C) is the pinned message in slack

src/Tribe/Admin/Notices.php
```diff
@@ -329,4 +331,60 @@ function maybe_display_paystack_notice() {
...
+	/**
+	 * Display AR Page notice for FSE theme detected.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function maybe_display_fse_ar_page_notice() {
+		// Bail if we aren't in Tickets > Settings.
+		if ( Settings::$settings_page_id !== tec_get_request_var( 'page' ) ) {
```
dpanta94, line 344: maybe we could utilize `tribe( Settings::class )->is_on_tab( 'tab' )` here instead ?

keep in mind that tec_get_request_var is not checking for frontend vs backend and if current_user_can. So the statement could become true for a frontend request as well by an unauthorized user.

src/Tribe/Admin/Notices.php
```diff
@@ -329,4 +331,60 @@ function maybe_display_paystack_notice() {
...
+		}
+
+		// Bail if not using block theme.
+		if ( ! wp_is_block_theme() ) {
+			return;
+		}
+
+		// Bail if the attendee registration page is already set or does not exist.
+		$id = Tribe__Settings_Manager::get_option( 'ticket-attendee-page-id', false );
+		if ( ! empty( $id ) || ! get_post_status( $id ) ) {
```
dpanta94, line 360: I would specifically check against `publish` So that would need to become:

```suggestion
		if ( ! empty( $id ) && 'publish' === get_post_status( $id ) ) {
```

So we bail if id exists and the status is publish only.

codingmusician: Meh. We just want to make sure it exists. I don't feel like we need to go as far as checking the status, checking if it has a shortcode, punctuation, spelling, etc.

dpanta94: Alright so it would still need to become like below otherwise you are bailing when the get_post_status returns false.

```suggestion
		if ( ! empty( $id ) && get_post_status( $id ) ) {
```

src/Tribe/Admin/Notices.php
dpanta94: Got it!

src/Tribe/Admin/Notices.php
codingmusician: If the `$id` is empty, we want to bail... OR if the `get_post_status() === false` we want to bail. I think if we used `&&`, it would throw a fatal if `$id` was `null`.

dpanta94: It wouldn't throw a fatal because then it would be empty as a result it would not be evaluating the get_post_status at all.

Why do we need to bail if get_post_status is false ? It means that the $id is present but no post status, so the post has been deleted. As a result the $id is equivalent to empty.

Following the above logic it would return 'trash' or 'draft' or 'private' as well so i think we should follow my initial comment [here](https://github.com/the-events-calendar/event-tickets/pull/3342#discussion_r1821559598)

rafsuntaskin (approved):

dpanta94 (approved):

PR #3347, opened 2024-10-28
SLR - fix issue 242

lucatume:
Fixes issue 242 from the issue sheet where creation and edit of a ticket using WooCommerce as provider would not be possible.

All the filtering applied by the Seating feature to the Block Editor relies on the `isUsingAssignedSeating` selector return value.
This fix ensures the return value will be consistently false if the current ticket provider is not Tickets Commerce.


rafsuntaskin (approved):

lucatume:
Build currently broken, I've run the failing tests locally and it's all passing.

PR #3367, opened 2024-11-01
Fix initial using seating state by considering license status

dpanta94:
### 🎫 Ticket

[ET-2264]

### 🗒️ Description

The issue becomes apparent on the FE specifically by the method [isUsingAssignedSeating](https://github.com/the-events-calendar/event-tickets/blob/127d60be346fd2c19d4ba167297974faa9b0e733/src/Tickets/Seating/app/blockEditor/store/selectors.js#L5).

When there is no seating license the GAC vs Seating radio controller is not present in the FE. So there is no way to change that property, resulting in wrong evaluations of the `isDisabled` filter for the "Create Ticket" button. It is being evaluated as ASC ticket instead of GAC while the ASC controls are hidden.

For testing use the below's snippet possibilities to test creating a new ticketable post type. Make sure in every case the block gives you the posibility to create a new ticket.

```php
add_filter( 'tec_tickets_seating_service_status', static function ( $_status, $backend_base_url ) {
	return new Service_Status( $backend_base_url, Service_Status::OK );
	// return new Service_Status( $backend_base_url, Service_Status::NO_LICENSE );
	// return new Service_Status( $backend_base_url, Service_Status::EXPIRED_LICENSE );
//	return new Service_Status( $backend_base_url, Service_Status::SERVICE_DOWN );
//	return new Service_Status( $backend_base_url, Service_Status::NOT_CONNECTED );
	// return new Service_Status( $backend_base_url, Service_Status::INVALID_LICENSE );
}, 10, 2 );
```

### ✔️ Checklist

[ET-2264]: https://stellarwp.atlassian.net/browse/ET-2264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/slr_integration/__snapshots__/Editor_Test__test_get_store_data__no license - new__0.snapshot.json
+++ b/tests/slr_integration/__snapshots__/Editor_Test__test_get_store_data__no license - new__0.snapshot.json
@@ -1,0 +1,30 @@
+{
+    "isUsingAssignedSeating": false,
+    "layouts": [
+        {
+            "id": "layout-1",
+            "name": "Test layout 1",
+            "seats": "1"
+        },
+        {
+            "id": "layout-2",
+            "name": "Test layout 2",
+            "seats": "2"
+        },
+        {
+            "id": "layout-3",
+            "name": "Test layout 3",
+            "seats": "3"
+        }
+    ],
+    "seatTypes": [],
+    "currentLayoutId": null,
+    "seatTypesByPostId": [],
+    "isLayoutLocked": false,
+    "eventCapacity": 0,
+    "serviceStatus": {
+        "ok": false,
+        "status": "no-license",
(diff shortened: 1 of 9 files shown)


bordoni (approved):

tests/slr_integration/__snapshots__/Editor_Test__test_get_store_data__no license - new__0.snapshot.json
```diff
@@ -0,0 +1,30 @@
+{
+    "isUsingAssignedSeating": false,
```
dpanta94, line 2: Important that only here is false. It leaves the rest of the business logic unaffected just like desired

rafsuntaskin (approved):

PR #3378, opened 2024-11-07
SLR - Fix Attendee seat assignment in some flows

lucatume:
[SL-233](https://stellarwp.atlassian.net/browse/SL-233)

Artifacts:
* tests
* [before](https://share.cleanshot.com/TwJRcwDj)
* [after](https://share.cleanshot.com/PjkyDsQQ)

This updates the code around seat assigment to Attendees to take into
account different stock reduction flows, hence Attendee generation
flows.

Depending on when the stock should be reduced and the Attendee
generated, the data required to assign Attendees might still be in the
session or not.

This updates the code to handle both instances and warmup caches used by
the Cart to assign Attendees in either cases.

[SL-233]: https://stellarwp.atlassian.net/browse/SL-233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Seating/Orders/Cart.php
+++ b/src/Tickets/Seating/Orders/Cart.php
@@ -116,8 +116,7 @@ class Cart {
 	 */
 	private function get_session_stack( string $token, int $object_id, int $ticket_id ): Generator {
 		if ( ! isset( $this->session_stacks[ $object_id ][ $ticket_id ] ) ) {
-			$reservations = $this->sessions->get_reservations_for_token( $token );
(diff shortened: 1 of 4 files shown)


rafsuntaskin (approved):
Assuming that we already added changelog in the previous PR for SL-233.

dpanta94 (approved):

PR #3383, opened 2024-11-11
SL-233 - pass 3

lucatume:
[SL-233](https://stellarwp.atlassian.net/browse/SL-233)

Fix another issue with timer and session found out during QA pass.

More details and demo in the issue.

[SL-233]: https://stellarwp.atlassian.net/browse/SL-233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Seating/Tables/Sessions.php
+++ b/src/Tickets/Seating/Tables/Sessions.php
@@ -32,7 +32,7 @@ class Sessions extends Table {
 	 *
 	 * @var string
 	 */
-	const SCHEMA_VERSION = '1.0.0';
+	const SCHEMA_VERSION = '1.1.0';
 
 	/**
 	 * The base table name, without the table prefix.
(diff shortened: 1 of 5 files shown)


src/Tickets/Seating/Tables/Sessions.php
```diff
@@ -120,6 +120,7 @@ protected function get_definition() {
 				`object_id` bigint(20) NOT NULL,
 				`expiration` int(11) NOT NULL,
 				`reservations` longblob DEFAULT '',
+				`expiration_lock` int(1) DEFAULT 0,
```
dpanta94, line 123: Maybe tinyint or boolean is a better option for the column type ?

dpanta94 (approved):

src/Tickets/Seating/Tables/Sessions.php
lucatume: Good idea, thank you.

PR #3384, opened 2024-11-11
[Booking Fees] Fee section should only display for Tickets Commerce.

redscar:
### 🎥 Artifacts 

WooCommerce - 
image
Tickets Commerce - 
image

### DIFF
--- a/src/Tickets/Commerce/Order_Modifiers/Admin/Order_Modifier_Fee_Metabox.php
+++ b/src/Tickets/Commerce/Order_Modifiers/Admin/Order_Modifier_Fee_Metabox.php
@@ -22,6 +22,7 @@ use TEC\Tickets\Registerable;
 use Tribe__Tickets__Admin__Views as Admin_Views;
 use Tribe__Tickets__Main as Main;
 use Tribe__Tickets__Ticket_Object as Ticket_Object;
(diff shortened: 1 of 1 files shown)


src/Tickets/Commerce/Order_Modifiers/Admin/Order_Modifier_Fee_Metabox.php
```diff
@@ -149,6 +150,11 @@ public function enqueue_order_modifiers_fee_scripts() {
 	 * @return void
 	 */
 	public function add_fee_section( int $post_id, ?int $ticket_id ): void {
+		$provider = Tribe__Tickets__Tickets::get_event_ticket_provider( $post_id );
+
+		if ( 'TEC\Tickets\Commerce\Module' !== $provider ) {
```
JPry, line 155: Two minor adjustments to how we import these other classes:

1. `Tribe__Tickets__Tickets` can be aliased to just `Tickets`
2. We should import the `TEC\Tickets\Commerce\Module` class. An alias here is optional, but it could be aliased to `Commerce_Module`. Then we use the `::class` constant to reference it.

```suggestion
		$provider = Tickets::get_event_ticket_provider( $post_id );

		if ( Module:class !== $provider ) {
```

PR #3408, opened 2024-11-18
Booking Fees - Fixed issue with serialized class.

redscar:
### 🗒️ Description

This fixes an issue where we were storing the items with a serialized class. This was due to us not calling `get()` on the precision value object.

### 🎥 Artifacts 

image
image

### DIFF
--- a/src/Tickets/Commerce/Order_Modifiers/Admin/Order_Modifier_Fee_Metabox.php
+++ b/src/Tickets/Commerce/Order_Modifiers/Admin/Order_Modifier_Fee_Metabox.php
@@ -222,8 +222,13 @@ class Order_Modifier_Fee_Metabox implements Registerable {
 		// Merge the automatic fee IDs into the ticket_order_modifier_fees array.
 		$ticket_order_modifier_fees = array_merge( $raw_data['ticket_order_modifier_fees'], $automatic_fee_ids );
 
+		$fee_ids = [];
 		// Ensure IDs are integers.
-		$fee_ids = array_map( 'absint', $ticket_order_modifier_fees );
+		foreach ( $ticket_order_modifier_fees as $item ) {
+			if ( is_object( $item ) && property_exists( $item, 'id' ) ) {
+				$fee_ids[] = absint( $item->id );
+			}
+		}
 
 		// Sync the relationships between the selected fees and the ticket.
 		$this->manager->sync_modifier_relationships( $fee_ids, [ $ticket->ID ] );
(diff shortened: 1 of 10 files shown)


dpanta94 (commented):
That looks good - lets cover this with tests as well and we are good.

Create testcase which creates an order with fees - pull the order object and ensure its fields and meta are of the expected instance

src/Tickets/Commerce/Order_Modifiers/Checkout/Abstract_Fees.php
```diff
@@ -319,13 +319,14 @@ protected function append_fees_to_cart( array $items, Value $subtotal ) {
...
-				'price'        => $fee['fee_amount'],
-				'sub_total'    => $fee['fee_amount'],
+				'price'        => $fee['fee_amount']->get(),
+				'sub_total'    => $fee['fee_amount']->get(),
 				'type'         => 'fee',
 				'fee_id'       => $fee['id'],
 				'display_name' => $fee['display_name'],
 				'ticket_id'    => '0',
 				'event_id'     => '0',
+				'quantity'     => 1,
```
redscar, line 329: I added quantity to fix an issue with the rest API erroring out while looking for the key.

src/Tickets/Commerce/Order_Modifiers/Admin/Order_Modifier_Fee_Metabox.php
```diff
@@ -222,8 +222,13 @@ protected function save_ticket_fee( Ticket_Object $ticket, array $raw_data ): vo
...
 		$ticket_order_modifier_fees = array_merge( $raw_data['ticket_order_modifier_fees'], $automatic_fee_ids );
 
+		$fee_ids = [];
 		// Ensure IDs are integers.
-		$fee_ids = array_map( 'absint', $ticket_order_modifier_fees );
+		foreach ( $ticket_order_modifier_fees as $item ) {
+			if ( is_object( $item ) && property_exists( $item, 'id' ) ) {
+				$fee_ids[] = absint( $item->id );
+			}
+		}
```
redscar, line 231: I had to refactor this section because `$ticket_order_modifier_fees` wasn't returning just an array.

It was returning 
```Array
  (
      [0] => stdClass Object
          (
              [id] => 25
              [modifier_type] => fee
              [sub_type] => flat
              [raw_amount] => 5.000000
              [slug] => test-fee
              [display_name] => TEC\Tickets\Commerce\Order_Modifiers\Orders_Test::test_this
              [status] => active
              [created_at] => 2024-11-19 15:03:21
              [start_time] => 
              [end_time] => 
              [updated_at] => 
              [meta_value] => all
          )
  
  )```

File: tests/order_modifiers_integration/Orders/Orders_Test.php [hunk 3]
```diff
@@ -0,0 +1,194 @@
...
+		// Associate the fee with the event.
+		$this->relationship_repository->insert(
+			new Relationships_Model(
+				[
+					'modifier_id' => $fee->id,
+					'post_id'     => $event_id,
+					'post_type'   => 'post',
+				]
+			)
+		);
```
redscar, line 80: I need to debug this section. I am attaching the fee to the post, but it isn't taking into consideration when calculating the subtotal/total. I think the issue is in the test though.

PR #3416, opened 2024-11-19
Booking Fees - Fixed failing tests  Split tests between Stripe and Paypal

redscar:
### 🗒️ Description

Fixed the failing tests. I also expanded to test for Stripe and Paypal.

### DIFF
--- a/src/Tickets/Commerce/Order_Modifiers/Checkout/Abstract_Fees.php
+++ b/src/Tickets/Commerce/Order_Modifiers/Checkout/Abstract_Fees.php
@@ -337,4 +337,17 @@ abstract class Abstract_Fees {
 
 		return $items;
 	}
+
+	/**
+	 * Resets the fees state and recalculates the subtotal to zero.
+	 *
+	 * This method clears the flag indicating that fees have been appended
+	 * and resets the subtotal to its default value of zero.
+	 *
+	 * @return void
+	 */
+	public function reset_fees_and_subtotal(): void {
+		self::$fees_appended = false;
+		$this->subtotal      = Value::create( 0 );
+	}
 }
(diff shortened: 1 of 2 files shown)


src/Tickets/Commerce/Order_Modifiers/Checkout/Abstract_Fees.php
```diff
@@ -337,4 +337,17 @@ protected function append_fees_to_cart( array $items, Value $subtotal ) {
...
+	 *
+	 * This method clears the flag indicating that fees have been appended
+	 * and resets the subtotal to its default value of zero.
+	 *
+	 * @return void
+	 */
+	public function reset_fees_and_subtotal(): void {
+		self::$fees_appended = false;
+		$this->subtotal      = Value::create( 0 );
+	}
```
redscar, line 352: I created this method because while testing the $fees_appended wasn't being reset. 

If there's another way I should handle, let me know!

tests/order_modifiers_integration/Orders/Orders_Test.php
```diff
@@ -68,22 +127,22 @@ public function it_calculates_order_totals_with_various_inputs(
...
+		// Create a ticket for the event with the specified price.
+		$ticket = $this->create_tc_ticket( $event_id, $ticket_price->get() );
+
 		// Associate the fee with the event.
 		$this->relationship_repository->insert(
 			new Relationships_Model(
 				[
 					'modifier_id' => $fee->id,
-					'post_id'     => $event_id,
+					'post_id'     => $ticket,
```
redscar, line 138: The bug I had was I was using the event Id instead of the ticket Id.

PR #3422, opened 2024-11-21
[Booking Fees] Booking Fee Controller tests

redscar:
Adding tests for Booking Fee Controllers


tests/order_modifiers_integration/Admin/Modifier_Admin_Handler_Test.php
```diff
@@ -0,0 +1,769 @@
...
+					'should_display_message' => false,
+				];
+			},
+		];
+	}
+
+	/**
+	 * @test
+	 */
+	public function it_should_delete_modifier_correctly(): void {
```
redscar, line 689: For this test, I tried to stay with the mindset of "using what we have available" so instead of manipulating the database, I used `handle_form_submission()` and then parsed the return URl to get the Modifier Id.

What are your thoughts on that?

src/Tickets/Commerce/Order_Modifiers/Modifier_Admin_Handler.php
```diff
@@ -333,7 +333,7 @@ public function handle_form_submission(): void {
 		}
 
 		// Get the raw POST data, and set the modifier ID.
-		$raw_data                      = tribe_get_request_vars();
+		$raw_data                      = tribe_get_request_vars( true );
```
redscar, line 336: I set this to true, because it was causing failures in the tests. `true` disables caching of the request_vars.

dpanta94: Thats good. Either use it with true or even better use tec_get_request_var or tec_get_request_var_raw so we dont grab the whole _REQUEST, but specifics.

tests/order_modifiers_integration/Admin/Modifier_Admin_Handler_Test.php
```diff
@@ -0,0 +1,769 @@
...
+		$assets = [
+			'tec-tickets-order-modifiers-table' => 'src/resources/js/admin/order-modifiers/table.js',
+		];
+
+		foreach ( $assets as $slug => $path ) {
+			yield $slug => [ $slug, $path ];
+		}
+	}
+
+	public function should_enqueue_assets_data_provider(): Generator {
```
dpanta94, line 49: All our assets should have this test since we should not enqueue everywhere

tests/order_modifiers_integration/Admin/Modifier_Admin_Handler_Test.php
```diff
@@ -0,0 +1,769 @@
...
+			$query_params['updated'] ?? null,
+			'The "updated" parameter should always be "1".'
+		);
+
+		// Assert `modifier_id` is greater than 0.
+		$this->assertGreaterThan(
+			0,
+			intval( $query_params['modifier_id'] ?? 0 ),
+			'The "modifier_id" parameter should be greater than 0.'
+		);
```
dpanta94, line 345: Shouldn't we make sure that the expected changes took place in the modifier ?

tests/order_modifiers_integration/Admin/Modifier_Admin_Handler_Test.php
dpanta94: Isn't handle_form_submission manipulating the DB ? Its ok to manipulate the DB as much as we want during a test case. (There are a few exceptions to the as much as we want).

So i would like to see during this test, modifiers being created, ensuring they are, deleted and enduring they are gone.

redscar: Instead of editing what I had, I decided to create another test that goes through the lifecycle. `it_should_handle_modifier_lifecycle_correctly`

src/Tickets/Commerce/Order_Modifiers/Modifier_Admin_Handler.php
redscar: Good point, since it was an easy change I moved towards `tec_get_request_var`.

Although, that brings out a question, is `tribe_get_request_vars` the correct call? The idea was that we passed all of the data to the strategy to figure out what it needed via `map_form_data_to_model`. Maybe the correct change is to remove the `$raw_data`, run `map_form_data_to_model()` and in that method we use the individual `tec_get_request_var`. What do you think?

dpanta94 (approved):

PR #3427, opened 2024-11-25
Fix fees calculations

dpanta94:
Fix fees calculations


src/Tickets/Commerce/Flag_Actions/Archive_Attendees.php
```diff
@@ -3,16 +3,9 @@
 namespace TEC\Tickets\Commerce\Flag_Actions;
 
 use TEC\Tickets\Commerce\Attendee;
-use TEC\Tickets\Commerce\Module;
```
JPry, line 6: These imports were unused, so I removed them while adding a new import.

JPry (changes requested):
Only a few minor tweaks, otherwise this looks good.

PR #3433, opened 2024-11-28
Add more snapshot tests for fee display

dpanta94:
Adding more snapshot tests for fees display. I need those in order to be able to reliably complete #3428 .

The diff of the snapshots will tell me if i have issues or not.

### DIFF
--- a/tests/order_modifiers_integration/Checkout/Fees_Test.php
+++ b/tests/order_modifiers_integration/Checkout/Fees_Test.php
@@ -149,7 +149,9 @@ class Fees_Test extends Controller_Test_Case {
 	 * @test
 	 */
 	public function it_should_calculate_fees_and_store_them_correctly_simple_math() {
-		$post = static::factory()->post->create();
+		$post = static::factory()->post->create(
+			[ 'post_title' => 'The Event' ],
+		);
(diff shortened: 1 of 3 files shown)


tests/order_modifiers_integration/Checkout/Fees_Test.php
```diff
@@ -273,6 +290,19 @@ public function it_should_calculate_fees_and_store_them_correctly_complex_math()
 		$cart_total    = $cart->get_cart_total();
 		$cart_subtotal = $cart->get_cart_subtotal();
 
+		$this->set_fn_return( 'wp_create_nonce', '1029384756' );
+		$this->assertMatchesHtmlSnapshot(
+			preg_replace(
```
lucatume, line 295: Why a `preg_replace` here? If you're using the AUTOINCREMENT setting, you should be able to use a plain `str_replace`.

dpanta94: the preg actually removes styles and scripts coming along with the shortcode. I dont want to test that, its being tests elsewhere. i am only interested in the math and the format here

lucatume (approved):

PR #3440, opened 2024-12-03
Fix/indexes length for modifiers

dpanta94:
In this PR i am making sure that:

No index is longer than 191 bytes (when its a text).
Only index what we need to index based on our where statements.
Change column name from object_id from relationships table since there is no object being referenced to id.
Fix queries (especially because of the previous change).

Code is covered by tests already made sure we covered all the changes as well.

### DIFF
--- a/src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Order_Modifiers.php
+++ b/src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Order_Modifiers.php
@@ -107,6 +107,7 @@ class Order_Modifiers extends Abstract_Custom_Table {
 
 		// Helper method to check and add indexes.
 		$results = $this->check_and_add_index( $results, 'tec_order_modifier_indx_slug', 'slug' );
+		$results = $this->check_and_add_index( $results, 'tec_order_modifier_indx_modifier_type', 'modifier_type' );
 		$results = $this->check_and_add_index( $results, 'tec_order_modifier_indx_status_modifier_type', 'status, modifier_type' );
 
 		return $results;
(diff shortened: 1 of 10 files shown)


src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Order_Modifiers.php
```diff
@@ -107,8 +107,8 @@ protected function after_update( array $results ): array {
 
 		// Helper method to check and add indexes.
 		$results = $this->check_and_add_index( $results, 'tec_order_modifier_indx_slug', 'slug' );
-		$results = $this->check_and_add_index( $results, 'tec_order_modifier_indx_status_modifier_type_slug', 'status, modifier_type, slug' );
```
redscar, line 110: This index was originally added to optimize coupon lookups, specifically for validating their criteria (status, modifier_type, slug).

We could always add this back in the future, I just wanted to call it out now.

dpanta94: We will not add an index now that might need in the future. Indexes should be decided based on our where statements.

Especially when you add multiple columns in an index we should make sure we are using where statemets with all the indexes in the where to actually take advantage of the index.

As a result having indexed slug and modifier_type with status we are covering that. Adding the other one is just making write queries slower. Nothing more.

redscar: I agree :100: . A lot of these indexes were introduced when we were still working on Coupons and Fees. So I'm glad your going through it again.

Camwyn (approved):

PR #3446, opened 2024-12-06
Feat/amend cr comments for 3430 v2

dpanta94:
Amends to Code review commnets of #3430


src/Tickets/Commerce/Order_Modifiers/Modifiers/Modifier_Abstract.php
```diff
@@ -147,33 +141,27 @@ public function update_modifier( array $data ): Model {
...
 	 */
-	public function get_modifier_by_id( int $modifier_id, string $modifier_type ): ?array {
+	public function get_modifier_by_id( int $modifier_id ): ?array {
 		$modifier_data = $this->repository->find_by_id( $modifier_id );
 		return $modifier_data ? $modifier_data->to_array() : null;
 	}
 
 	/**
-	 * Finds a modifier by its slug.
+	 * Finds a modifier by its display name.
```
redscar, line 164:
```suggestion
	 * Search modifiers by criteria.
```
I'm not sure of the exact text, but this doesn't search by just display name.

src/Tickets/Commerce/Order_Modifiers/Modifier_Admin_Handler.php
```diff
@@ -279,23 +280,26 @@ protected function render_edit_view( Modifier_Manager $manager, array $context )
 		// Get modifier ID from the context.
 		$modifier_id = (int) $context['modifier_id'];
 
-		// Merge the modifier data into the context to be passed to the form rendering logic.
-		// If a valid modifier ID is provided, fetch the data to populate the form.
-		if ( $modifier_id > 0 ) {
-			$modifier_data = $this->get_modifier_data_by_id( $modifier_id );
+		if ( 0 > $modifier_id ) {
```
JPry, line 283: Should this be `>=`?

dpanta94: No for new modifiers being added

dpanta94: Check also tests that are not failing

src/Tickets/Commerce/Order_Modifiers/Table_Views/Order_Modifier_Table.php
```diff
@@ -164,7 +164,7 @@ protected function column_default( $item, $column_name ) {
 		$method = 'render_' . $column_name . '_column';
 
 		// If a specific method exists for the column, call it.
-		if ( method_exists( $this, $method ) ) {
+		if ( method_exists( $this, $method ) && is_callable( [ $this, $method ] ) ) {
```
JPry, line 167: My comment on the other PR may have been missed, but I think we could do with *only* `is_callable()` here. That way, if `__call()` is implemented instead of an actual method definition, this will return the correct result.

```suggestion
		if ( is_callable( [ $this, $method ] ) ) {
```

src/Tickets/Commerce/Order_Modifiers/Modifier_Admin_Handler.php
JPry: The previous condition was `$modifier_id > 0`. The inverted condition should be `0 >= $modifier_id`, since `0` is also invalid.

dpanta94: I explained that 0 is for new modifiers. Please test it out in your local to understand it

src/Tickets/Commerce/Flag_Actions/Flag_Action_Handler.php
```diff
@@ -69,6 +74,28 @@ public function register() {
...
+	 * @param WP_Post $order The order to prepare.
+	 *
+	 * @return WP_Post
+	 */
+	public function prepare_order_for_email( WP_Post $order ) {
+		if ( empty( $order->items ) || ! is_array( $order->items ) ) {
+			return $order;
+		}
+
+		$order->items = array_filter( $order->items, fn ( $item ) => $this->is_ticket( $item['product_id'] ) );
```
JPry, line 96: `is_ticket()` need the entire `$item` array. This change caused a fatal error on my local that interrupted the process of placing an order.

Also, while doing a debug, it does not appear that the `product_id` property exists.

```suggestion
		$order->items = array_filter( $order->items, fn ( $item ) => $this->is_ticket( $item ) );
```

dpanta94: Nice this also needs tests

src/Tickets/Commerce/Order_Modifiers/Table_Views/Order_Modifier_Table.php
dpanta94: This actually breaks the tests - so ill keep `method_exists` as well.

src/Tickets/Commerce/Order_Modifiers/Modifiers/Modifier_Abstract.php
redscar: Right now I think we can get away with just searching by display name. 

Looking at the Figma https://www.figma.com/design/BPWocsWvjojTaiip1ivW5A/Coupons-%2B-Fees?node-id=0-1&node-type=canvas&t=pyvfWQJNJFjcqbW6-0 we just have the search based off of the display name.

PR #3447, opened 2024-12-09
Fix/add support for series passes

dpanta94:
Adding and testing support for Fees + Series


src/Tickets/Commerce/Order_Modifiers/Checkout/Gateway/PayPal/Fees.php
```diff
@@ -10,7 +10,7 @@
  * @package TEC\Tickets\Commerce\Order_Modifiers\Checkout\Gateway\Paypal
  */
 
-namespace TEC\Tickets\Commerce\Order_Modifiers\Checkout\Gateway\Paypal;
+namespace TEC\Tickets\Commerce\Order_Modifiers\Checkout\Gateway\PayPal;
```
Camwyn, line 13: `capital_P_dangit`? 😆

dpanta94: I know - i was looking like crazy to understand why my tests were failing...

PR #3470, opened 2025-01-06
Prevent duplicated orders and attendees as a result

dpanta94:
### 🎫 Ticket

[ET-2279]

### 🗒️ Description

Description in video: https://www.loom.com/share/43b3d371066c40a2b4c341e63044c8aa

### 🎥 Artifacts 

Solution in video: https://www.loom.com/share/9342174d6cee48efaf7a4d3bec06abff

### ✔️ Checklist

[ET-2279]: https://stellarwp.atlassian.net/browse/ET-2279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Commerce/Order.php
```diff
@@ -514,6 +532,92 @@ public function create( Gateway_Interface $gateway, $args ) {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param Gateway_Interface $gateway           The gateway to use to create the order.
+	 * @param array             $args              The arguments to create the order.
+	 * @param ?int              $existing_order_id The ID of an existing order to update.
+	 *
+	 * @return false|WP_Post
+	 */
+	public function upsert( Gateway_Interface $gateway, array $args, ?int $existing_order_id = null ) {
```
redscar, line 546: Would we be better off making the method name clearer? Maybe renaming it to `save_order`? I'm just thinking at first glance I may not realize this should be used for creating orders moving forward.

lucatume: The name `upsert` is pretty standard in place of `updat_or_insert`, it could be made more readable by renaming it to `update_or_insert`, but we're using `upsert` elsewhere in the code. I think "Order" is inferred from the fact that it's a method on the Order object, it's typical use sould be `$order->upsert()` which seems readable enough to me.

redscar: I agree that `upsert` makes sense as a term on its own, especially when we consider it in the context of update-or-insert. However, my concern lies with how this method interacts with the existing `create` method, which is called within `upsert`.

This creates ambiguity:

- When should I use create versus upsert?
- If upsert calls create internally, does that mean upsert should always be the default for creating orders moving forward?

If upsert is the primary method for creating or updating orders, should we consider:
- Deprecating the create method or limiting its direct use to internal logic? (Though I recognize this might be challenging since create is already public.)
- Adding clearer documentation for both methods to explain when each should be used?
- Renaming `upsert` to something more descriptive `create_or_update_order` or `save_order`, to make it more intuitive for developers to prefer it over create?

tests/commerce_integration/TEC/Tickets/Commerce/Order_Test.php
```diff
@@ -0,0 +1,98 @@
...
+		$this->assertInstanceof( WP_Post::class, $order_5 );
+		$this->assertInstanceof( WP_Post::class, $order_6 );
+
+		// They should NOT be the same order!
+		$this->assertTrue( $order_4->ID !== $order_5->ID );
+		$this->assertTrue( $order_4->ID !== $order_6->ID );
+		$this->assertTrue( $order_5->ID !== $order_6->ID );
+
+		// They should have the same totals though.
+		$this->assertEquals( 10, $order_4->total_value->get_decimal() );
```
lucatume, line 60: Use `assertSame` and the correct type here.

src/Tickets/Commerce/Order.php
lucatume: In the `create` method doc-block add the `@internal` annotation and propose the use of `upsert`.

stratease, line 546:
```suggestion
	public function upsert( Gateway_Interface $gateway, array $args ) {
```
Typically the upsert determines whether the object in question has the fields required to update versus insert. In this case that would be inferred from the `$args` param. I wouldn't leave the logic outside the params, it moves some of the logic for the Order details outside the upsert, and isn't very obvious to me that it would be required me to pass that optional order ID param. 

In other words, if I don't always pass that optional order ID it will always create, thus no longer being an `upsert` .

I think it should be moved inside the method, like...
```
$id = $args['id'] ?? 0;
if($id > 0) {
// Do insert 
} else {
// Do update 
}
```

stratease (changes requested):
Bit of a design change.

src/Tickets/Commerce/Order.php
```diff
@@ -514,6 +532,92 @@ public function create( Gateway_Interface $gateway, $args ) {
...
+	/**
+	 * Filters the values and creates a new Order with Tickets Commerce or updates an existing one.
+	 *
+	 * @since TBD
+	 *
+	 * @param Gateway_Interface $gateway           The gateway to use to create the order.
+	 * @param array             $args              The arguments to create the order.
+	 * @param ?int              $existing_order_id The ID of an existing order to update.
+	 *
+	 * @return false|WP_Post
```
lucatume, line 544: Add comment about return type here.

src/Tickets/Commerce/Order.php
```diff
@@ -514,6 +532,92 @@ public function create( Gateway_Interface $gateway, $args ) {
...
+		if ( empty( $updated[ $existing_order_id ] ) ) {
+			/**
+			 * It seems like the $existing_order_id no longer exists or failed to be updated. Let's create a new one instead.
+			 *
+			 * BE AWARE: THe variable $args here is not passed through the update filters since its going to pass through the create filters.
+			 */
+			return $this->create( $gateway, $args );
+		}
+
+		return tec_tc_get_order( $existing_order_id );
```
lucatume, line 618: Make sure the return is what you expect it to be.
It's fine to just forward the return value from it, but take the possible return values into account in the `@return` annotation.

lucatume (changes requested):

src/Tickets/Commerce/Order.php
Camwyn: `tec_tc_get_order` returns WP_Post|**null** with just the ID as a param - maybe check for empty and return false to ensure the return matches the expected.

Or grab from the repository and return.

src/Tickets/Commerce/Abstract_Order.php
```diff
@@ -95,7 +95,7 @@ public function get_purchaser_data( $data ) {
 		if ( is_user_logged_in() ) {
 			$user                              = wp_get_current_user();
 			$purchaser['purchaser_user_id']    = $user->ID;
-			$purchaser['purchaser_full_name']  = $user->first_name . ' ' . $user->last_name;
+			$purchaser['purchaser_full_name']  = trim( $user->first_name . ' ' . $user->last_name );
```
Camwyn, line 98: If we need to trim this, do we need to trim the next two as well?

src/Tickets/Commerce/Order.php
```diff
@@ -466,7 +471,22 @@ static function ( $item ) {
...
+			)->first_id();
+
+			if ( ! $existing_order_id || ! is_int( $existing_order_id ) ) {
+				$existing_order_id = null;
+			}
+		}
+
+		$order_args['id'] = $existing_order_id;
+
+		$order = $this->upsert( $gateway, $order_args, $existing_order_id );
```
codingmusician, line 489: `upsert()` is not set up to receive `$existing_order_id`

src/Tickets/Commerce/Abstract_Order.php
dpanta94: we are doing the trim here because of the empty space added regardless of the last_name value. That's not the case for the rest :D

PR #3474, opened 2025-01-08
Fix/duplicated orders and attendees

dpanta94:
### 🎫 Ticket

[ET-2280]
[ET-2282]

### 🗒️ Description

The issues we have identified regarding attendee duplications are:

Issue 1: Duplicating attendees on status transition without considering if attendees are already created.
Issue 2: Duplicating orders and as a result attendees when at least one failed payment try.
Issue 3: Duplicating attendees because of race condition. Webhook running prior Ajax.

This PR is dealing with Issue 1 and 2.

Description in video: https://www.loom.com/share/43b3d371066c40a2b4c341e63044c8aa

### 🎥 Artifacts 

Solution in video: https://www.loom.com/share/9342174d6cee48efaf7a4d3bec06abff

### ✔️ Checklist
- [x] Ran `npm run changelog` to add changelog file(s). More info [here](https://docs.theeventscalendar.com/developer/git/changelogs/#process)
- [ ] Code is covered by **NEW** `wpunit` or `integration` tests.
- [x] Code is covered by **EXISTING** `wpunit` or `integration` tests.
- [x] Are all the **required** tests passing?
- [x] Automated code review comments are addressed.
- [ ] Have you added Artifacts?
- [x] Check the base branch for your PR.
- [x] Add your PR to the project board for the release.

[ET-2280]: https://stellarwp.atlassian.net/browse/ET-2280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ET-2282]: https://stellarwp.atlassian.net/browse/ET-2282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3477, opened 2025-01-09
Introduce Lock Order Mechanism

dpanta94:
### 🎫 Ticket

[ET-2279]

### 🗒️ Description

This PR is work on top of #3474 

The issues we have identified regarding attendee duplications are:

Issue 1: Duplicating attendees on status transition without considering if attendees are already created.
Issue 2: Duplicating orders and as a result attendees when at least one failed payment try.
Issue 3: Duplicating attendees because of race condition. Webhook running prior Ajax.

This PR is dealing with Issue 3

### ✔️ Checklist

[ET-2279]: https://stellarwp.atlassian.net/browse/ET-2279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Commerce/Provider.php
```diff
@@ -100,6 +102,100 @@ public function register() {
...
+		$order = tec_tc_get_order( $order_id );
+
+		if ( ! $order || ! $order instanceof WP_Post || ! $order->ID ) {
+			return;
+		}
+
+		// The order is already there!
+		if ( $order->post_status === $new_status_wp_slug ) {
+			return;
+		}
```
rafsuntaskin, line 132: If the locking mechanism is working then we don't need to check it here again. The `can_apply_to` method already doing this same thing and is called by the `modify_status` method.

rafsuntaskin (changes requested):

tests/_bootstrap.php
```diff
@@ -40,3 +42,44 @@
...
+		);
+	}, true );
+}
+
+/**
+ * Disable the fake lock mechanism for Order instances.
+ *
+ * @return void
+ */
+function tests_et_disable_fake_lock_mechanism() {
```
lucatume, line 80: This seems too complicated.
The lock mechanism is based on setting the value of the `post_content_filtered` to the lock key.
To simulate, in tests, a locked order, you can just set the value of `post_content_filtered` to a value that is not empty.
To simulate an unlocked order, you can set the value to an empty string.
Why not use this simpler approach?
Mocking the Order `lock_order` and `unlock_order` methods does not actually cover them.

src/Tickets/Commerce/Gateways/Stripe/Webhooks/Handler.php
```diff
@@ -129,6 +129,31 @@ public static function get_handler_method_for_event( $type ) {
 	 * @return bool|WP_Error|null
 	 */
 	public static function update_order_status( \WP_Post $order, Commerce_Status\Status_Interface $status, array $metadata = [] ) {
+		if ( tribe( Order::class )->is_order_locked( $order->ID ) || ! tribe( Order::class )->is_checkout_completed( $order->ID ) ) {
+
+			if ( function_exists( 'as_enqueue_async_action' ) ) {
```
lucatume, line 134: This method call is already happening in the context of handling a request from the Payment Gateway: it's already a background (i.e. not user-triggered) request, why queue an action and add further delay and complexity?

lucatume (commented):
I think things could be decomplicated and am puzzled by the use of AS.

src/Tickets/Commerce/Provider.php
dpanta94: The locking mechanism is working - this check aims to mark the action as succesfull instead of protecting as from something else.

If we didnt check this here, the action would fail during modify_status and as a result reschedule itself.

src/Tickets/Commerce/Provider.php
```diff
@@ -100,6 +102,100 @@ public function register() {
...
+
+		// We have retried too many times, lets fail.
+		if ( $current_try > 9 ) {
+			// AS catches exception and uses them as the fail message in the action management screen.
+			throw new Exception( __( 'Action failed after too many retries.', 'event-tickets' ) );
```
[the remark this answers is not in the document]
dpanta94: These are all logged by the Action scheduler though.

src/Tickets/Commerce/Gateways/Stripe/Webhooks/Handler.php
```diff
@@ -129,6 +129,31 @@ public static function get_handler_method_for_event( $type ) {
 	 * @return bool|WP_Error|null
 	 */
 	public static function update_order_status( \WP_Post $order, Commerce_Status\Status_Interface $status, array $metadata = [] ) {
+		if ( tribe( Order::class )->is_order_locked( $order->ID ) || ! tribe( Order::class )->is_checkout_completed( $order->ID ) ) {
```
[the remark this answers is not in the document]
dpanta94: Removed one check from there so no longer needed

src/Tickets/Commerce/Gateways/Stripe/Webhooks/Handler.php
```diff
@@ -129,6 +129,31 @@ public static function get_handler_method_for_event( $type ) {
...
+				);
+			}
+
+			/**
+			 * We can't return WP_Error because that will make Stripe think that
+			 * we failed to process the Webhook and as a result will resend it.
+			 *
+			 * Returning bool is the best option here. False since we didn't update, but we will!
+			 */
+			return false;
```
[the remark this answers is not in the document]
dpanta94: Yes, return false is still a 200 return code. Stripe would stop bothering us as a result.

What happens if `as_enqueue_async_action` doesn't exist. That made me consider when that would happen and i decided to remove the check to the functions existence altogether.

The function should exist since we are loading in Common. If it doesnt that means a new version of AS is being loaded which dosnt include the method anymore. We should become aware asap.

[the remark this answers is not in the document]
dpanta94: Now it has changed :D

src/Tickets/Commerce/Gateways/Stripe/Hooks.php
```diff
@@ -61,6 +67,52 @@ protected function add_filters() {
...
+		$order = tec_tc_get_order( $order_id );
+
+		if ( ! $order || ! $order instanceof WP_Post || ! $order->ID ) {
+			return;
+		}
+
+		$pending_webhooks = get_post_meta( $order->ID, '_tec_tickets_commerce_stripe_webhook_pending' );
+
+		// On multiple checkout completes, make sure we dont process the same webhook twice.
+		delete_post_meta( $order->ID, '_tec_tickets_commerce_stripe_webhook_pending' );
```
bordoni, line 89: Store this meta key somewhere. I like something Protected with an access public in the future if we ever need it.

src/Tickets/Commerce/Gateways/Stripe/Hooks.php
```diff
@@ -61,6 +67,52 @@ protected function add_filters() {
...
+
+		foreach ( $pending_webhooks as $pending_webhook ) {
+			if ( ! ( is_array( $pending_webhook ) && isset( $pending_webhook['new_status'], $pending_webhook['metadata'], $pending_webhook['old_status'] ) ) ) {
+				continue;
+			}
```
bordoni, line 94: Split this into two individual conditionals to make it simpler to read.

src/Tickets/Commerce/Gateways/Stripe/Hooks.php
```diff
@@ -61,6 +67,52 @@ protected function add_filters() {
...
+	 * @param int $order_id The order ID.
+	 *
+	 * @throws Exception If the action fails after too many retries.
+	 */
+	public function process_async_stripe_webhook( int $order_id ): void {
+		$order = tec_tc_get_order( $order_id );
+
+		if ( ! $order || ! $order instanceof WP_Post || ! $order->ID ) {
+			return;
+		}
```
bordoni, line 84: Split this into 3  separate conditionals to make it simpler to ready.

src/Tickets/Commerce/Gateways/Stripe/Webhooks/Handler.php
```diff
@@ -129,6 +129,27 @@ public static function get_handler_method_for_event( $type ) {
...
+
+			add_post_meta(
+				$order->ID,
+				'_tec_tickets_commerce_stripe_webhook_pending',
+				[
+					'new_status' => $status->get_wp_slug(),
+					'metadata'   => $metadata,
+					'old_status' => $order->post_status,
+				]
+			);
```
bordoni, line 142: If this is something we will add manually like this, create a method to enforce the array integrity.

src/Tickets/Commerce/Order.php
```diff
@@ -28,6 +37,35 @@ class Order extends Abstract_Order {
...
+	protected const CHECKOUT_COMPLETED_META = '_tec_tc_checkout_completed';
+
+	/**
+	 * Meta key for the order lock status.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	protected const ORDER_LOCK_KEY = 'post_content_filtered';
```
bordoni, line 56: If we want to hijack a particular column, I would call it a column instead of Key, which kind alludes to Meta Keys.

I would rather hijack `ping_status` which is a `varchar(20)`, which can speed queries depending on the MySQL engine being used, since `post_content_filtered` is a `longtext`.

bordoni (changes requested):

src/Tickets/Commerce/Gateways/Stripe/Hooks.php
dpanta94: ill do a public const - thats equivalent to what you are saying i think.
write => no
read => yes

src/Tickets/Commerce/Order.php
dpanta94: the key is 34 chars - considering for mb4 we need at least varchar( 132 )

src/Tickets/Commerce/Gateways/Stripe/Webhooks.php
```diff
@@ -89,6 +98,56 @@ public function get_merchant(): Abstract_Merchant {
...
+	 *
+	 * @param int    $order_id   Order ID.
+	 * @param string $new_status New status.
+	 * @param string $old_status Old status.
+	 * @param array  $metadata   Metadata.
+	 *
+	 * @return void
+	 */
+	public function add_pending_webhook( int $order_id, string $new_status, string $old_status, array $metadata = [] ): void {
+		add_post_meta(
```
bordoni, line 114: why not return the MID or false?

src/Tickets/Commerce/Gateways/Stripe/Webhooks.php
```diff
@@ -89,6 +98,56 @@ public function get_merchant(): Abstract_Merchant {
...
+	 * Delete the pending webhooks for an order.
+	 *
+	 * @since TBD
+	 *
+	 * @param int $order_id Order ID.
+	 *
+	 * @return void
+	 */
+	public function delete_pending_webhooks( int $order_id ): void {
+		delete_post_meta( $order_id, self::PENDING_WEBHOOKS_KEY );
```
bordoni, line 148: why not return if the deletion was successful?

src/Tickets/Commerce/Order.php
bordoni: `post_password`?

src/Tickets/Commerce/Gateways/Stripe/Webhooks.php
dpanta94: Because currently we dont care - we are not going to do anything differently succeeded or failed.

src/Tickets/Commerce/Order.php
dpanta94: ill try it but i dont like it because it is actually being used by WP. Dont want to end up facing weird uncalculated issues.

bordoni (approved):

PR #3479, opened 2025-01-10
Update order modifier queries

JPry:
### 🎫 Ticket

[ET-2268]

### 🗒️ Description

This PR makes the following changes to the order modifier database queries:

1. It makes correct use of the `QueryBuilder` objects to create the queries instead of relying on direct strings
2. Some of the main queires now allow for more parameters, including:
  - `limit` - limit the number of results returned, with a default of 10.
  - `order` - specify ordering ASC or DESC.
  - `orderby` - specify what column to order the results by. Defaults to `id`.
  - `offset` - specify how many results to skip before returning results.
  - `page` - specify what page of results to retrieve. Overrides `offset`.
  - `status` - specify an array of statuses for the results
3. The Fees admin table uses the pagination features of the updated queries
4. There are tests in place to ensure the correct results are provided.

### ✔️ Checklist


tests/order_modifiers_integration/Repositories/Fees.php
```diff
@@ -0,0 +1,80 @@
...
+	 * @return void
+	 */
+	public function should_find_all_fees() {
+		$this->create_modifiers();
+
+		$repo = Factory::get_repository_for_type( $this->modifier_type );
+
+		// Test that we get the correct count of all fees.
+		$count = $repo->get_search_count();
+		$this->assertEquals( 20, $count );
```
dpanta94, line 74: I see 5 modifiers being inserted twice, so 10. why is the assertion for 20 ?

tests/order_modifiers_integration/Repositories/Fees.php
```diff
@@ -0,0 +1,80 @@
...
+
+			$this->upsert_order_modifier_for_test( $data );
+		}
+	}
+
+	/**
+	 * @test
+	 * @return void
+	 */
+	public function should_find_all_fees() {
```
dpanta94, line 67: Are we sure that the most crucial functionality introduced is covered by this test case ? If not, could we become please ?

Camwyn (changes requested):
Dimi brought up some other things in Slack, so I'm not going to cover that here.

src/Tickets/Commerce/Order_Modifiers/Table_Views/Order_Modifier_Table.php
```diff
@@ -104,39 +104,34 @@ public function __construct(
...
-		$data = $this->modifier->find_by_search(
+		$this->items = $this->modifier->find_by_search(
 			[
-				'search_term'   => $search,
-				'orderby'       => $orderby,
-				'order'         => $order,
-				'modifier_type' => $this->modifier->get_modifier_type(),
+				'search_term' => $search,
+				'orderby'     => sanitize_text_field( tribe_get_request_var( 'orderby', 'display_name' ) ),
+				'order'       => sanitize_text_field( tribe_get_request_var( 'order', 'asc' ) ),
```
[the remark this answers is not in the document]
JPry: Thanks, I'll change those. This was just moving these statements around, but I'm glad it revealed that we should be doing something differently here.

tests/order_modifiers_integration/Repositories/Fees.php
JPry: I intended to add more, good catch.

src/Tickets/Commerce/Order_Modifiers/Table_Views/Order_Modifier_Table.php
```diff
@@ -104,39 +104,34 @@ public function __construct(
...
-		$per_page     = $this->get_items_per_page( $this->modifier->get_modifier_type() . '_per_page', 10 );
-		$current_page = $this->get_pagenum();
-		$total_items  = count( $data );
-
-		$data = array_slice( $data, ( $current_page - 1 ) * $per_page, $per_page );
-
-		// Set the items for the table.
-		$this->items = $data;
+		// Get the total number of items.
+		$total_items = $this->modifier->find_count_by_search( [ 'search_term' => $search ] );
```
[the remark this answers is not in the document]
Camwyn: Interesting.

Since we're doing a limit query and then an "unlimited" one - can we just do one unlimited one and return a limited set?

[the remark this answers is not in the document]
JPry: I'm up for doing it that way instead.

I don't know if it would make any difference, but I considered trying to return only a single column instead of all of the results, but I don't know if that would make a substantial difference.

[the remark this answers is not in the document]
Camwyn: What you have can do for now but we should explore this a bit more when we're not over deadline.

tests/order_modifiers_integration/Coupons/Create_Coupon_Modifiers_Test.php
```diff
@@ -26,6 +26,9 @@ class Create_Coupon_Modifiers_Test extends Order_Modifiers_TestCase {
 	 * @throws InvalidArgumentException
 	 */
 	public function does_table_render_correctly() {
+		// Skip this test until the coupon feature is enabled.
+		$this->markTestSkipped( 'Pending the coupon feature being enabled.' );
```
dpanta94, line 30: Why skip ? This was working. If this now fails, there could be a valid reason.

src/Tickets/Commerce/Order_Modifiers/Table_Views/Order_Modifier_Table.php
```diff
@@ -146,6 +141,63 @@ public function prepare_items() {
...
+	 * @return int The total number of items.
+	 */
+	protected function setup_items_with_search( string $search, array $params, int $per_page, int $current_page ): int {
+		// Fetch the data from the modifier class, including sorting.
+		$params['search_term'] = $search;
+		$this->items           = $this->modifier->find_by_search( $params );
+
+		// Get the total number of items.
+		if ( count( $this->items ) < $per_page && $current_page === 1 ) {
+			$total_items = count( $this->items );
```
Camwyn, line 163: can just return here

Camwyn: ...unless we're going to filter the value?

src/Tickets/Commerce/Order_Modifiers/Table_Views/Order_Modifier_Table.php
```diff
@@ -146,6 +141,63 @@ public function prepare_items() {
...
+		} else {
+			$total_items = $this->modifier->find_count_by_search(
```
Camwyn, line 165: then we can return here - without the else...

src/Tickets/Commerce/Order_Modifiers/Table_Views/Order_Modifier_Table.php
```diff
@@ -146,6 +141,63 @@ public function prepare_items() {
...
+	 *
+	 * @return int The total number of items.
+	 */
+	protected function setup_items( array $params, int $per_page, int $current_page ): int {
+		$applied_to  = [ 'per', 'all' ];
+		$this->items = $this->modifier->get_modifier_by_applied_to( $applied_to, $params );
+
+		// Get the total number of items.
+		if ( count( $this->items ) < $per_page && $current_page === 1 ) {
+			$total_items = count( $this->items );
```
Camwyn, line 193: same as above on the returning

src/Tickets/Commerce/Order_Modifiers/Repositories/Order_Modifiers.php
```diff
@@ -185,46 +188,45 @@ public function find_by_id( int $id ): Order_Modifier {
...
-		$defaults = [
-			'search_term' => '',
-			'orderby'     => 'display_name',
-			'order'       => 'asc',
-		];
-
 		// Merge passed arguments with defaults.
-		$args = array_merge( $defaults, $args );
+		$args       = wp_parse_args( $args, $this->get_default_query_params() );
+		$valid_args = $this->get_valid_params( $args );
```
Camwyn, line 206: Not huge, but better to use one term - `params` or `args` (in this case...`args`?)

JPry:
@Camwyn @dpanta94 This is ready for another look

From slack:

> query 48 is used to count ALL the fees => that should not happen if the primary query returns less than 10 fees (or whatever the limit is)

That was a good idea, and I updated the code to behave this way. The second query is skipped when the number of fees retrieved is less than the limit.

> queries 49-52 are then individual meta queries to determine if each fee is applied per ticket or to all tickets => that should be avoided and it will be more performant if we grab that info with our primary query using a JOIN

Another good idea, I updated the code to perform the JOIN.

> as a result, current code would perform 12 queries in every page reload in order to display 10 fees.
> 12 = primary + count + 1 meta query for each fee
> when it could do
> 1 if fees less than limit or 2 if fees retrieved are equal to the limit

The code has been updated so that it will perform only one query if less than the limit, and 2 if it meets the limit.

The only exception is searching, but that is something we can add on in the next iteration.

tests/order_modifiers_integration/Repositories/Fees.php
JPry: I added some more tests, and I plan to add even more in a follow-up.

dpanta94 (approved):

PR #3485, opened 2025-01-14
Enriches ticket management hooks

dpanta94:
Enriches ticket management hooks. As a result, external plugins can offer their functionality with hooks rather than needing access to an external API.

It is backwards compatible since only the waitlists feature ( which is unreleased ) is going to take advantage of the new parameters passed through the hooks for now.

### DIFF
--- a/src/modules/data/blocks/ticket/sagas.js
+++ b/src/modules/data/blocks/ticket/sagas.js
@@ -827,7 +827,7 @@ export function* updateTicket( action ) {
 				saleEndDate,
 				saleEndDateInput,
 				saleEndDateMoment,
-			}
+			};
 
 			yield all( [
 				put( actions.setTicketDetails( clientId, ticketDetails ) ),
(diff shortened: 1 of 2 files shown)


codingmusician:
I'm not sure why the lint tests are no longer required. They should be.

src/modules/data/blocks/ticket/sagas.js
```diff
@@ -794,35 +800,37 @@ export function* updateTicket( action ) {
...
+				capacity,
+				salePriceChecked,
+				salePrice,
+				saleStartDate,
+				saleStartDateInput,
+				saleStartDateMoment,
+				saleEndDate,
+				saleEndDateInput,
+				saleEndDateMoment,
+			}
```
codingmusician, line 830:
```suggestion
			};
```

codingmusician (changes requested):

src/modules/data/blocks/ticket/sagas.js
dpanta94: Good catch!

codingmusician (approved):

PR #3489, opened 2025-01-15
Include status check as part of the transaction

dpanta94:
### 🎫 Ticket

[ET-2281]

### 🗒️ Description

I am thinking that including the can_apply method in the DB transaction is the correct approach here.

What i suspect was still happening:

1) 2 requests very close to each other - probably a diff of some ms.
2) both would evaluate that the status could_apply_to => true ( e.g. both of them from pending to completed )
3) Since they had a slight difference, each one of them would => lock the order, update the order status using wp_update_post which DOES NOT care if the status actually changed before triggering the post status transition sequence, unlock the order.

What i am trying to do over here:

1) Create a transaction - so all of the following queries fail or succeed together
2) Lock the order - will fail quietly if order already locked.
3) Select the current status of the locked order. - will fail quietly if fail to select because ...
4) Evaluate if new status can be applied on top of current status - considering wp_errors which seems we were missing for a while. - will fail quietly if not.
5) Update the order.
6) Unlock the order.
7) Commit the transaction.

### 🎥 Artifacts 

All the tests passing while modified method is being used by them is artifact enough for such a low level change.
Also added dedicated tests.

### ✔️ Checklist

[ET-2281]: https://stellarwp.atlassian.net/browse/ET-2281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Commerce/Order.php
+++ b/src/Tickets/Commerce/Order.php
@@ -2,22 +2,22 @@
 /**
  * Tickets Commerce Order
  *
- * @since 5.1.9
+ * @since   5.1.9
  *
  * @package TEC\Tickets\Commerce
  */
 
 namespace TEC\Tickets\Commerce;
 
+use TEC\Common\StellarWP\DB\Database\Exceptions\DatabaseQueryException;
+use TEC\Common\StellarWP\DB\DB;
 use TEC\Tickets\Commerce\Gateways\Contracts\Gateway_Interface;
+use TEC\Tickets\Commerce\Status\Pending;
 use TEC\Tickets\Commerce\Status\Refunded;
 use TEC\Tickets\Commerce\Status\Reversed;
 use TEC\Tickets\Commerce\Utils\Value;
 use Tribe__Date_Utils as Dates;
 use WP_Post;
-use TEC\Tickets\Commerce\Status\Pending;
-use TEC\Common\StellarWP\DB\DB;
-use TEC\Common\StellarWP\DB\Database\Exceptions\DatabaseQueryException;
 
 /**
  * Class Order
@@ -291,18 +291,20 @@ class Order extends Abstract_Order {
 				'item_scheduled'           => __( 'Order scheduled.', 'event-tickets' ),
 				'item_updated'             => __( 'Order updated.', 'event-tickets' ),
 				'item_link'                => _x( 'Order Link', 'navigation link block title', 'event-tickets' ),
-				'item_link_description'    => _x( 'A link to an order.', 'navigation link block description', 'event-tickets' ),
+				'item_link_description'    => _x( 'A link to an order.',
+					'navigation link block description',
+					'event-tickets' ),
 			],
 		];
 
 		/**
 		 * Filter the arguments that craft the order post type.
 		 *
-		 * @see   register_post_type
 		 * @since 5.1.9
 		 *
 		 * @param array $post_type_args Post type arguments, passed to register_post_type()
 		 *
+		 * @see   register_post_type
 		 */
 		$post_type_args = apply_filters( 'tec_tickets_commerce_order_post_type_args', $post_type_args );
 
@@ -356,13 +358,13 @@ class Order extends Abstract_Order {
 	 *
 	 * @since 5.1.9
 	 *
-	 * @throws \Tribe__Repository__Usage_Error
-	 *
 	 * @param int    $order_id    Which order ID will be updated.
 	 * @param string $status_slug Which Order Status we are modifying to.
 	 * @param array  $extra_args  Extra repository arguments.
 	 *
 	 * @return bool|\WP_Error
+	 * @throws \Tribe__Repository__Usage_Error
+	 *
 	 */
 	public function modify_status( $order_id, $status_slug, array $extra_args = [] ) {
 		$status = tribe( Status\Status_Handler::class )->get_by_slug( $status_slug );
@@ -371,14 +373,15 @@ class Order extends Abstract_Order {
 			return false;
 		}
 
-		$this->start_transaction();
+		DB::beginTransaction();
 
 		// During this operations - the order should be locked!
 		$locked = $this->lock_order( $order_id );
 
 		// If we were unable to lock the order, bail.
 		if ( ! $locked ) {
-			$this->rollback_transaction();
+			DB::rollback();
+
 			return false;
 		}
 
@@ -395,12 +398,14 @@ class Order extends Abstract_Order {
 			);
 		} catch ( DatabaseQueryException $e ) {
 			// The query should be failing silently.
-			$this->rollback_transaction();
+			DB::rollback();
+
 			return false;
 		}
 
 		if ( ! $current_status_wp_slug ) {
-			$this->rollback_transaction();
+			DB::rollback();
+
 			return false;
 		}
 
@@ -408,7 +413,8 @@ class Order extends Abstract_Order {
 
 		$can_apply = $status->status_can_apply_to_status( $current_status, $status, $order_id );
 		if ( ! $can_apply || is_wp_error( $can_apply ) ) {
-			$this->rollback_transaction();
+			DB::rollback();
+
 			return $can_apply;
 		}
 
@@ -427,7 +433,7 @@ class Order extends Abstract_Order {
 
 		$this->unlock_order( $order_id );
 
-		$this->commit_transaction();
+		DB::commit();
 
 		// After modifying the status we add a meta to flag when it was modified.
 		if ( $updated ) {
@@ -472,15 +478,15 @@ class Order extends Abstract_Order {
 	 * @since 5.1.9
 	 * @since TBD Now it will only create one order per cart hash. Every next time it will update the existing order.
 	 *
+	 * @return false|WP_Post
 	 * @throws \Tribe__Repository__Usage_Error
 	 *
-	 * @return false|WP_Post
 	 */
 	public function create_from_cart( Gateway_Interface $gateway, $purchaser = null ) {
 		$cart = tribe( Cart::class );
 
-		$items      = $cart->get_items_in_cart();
-		$items      = array_filter( array_map(
+		$items = $cart->get_items_in_cart();
+		$items = array_filter( array_map(
 			static function ( $item ) {
 				/** @var Value $ticket_value */
 				$ticket_value         = tribe( Ticket::class )->get_price_value( $item['ticket_id'] );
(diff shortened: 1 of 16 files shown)


rafsuntaskin:
> both would evaluate that the status could_apply_to => true.

I believe this was the original issue we had before introducing the locking mechanism. So, where is the gap now? Aren't we checking the `order->checkout_completed`? That should send the Payment intent handler to process async at a delayed time, right?

rafsuntaskin:
I believe the locking mechanism is having the same issue as the order `can_apply_to` method? Simultaneous multi threads are both reading the same value from DB. Which is causing them both to find the order unlocked at the same time and locking them at the same time.

tests/_bootstrap.php
```diff
@@ -42,3 +41,17 @@
 // By default, do not enable the Custom Tables v1 implementation in tests.
 putenv( 'TEC_CUSTOM_TABLES_V1_DISABLED=1' );
 $_ENV['TEC_CUSTOM_TABLES_V1_DISABLED'] = 1;
+
+function tec_tickets_tests_fake_transactions_enable() {
+	uopz_set_return( Order::class, 'start_transaction', true, false );
```
lucatume, line 46: I understand the need to control the transaction process, but the Order class should not have transaction-related methods like `start_transaction`. `rollback_transaction` and `commit_transaction`. They are just passthroughs to the `DB` class, that, defining those magically with `__callStatic`, makes mocking them impossible.
This is a case where I think production code is being affected by testing needs.
The solution is not to mock them.

Create a test case dedicated to this that will:
* run `commit` on `setUp`
* run the test and note the affected post IDs
* clean up the affected post IDs on `tearDown`.

src/Tickets/Commerce/Order.php
```diff
@@ -371,21 +371,49 @@ public function modify_status( $order_id, $status_slug, array $extra_args = [] )
 			return false;
 		}
 
-		$can_apply = $status->can_apply_to( $order_id, $status );
-		if ( ! $can_apply ) {
-			return $can_apply;
-		}
-
-		$args = array_merge( $extra_args, [ 'status' => $status->get_wp_slug() ] );
+		$this->start_transaction();
```
lucatume, line 374: The logic holds here, but the Order class, as I've explained in my other comment, should not have transaction-related thin methods: managing transactions is the `DB` class' responsibility.

A test case can be written to avoid mocking transaction code and actually using transactions (see my other comment).

lucatume (changes requested):

src/Tickets/Commerce/Status/Status_Abstract.php
```diff
@@ -148,6 +148,30 @@ public function can_apply_to( $order, $new_status ) {
...
+	/**
+	 * Whether a Status Interface can be applied on top of another Status Interface.
+	 *
+	 * @since TBD
+	 *
+	 * @param self $current_status The current status.
+	 * @param self $new_status     The new status.
+	 * @param ?int $order_id       Which order we are testing against.
+	 *
+	 * @return bool
```
lucatume, line 160:
```suggestion
	 * @return bool Whether the new status can be applied to the current status.
```

src/Tickets/Commerce/Status/Pending.php
```diff
@@ -60,8 +60,31 @@ public function get_name() {
...
+	 * @since TBD
+	 *
+	 * @param self $current_status The current status.
+	 * @param self $new_status     The new status.
+	 * @param ?int $order_id       Which order we are testing against.
+	 *
+	 * @return bool|WP_Error
+	 */
+	public function status_can_apply_to_status( $current_status, $new_status, $order_id = null ) {
+		return $this->can_apply_to_pending_status( parent::status_can_apply_to_status( $current_status, $new_status, $order_id ), $order_id );
```
lucatume, line 78: Very long line: break it down by collecting the first return value in a variable or by breaking lines:
```suggestion
		return $this->can_apply_to_pending_status(
		    parent::status_can_apply_to_status( $current_status, $new_status, $order_id ), 
		    $order_id 
		);
```

src/Tickets/Commerce/Order.php
```diff
@@ -745,18 +779,19 @@ public function get_gateway_label( $order ) {
...
 	 *
 	 * @param int  $post_id    A post ID.
 	 *
 	 * @param int  $error_code The current error code.
 	 * @param bool $redirect   Whether to really redirect or not.
+	 *
+	 * @todo  Deprecate tpp_error
+	 *
+	 * @see   \Tribe__Tickets__Commerce__PayPal__Errors for error codes translations.
+	 * @todo  Determine if redirecting should be something relegated to some other method, and here we just actually
```
lucatume, line 792:
```suggestion
	 * @todo  Determine if redirecting should be something relegated to some other method, and here we only generate
```

src/Tickets/Commerce/Order.php
```diff
@@ -488,12 +522,12 @@ static function ( $item ) {
 		 * @param Value             $subtotal  The calculated subtotal of the cart items.
 		 * @param Gateway_Interface $gateway   The payment gateway used for the order.
 		 * @param ?array            $purchaser { Purchaser details.
-		 *    @type ?int    $purchaser_user_id    The purchaser user ID.
-		 *    @type ?string $purchaser_full_name  The purchaser full name.
-		 *    @type ?string $purchaser_first_name The purchaser first name.
-		 *    @type ?string $purchaser_last_name  The purchaser last name.
-		 *    @type ?string $purchaser_email      The purchaser email.
-		 * }
+		 * @type ?int $purchaser_user_id The purchaser user ID.
```
lucatume, line 525: These will need indentation to be restored to the previous state.

lucatume (changes requested):
Minor doc-block changes.
Dedicated tests? 
I know the feature is implicitly covered by other code, but I think having dedicated tests for the lock would prevent regressions.

lucatume (commented):
Nice work; ping me again when all tests pass.

src/Tickets/Commerce/Order.php
```diff
@@ -298,11 +298,11 @@ public function register_post_type() {
 		/**
 		 * Filter the arguments that craft the order post type.
 		 *
-		 * @see   register_post_type
 		 * @since 5.1.9
 		 *
 		 * @param array $post_type_args Post type arguments, passed to register_post_type()
 		 *
+		 * @see   register_post_type
```
Camwyn, line 305: Why did we move this?

src/Tickets/Commerce/Order.php
```diff
@@ -745,18 +779,19 @@ public function get_gateway_label( $order ) {
...
-	 *
-	 * @todo  Deprecate tpp_error
-	 *
-	 * @see   \Tribe__Tickets__Commerce__PayPal__Errors for error codes translations.
 	 * @since 5.1.9
 	 *
 	 * @param int  $post_id    A post ID.
 	 *
 	 * @param int  $error_code The current error code.
 	 * @param bool $redirect   Whether to really redirect or not.
```
Camwyn, line 787:
```suggestion
	 * @param int  $post_id    A post ID.
	 * @param int  $error_code The current error code.
	 * @param bool $redirect   Whether to really redirect or not.
```
No line break here

src/Tickets/Commerce/Order.php
```diff
@@ -1054,14 +1125,27 @@ public function is_order_locked( int $order_id ): bool {
...
-				DB::prepare(
-					'SELECT meta_value FROM %i WHERE post_id = %d AND meta_key = %s',
-					DB::prefix( 'postmeta' ),
-					$order_id,
-					static::CHECKOUT_COMPLETED_META
-				)
+			/**
+			 * Filters whether the checkout is completed.
+			 *
+			 * Direct query to overcome object cache.
```
Camwyn, line 1131:
```suggestion
			 * Does a direct query to overcome object cache.
```

src/Tickets/Commerce/Order.php
```diff
@@ -1054,14 +1125,27 @@ public function is_order_locked( int $order_id ): bool {
...
+			 *
+			 * @since TBD
+			 *
+			 * @param bool $is_completed Whether the checkout is completed.
+			 * @param int  $order_id     The order ID.
+			 */
+			return apply_filters(
```
Camwyn, line 1138: Personal preference is to do the query before the hook and then just pass the value to the hook.

src/Tickets/Commerce/Status/Status_Abstract.php
```diff
@@ -148,6 +148,29 @@ public function can_apply_to( $order, $new_status ) {
...
+	 *
+	 * @param self $new_status The new status.
+	 * @param ?int $order_id   Which order we are testing against.
+	 *
+	 * @return bool Whether the new status can be applied to the current status.
+	 */
+	public function can_transition_to_status( $new_status, $order_id = null ) {
+		if ( $this->get_wp_slug() === $new_status->get_wp_slug() ) {
+			// Allow from refunded to refunded in order to support multiple refunds.
+			return 'refunded' === $this->get_slug();
```
Camwyn, line 164: Shouldn't this really live in the Refunded status class?
The abstract class should have no knowledge if its children. The children should override this value.

Camwyn: Move that and this whole function becomes
`return $this->is_final()`

src/Tickets/Commerce/Status/Pending.php
```diff
@@ -60,8 +60,37 @@ public function get_name() {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param self $new_status     The new status.
+	 * @param ?int $order_id       Which order we are testing against.
+	 *
+	 * @return bool|WP_Error Whether the new status can be applied to the current status.
+	 */
+	public function can_transition_to_status( $new_status, $order_id = null ) {
+		return $this->can_transition_from_pending_to_status(
```
Camwyn, line 77: Why put this in a second function? It could all live in here, overriding the parent value.

Camwyn (changes requested):
Some questions

src/Tickets/Commerce/Order.php
dpanta94: Because Luca's IDE does these weird stuff on its own.

src/Tickets/Commerce/Status/Pending.php
dpanta94: undid all of that

src/Tickets/Commerce/Status/Status_Abstract.php
dpanta94: this was in abstact already - fixing that for my changes wont do it for existing code.

src/Tickets/Commerce/Status/Trashed.php
```diff
@@ -28,6 +28,19 @@ class Trashed extends Status_Abstract {
...
+	/**
+	 * Gets the slug of this status in WordPress.
+	 *
+	 * Since this is a core status, we do need to match.
+	 *
+	 * @since TBD
+	 *
+	 * @return string
+	 */
+	public function get_wp_slug() {
```
Camwyn, line 40: This should be unnecessary? It's identical to the [one in the abstract](https://github.com/the-events-calendar/event-tickets/pull/3489/files#diff-93d790a4d112780d1f419e952b6d8f36299dd3c503c04f6fe172b5ff5bddda1eR57).

Camwyn (changes requested):
One small observation.

src/Tickets/Commerce/Status/Trashed.php
dpanta94: you are confusing it with get_slug

Camwyn: Nope, you're right, sorry.

PR #3491, opened 2025-01-17
Make increase sales action idempotent

dpanta94:
[ET-2286]

Makes the increase sales action idempotent.

As a result is self aware whether the current order has increased the sales for which tickets and how many times for each.
Avoiding redoing the same sales increase on every consecutive fire of said action.

[ET-2286]: https://stellarwp.atlassian.net/browse/ET-2286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changelog/fix-increase-sales-action-to-idempotent
+++ b/changelog/fix-increase-sales-action-to-idempotent
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Ticket sales will be counted correctly during status transitions of the orders they belong to. [ET-2286]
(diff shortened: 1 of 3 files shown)


codingmusician (approved):
Looks solid.

rafsuntaskin:
@dpanta94 do we know why only increase sales were firing twice but not the attendee generation?

https://github.com/the-events-calendar/event-tickets/blob/8249bedf26a82be8eb232bbe50742f389d2f664e/src/Tickets/Commerce/Status/Action_Required.php#L35

dpanta94:
@rafsuntaskin i had already made that idempotent in previous PR in IMP.

See example about prod code in slack here https://lw.slack.com/archives/C0714HQSMEW/p1737128755307849?thread_ts=1736950314.961549&cid=C0714HQSMEW

PR #3492, opened 2025-01-17
Fix paginated search for fees

JPry:
### 🎫 Ticket

[ET-2268]

### 🗒️ Description

Fixes two issues:

1. An issue when trying to search for Booking Fees after the first page in the admin area. This ensures that the results properly handle pagination when searching by accounting for the last page of searched items.
2. An issue when **not** using search where all pages of results would show the first page of items. The query has been updated to properly account for pagination.

### 🎥 Artifacts 

**Before this change:**

https://github.com/user-attachments/assets/f99d6ad5-baef-4b65-aae1-90f207ed3f6e

**After this change:**

https://github.com/user-attachments/assets/f5ae8c75-56e2-475a-a675-bc801e1d1b31

### ✔️ Checklist

[ET-2268]: https://stellarwp.atlassian.net/browse/ET-2268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
+++ b/tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
@@ -95,6 +95,7 @@ class Create_Fees_Modifiers_Test extends Order_Modifiers_TestCase {
 	}
 
 	protected function get_table_class_instance(): Order_Modifier_Table {
+		tribe()->singleton( Fee_Table::class );
 		return tribe( Fee_Table::class );
 	}
 }
--- a/tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
+++ b/tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
@@ -4,6 +4,7 @@ namespace TEC\Tickets\Commerce\Order_Modifiers\Fees;
 
 use TEC\Tickets\Commerce\Order_Modifiers\Modifier_Admin_Handler;
 use TEC\Tickets\Commerce\Order_Modifiers\Table_Views\Fee_Table;
+use TEC\Tickets\Commerce\Order_Modifiers\Table_Views\Order_Modifier_Table;
 use Tribe\Tests\Traits\With_Uopz;
 use Tribe\Tickets\Test\Testcases\Order_Modifiers_TestCase;
 
@@ -87,9 +88,13 @@ class Create_Fees_Modifiers_Test extends Order_Modifiers_TestCase {
 
 		ob_start();
 
-		tribe( Fee_Table::class )->prepare_items();
+		$this->get_table_class_instance()->prepare_items();
 		tribe( Modifier_Admin_Handler::class )->render_tec_order_modifiers_page();
 
 		return ob_get_clean();
+	}
+
+	protected function get_table_class_instance(): Order_Modifier_Table {
+		return tribe( Fee_Table::class );
 	}
 }
--- a/tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
+++ b/tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
@@ -2,9 +2,13 @@
 
 namespace TEC\Tickets\Commerce\Order_Modifiers\Fees;
 
+use TEC\Tickets\Commerce\Order_Modifiers\Modifier_Admin_Handler;
+use Tribe\Tests\Traits\With_Uopz;
 use Tribe\Tickets\Test\Testcases\Order_Modifiers_TestCase;
 
 class Create_Fees_Modifiers_Test extends Order_Modifiers_TestCase {
+
+	use With_Uopz;
 
 	/**
 	 * The type of order modifier being tested (fee).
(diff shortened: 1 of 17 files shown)


bordoni (approved):

dpanta94 (changes requested):
there are tests failing

tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
```diff
@@ -39,4 +43,66 @@ public function does_edit_screen_render_with_no_data() {
...
+			'wp_redirect',
+			function( $location ) use ( $test ) {
+				$test->assertEquals( '?paged=1', $location );
+				return true;
+			},
+			true
+		);
+
+		// Prevent exit.
+		uopz_allow_exit( false );
```
dpanta94, line 96: Why ? This is going to silence tests that are failing. We have raised this discussion previously, please read here https://lw.slack.com/archives/C01H7Q57P1C/p1732802261942589

As suggested in the above discussion, use tribe_exit instead in your prod code.

tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
```diff
@@ -39,4 +43,66 @@ public function does_edit_screen_render_with_no_data() {
...
+
+		ob_start();
+
+		tribe( Modifier_Admin_Handler::class )->render_tec_order_modifiers_page();
+		$unset();
+
+		// Allow exit again.
+		uopz_allow_exit( true );
```
dpanta94, line 104: This is not the correct way to revert this - if your test inside the mocked wp_redirect this will never be executed and we will never see in the CI that the suite is failing

tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
```diff
@@ -39,4 +43,66 @@ public function does_edit_screen_render_with_no_data() {
...
+			true
+		);
+
+		// Prevent exit.
+		uopz_allow_exit( false );
+
+		ob_start();
+
+		tribe( Modifier_Admin_Handler::class )->render_tec_order_modifiers_page();
+		$unset();
```
dpanta94, line 101: No need really - the trait will do that in an `@after`

dpanta94 (changes requested):

tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
```diff
@@ -39,4 +43,66 @@ public function does_edit_screen_render_with_no_data() {
...
+			'order_modifier_display_name' => 'XXXX Unique Fee',
+		] )->id;
+
+		unset( $_REQUEST['s'], $_REQUEST['paged'], $_REQUEST['id'], $_REQUEST['edit'], $_POST['edit'] );
+
+		$_REQUEST = [
+			'modifier' => $this->modifier_type,
+			's'        => 'Unique Fee',
+			'paged'    => 3,
+		];
```
dpanta94, line 82: Paged here is 3 but check the snapshot its still 1.

tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
```diff
@@ -39,4 +43,66 @@ public function does_edit_screen_render_with_no_data() {
...
+		// We expect wp_redirect to be called.
+		$test  = $this;
+		$unset = $this->set_fn_return(
+			'wp_redirect',
+			function( $location ) use ( $test ) {
+				$test->assertEquals( '?paged=1', $location );
+				return true;
+			},
+			true
+		);
```
dpanta94, line 93: This assertion does not tell us if the wp_redirect funciton was actually called. Localize a variable inside the callback by referrence and then after it the function should have been called make the assertion.

e.g.

```php
$locaiton = null;
$this->set_fn_return(
			'wp_redirect',
			function( $redirect_to ) use ( $location ) {
				$location = $redirect_to;
				return true;
			},
			true
		);
		
		
.... execute what should call wp_redirect here...

$this->assert( $location ); // now we are sure what we expect was called!
```

tests/order_modifiers_integration/Fees/__snapshots__/Create_Fees_Modifiers_Test__does_table_render_correctly_with_search_and_page__0.snapshot.html
```diff
@@ -0,0 +1,49 @@
...
+
+				<div class="alignleft actions bulkactions">
+					</div>
+			<div class='tablenav-pages one-page'><span class="displaying-num">1 item</span>
+<span class='pagination-links'><span class="tablenav-pages-navspan button disabled" aria-hidden="true">&laquo;</span>
+<span class="tablenav-pages-navspan button disabled" aria-hidden="true">&lsaquo;</span>
+<span class="paging-input"><label for="current-page-selector" class="screen-reader-text">Current Page</label><input class='current-page' id='current-page-selector' type='text'
+					name='paged' value='1' size='1' aria-describedby='table-paging' /><span class='tablenav-paging-text'> of <span class='total-pages'>1</span></span></span>
```
dpanta94, line 15: see here the paged value is 1

dpanta94 (changes requested):
Seems like the snapshot is wrong :(

tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
JPry: @dpanta94 

> use tribe_exit instead in your prod code.

This isn't possible in this scenario. This happens when the table detects that a page is requested that doesn't exist; it calls `wp_redirect()` (if headers have not been sent) and that calls `exit`.

This is part of the native functionality of the `WP_List_Table` class.

tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
JPry: Ok, what would be the correct way to revert this?

tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
JPry: Yes, that's on purpose. That's part of what we're testing.

dpanta94 (approved):

redscar (approved):

redscar (approved):

codingmusician (approved):

PR #3494, opened 2025-01-20
Fix/simplify queries

dpanta94:
### 🎫 Ticket

[TICKET_ID]

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Commerce/Order_Modifiers/Modifiers/Modifier_Abstract.php
+++ b/src/Tickets/Commerce/Order_Modifiers/Modifiers/Modifier_Abstract.php
@@ -182,29 +182,32 @@ abstract class Modifier_Abstract implements Modifier_Strategy_Interface {
 	}
 
 	/**
+	 * Get modifiers along with their applied to meta key.
+	 *
+	 * @since TBD
+	 *
+	 * @param array $params               Additional parameters to filter the results.
+	 * @param bool  $with_applied_to_meta Whether to include the applied to meta key.
+	 *
+	 * @return array The modifiers that were found.
+	 */
+	public function get_modifiers( array $params = [], bool $with_applied_to_meta = true ): array {
+		return $this->repository->get_modifiers( $params, $with_applied_to_meta );
+	}
+
+	/**
 	 * Finds a modifier by its display name.
 	 *
 	 * @since 5.18.0
+	 * @deprecated TBD Use `get_modifiers` instead.
 	 *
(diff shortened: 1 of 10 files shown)


tests/order_modifiers_integration/Fees/Create_Fees_Modifiers_Test.php
```diff
@@ -85,45 +85,13 @@ protected function get_table_display_with_search(): string {
 		$_REQUEST = [
 			'modifier' => $this->modifier_type,
 			's'        => 'Unique Fee',
-			'paged'    => 3,
+			'paged'    => 1,
```
JPry, line 88: We shouldn't be changing this test. The point is to test against a previously-identified scenario where incorrect results were returned. Testing for the first page of results is fine, but querying for the 3rd page should actually return the first page of results. That's why `paged` is `3` here.

src/Tickets/Commerce/Order_Modifiers/Modifiers/Modifier_Abstract.php
```diff
@@ -182,29 +182,32 @@ public function get_modifier_by_applied_to( array $applied_to, array $params = [
...
+	 * @deprecated TBD Use `get_modifiers` instead.
 	 *
 	 * @param array $search Parameters to search Order Modifiers by.
 	 *
-	 * @return int The count of the modifiers found.
+	 * @return array The modifier data.
 	 */
-	public function find_count_by_search( array $search ): int {
-		return $this->repository->get_search_count( $search );
+	public function find_by_search( array $search ): array {
```
JPry, line 208: We don't need to deprecate this, as it's being added in this release. We can just remove it.

src/Tickets/Commerce/Order_Modifiers/Repositories/Order_Modifiers.php
```diff
@@ -203,38 +205,8 @@ public function find_by_id( int $id ): Order_Modifier {
...
-		// Add the query offset.
-		if ( ! empty( $valid_args['offset'] ) ) {
-			$query->offset( $valid_args['offset'] );
-		}
-
-		// Set the modifier type.
-		$query = $query->where( 'modifier_type', $this->modifier_type );
-
-		return $query->getAll() ?? [];
+		_deprecated_function( __METHOD__, 'TBD', 'get_modifiers' );
```
JPry, line 208: Same as above, this was added in this release so we don't need to deprecate; we can just remove it.

JPry (changes requested):

PR #3522, opened 2025-02-03
Ticket-agnostic cart

JPry:
### 🎫 Ticket

[ET-2200]

### 🗒️ Description

This PR updates the cart component to be able to handle non-ticket items (e.g. an agnostic cart).

The main changes are:

* Create a new class `\TEC\Tickets\Commerce\Cart\Agnostic_Cart`, which implements the `Cart_Interface` interface.
* A new class `\TEC\Tickets\Commerce\Cart\Cart_Item` was created to handle the cart items. This class implements `ArrayAccess` with a few additional methods, and an assurance of some common properties.
* The new `Agnostic_Cart` class becomes the default cart class (see `\TEC\Tickets\Commerce\Cart::get_repository()`).
* The `Agnostic_Cart` class has its own unit test suite similar to the `Unmanaged_Cart` suite to ensure it works as expected.
* Except for tests, wherever we referenced the prior cart class, `Unmanaged_Cart`, we should now be referencing `Cart_Interface` instead.
* A number of methods from the `Unmanaged_Cart` class were moved over to the `Abstract_Cart` parent class.
* Some of the logic for whether an item can be added to the cart was moved to the `Tribe__Tickets__Tickets_Handler` class.
* Within the `get_cart_total()` and `get_cart_subtotal()` methods, I implemented logic to prevent unnecessary duplicate calculations. During the process of working through this code, I discovered that the `get_cart_total()` method is called over 30 times within a single request, and the old logic allowed for those calculations to be performed fresh each time, even if the cart contents had not changed. This new logic handles when the cart contents *have* changed, and will ensure calculations are updated afterwards.
* The `Abstract_Cart::get_items_in_cart()` and `Commerce\Cart::get_items_in_cart()` methods have been updated to allow getting items of a particular type, or all items. The default behavior is to only get `ticket` items to preserve the current functionality in other places without needing to modify everything within this PR.

Template file updates:

* `src/views/v2/commerce/checkout/cart/ticket.php` - I added this file to handle the display of ticket items in the cart.
* `src/views/v2/commerce/checkout/cart/items.php` - This template file was updated to handle the display of ticket vs. non-ticket items. For the time being, the old `item.php` tempalte is identical to the `ticket.php` template, but now it will be easy to handle the display of non-ticket items in the cart.

There are some new hooks added:

* `tec_tickets_commerce_cart_cookie_expiration` **filter** - allows filtering the cookie expiration time for the cart.
* `tec_tickets_commerce_cart_repo_prepare_data` **filter** - allows filtering the data within the cart repository class (e.g. inside `Abstract_Cart::prepare_data()`).
* `tec_tickets_commerce_cart_process` **action** - allows access to the cart object and the `$data` array within the `process()` method. This is where additional types of items can be hooked into the cart.
* `tec_tickets_commerce_cart_add_full_item_params` **filter** - allows filtering the full item parameters before calculations are performed.
* `tec_tickets_commerce_cart_transient_expiration` **filter** - allows filtering the transient expiration time for the cart.

There are some hooks that were deprecated:

* `tec_tickets_commerce_cart_expiration` - use `tec_tickets_commerce_cart_cookie_expiration` instead.

### ✔️ Checklist

[ET-2200]: https://stellarwp.atlassian.net/browse/ET-2200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Commerce/Cart.php
```diff
@@ -409,18 +397,18 @@ public function maybe_delete_expired_products() {
 			return;
 		}
 
-		$transient = get_transient( $transient_key );
-
-		// Bail if we have no data to delete.
-		if ( empty( $transient ) ) {
+		// Bail if ET+ is not in place.
+		if ( ! class_exists( Meta_Storage::class ) ) {
```
redscar, line 401: Do you want to switch this to 

`if ( did_action('tec_tickets_plus_attendee_bind_implementations') === 0 ) {`?

src/Tickets/Commerce/Cart.php
```diff
@@ -679,4 +624,78 @@ public static function get_cart_hash_cookie_name(): string {
...
+	 * @since 5.1.9
+	 * @depecated Use `get_transient_key()` via the Cart trait instead.
+	 *
+	 * @param string $id The cart id.
+	 *
+	 * @return string The transient name.
+	 */
+	public static function get_transient_name( $id ) {
+		_deprecated_function( __METHOD__, 'TBD', 'TEC\Tickets\Commerce\Traits\Cart::get_transient_key' );
+		return Commerce::ABBR . '-cart-' . md5( $id ?? '' );
```
redscar, line 699: Should you call the trait here as a passthrough until we remove this method? That way the data stays in sync.

JPry: @redscar It's just a string concatenation, and the trait does the same thing. I just thought it made more sense for using this where it is needed to put it in a trait rather than relying on static methods.

src/Tickets/Commerce/Cart.php
JPry: I'd be up for changing it, but we're also directly using the class we're checking after this statement. So I think it would be safer to keep it as-is.

src/Tickets/Commerce/Cart/Unmanaged_Cart.php
```diff
@@ -14,65 +16,21 @@
...
-
 	/**
 	 * @var array|null The list of items, null if not retrieved from transient yet.
 	 */
 	protected $items = null;
 
 	/**
-	 * {@inheritDoc}
-	 */
-	public function has_public_page() {
```
JPry, line 30: Methods and properties removed from this class were added to the `Abstract_Cart` class.

src/Tickets/Commerce/Cart/Unmanaged_Cart.php
```diff
@@ -113,48 +84,73 @@ public function get_items() {
...
+	 *
+	 * @since 5.1.9
+	 *
+	 * @return bool|int The number of products in the cart (regardless of the products quantity) or `false`
+	 *
 	 */
 	public function has_items() {
 		$items = $this->get_items();
 
+		// When we don't have items, return false.
```
JPry, line 138: This was added to ensure the method behaves as indicated by the interface.

src/Tickets/Commerce/Cart/Agnostic_Cart.php
```diff
@@ -0,0 +1,525 @@
...
+		);
+	}
+
+	/**
+	 * Add the ticket parameters to the item in the cart.
+	 *
+	 * @param array $item The item in the cart.
+	 *
+	 * @return ?array The item in the cart with the full set of parameters.
+	 */
```
redscar, line 409:
```suggestion
	/**
	 * Add the ticket parameters to the item in the cart.
	 *
	 * @since TBD
	 *
	 * @param array $item The item in the cart.
	 *
	 * @return ?array The item in the cart with the full set of parameters.
	 */
```

src/Tickets/Commerce/Cart/Agnostic_Cart.php
```diff
@@ -0,0 +1,525 @@
...
+		return true;
+	}
+
+	/**
+	 * Add the full set of parameters to the items in the cart.
+	 *
+	 * @param array $items The items in the cart.
+	 *
+	 * @return array The items in the cart with the full set of parameters.
+	 */
```
redscar, line 372:
```suggestion
	/**
	 * Add the full set of parameters to the items in the cart.
	 *
	 * @since TBD
	 *
	 * @param array $items The items in the cart.
	 *
	 * @return array The items in the cart with the full set of parameters.
	 */
```

src/Tickets/Commerce/Cart.php
```diff
@@ -15,6 +20,8 @@
  */
 class Cart {
 
+	use Cart_Trait;
+
 	/**
 	 * Which URL param we use to identify a given page as the cart.
```
redscar, line 26: This file needs the `@since` to be updated in the docblocks. I see a lot of them not touched.

src/Tickets/Commerce/Cart.php
```diff
@@ -313,10 +303,25 @@ public function set_cart_hash_cookie( $value = '' ) {
 		 * Filters the life span of the Cart Cookie.
 		 *
 		 * @since 5.1.9
+		 * @since TBD Deprecated in favor of `tec_tickets_commerce_cart_cookie_expiration`.
+		 *
```
redscar, line 307: Do we need to add `@deprecated`?

redscar (approved):

src/Tickets/Commerce/Cart.php
JPry: Wouldn't hurt. I'll add it.

src/Tickets/Commerce/Cart.php
JPry: Thanks, I added updated `@since` tags to things that had notable changes.

Camwyn (commented):
comment

Camwyn:
Ignore the rando comment/request/approval above. Jeremy and I were testing something.

redscar (approved):

src/Tickets/Commerce/Cart.php
```diff
@@ -409,18 +418,18 @@ public function maybe_delete_expired_products() {
...
-		// Bail if we have no data to delete.
-		if ( empty( $transient ) ) {
+		// Bail if ET+ is not in place.
+		if ( ! class_exists( Meta_Storage::class ) ) {
 			return;
 		}
 
-		// Bail if ET+ is not in place.
-		if ( ! class_exists( 'Tribe__Tickets_Plus__Meta__Storage' ) ) {
+		$storage = new Meta_Storage();
```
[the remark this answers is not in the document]
Camwyn: @bordoni - should we alter this to use the singleton?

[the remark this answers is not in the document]
JPry: This was old code; I only changed it to be an alias instead of the long name.

I think this whole method should be moved to ET+. There's nothing here that really should be in ET.

With that said, I think that can be a separate set of PRs.

src/Tickets/Commerce/Cart/Abstract_Cart.php
```diff
@@ -17,54 +20,85 @@
...
+	/**
+	 * Cart subtotal.
+	 *
+	 * This should be the total of items in the cart without any additional calculations.
+	 *
+	 * @since TBD
+	 *
+	 * @var ?float
+	 */
+	protected ?float $cart_subtotal = null;
```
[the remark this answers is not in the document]
JPry: Old code, trying not to break anything.

src/Tickets/Commerce/Cart/Unmanaged_Cart.php
```diff
@@ -163,7 +158,13 @@ public function has_item( $item_id ) {
...
-	 * {@inheritdoc}
+	 * Adds a specified quantity of the item to the cart.
+	 *
+	 * @since 5.1.9
+	 *
+	 * @param int|string $item_id    The item ID.
+	 * @param int        $quantity   The quantity to add.
+	 * @param array      $extra_data Extra data to save to the item.
 	 */
 	public function add_item( $item_id, $quantity, array $extra_data = [] ) {
```
Camwyn, line 169: Do we lose anything by simplifying this?
```php
public function add_item( $item_id, $quantity, array $extra_data = [] ) {
	// cast immediately - we only use it once anyway.
	$current_quantity = (int) $this->has_item( $item_id ); 

	// if current quantity is 0, it won't hurt to just add it. If it's a negative number, we have other issues
	$new_quantity = (int) $quantity + $current_quantity; 

	// Remove the unneccessary max() calculation - since we really only care (and use it) if it's above 0.

	if ( 0 < $new_quantity ) {
		$item['ticket_id'] = $item_id;
		$item['quantity']  = $new_quantity;
		$item['extra']     = $extra_data;

		$this->items[ $item_id ] = $item;
	} else {
		// PS: I hate that we call remove from add which then calls add...
		$this->remove_item( $item_id );
	}
}```

File: src/Tickets/Commerce/Cart/Unmanaged_Cart.php [hunk 12]
```diff
@@ -188,7 +189,12 @@ public function add_item( $item_id, $quantity, array $extra_data = [] ) {
...
-	 * {@inheritdoc}
+	 * Removes an item from the cart.
+	 *
+	 * @since 5.1.9
+	 *
+	 * @param int|string $item_id  The item ID.
+	 * @param null|int   $quantity The quantity to remove.
 	 */
 	public function remove_item( $item_id, $quantity = null ) {
 		if ( null !== $quantity ) {
```
Camwyn, line 200: `is_numeric()` ?

src/Tickets/Commerce/Cart/Unmanaged_Cart.php
```diff
@@ -259,9 +276,18 @@ public function process( array $data = [] ) {
...
+	 * @param string|int $item_id     The item ID.
+	 * @param int        $quantity    The quantity of the item. If the item exists, this quantity will override
+	 *                                the previous quantity. Passing 0 will remove the item from the cart entirely.
+	 * @param array      $extra_data  Extra data to save to the item.
+	 *
+	 * @return void
 	 */
-	public function prepare_data( array $data = [] ) {
-		return $data;
+	public function upsert_item( $item_id, int $quantity, array $extra_data = [] ) {
```
Camwyn, line 290: This is now a useless function.

src/views/v2/commerce/checkout/cart/items.php
```diff
@@ -11,34 +11,60 @@
...
+				$this->template(
+					'checkout/cart/item',
+					[
+						'section' => $section,
+						'post'    => $post,
+						'item'    => $item,
+					]
+				);
+		}
+	}
```
Camwyn, line 68:
```suggestion
		$item_type = $item['type'] ?? 'ticket';

$path = $item_type === 'ticket' ? 'checkout/cart/ticket' : 'checkout/cart/item';

$this->template(
	$path,
	[
		'section' => $section,
		'post'    => $post,
		'item'    => $item,
	]
);
```

Camwyn (changes requested):

JPry:
@Camwyn Those comments you made are for the `Unmanaged_Cart` class, which I'm trying to leave alone as much as possible. Do we want to change it further, or leave its functionality as-is?

src/Tickets/Commerce/Cart/Unmanaged_Cart.php
JPry: Yes, it's maintaining the way this class has functioned with minimal changes. All the real changes are in the `Agnostic_Cart` class.

Camwyn:
> @Camwyn Those comments you made are for the `Unmanaged_Cart` class, which I'm trying to leave alone as much as possible. Do we want to change it further, or leave its functionality as-is?

Oh dang, no leave it as is.

Camwyn (approved):

src/Tickets/Commerce/Cart.php
```diff
@@ -228,17 +224,12 @@ public function generate_cart_order_hash( $salt = '' ): string {
...
+		$cart_hash        = $this->get_repository()->get_hash();
 		$hash_from_cookie = sanitize_text_field( $_COOKIE[ static::get_cart_hash_cookie_name() ] ?? '' );
 
-		if (
-			strlen( $hash_from_cookie ) === $cart_hash_length
-		) {
-			$cart_hash = $hash_from_cookie;
-
-			$cart_hash_transient = get_transient( static::get_transient_name( $cart_hash ) );
+		if ( strlen( $hash_from_cookie ) === $cart_hash_length ) {
```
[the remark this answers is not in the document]
JPry: 🤷🏻‍♂️ I took a stab at reworking the logic here to make more sense, but there were too many unknowns (to me) about why it works the way it does right now. So I decided to clean up formatting a little bit, but otherwise leave the logic intact.

It seems to me that this is checking to make sure the cart hash length matches the length that's set at the beginning of the method, but I'm not sure what it's supposed to be checking. Maybe to ensure it isn't an empty string? Maybe to be sure it doesn't contain some kind of garbage value? In the latter case, it has the weakness you mentioned: any string of the same length will pass.

PR #3532, opened 2025-02-06
Feat/ticket actions

dpanta94:
### 🎫 Ticket

[ETP-975]

### 🗒️ Description

Adds some Hooks for our tickets!

This feature is needed as part of the Waitlist feature. The reason is the waitlist will need to process waitlist subscribers exactly based on these hooks we are adding here.

A quick description about them:

Listening for stock changes is being done through the updated meta data. All of our tickets are using the meta key `_stock` for that. Since this is a SINGLE field that way is ideal. As long as the rest of our codebase keeps the field updated (it should happen already) the listener is working ok.

Listening for ticket date start and date end is a bit more complicated.

We want to avoid having to query all of our tickets every X amount of time to determine this. => Why ? Not required load and not reliable. e.g. a ticket goes on sale in the X meantime we didnt check.

Additionally, this is a personal opinion, but the whole WP ecosystem is event based. So i like going with events as well.

The functionality works as:

A single action is bound to the date time each ticket start date will be reached. The same happens for the ticket's end date. Whenever each one of those datetimes is updated the actions are updated as well.

That allows when the action fires, anyone who listens for it to take action.

The code has been fully tested for what is being added. Also performed a test to evaluate the performance of such a process.

### ✔️ Checklist

[ETP-975]: https://stellarwp.atlassian.net/browse/ETP-975?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tickets/Ticket_Actions.php
+++ b/src/Tickets/Ticket_Actions.php
@@ -160,7 +160,7 @@ class Ticket_Actions extends Controller_Contract {
 	 *
 	 * @since TBD
 	 *
-	 * @param int    $ticket_id The ticket id.
+	 * @param int $ticket_id The ticket id.
 	 *
 	 * @return void
 	 */
@@ -305,8 +305,8 @@ class Ticket_Actions extends Controller_Contract {
 		$its_happening = true;
 
 		$prefix = $is_start ? 'start' : 'end';
-
 		$method = "{$prefix}_date";
+
 		$timestamp = $ticket->$method();
 
 		if ( ! $timestamp ) {
--- a/src/Tickets/Ticket_Actions.php
+++ b/src/Tickets/Ticket_Actions.php
@@ -1,0 +1,468 @@
+<?php
+/**
+ * The Ticket Actions controller.
+ *
+ * @since TBD
+ * @package TEC\Tickets
+ */
+
+namespace TEC\Tickets;
+
+use TEC\Common\Contracts\Provider\Controller as Controller_Contract;
+
+use Tribe__Tickets__Tickets as Tickets;
+use Tribe__Tickets__Ticket_Object as Ticket_Object;
+use TEC\Tickets\Commerce\Ticket;
+use WP_Post;
+use Exception;
+
+/**
+ * Class Ticket_Actions.
+ *
+ * @since TBD
+ * @package TEC\Tickets
+ */
+class Ticket_Actions extends Controller_Contract {
+	/**
+	 * The action that will be fired when a ticket's start time is almost reached or reached or just reached briefly in the past.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const TICKET_START_SALES_HOOK = 'tec_tickets_ticket_start_sales';
+
+	/**
+	 * The action that will be fired when a ticket's end time is almost reached or reached or just reached briefly in the past.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const TICKET_END_SALES_HOOK = 'tec_tickets_ticket_end_sales';
+
+	/**
+	 * The action scheduler group for ticket actions.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	public const AS_TICKET_ACTIONS_GROUP = 'tec_tickets_ticket_actions';
+
+	/**
+	 * Registers the controller by subscribing to front-end hooks and binding implementations.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	protected function do_register(): void {
+		add_action( 'tec_tickets_ticket_upserted', [ $this, 'sync_ticket_dates_actions' ], PHP_INT_MAX );
+		add_action( 'added_post_meta', [ $this, 'sync_rsvp_dates_actions' ], PHP_INT_MAX, 3 );
+		add_action( 'updated_postmeta', [ $this, 'sync_rsvp_dates_actions' ], PHP_INT_MAX, 3 );
+		add_action( 'added_post_meta', [ $this, 'fire_stock_update_action' ], PHP_INT_MAX, 3 );
+		add_action( 'updated_postmeta', [ $this, 'fire_stock_update_action' ], PHP_INT_MAX, 3 );
+		add_action( self::TICKET_START_SALES_HOOK, [ $this, 'fire_ticket_start_date_action' ], 10, 2 );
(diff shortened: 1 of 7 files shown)


src/Tickets/Ticket_Actions.php
```diff
@@ -0,0 +1,464 @@
...
+
+		$event = $ticket->get_event();
+
+		if ( ! $event instanceof WP_Post || 1 > $event->ID ) {
+			// Parent event, no longer exists.
+			return;
+		}
+
+		$cache     = tribe_cache();
+		$cache_key = __METHOD__ . $ticket_id;
```
lucatume, line 247: This is an ephemeral cache, still, though: per the whole request lifecycle.

Is there a case where this should be invalidated in the context of the same request? E.g. if the action hooked to `shutdown` is removed with `remove_action`?

Next question: why not use a `has_action` check to see if this class has already hooked on `shutdown`?
This would avoid the hooking of a Closure (few lines below) that then cannot be unhooked by other code.

E.g. you could save the ticket IDs to fire the `$this->sync_ticket_date_actions` for in an array property on this class, hook **once** on `shutdown` and fire the logic for each ticket ID that requires it.
This would allow for a check like `has_action('shutdown', [$this, 'sync_all_ticket_dates_actions'])` and a similar `remove_action` call.
The method `sync_all_ticket_dates_actions` would do nothing if there are no ticket IDs, it could be hooked in `do_register` and unhooked in `unregister` cleaning up the API a bit.
And then you would not need the cache.

src/Tickets/Ticket_Actions.php
```diff
@@ -0,0 +1,464 @@
...
+			// What is going on ? We cant work with that...
+			return;
+		}
+
+		// Unschedule first pre-existing actions.
+		as_unschedule_action( self::TICKET_START_SALES_HOOK, [ $ticket_id ], self::AS_TICKET_ACTIONS_GROUP );
+		as_unschedule_action( self::TICKET_END_SALES_HOOK, [ $ticket_id ], self::AS_TICKET_ACTIONS_GROUP );
+
+		if ( time() > $end_timestamp ) {
+			// The ticket has already ended. Do nothing.
```
lucatume, line 383: "The ticket [sale] has already ended." - correct?
The **ticket** cannot really end.

src/Tickets/Ticket_Actions.php
```diff
@@ -0,0 +1,464 @@
...
+	 *
+	 * @since TBD
+	 *
+	 * @param int $ticket_id       The ticket ID.
+	 * @param int $start_timestamp The ticket start date.
+	 *
+	 * @return void
+	 */
+	protected function schedule_date_start_action( int $ticket_id, int $start_timestamp ): void {
+		$now              = time();
```
lucatume, line 402: The `time` function returns the timestamp in seconds; it's a rounding of the microseconds time.

Would it make sense to get and store the `$now` value when this class is constructed, in the `__construct` method? 
If I get the code correctly, "now" is really modeling "the moment the current request started" which is not really variable.

If you keep fetching `time()` in each method, then you have to test for three possible states depending on when the rounding happens and how fast the machine is:
* no method executes
* some method do not execute at `t`, some method execute at `t+1` as the machine processes code an the rounding happen
* all methods execute

While the code is likely already robust to the second case, it could entirely be avoided collecting `time()` once, on `__construct`.

lucatume (changes requested):
The code is good, I've proposed some changes that would not horribly alter its nature.

src/Tickets/Ticket_Actions.php
```diff
@@ -0,0 +1,464 @@
...
+			$tickets_handler->key_start_time,
+			$tickets_handler->key_end_date,
+			$tickets_handler->key_end_time,
+		];
+
+		if ( ! in_array( $meta_key, $keys_of_interest, true ) ) {
+			return;
+		}
+
+		if ( 'tribe_rsvp_tickets' !== get_post_type( $ticket_id ) ) {
```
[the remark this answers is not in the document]
dpanta94: I would like to avoid that as well but i dont want to init a class or use the service locator to grab it.

Currently is a public dynamic property of a class when it could be a const.

https://github.com/the-events-calendar/event-tickets/blob/15f56a9a3d97f5cdd8f89835d692b2533981d887/src/Tribe/RSVP.php#L68

https://github.com/the-events-calendar/event-tickets/blob/ff539ea16f4e02f8ceea0b5ea25e2da88b4b9495/src/Tickets/Ticket_Cache_Controller.php#L232

src/Tickets/Ticket_Actions.php
```diff
@@ -0,0 +1,464 @@
...
+
+	/**
+	 * Registers the controller by subscribing to front-end hooks and binding implementations.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	protected function do_register(): void {
+		add_action( 'tec_tickets_ticket_upserted', [ $this, 'sync_ticket_dates_actions' ], PHP_INT_MAX );
```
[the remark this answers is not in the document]
dpanta94: Makes sense!

src/Tickets/Ticket_Actions.php
```diff
@@ -0,0 +1,464 @@
...
+		add_action( 'added_post_meta', [ $this, 'sync_rsvp_dates_actions' ], PHP_INT_MAX, 3 );
```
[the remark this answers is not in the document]
dpanta94: Very good observation! Thank you. Doing that change!

src/Tickets/Ticket_Actions.php
dpanta94: > Is there a case where this should be invalidated in the context of the same request? E.g. if the action hooked to shutdown is removed with remove_action?

Not that i can think of.

But the rest of your comment makes very much sense so i am going to test it out. Very good idea regardless of if it works or not. Coming back to this soon.

src/Tickets/Ticket_Actions.php
dpanta94: In the constructor i feel like that would make testing the functionality too complex. Also it might introduce an not needed complexity in the future.

We can keep the processes into the same time by passing `time()` around as a local var. So its set once at the start of every operation and its steady for that operation.

src/Tickets/Ticket_Actions.php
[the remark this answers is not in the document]
lucatume: I'm neutral to this. Your argument makes sense: this _should_ be a `public static` property, at least.

lucatume (approved):

PR #3534, opened 2025-02-07
Support UTC timezone

dpanta94:
Support UTC timezone. I know offsets can be a serious pain. but UTC itself is not.

Also our tests work on UTC, and when its not supported for time sensitive tests its an issue.


andrasguseo (approved):
Looks logical.

PR #3571, opened 2025-02-19
ET-2198/ET-2199 Coupon code entry and validation

JPry:
- **Add checkout controller to the main controller**
- **Add initial JS files**
- **Create Type trait**
- **DELETE ME temporary revert coupon JS**
- **Move coupons before fees**
- **Add multiplication for Precision value**
- **Update the HTML for the coupon form**
- **Use correct properties for the start/end dates**
- **Add method get_discount_amount() for a coupon model**
- **Initial updates to the coupons API and checkout JS**
- **Initial styling of the Coupon form**
- **Start working on modern asset for chekcout**
- **Update snapshots with new form markup**
- **Order imports alphabetically**
- **Add is_coupon() to Type trait**
- **Fix action typo**
- **Add coupons as properties and cart values**
- **Allow test create_order() to accept more arguments**
- **Add tests for coupons**
- **Add stripe controller**
- **Allow getting the raw integer value from the precision value**
- **Add initial tests for the Coupons API**
- **Update API class with corrected logic and return values**

### 🎫 Ticket

[ET-2198]
[ET-2199]

### ✔️ Checklist


src/Tickets/Commerce/Order_Modifiers/Traits/Coupons.php
```diff
@@ -22,24 +23,35 @@
...
 
 	/**
-	 * The repository for interacting with the order modifiers table.
+	 * Determine if a coupon is valid.
 	 *
-	 * @since 5.18.0
+	 * @param string $slug The coupon slug.
 	 *
-	 * @var Coupons_Repository
+	 * @return bool
```
redscar, line 30: Missing `@since`

src/Tickets/Commerce/Shortcodes/Checkout_Shortcode.php
```diff
@@ -42,12 +42,11 @@ class Checkout_Shortcode extends Shortcode_Abstract {
 	 */
 	public function setup_template_vars() {
 		$cart          = tribe( Cart::class );
-		$items         = $cart->get_items_in_cart( true );
 		$cart_subtotal = Value::create( $cart->get_cart_subtotal() ?? 0 );
 		$cart_total    = Value::create( $cart->get_cart_total() ?? 0 );
+		$items         = $cart->get_repository()->update_items_with_subtotal( $cart->get_items_in_cart( true, 'all' ) );
```
redscar, line 47: Need to update `@since` of method.

src/Tickets/Commerce/Cart/Abstract_Cart.php
```diff
@@ -108,23 +122,27 @@ public function get_items_in_cart( $full_item_params = false, string $type = 'ti
 	 * Get the total value of the cart, including additional values such as fees or discounts.
 	 *
 	 * This method calculates the total by first computing the subtotal from all items in the cart,
-	 * and then applying any additional values (e.g., fees or discounts) provided via the `tec_tickets_commerce_get_cart_additional_values` filter.
+	 * and then applying any additional values (e.g., fees or discounts) provided via the
+	 * `tec_tickets_commerce_get_cart_additional_values` filter.
 	 *
-	 * @since 5.18.0 Refactored logic, to include a new filter.
 	 * @since 5.10.0
+	 * @since 5.18.0 Refactored logic, to include a new filter.
```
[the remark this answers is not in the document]
JPry: Good idea

src/Tickets/Commerce/Cart/Abstract_Cart.php
```diff
@@ -162,22 +182,33 @@ public function get_cart_total() {
 	 * @return float The subtotal of the cart.
 	 */
 	public function get_cart_subtotal(): float {
```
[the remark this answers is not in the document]
JPry: I updated it now.

src/Tickets/Commerce/Traits/Type.php
```diff
@@ -0,0 +1,82 @@
...
+	 * "coupon". If the type key is not set, or if it is set to something other
+	 * than "coupon", this will return false.
+	 *
+	 * @since TBD
+	 *
+	 * @param array $thing The thing to check.
+	 *
+	 * @return bool Whether the thing is a coupon.
+	 */
+	protected function is_coupon( array $thing ): bool {
```
Camwyn, line 32: Why do we need this _and_ the one in Coupon above? Too much specificity here?

src/resources/js/v2/tickets-commerce.js
```diff
@@ -204,5 +219,187 @@ tribe.tickets.commerce = {};
...
+			'text/html',
+		).body.textContent;
+
+		$couponValueElement.text( unescapedDiscount );
+	};
+
+	obj.bindCouponApply = function() {
+		let ajaxInProgress = false;
+
+		// $document.on( 'click', obj.selectors.couponApplyButton, applyCoupon );
```
Camwyn, line 256: remove?

src/resources/js/v2/tickets-commerce.js
```diff
@@ -204,5 +219,187 @@ tribe.tickets.commerce = {};
...
+					if ( response.success ) {
+						// Hide input and button, show applied coupon.
+						$( obj.selectors.couponInputContainer ).hide();
+
+						// Display coupon value and discount.
+						obj.updateCouponDiscount( response.discount );
+						obj.updateTotalPrice( response.cart_amount );
+						$( obj.selectors.couponAppliedSection ).show();
+					} else {
+						$errorMessage.text( response.message || 'Invalid Coupon Code' ).show();
```
Camwyn, line 317: i18n?

src/resources/js/v2/tickets-commerce.js
```diff
@@ -204,5 +219,187 @@ tribe.tickets.commerce = {};
...
+			const couponValue = $( obj.selectors.couponInput ).val().trim();
+			const nonce = $( obj.selectors.nonce ).val();
+			const $errorMessage = $( obj.selectors.couponError );
+
+			// Hide the error message initially.
+			$errorMessage.hide();
+
+			// Ensure the coupon is not empty.
+			if ( ! couponValue ) {
+				$errorMessage.text( 'Coupon code cannot be empty.' ).show();
```
Camwyn, line 285: i18n?

src/views/v2/commerce/checkout/cart/item.php
```diff
@@ -35,6 +35,18 @@
...
 
+/**
+ * Filters whether the current item should be skipped in the checkout.
+ *
+ * @since TBD
+ *
+ * @param bool  $should_skip Whether the item should be skipped or not.
+ * @param array $item        The item to be checked.
+ */
+if ( apply_filters( 'tec_tickets_checkout_should_skip_item', false, $item ) ) {
```
Camwyn, line 46: Why wait until the last minute in the template to filter this? Add a filtered prop to the item.

Camwyn: Also - check the comment above the use - it's on the wrong line. (L31)

src/views/v2/commerce/checkout/order-modifiers/coupons.php
```diff
@@ -3,32 +3,77 @@
...
+
+$apply_button_classes = [
+	'tec-tickets__commerce-checkout-cart-coupons__apply-button',
+	'tribe-common-c-btn',
+];
+
+$coupon_items = array_filter( $items, fn( $item ) => 'coupon' === ( $item['type'] ?? '' ) );
+
+// If we have any coupons, always use the first one. There *shouldn't* be more than one.
+$coupon = ! empty( $coupon_items ) ? $coupon_items[ array_key_first( $coupon_items ) ] : [];
```
Camwyn, line 29: `array_shift`? We never use `$coupon_items` again...

src/views/v2/commerce/checkout/order-modifiers/coupons.php
```diff
@@ -3,32 +3,77 @@
...
+
+// Determine the discount to display.
+$discount = '';
+if ( isset( $coupon['sub_total'] ) && $coupon['sub_total'] instanceof Value ) {
+	$discount = $coupon['sub_total']->get_currency();
+}
+
+// Sections that should be hidden based on the coupon state.
+$hide_input   = ! empty( $coupon );
+$hide_applied = empty( $coupon );
```
Camwyn, line 39: Why two values that are opposites - just use one and compare `$hide_input` or `! $hide_input`

Or, (actually, _also_) add a class instead of an inline-style. Use `tribe_classes()` with conditionals.

i.e.
```
tribe_classes(
    [
        'tec-tickets__commerce-checkout-cart-coupons' => true
        'hide-that-sucker' => empty( $coupon )
    ]
);

File: src/views/v2/commerce/checkout/order-modifiers/coupons.php [hunk 12]
```diff
@@ -3,32 +3,77 @@
...
+			<span class="tec-tickets__commerce-checkout-cart-coupons__applied-value">
+				<?php echo esc_html( $coupon['display_name'] ?? '' ); ?>
+			</span>
+			- <?php esc_html_e( 'Discount:', 'event-tickets' ); ?>
+			<span class="tec-tickets__commerce-checkout-cart-coupons__applied-discount">
+				<?php echo esc_html( $discount ); ?>
+			</span>
+		</span>
+		<button class="tec-tickets__commerce-checkout-cart-coupons__remove-button" type="button">
+			<?php echo esc_html_x( 'X', 'text for button to remove an applied coupon', 'event-tickets' ); // @todo use an ICON instead. @see https://github.com/the-events-calendar/event-tickets/pull/3430#discussion_r1872545653 ?>
```
Camwyn, line 76: Get this `@TODO` out of here and do it.

Camwyn: https://github.com/the-events-calendar/tribe-common/blob/8de5260664aef59e4d826bfdfec0a283c93bea31/src/resources/images/icons/close.svg

Camwyn (changes requested):

src/Tickets/Commerce/Order_Modifiers/API/Coupons.php
```diff
@@ -90,7 +101,7 @@ protected function register_routes(): void {
 			'/coupons/validate',
 			[
 				'methods'             => Server::CREATABLE,
-				'callback'            => [ $this, 'validate_coupon' ],
+				'callback'            => fn( Request $request ) => $this->validate_coupon( $request ),
```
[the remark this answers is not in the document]
JPry: We could use the array syntax, but this allows for keeping the methods `protected` rather than `public`. I prefer to have as few `public` methods as possible, as it keeps things much simpler if/when we need to refactor later.

src/Tickets/Commerce/Order_Modifiers/API/Coupons.php
```diff
@@ -537,4 +553,23 @@ protected function get_purchaser_information( Request $request ) {
...
+	 * Get the coupon slug from the request object and validate it.
+	 *
+	 * @since TBD
+	 *
+	 * @param Request $request The request object.
+	 *
+	 * @return string The coupon slug.
+	 * @throws Exception If the coupon slug is invalid.
+	 */
+	protected function get_and_validate_coupon_slug( Request $request ): string {
```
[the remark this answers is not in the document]
JPry: I might have had a good reason at the time, but changed it later. Good call to only pass the slug.

src/Tickets/Commerce/Order_Modifiers/Checkout/Coupons.php
```diff
@@ -101,4 +145,88 @@ public function localize_asset(): void {
...
+	public function attach_coupons_to_order_object( array $properties ): array {
+		// We shouldn't have an order with no items, but let's just be safe.
+		$items = $properties['items'] ?? [];
+		if ( empty( $items ) ) {
+			return $properties;
+		}
+
+		// Separate coupons and non-coupons.
+		$coupons = array_filter( $items, fn( $item ) => $this->is_coupon( $item ) );
+		$items   = array_filter( $items, fn( $item ) => ! $this->is_coupon( $item ) );
```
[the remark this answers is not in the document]
JPry: Sure, that's a reasonable alternative. But it doesn't have the nice symmetry! 😅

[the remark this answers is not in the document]
JPry: On second thought, while I was reading up on [`array_diff`](https://www.php.net/manual/en/function.array-diff.php), I encountered this:

> Two elements are considered equal if and only if `(string) $elem1 === (string) $elem2`. That is, when the [string representation](https://www.php.net/manual/en/language.types.string.php#language.types.string.casting) is the same.

Since we're working with arrays, I think it might be better to stick with the current filtering.

src/Tickets/Commerce/Order_Modifiers/Checkout/Gateway/Stripe/Coupons.php
```diff
@@ -0,0 +1,105 @@
...
+			} catch ( Exception $e ) {
+				continue;
+			}
+
+			$items[ $id ] = [
+				'id'           => $id,
+				'type'         => 'coupon',
+				'coupon_id'    => $coupon->id,
+				'price'        => $coupon->raw_amount,
+				'sub_total'    => -1 * $coupon->get_discount_amount( $subtotal->get_float() ),
```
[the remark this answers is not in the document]
JPry: That's a good point. I think it would be reasonable to put the negative into `$coupon->get_discount_amount()`, and then use `abs()` if we need to display the number as positive.

src/Tickets/Commerce/Order_Modifiers/Traits/Coupons.php
```diff
@@ -48,49 +62,141 @@ protected function is_coupon_slug_valid( string $slug ): bool {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param Coupon $coupon The coupon to check.
+	 *
+	 * @return bool Whether the coupon is within its start and end date range.
+	 *
+	 * @throws Exception If the coupon end date has passed or the coupon start date is in the future.
+	 */
+	protected function is_coupon_within_date_ranage( Coupon $coupon ): bool {
```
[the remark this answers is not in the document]
JPry: Whoops 😆

src/Tickets/Commerce/Order_Modifiers/Traits/Coupons.php
```diff
@@ -48,49 +62,141 @@ protected function is_coupon_slug_valid( string $slug ): bool {
...
+		// Current time stamp for checking start and end dates.
+		$current_time = time();
+
+		// If the coupon end date has passed, it is invalid.
+		if ( null !== $coupon->end_time && strtotime( $coupon->end_time ) < $current_time ) {
+			throw new Exception( 'Coupon end date has passed' );
+		}
+
+		// If the coupon start date is in the future, it is invalid.
+		if ( null !== $coupon->start_time && strtotime( $coupon->start_date ) > $current_time ) {
```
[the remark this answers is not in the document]
JPry: 🤦🏻‍♂️ It's start and end `time`.

This was wrong from our early work on this code (or we changed it at some point), and I clearly only fixed the first error. Good catch!

src/Tickets/Commerce/Traits/Type.php
JPry: They do different things, but that's a good point. There should only be 1.

src/views/v2/commerce/checkout/order-modifiers/coupons.php
JPry: You know, I spotted this at one point, but then I forgot about it. It's so far off the right side of the line that I missed it when scrolling through the file later 🤦🏻‍♂️ 

Thanks for the link, I'll update it with the SVG.

bordoni (changes requested):
I would like for us to stop doing cart calculation in float, that is a recipe for disaster.

When handling money, we always store cents never anything with the Dollar amount as `(float)`.

E.g.:
* `1 USD` = `(int) 100`
* `0.2 USD` = `(int) 20`
* `23034.43 USD` = `(int) 2303443`

This avoids problems on calculations as well as using `number_format()`.

Please review this code to ensure we are not creating calculations of price or percentages with floats.

src/views/v2/commerce/checkout/cart/item.php
JPry: > Add a filtered prop to the item.

@Camwyn Could be a little more specific about what you mean here? Or perhaps an example of where would be best to do that?

Camwyn: When we assemble `$items` for the template, add a `tec_tickets_checkout_should_skip_item` to each item and filter it there, rather than here in the template.

Filters like this don't belong in a template. The template should just check it (as a passed "prop") and bail if needed.

Camwyn (commented):
I don't see anything jumping out at me...
Did we test this with non-US currency _and_ a currency that has more/less than 2 decimals?

Camwyn (changes requested):
Don't recreate the wheel though

src/resources/postcss/tickets-commerce/_checkout.pcss
```diff
@@ -7,6 +7,10 @@
 .tribe-common,
 .event-tickets {
 
+	.tribe-hidden {
```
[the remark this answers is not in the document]
JPry: @Camwyn I added this because I didn't see it being applied on the checkout page. And I just confirmed locally, when I remove this group and re-build the styles, it shows

Screenshot 2025-02-21 at 12 33 40

Camwyn (approved):
Approved pending @bordoni 's sign-off

bordoni (approved):

PR #3573, opened 2025-02-19
Amend listener for RSVP date start/end

dpanta94:
Continuing work of PRs:

1) https://github.com/the-events-calendar/event-tickets/pull/3532
2) https://github.com/the-events-calendar/event-tickets/pull/3551

I noticed that the RSVP would not create start/end actions during its initial creation. That was because for our code, the WP_Post that would was created it was not yet a Ticket_Object.

So i am sending every request to sync RSVP dates to be performed during the request's end.

The conditionals being removed are already happening in `sync_ticket_dates_actions` - so the only thing that is affected by said change is that more IDs would be eligible for date sync during the request's end. But they will be correctly be filtered based on if they are Tickets or not.

## Why wasn't that caught by tests?

The reason is simple: order of meta keys is different for tests vs production.

see in tests: https://github.com/the-events-calendar/event-tickets/blob/4fbfb79fc4e487ef99517825a2fbddbdf57869e1/tests/_support/Commerce/RSVP/Ticket_Maker.php#L46

`_tribe_rsvp_for_event` key comes before start/end dates, while in production comes later.

As a result our `load_ticket_object` method would fail [here](https://github.com/the-events-calendar/event-tickets/blob/4fbfb79fc4e487ef99517825a2fbddbdf57869e1/src/Tribe/Tickets.php#L633) because an event would not be found for the ticket in production, but in tests would!

### DIFF
--- a/src/Tickets/Ticket_Actions.php
+++ b/src/Tickets/Ticket_Actions.php
@@ -109,7 +109,7 @@ class Ticket_Actions extends Controller_Contract {
 		add_action( 'update_post_meta', [ $this, 'pre_update_listener' ], 1000, 3 );
 		add_action( 'added_post_meta', [ $this, 'meta_keys_listener' ], 1000, 4 );
 		add_action( 'updated_postmeta', [ $this, 'meta_keys_listener' ], 1000, 4 );
-		add_action( 'tec_shutdown', [ $this, 'sync_rsvp_dates_for_all' ] );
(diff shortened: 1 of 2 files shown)


Camwyn (approved):
Trait needs a docblock.

Camwyn:
> Trait needs a docblock.

No, that's in tests, nvm

PR #3580, opened 2025-02-21
Resolve Stripe Payments failing and stock resolution.

bordoni:
### 🎫 Ticket

[ET-2306] [ET-2307] [ET-2300] [TECTRIA-796]

### 🗒️ Description

Please watch the video, it's a bit long but it explains the problem.

### 🎥 Artifacts 
https://i.bordoni.me/zd8vqRh7

[ET-2306]: https://stellarwp.atlassian.net/browse/ET-2306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-2307]: https://stellarwp.atlassian.net/browse/ET-2307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ET-2300]: https://stellarwp.atlassian.net/browse/ET-2300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TECTRIA-796]: https://stellarwp.atlassian.net/browse/TECTRIA-796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Commerce/Gateways/Stripe/Status.php
```diff
@@ -27,6 +27,8 @@ class Status {
 	/**
 	 * Order Status in Stripe for when the payment intent is first created or when payment is denied.
 	 *
+	 * Deprecated in favor of REQUIRES_PAYMENT_METHOD.
```
andrasguseo, line 30: Should we add a date/release when this was deprecated?

bordoni: Not really deprecated yet, its a Stripe thing. Once it gets moved I will actually move it.

src/Tickets/Commerce/Gateways/Stripe/REST/Order_Endpoint.php
```diff
@@ -210,10 +267,15 @@ public function handle_update_order( WP_REST_Request $request ) {
 		$messages         = $this->get_error_messages();
 		$gateway_order_id = $request->get_param( 'order_id' );
 
-		$order = tec_tc_orders()->by_args( [
-			'status' => 'any',
-			'gateway_order_id' => $gateway_order_id,
-		] )->first();
+		$order = tec_tc_orders()->by_args(
```
dpanta94, line 270: Lock system here as well please!

src/Tickets/Commerce/Gateways/Stripe/Status.php
```diff
@@ -163,4 +165,35 @@ public function convert_to_commerce_status( $stripe_status ) {
...
+
+		$last_payment_error = $payment_intent['last_payment_error'] ?? null;
+
+		$has_valid_error = ! empty( $last_payment_error['decline_code'] ) || ! empty( $last_payment_error['type'] );
+
+		if ( ! $last_payment_error || ! $has_valid_error ) {
+			return tribe( Commerce_Status\Status_Handler::class )->get_by_slug( $statuses[ $stripe_status ] );
+		}
+
+		return tribe( Commerce_Status\Denied::class );
```
dpanta94, line 197: Do any additional checks you need to do and then reuse the method `convert_to_commerce_status`. I cant see the reason we would have duplicate logic here. Adds a maintanability dept and easier to update one and not the other...

src/Tickets/Commerce/Order.php
```diff
@@ -358,22 +359,39 @@ public static function get_flag_action_marker_meta_key( $flag, Status\Status_Int
...
+		if ( $required_previous_status ) {
+			$order = tec_tc_get_order( $order_id );
+			foreach ( $required_previous_status as $required_status ) {
+				// If the required status is present we skip that status.
+				if ( isset( $order->status_log[ $required_status->get_slug() ] ) ) {
+					continue;
+				}
+
+				// Recursively call the modify_status method to ensure all required statuses are met.
+				$this->modify_status( $order_id, $required_status, $extra_args );
```
dpanta94, line 387: Wow this seems dangerous... I think this is trying to patch a whole to our system rather than fixing our system. I dont feel confident with this. Why do the order needs to pass through the required statuses ? What is our system doing wrong if it doesn't ?

src/Tickets/Commerce/Order.php
```diff
@@ -358,22 +359,39 @@ public static function get_flag_action_marker_meta_key( $flag, Status\Status_Int
...
+			}
+		}
+
+		// we specifically only lock the order transactions after the required statuses have been met.
```
dpanta94, line 391: The comment is not correct the order is being locked for each update to the required statuses...

Also we could end up in a loop calling this method recursively

src/Tickets/Commerce/Order.php
```diff
@@ -691,6 +709,7 @@ public function upsert( Gateway_Interface $gateway, array $args ) {
 		$locked = $this->lock_order( $existing_order_id );
 
 		if ( ! $locked ) {
+			$this->unlock_order( $existing_order_id );
```
dpanta94, line 712: Since its not locked why are you unlocking ?

src/Tickets/Commerce/Order.php
dpanta94, line 387: Coming back to this after having completed the CR --- I see that only Pending is required for going to a Completed order.

The first question is why ?

If it is about the flag actions that are fired for pending, why dont we fire them for the completed transition as well but idempotent and avoid the dangerous recurrency here ?

src/Tickets/Commerce/Gateways/Stripe/Payment_Intent_Handler.php
```diff
@@ -13,87 +13,120 @@
...
+		} else {
+			// If it all fails lets create a new one.
+			$payment_intent = Payment_Intent::create_from_cart( tribe( Cart::class ) );
+
+			if ( isset( $payment_intent['id'] ) && empty( $payment_intent['errors'] ) ) {
+				$this->store_payment_intent_cookie( $payment_intent['id'] );
+			}
+		}
+
+		$this->set( $payment_intent );
```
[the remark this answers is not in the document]
bordoni: REsolved.

src/Tickets/Commerce/Gateways/Stripe/Hooks.php
```diff
@@ -92,6 +99,10 @@ public function process_async_stripe_webhook( int $order_id ): void {
 			return;
 		}
 
+		if ( time() < $order->on_checkout_hold ) {
+			return;
+		}
+
```
dpanta94, line 105: This should not be here. Why wait forcefully for the timeout to expire if the checkout_completed has been fired already ?

src/views/v2/commerce/checkout.php
```diff
@@ -34,7 +36,7 @@
 <div class="tribe-common event-tickets">
 	<section
 		class="tribe-tickets__commerce-checkout"
-		<?php tribe_attributes( $attributes ); ?>
+		<?php tribe_attributes( $attributes ); ?>90
```
dpanta94, line 39:
```suggestion
		<?php tribe_attributes( $attributes ); ?>
```

src/Tickets/Commerce/Admin_Tables/Orders_Table.php
```diff
@@ -371,6 +371,7 @@ public function column_status( $item ) {
 					$dashicon = 'undo';
 					break;
 				case 'failed':
+				case 'denied':
```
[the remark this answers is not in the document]
bordoni: It was causing some errors.

src/Tickets/Commerce/Gateways/Stripe/Hooks.php
```diff
@@ -328,6 +339,122 @@ public function maybe_create_stripe_payment_intent() {
...
+				'gateway_payload'  => $payment_intent,
+				'gateway_order_id' => $payment_intent['id'],
+			]
+		);
+
+		// If we get a success status, we redirect to the success page.
+		if ( Completed::SLUG === $new_status->get_slug() ) {
+			wp_safe_redirect( $success_url );
+			exit;
+		}
```
rafsuntaskin, line 399: This can be problematic, for async payment methods Stripe will send `pending` status which will later be `completed` by webhooks. But for that case the user will be stuck on the checkout page and continue to create pending orders.

PR #3592, opened 2025-02-25
Fix/add test coverage

dpanta94:
### 🎫 Ticket

Fixes against the original [PR](https://github.com/the-events-calendar/event-tickets/pull/3580)

### DIFF
--- a/src/Tickets/Commerce/Gateways/Stripe/Hooks.php
+++ b/src/Tickets/Commerce/Gateways/Stripe/Hooks.php
@@ -79,14 +79,14 @@ class Hooks extends \TEC\Common\Contracts\Service_Provider {
 	 * Process the async stripe webhook.
 	 *
 	 * @since 5.18.1
-	 * @since TBD Added the $try parameter.
+	 * @since TBD Added the $retry parameter.
 	 *
 	 * @param int $order_id The order ID.
-	 * @param int $try      The number of times this has been tried.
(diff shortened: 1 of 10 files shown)


src/Tickets/Commerce/Gateways/Stripe/REST/Order_Endpoint.php
```diff
@@ -159,20 +157,6 @@ public function handle_create_order( WP_REST_Request $request ) {
...
-					'id' => $order->ID,
-				]
-			)
-			->set_args(
-				[
-					'gateway_payload'  => $payment_intent,
-					'gateway_order_id' => $payment_intent['id'],
-				]
-			)
-			->save();
```
bordoni, line 174: Without this we loose data.

src/Tickets/Commerce/Gateways/Stripe/Hooks.php
```diff
@@ -99,10 +99,6 @@ public function process_async_stripe_webhook( int $order_id ): void {
 			return;
 		}
 
-		if ( time() < $order->on_checkout_hold ) {
-			return;
-		}
```
bordoni, line 104: Without this check here it will still execute if the is a hold on the order I want to ensure it doesnt happen.

bordoni (changes requested):

PR #3612, opened 2025-02-27
Coupons – fix location of negative sign

JPry:
### 🎫 Ticket

QA Issue spreadsheet – Item `#003`

### 🗒️ Description

This PR adjusts the display of the coupon discount so that the negative sign is separate from the number. For currencies with the symbol before the number, this means that the `-` sign will be output before the currency symbol instead of after it.

As part of this PR, I took the opportunity to add some enhancements to the `Currency_Value` class and the `Legacy_Value_Factory` class.

#### Updates

After some flaws in the coupon calculations were pointed out, I added the following changes:

1. I updated the cart to calculate coupons (and any other dynamic items) within the `get_cart_total()` method. These were previously being done in the `get_cart_subtotal()` method. The flaw in logic was that the amount generated from the `get_cart_subtotal()` calculation was used to determine the value that the coupon displayed, so it was essentially being calculated twice.
2. I added a method to the `Models/Coupon` class to add/remove itself to/from the cart, given a `$cart` object to work with. This ensures consistency in generating a unique ID to ensure there isn't any conflict with any other item IDs.
3. I updated the `Precision_Value::create()` method to allow for passing custom symbols when it is used. Previously, the *only* option was to use the default values.
4. I added a method `to_currency_value()` to the `Legacy_Value_Factory` class.
5. I enhanced the unit tests and included additional tests for the Coupons REST API. I also added more calculation testing based on the coupons being added to the cart (Fun side note: I let AI generate some of the testing data, and I had to go back and correct some of the numbers it came up with. Turns out AI might not know how to handle floats and rounding as well as we do!).

### 🎥 Artifacts 

**Before:**

Screenshot 2025-02-27 at 16 19 25

**After (old – wrong calculations):**
Screenshot 2025-02-27 at 16 20 40

**After (NEW – correct calculations):**

Screenshot 2025-03-05 at 12 10 32

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier.php
+++ b/src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier.php
@@ -85,6 +85,11 @@ class Order_Modifier extends Model implements ModelCrud, ModelFromQueryBuilderOb
 			return;
 		}
 
+		// If we don't have the raw amount set, nothing else to do.
+		if ( ! array_key_exists( 'raw_amount', $this->attributes ) ) {
+			return;
+		}
+
 		if ( $this->attributes['raw_amount'] instanceof Percent_Value ) {
 			return;
 		}
--- a/src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier.php
+++ b/src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier.php
@@ -71,6 +71,28 @@ class Order_Modifier extends Model implements ModelCrud, ModelFromQueryBuilderOb
 	protected static string $order_modifier_type;
 
 	/**
+	 * Constructor.
+	 *
+	 * @since 1.0.0
+	 *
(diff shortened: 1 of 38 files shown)


bordoni (approved):

src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier.php
```diff
@@ -70,6 +70,33 @@ class Order_Modifier extends Model implements ModelCrud, ModelFromQueryBuilderOb
...
+	 * Constructor.
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param array<string,mixed> $attributes Attributes.
+	 */
+	public function __construct( array $attributes = [] ) {
+		parent::__construct( $attributes );
+
+		// If we have a percent sub_type, we need to ensure the raw_amount is a Percent_Value.
```
Camwyn, line 83: This comment feels like it belongs with line 97?

PR #3644, opened 2025-03-07
Add percent and currency symbols to coupon/fee input

JPry:
### 🎫 Ticket

QA Issue `#009`

### 🗒️ Description

This updates the input fields for Coupons and Fees to behave dynamically. The amount field will now show a `%` symbol and a currency symbol in the field based on the Type field.

Per [Stephen's comment](https://github.com/the-events-calendar/event-tickets/pull/3644#issuecomment-2706655820) on my original attempt, I've updated this to use the [imask](https://imask.js.org) library to power the display. This allows for a better visual experience, while also mitigating the currency issues mentioned in review comments.

While working on this, I noticed a small issue with the default values that were set for the `Currency_Value` class, which I fixed and added to this PR. 

### 🎥 Artifacts 

**Original video:**

https://github.com/user-attachments/assets/2e9046b4-8cbc-4c81-89ae-bae7412f8f2e

**Updated video:** https://docs.google.com/videos/d/1KgpbVSUgHnxXggnhfyhHMibmJQIWcb0wMQVTpiY30qo/edit?usp=sharing

### ✔️ Checklist


src/resources/js/admin/order-modifiers/amount-field.js
```diff
@@ -0,0 +1,120 @@
...
+		const $typeSelect = $( selectors.type );
+		return $typeSelect.val();
+	};
+
+	const addAmountExtras = () => {
+		const $amount = $( selectors.amount );
+		let value = Number.parseFloat( $amount.val() || 0 ).toFixed( i18n.precision );
+
+		// If the decimal separator is not a period, replace it.
+		if ( '.' !== i18n.decimalSeparator ) {
```
Camwyn, line 37: How does this handle currencies where the comma and period are swapped?
i.e. $ 1.234,56 pesos in Argentina

src/admin-views/order_modifiers/coupon-edit.php
```diff
@@ -99,9 +99,8 @@ class="tribe-field"
 						id="order_modifier_amount"
 						class="tribe-field tec_order_modifier_amount_field"
 						maxlength="9"
-						step="0.01"
+						step="0.1"
```
Camwyn, line 102: Are we OK with the fact that changing this to `.1` means I can't enter `$0.25`?

Camwyn: This change isn't in the description, nor the sheet, so I need to ask 😉

Camwyn (commented):
Some questions...

src/resources/js/admin/order-modifiers/amount-field.js
JPry: The comma would be stripped out in this field.

rafsuntaskin: @JPry what stephen is looking for is whether this field respects currency formatting or not?. Like in the example shown `.` is used for thousand separator and `,` is used for decimal separators.

Camwyn: Right - so would we get `$ 1.234.56`? That's not all right...
Might be OK if we got `$ 1234.56` and it was then reformatted, but then we're just adding more steps.

Camwyn: Honestly, after watching the video I am unsure this is "better" as the value format keeps switching on save/entry.

A better solution might be an input mask or an [input prefix/suffix](https://designsystem.digital.gov/components/input-prefix-suffix/)?

Camwyn: In the end, check with design. 😉

Camwyn:
Honestly, after watching the video I am unsure this is "better" as the value format keeps switching on save/entry.

A better solution might be an input mask or an [input prefix/suffix](https://designsystem.digital.gov/components/input-prefix-suffix/)?

In the end, check with design. 😉

package.json
```diff
@@ -96,6 +96,7 @@
     "dotenv": "^16.3.1",
     "gettext-parser": "1.3.0",
     "husky": "^1.3.1",
+    "imask": "^7.6.1",
```
Camwyn, line 99: Why? You don't import this anywhere. 
Honestly though, doing this and an import in the js seems less impactful than creating a new asset class to import from the `unpkg` URL. (and it version locks!)

JPry:
@Camwyn I removed `imask` from the `package.json` file.

Camwyn (approved):

PR #3652, opened 2025-03-10
Correctly handle coupon usage updates

JPry:
### 🎫 Ticket

QA spreadsheet – items `#027` and `#028`.

### 🗒️ Description

This updates the coupon logic to ensure each use of the coupon is properly counted.

Notable changes:

1. 3 new classes in the `Order_Modifiers/Flag_Actions/` directory:
	1. `Abstract_Coupon_Usage` – base class for other flag action classes 
	2. `Decrease_Coupon_Usage` – fires on the `increase_stock` hook, and handles decreasing the usage of a coupon
	3. `Increase_Coupon_Usage` – fires on the `decrease_stock` hook, and handles increasing the usage of a coupon
2. 3 new methods added to the `Coupons` trait. 1 to directly retrieve a meta object, and the others to add/remove uses
3. I updated the table class to make use of the `Coupons` trait.
4. While looking at the cart, I noticed that the count of items in the cart would include the coupon count. This isn't correct, so I tweaked the template file that shows the quantity
5. I updated `Order_Modifier_Meta` to override the `isPropertyTypeValid()` method. This allows for multiple valid types of data for a given property.
6. Makes a small tweak to the JS that handles applying the coupon to provide the actual message returned from the Ajax request.
7. As part of updating the tests, I noticed that the pagination wasn't present here, so I included the fixes for issue `#028` here also.
8. I noticed that the `amount-field.js` file was being included on all order modifier pages, not just the edit pages, so I changed how that is loaded.
9. I moved the `render_display_name_column()` method from each of the child table classes to the parent abstract class.

### 🎥 Artifacts 

https://github.com/user-attachments/assets/38381c1d-e681-40fa-a486-779a34f31666

### ✔️ Checklist


src/resources/js/v2/tickets-commerce.js
```diff
@@ -364,10 +364,9 @@ tribe.tickets.commerce = {};
 						$inputContainer.removeClass( hiddenName );
 					}
 				},
-				error() {
-					$errorMessage
-						.text( tecTicketsCommerce.i18n.couponApplyError )
-						.removeClass( hiddenName );
+				error( response ) {
+					const msg = response.responseJSON?.message || tecTicketsCommerce.i18n.couponApplyError;
```
dpanta94, line 368:
```suggestion
					const msg = response?.responseJSON?.message || tecTicketsCommerce.i18n.couponApplyError;
```

tests/order_modifiers_integration/Coupons/Create_Coupon_Modifiers_Test.php
```diff
@@ -31,6 +37,58 @@ public function does_table_render_correctly() {
...
+			}
+
+			// For the first 10 coupons, use them.
+			if ( $i < 10 ) {
+				$this->add_coupon_use( $modifier->id, $i + 1 );
+			}
+		}
+
+		// Clear the request data to avoid conflicts with other tests.
+		unset( $_REQUEST['id'], $_REQUEST['edit'], $_POST['modifier_id'], $_POST['edit'] );
```
dpanta94, line 78: This should be done in an `@after` or `@before` or both method.

tests/order_modifiers_integration/Coupons/Create_Coupon_Modifiers_Test.php
```diff
@@ -31,6 +37,58 @@ public function does_table_render_correctly() {
...
+
+		$modifier_admin_handler = tribe( Modifier_Admin_Handler::class );
+		$_REQUEST               = [
+			'modifier' => $this->modifier_type,
+		]; // phpcs:ignore WordPress.Security.NonceVerification
```
dpanta94, line 83: Maybe clean this as well ?

tests/order_modifiers_integration/Coupons/Create_Coupon_Modifiers_Test.php
```diff
@@ -31,6 +37,58 @@ public function does_table_render_correctly() {
...
+
+		ob_start();
+		$this->get_table_class_instance()->prepare_items();
+		$modifier_admin_handler->render_tec_order_modifiers_page();
+
+		$this->assertMatchesHtmlSnapshot( ob_get_clean() );
```
dpanta94, line 89: No IDs in the snapshot that could change and as a result we need to `str_replace` ?

dpanta94 (changes requested):

tests/order_modifiers_integration/Checkout/__snapshots__/Coupons_Test__it_should_calculate_coupons_simple_math__0.snapshot.html
```diff
@@ -323,7 +323,7 @@ <h4 class="tribe-common-h4 tribe-common-h--alt tribe-tickets__commerce-checkout-
 	</div>
 </div>
 <div class="tribe-tickets__commerce-checkout-cart-footer-quantity">
-	<span class="tribe-tickets__commerce-checkout-cart-footer-quantity-label">Quantity: </span><span class="tribe-tickets__commerce-checkout-cart-footer-quantity-number">21</span></div>
+	<span class="tribe-tickets__commerce-checkout-cart-footer-quantity-label">Quantity: </span><span class="tribe-tickets__commerce-checkout-cart-footer-quantity-number">20</span></div>
```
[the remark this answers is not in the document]
JPry: Yes, this was an expected change. That's item 4 in the description.

tests/order_modifiers_integration/Coupons/Create_Coupon_Modifiers_Test.php
JPry: @dpanta94 This was copied from your code 😅

tests/order_modifiers_integration/Checkout/Coupons_Test.php
```diff
@@ -316,4 +327,167 @@ public function it_should_calculate_coupons_complex_math() {
...
+				[
+					'order_modifier_id' => $coupon->id,
+					'meta_key'          => 'coupons_available',
+					'meta_value'        => $limit,
+				]
+			)
+		);
+
+		// Ensure the usage limit is set correctly.
+		Assert::assertEquals( $limit, $this->get_coupon_usage_limit( $coupon->id ) );
```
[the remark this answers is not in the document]
JPry: @dpanta94 I switched to using `Assert::` instead of `$this->` because the base classes aren't being picked up in the IDE, so autocomplete doesn't work. Using `Assert::`, all the methods will autocomplete.

It's not as relevant when using `->assertEquals()`, but for some of other assertions I don't use as often, it's much nicer to use autocomplete rather than looking up the available methods.

tests/order_modifiers_integration/Coupons/Create_Coupon_Modifiers_Test.php
JPry: No, we don't display the IDs in the table. We display the coupon slug, which has a number in it, but that number is an increment from when the coupon is generated, not the ID that goes into the DB.

Here's a sample row in the table (with whitespace added for readability):

```html
<tr>
	<td class='display_name column-display_name has-row-actions column-primary' data-colname="Coupon Name">Modifier
		Display Name
		<button type="button" class="toggle-row">
			<span class="screen-reader-text">Show more details</span>
		</button>
	</td>
	<td class='slug column-slug' data-colname="Code">test_coupon_10</td>
	<td class='raw_amount column-raw_amount' data-colname="Amount">$10.00</td>
	<td class='used column-used' data-colname="Used">0</td>
	<td class='remaining column-remaining' data-colname="Remaining">-</td>
	<td class='status column-status' data-colname="Status">Active</td>
</tr>
```

tests/order_modifiers_integration/Coupons/Create_Coupon_Modifiers_Test.php
JPry: Updated

src/views/v2/commerce/checkout/cart/footer/quantity.php
```diff
@@ -9,18 +9,28 @@
...
 
+use TEC\Tickets\Commerce\Module;
+
+$items = array_filter(
+	$items,
+	function ( $item ) {
+		return 'ticket' === ( $item['type'] ?? 'ticket' );
+	}
+);
+
```
Camwyn, line 33: The quantity is not escaped (below the modifications). Do your sum/pluck here into a variable and escape it below (absint?)

Also - array_sum will convert to count() [if any of the values is a string](https://onlinephp.io?s=s7EvyCjg5VIpS8wpTVWwVYjm5eI01AESRiBCPT8vVZ2XK9YayAai1OSMfIXEoqLEyvji0lwNiCZNBT0FpZg8JagaIMJhmDEhg6wB&v=8.2.20).

All template values are filterable, do not trust them!

src/Tickets/Commerce/Order_Modifiers/Modifier_Admin_Handler.php
```diff
@@ -199,6 +199,19 @@ public function is_on_page(): bool {
...
+	 * Defines whether the current page is the Event Tickets Order Modifiers edit page.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool True if on the Order Modifiers edit page, false otherwise.
+	 */
+	protected function is_on_edit_page(): bool {
+		$is_edit = tribe_is_truthy( tec_get_request_var( 'edit', '0' ) );
+
+		return $this->is_on_page() && $is_edit;
```
dpanta94, line 212: put `$is_edit` first for faster fail fast :D

tests/order_modifiers_integration/Coupons/Create_Coupon_Modifiers_Test.php
```diff
@@ -77,10 +76,17 @@ public function does_table_render_correctly_with_uses() {
...
+		$filter = static fn() => 20;
+		add_filter( 'coupon_per_page', $filter );
+
 		ob_start();
 		$this->get_table_class_instance()->prepare_items();
 		$modifier_admin_handler->render_tec_order_modifiers_page();
 
+		// Remove the filter
+		remove_filter( 'coupon_per_page', $filter );
+
```
dpanta94, line 89: Filters should be reset on their own before every test case. That doesn't hurt anything just an info :D

dpanta94 (approved):

PR #3665, opened 2025-03-13
Fix PayPal errors when using coupons

JPry:
### 🎫 Ticket

QA Issue `031`

### 🗒️ Description

This fixes the PayPal errors that are generated when using coupons. Notably:

1. Adds a custom controller to handle adding Coupons to the PayPal order data.
    1. PayPal doesn't accept negative values for items, so we have to handle this a different way. I'm using a solution that I [found here](https://www.paypal-community.com/t5/REST-APIs/Applying-Discounts-to-Payment-Orders/td-p/1761738).
    1. I added a filter and a bit of extra logic to the PayPal `Client` class. This facilitates the changes described above.
1. I updated the `tickets-commerce.js` script to not error when PayPal is in use. This was the cause of the console error where we tried to query a window object that didn't exist
1. I updated the Coupons API to only require the Stripe payment intent when stripe is being used
1. I also fixed a PHP 7.4 error in the template file where using array destructring triggered an error because the array had string keys instead of numeric keys.

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Commerce/Gateways/PayPal/Client.php
+++ b/src/Tickets/Commerce/Gateways/PayPal/Client.php
@@ -503,6 +503,14 @@ class Client {
 						'currency_code' => Arr::get( $unit, 'currency' ),
 					],
 				];
+
+				// If we have extra breakdown provided, merge into the item breakdown.
+				if ( array_key_exists( 'extra_breakdown', $unit ) ) {
+					$purchase_unit['amount']['breakdown'] = array_merge(
+						$purchase_unit['amount']['breakdown'],
+						$unit['extra_breakdown']
+					);
+				}
 			}
 
 			if ( ! empty( $unit['tax_id'] ) ) {
--- a/src/Tickets/Commerce/Gateways/PayPal/REST/Order_Endpoint.php
+++ b/src/Tickets/Commerce/Gateways/PayPal/REST/Order_Endpoint.php
@@ -128,6 +128,16 @@ class Order_Endpoint extends Abstract_REST_Endpoint {
 			$unit['items'][] = $this->get_unit_data( $item, $order );
 		}
 
+		/**
+		 * Filters the unit data for the order before sending it to PayPal.
+		 *
+		 * @since TBD
+		 *
+		 * @param array   $unit  The structured data for the order.
+		 * @param WP_Post $order The current order object.
+		 */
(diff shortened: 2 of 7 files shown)


src/Tickets/Commerce/Gateways/PayPal/Client.php
```diff
@@ -499,10 +499,18 @@ public function create_order( array $units = [] ) {
 				$purchase_unit['items']               = $items;
 				$purchase_unit['amount']['breakdown'] = [
 					'item_total' => [
-						'value'         => Arr::get( $unit, 'value' ),
+						'value'         => $unit['item_value'] ?? $unit['value'],
 						'currency_code' => Arr::get( $unit, 'currency' ),
 					],
```
redscar, line 504: Small suggestion, Since we had trouble understanding how the breakdown worked, I would link to the PayPal documentation section so that our future selves have any easy to find reference. https://developer.paypal.com/docs/api/payments/v2/#definition-amount_breakdown

tests/order_modifiers_integration/Checkout/Gateway/PayPal/Coupons_Test.php
```diff
@@ -0,0 +1,166 @@
...
+		Assert::assertEquals( 2, $purchase_unit['items'][0]['quantity'], 'The quantity should be 2' );
+		Assert::assertArrayHasKey( 'breakdown', $purchase_unit['amount'], 'The amount should have a breakdown of items.' );
+		Assert::assertArrayHasKey( 'item_total', $purchase_unit['amount']['breakdown'], 'The breakdown should have an item total' );
+		Assert::assertArrayHasKey( 'discount', $purchase_unit['amount']['breakdown'], 'The breakdown should have a discount' );
+
+		// The item total should be 11.28 * 2 = 22.56 (as a string).
+		Assert::assertEquals( '22.56', $purchase_unit['amount']['breakdown']['item_total']['value'] );
+
+		// The discount should be 17.3% * 2 * 11.28 = 3.90 (as a string).
+		Assert::assertEquals( '3.90', $purchase_unit['amount']['breakdown']['discount']['value'] );
```
redscar, line 110: I think adding one more Assert to make sure that the item_total value + discount value = the complete total would also be a good check.

redscar (approved):

PR #3681, opened 2025-03-20
Fix input display for pecent to use correct thousand/decimal settings

JPry:
### 🎫 Ticket

QA issue `041`

### 🗒️ Description

This fixes the display and handling of Percent values in the "Amount" input field for Coupons and Fees. Previously, if the site settings use `,` for the decimal separator instead of `.`, the amount will not render correctly. For example, a `10%` value entered would be converted to `1000%` after saving.

The reason for this behavior is because the `Percent_Value` class wasn't handling number formats correctly, and was instead *always* using a `.` for the decimal separator (and `,` for the thousands separator).

This PR does the following to fix this issue:

1. Adds a new `Digit_Separators` trait that can be used across value classes
1. Applies the `Digit_Separators` trait to both `Currency_Value` and `Percent_Value`
1. Updates the `get_as_string()` method in `Percent_Value` to use the corrected format.
1. Updates the controller that sets the default values for thousands and decimal separators.
1. Adds new unit tests that cover the changes

#### Before

https://github.com/user-attachments/assets/19f0064f-f094-43a6-84e6-fdc2f97086ff

#### After

https://github.com/user-attachments/assets/39306bac-a6f4-4815-96fd-21911cafb987

### ✔️ Checklist

### DIFF
--- a/changelog/fix-percent-input-display
+++ b/changelog/fix-percent-input-display
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix the input field for Fees (and Coupons) when the type is Percent and the thousands/decimal separators are set to "." and ","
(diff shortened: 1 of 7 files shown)


bordoni (approved):
I like the change.s.

PR #3700, opened 2025-03-29
Prevent fatals when activating ET with new TEC Classy editor

JPry:
### 🎫 Ticket

Related to [TEC-5416]

### ✔️ Checklist
- [ ] Ran `npm run changelog` to add changelog file(s). More info [here](https://docs.theeventscalendar.com/developer/git/changelogs/#process)
- [ ] Code is covered by **NEW** `wpunit` or `integration` tests.
- [ ] Code is covered by **EXISTING** `wpunit` or `integration` tests.
- [ ] Are all the **required** tests passing?
- [ ] Automated code review comments are addressed.
- [ ] Have you added Artifacts?
- [ ] Check the base branch for your PR.
- [ ] Add your PR to the project board for the release.

[TEC-5416]: https://stellarwp.atlassian.net/browse/TEC-5416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3701, opened 2025-04-01
Prevent fatals when activating/using ET with Classy editor

JPry:
### 🎫 Ticket

Related to [TEC-5416]

Replaces #3700.

### ✔️ Checklist

[TEC-5416]: https://stellarwp.atlassian.net/browse/TEC-5416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/commerce_integration/_bootstrap.php
+++ b/tests/commerce_integration/_bootstrap.php
@@ -1,6 +1,7 @@
 <?php
 
-use \TEC\Tickets\Commerce\Provider as Commerce_Provider;
+use TEC\Events\Classy\Controller as Classy_Controller;
+use TEC\Tickets\Commerce\Provider as Commerce_Provider;
 use Tribe\Tickets\Promoter\Triggers\Dispatcher;
 
 $tec_support = dirname( __DIR__, 3 ) . '/the-events-calendar/tests/_support';
@@ -9,7 +10,10 @@ Codeception\Util\Autoload::addNamespace( 'Tribe\Events\Test', $tec_support );
 // Let's make sure Commerce is enabled.
 putenv( 'TEC_TICKETS_COMMERCE=1' );
 putenv( 'TEC_DISABLE_LOGGING=1' );
-tribe_register_provider( Commerce_Provider::class );
+
+/** @var Tribe__Container $container */
+$container = tribe();
+$container->register( Commerce_Provider::class );
 
 // Let's make sure to set rewrite rules.
 global $wp_rewrite;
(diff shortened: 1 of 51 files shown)


JPry:
> So far so good, I've been nitpicking.

@lucatume I welcome any and all nitpicks.

tests/commerce_integration/_bootstrap.php
```diff
@@ -27,3 +31,8 @@
 
 // Disconnect Promoter to avoid license-related notices.
 remove_action( 'tribe_tickets_promoter_trigger', [ tribe( Dispatcher::class ), 'trigger' ] );
+
+// Ensure Classy is disabled for these tests.
```
lucatume, line 35: This seems like a left-over from debug.
But I would favor this approach over the skipping of tests if Classy is active.
My thinking it that any logic, be that front-end or back-end, affected by Classy work will be tested in tests dedicated to it, so it's safe for tests in these suites to run without Classy, and keep testing the current implementation, for now.

tests/ft_integration/TEC/Tickets/Flexible_Tickets/Series_Passes/BaseTest.php
```diff
@@ -67,6 +68,10 @@ public function should_disable_tickets_and_rsvps_for_series(): void {
 	 * @test
 	 */
 	public function should_not_replace_tickets_block_on_post(): void {
+		if ( tec_using_classy_editor() ) {
```
lucatume, line 71: As per my previous comment, I think Classy should just be disabled in these tests, removing the need to skip tests if it's enabled.

tests/ft_integration/Tribe/Tickets/__snapshots__/MetaboxTest__test_get_panels__post with RSVP__0.snapshot.html
```diff
@@ -6,19 +6,7 @@
 >
 	<div class="tribe_sectionheader ticket_list_container">
 		<div class="ticket_table_intro">
-			<button
```
lucatume, line 9: This looks like a side-effect of Classy being enabled. Disabling Classy for the suite, these should not change.

lucatume (changes requested):
The guard logic looks good, but I would not alter the existing suites, just disable Classy in them.

PR #3722, opened 2025-04-11
Include an initial version of the Square provider

bordoni:
For a video demo (which contains some sensitive data), go to the following PR:

https://github.com/the-events-calendar/whodat/pull/43

---

@dpanta94 my intention was to get the connection setup as fast as possible here so we can start playing around with POS and engaging with setting up tests for what is here currently.

I am going to fix all the PHPCS comments before you start on Monday, but I will add tests early next week. So feel free to merge this PR to get your work done.


src/Tickets/Commerce/Gateways/Square/Assets.php
```diff
@@ -0,0 +1,63 @@
...
+	/**
+	 * Binds and sets up implementations.
+	 *
+	 * @since TBD
+	 */
+	public function register() {
+		$plugin = Tribe__Tickets__Main::instance();
+
+		// Register the Square integration script
+		tribe_asset(
```
dpanta94, line 24: Use TEC\Common\Asset instead

src/Tickets/Commerce/Gateways/Square/Assets.php
```diff
@@ -0,0 +1,63 @@
...
+	/**
+	 * Determines if we should enqueue admin assets.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	public function should_enqueue_admin(): bool {
+		$admin_page = tribe_get_request_var( 'page', false );
+		$tab = tribe_get_request_var( 'tab', false );
```
dpanta94, line 58: Convert to early return approach please and use wp context like `is_admin`. Currently if i simply provide the params in the URL the script would be enqueued

src/Tickets/Commerce/Gateways/Square/Controller.php
```diff
@@ -0,0 +1,216 @@
...
+		$this->container->singleton( Gateway::class );
+		$this->container->singleton( Merchant::class );
+		$this->container->singleton( WhoDat::class );
+		$this->container->singleton( Order::class );
+		$this->container->singleton( Settings::class );
+		$this->container->singleton( REST::class );
+		$this->container->singleton( Assets::class );
+
+		$this->add_actions();
+		$this->add_filters();
```
dpanta94, line 46: Why are these here and not in smaller controllers ?

src/Tickets/Commerce/Gateways/Square/Controller.php
```diff
@@ -0,0 +1,216 @@
...
+
+	/**
+	 * Register the REST endpoints for Square.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function register_endpoints(): void {
+		$this->container->make( REST::class )->register_endpoints();
```
dpanta94, line 141: Hook that should be registered and live in the REST controller instead. Avoid code duplication which simply redirects to the real method....

dpanta94 (changes requested):

PR #3726, opened 2025-04-11
Alternate build mode: development

pattihis:
Till we have Tyson, this is needed to be able to inspect React components.

Usage:

`npm run before-zip`
or
`npm run before-zip:dev`


Camwyn:
Implementing Tyson, so we're removing this file completely.

PR #3774, opened 2025-04-29
Fix translations called too early

Camwyn:
All of these changes address `Notice: Function _load_textdomain_just_in_time was called incorrectly. Translation loading for the event-tickets domain was triggered too early.` issues I have encountered while trying to style an admin page.


src/Tickets/Commerce/Order_Modifiers/Controller.php
```diff
@@ -124,7 +124,10 @@ public function do_register(): void {
 		$this->container->singleton( Coupon::class );
 
 		$this->register_flag_actions();
-		$this->run_deprecated_coupon_filter();
+
+		add_action( 'init', function() {
```
Camwyn, line 128: Note, this is a deprecated private function below.

PR #3794, opened 2025-05-08
Include a second layer of security around received WhoDat Webhooks

bordoni:
Look at the WhoDat PR for more context.

### DIFF
--- a/src/Tickets/Commerce/Gateways/Square/Merchant.php
+++ b/src/Tickets/Commerce/Gateways/Square/Merchant.php
@@ -191,11 +191,7 @@ class Merchant extends Abstract_Merchant {
 	public function get_whodat_signature(): ?string {
 		$data = get_option( $this->get_signup_data_key() );
 
-		if ( empty( $data['whodat_signature'] ) ) {
-			return null;
-		}
-
-		return $data['whodat_signature'];
+		return $data['whodat_signature'] ?? null;
 	}
 
 	/**
--- a/src/Tickets/Commerce/Gateways/Square/Merchant.php
+++ b/src/Tickets/Commerce/Gateways/Square/Merchant.php
@@ -182,6 +182,23 @@ class Merchant extends Abstract_Merchant {
 	}
 
 	/**
+	 * Returns the Whodat signature stored for server-side transactions.
+	 *
+	 * @since TBD
+	 *
+	 * @return ?string
+	 */
+	public function get_whodat_signature(): ?string {
+		$data = get_option( $this->get_signup_data_key() );
+
+		if ( empty( $data['whodat_signature'] ) ) {
+			return null;
+		}
+
+		return $data['whodat_signature'];
+	}
+
+	/**
 	 * Get the account ID (same as merchant ID for Square).
 	 *
 	 * @since TBD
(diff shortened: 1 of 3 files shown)


src/Tickets/Commerce/Gateways/Square/REST/Webhook_Endpoint.php
```diff
@@ -109,20 +109,23 @@ public function get_path(): string {
...
 
 		// Get the webhook secret key from the URL.
 		$secret_key = $request->get_param( Webhooks::PARAM_WEBHOOK_KEY );
 
+		// Get the whodat signature from the request header.
+		$whodat_signature = $request->get_header( 'X-WhoDat-Signature' );
 
-
-		if ( ! tribe( Webhooks::class )->verify_signature( $secret_key ) ) {
+		if ( ! $webhook->verify_signature( $secret_key ) || ! $webhook->verify_whodat_signature( $whodat_signature ) ) {
```
dpanta94, line 120:
```suggestion
		if ( ! ( $webhook->verify_signature( $secret_key ) && $webhook->verify_whodat_signature( $whodat_signature ) ) ) {
```

src/Tickets/Commerce/Gateways/Square/Merchant.php
```diff
@@ -181,6 +181,23 @@ public function get_merchant_id(): string {
...
+	 * @return ?string
+	 */
+	public function get_whodat_signature(): ?string {
+		$data = get_option( $this->get_signup_data_key() );
+
+		if ( empty( $data['whodat_signature'] ) ) {
+			return null;
+		}
+
+		return $data['whodat_signature'];
```
dpanta94, line 198:
```suggestion
		return $data['whodat_signature'] ?? null;
```

src/Tickets/Commerce/Gateways/Square/Webhooks.php
```diff
@@ -265,6 +265,26 @@ public function verify_signature( $received_secret_key ) {
...
+	 * @return bool Whether the signature is valid.
+	 */
+	public function verify_whodat_signature( $received_secret_key ) {
+		$signature = tribe( Merchant::class )->get_whodat_signature();
+
+		if ( empty( $signature ) ) {
+			return false;
+		}
+
+		return $signature === $received_secret_key;
```
dpanta94, line 285:
```suggestion
		return $signature && $signature === $received_secret_key;
```

dpanta94 (changes requested):
minor comments and phpcs

src/Tickets/Commerce/Gateways/Square/Webhooks.php
dpanta94, line 285: I'd like to see a runtime evaluation of the signature. The alternative approach is a good idea, which i can support.

Simply storing once the signature though and just evaluating that they are equal after that is not what i would choose.

A runtime evaluation which would also involve the merchant id would be great. (So that if the merchant id changes in the local wp site but an event arrives for the old merchant id we had stored in whoDat, the eval should fail)

dpanta94 (changes requested):

src/Tickets/Commerce/Gateways/Square/Webhooks.php
```diff
@@ -265,6 +264,30 @@ public function verify_signature( $received_secret_key ) {
...
+	 * @param string $payload       The payload from the request.
+	 * @param string $received_hash The hash from the request.
+	 *
+	 * @return bool Whether the signature is valid.
+	 */
+	public function verify_whodat_signature( string $payload, string $received_hash ) {
+		$signature        = tribe( Merchant::class )->get_whodat_signature();
+		$notification_url = $this->get_webhook_endpoint_url();
+
+		if ( empty( $signature ) || empty( $notification_url ) || empty( $payload ) ) {
```
dpanta94, line 281: No need for empty check - we can just do 
```suggestion
		if ( ! ( $signature && $notification_url && $payload ) ) {
```

PR #3804, opened 2025-05-10
Common admin page styles

pattihis:
### 🎫 Ticket

[ET-2339]

### 🗒️ Description

This PR resolves issues [reported during QA](https://docs.google.com/spreadsheets/d/1NzXAjMpzE7Cum9rOix2ctIjEsdpR_YgFLLMKvnGtzkk/edit?gid=619168414#gid=619168414) of our Onboarding Wizard. Issues resolved:
- 28 Preserve settings between steps when updating 
- 42 Centralize common styles
- 43 Add IAN to onboarding landing page

Goes along with:
https://github.com/the-events-calendar/the-events-calendar/pull/5134
https://github.com/the-events-calendar/tribe-common/pull/2560

[ET-2339]: https://stellarwp.atlassian.net/browse/ET-2339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Tickets/Admin/Onboarding/API.php
```diff
@@ -215,8 +217,8 @@ public function update_wizard_settings( $params ): bool {
 			$params['_wpnonce']
 		);
 
-		// Add a snapshot of the data from the last request.
-		$settings['last_send'] = $params;
+		// Merge the new params with existing last_send data instead of overwriting.
```
Camwyn, line 220: Why is this necessary? All the data is sent every time - in its current state. How will this interact with my backing up a step and changing things?

Camwyn (changes requested):
Unsure about that merge...

src/Tickets/Admin/Onboarding/API.php
pattihis: Exactly, not every tab was sending all needed data. I saw communication tab overwriting the currency with default `USD` because it didn't include it in its data. Even though previous step had properly saved currency in database as `EUR`.

By merging the params from the previous step to the next, we ensure that all of them are carried forward. This is the fix for #28 and it allows a tab to submit just one relevant param. Any skipped ones will be taken from previous tabs.

Camwyn: That sounds like the store is not recording things correctly.

I believe the whole tab system is getting a once-over in the next iteration though.

Camwyn (approved):

PR #3808, opened 2025-05-12
ET-2310 Add bail checks to avoid fatal

sdokus:
### 🎫 Ticket

[ET-2310]

### 🗒️ Description

If a Tickets Commerce attendee exists on an event, but then Tickets Commerce is disabled and the event is switched to WooCommerce, a fatal is thrown because `TEC\Tickets\Commerce\tec_tc_attendees()` appears as an undefined function. This is a simple approach to bail in the two places where the fatal is being triggered if Tickets Commerce is disabled. 

### 🎥 Artifacts 
Before:
CleanShot 2025-05-12 at 13 08 50

After: 
CleanShot 2025-05-12 at 13 05 21

### ✔️ Checklist
- [X] Ran `npm run changelog` to add changelog file(s). More info [here](https://docs.theeventscalendar.com/developer/git/changelogs/#process)

[ET-2310]: https://stellarwp.atlassian.net/browse/ET-2310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rafsuntaskin (dismissed):
If the TicketsCommerce provider is not enabled then ideally it should not call the methods of that provider.

So, the `get_attendee` or `get_attendees_by_id` should not be called if TC is not active. Let's try to achieve that.

src/Tribe/Tickets.php
```diff
@@ -1658,6 +1658,11 @@ public static function get_attendees_from_modules( $attendees, $post_id = 0 ) {
 					continue;
 				}
 
+				// Skip if the provider is Tickets Commerce but the module is disabled.
+				if ( $provider instanceof TEC\Tickets\Commerce\Module && ! tec_tickets_commerce_is_enabled() ) {
+					continue;
+				}
```
rafsuntaskin, line 1664: Look into these [lines](https://github.com/the-events-calendar/event-tickets/blob/93792c7774e9c55ad82206a79d7a4593293a1f14/src/Tribe/Tickets.php#L1655-L1660) above. Here it should return false for `tribe_tickets_get_ticket_provider` for disabled TicketsComemrce provider.

Camwyn:
tests are failing

src/template-tags/tickets.php
```diff
@@ -822,8 +824,14 @@ function tribe_tickets_get_event_ids( $id ) {
 	function tribe_tickets_get_ticket_provider( $id ) {
 		/** @var Tribe__Tickets__Data_API $data_api */
 		$data_api = tribe( 'tickets.data_api' );
+		$provider = $data_api->get_ticket_provider( $id );
+		
+		// Skip if the provider is Tickets Commerce but the module is disabled.
+		if ( $provider instanceof TEC\Tickets\Commerce\Module && ! tec_tickets_commerce_is_enabled() ) {
```
JPry, line 830: Minor: You should add a `use` statement to the top of the file for `TEC\Tickets\Commerce\Module`, and then you can use the shortened name here instead of the fully-qualified name.

Camwyn (approved):

PR #3852, opened 2025-05-22
Feat/automatic refund order

dpanta94:
### 🗒️ Description

Enables automatic refund of an incoming order which would exceed our available stock.


bordoni (commented):
I like it overall, but I would have loved to hook later into the Action Flag that takes stock out and if at that point we have zero just refund the order.

bordoni (approved):

PR #3862, opened 2025-05-28
Feat/app fee

dpanta94:
Application fee implementation


src/Tickets/Commerce/Gateways/Square/Application_Fee.php
```diff
@@ -0,0 +1,71 @@
...
+			return Value::create();
+		}
+
+		// Check if merchant is in the US - Square application fees only apply to US sales.
+		$merchant         = tribe( Merchant::class );
+		$merchant_country = $merchant->get_merchant_country();
+
+		if ( 'US' !== $merchant_country ) {
+			return Value::create();
+		}
```
dpanta94, line 55: Should we change the order here ? So that non-US merchant dont even have their license validated through a remote API call ?

PR #3868, opened 2025-05-30
Feat/test coverage square

dpanta94:
Adding test coverage for Square integration!


src/Tickets/Commerce/Gateways/Square/Ajax.php
```diff
@@ -90,25 +90,20 @@ public function unregister(): void {
...
 
-			if ( ! empty( $connect_url ) ) {
-				Commerce_Settings::delete( 'tickets_commerce_gateways_square_remotely_disconnected_%s' );
-				wp_send_json_success( [ 'url' => $connect_url ] );
-				return;
-			}
-		} catch ( \Exception $e ) {
-			wp_send_json_error( [ 'message' => $e->getMessage() ] );
+		if ( ! $connect_url ) {
+			wp_send_json_error( [ 'message' => __( 'Failed to generate connection URL.', 'event-tickets' ) ], 400 );
```
lucatume: Is this a 400? As in  "is this a user error?"; I'm asking as I'm not sure. It might be, but it _looks_ like a server error.

src/Tickets/Commerce/Gateways/Square/Webhooks.php
```diff
@@ -364,44 +331,47 @@ public function ajax_register_webhook(): void {
...
 			);
+			return;
 		}
 
 		// Register new webhook.
 		$response = $this->register_webhook_endpoint();
 
 		if ( is_wp_error( $response ) ) {
-			wp_send_json_error( $response );
+			wp_send_json_error( $response, 400 );
```
lucatume: Here as well, I'm not sure this is a user error, is it? Should this be a 500?

src/Tickets/Commerce/Gateways/Square/Webhooks.php
```diff
@@ -364,44 +331,47 @@ public function ajax_register_webhook(): void {
...
-		if (
-			! current_user_can( 'manage_options' )
-		) {
+		if ( ! current_user_can( 'manage_options' ) ) {
 			wp_send_json_error(
 				[
 					'message' => __( 'You do not have permission to perform this action.', 'event-tickets' ),
-				]
+				],
+				403
```
lucatume: Should this be 401 (Unauthorized) over 403 (Forbidden)?
The operation is not forbidden, it's the user that cannot peform the action.

src/Tickets/Commerce/Gateways/Square/Webhooks.php
lucatume: There are other places in the code where this is happening, I'm not pointing them all out, but please give a pass  over the HTTP return codes to make sure they reflect what happened and where the error is coming from.
When I say "user" I mean "API user" (likely the plugin itself), not the end user. The error codes might already be correct.

src/Tickets/Commerce/Models/Order_Model.php
```diff
@@ -58,7 +58,7 @@ protected function build_properties( $filter ) {
 			$events_in_order  = (array) Arr::get( $post_meta, [ Order::$events_in_order_meta_key ] );
 			$tickets_in_order = (array) Arr::get( $post_meta, [ Order::$tickets_in_order_meta_key ] );
 
-			$on_checkout_hold_timeout = (int) Arr::get( $post_meta, [ Order::ON_CHECKOUT_SCREEN_HOLD_META, 0 ], 0 );
+			$on_checkout_hold_timeout = strtotime( Arr::get( $post_meta, [ Order::ON_CHECKOUT_SCREEN_HOLD_META, 0 ], 0 ) );
```
lucatume: This needs some defensive coding, maybe using a default value to "now". The function `strtotime` will raise a warning if the input is not a valid date, and you cannot trust the metadata to be correct when you read it. Or to be there at all (e.g. if some cleanup plugin flagged the meta as not required anymore or somehow it was removed from the database).

lucatume: The return value of `strtotime(0)` if `false`; is this expected?

tests/_support/Traits/With_WhoDat_Mocks.php
```diff
@@ -0,0 +1,52 @@
...
+	use With_Uopz;
+
+	/**
+	 * Mock the methods of the WhoDat class.
+	 *
+	 * @before
+	 *
+	 * @return void
+	 */
+	public function mock_methods() {
```
lucatume: Very generic method names for a trait. While a developer can alias the names of imported methods, it will add difficulty to the readability of the code. What about calling this "mock_whodat_methods" instead?  Or something similar. Think about this trait potentially being used in other tests and how the IDE suggestion when one types "whodat" helps.

tests/_support/Traits/With_WhoDat_Mocks.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Tribe\Tickets\Test\Traits;
+
+use Tribe\Tests\Traits\With_Uopz;
+use TEC\Tickets\Commerce\Gateways\Square\WhoDat;
+use TEC\Tickets\Commerce\Gateways\Contracts\Abstract_WhoDat;
+use RuntimeException;
+
+trait With_WhoDat_Mocks {
```
lucatume: Nitpick, the "With" prefix comes from Laravel's convention (and I've used it myself), but I think "use WhoDat_Mocks" reads better without the prefix.
The trait "with" prefix comes form Laravel's eloquent test case builders where the "with" prefix allows for almost plain-English naming of the resulting test case.

tests/square_integration/Assets_Test.php
```diff
@@ -0,0 +1,96 @@
...
+
+		$_REQUEST[ Payments_Tab::$key_current_section_get_var ] = Gateway::get_key();
+
+		$this->assertTrue( $controller->is_square_section() );
+	}
+
+	/**
+	 * @test
+	 */
+	public function it_should_correctly_determine_if_fe_assets_should_be_enqueued(): void {
```
lucatume: Today you know "fe" is front-end, me and you will forget and will have to debug this test. Help our future selves and  avoid short names or acronyms  in test names.

PR #3907, opened 2025-06-12
Update to prevent Potential Fatal on earlier MySQL verisons

bordoni:
- Bump schema version from 1.1.0 to 1.2.0
- Add migration method for MySQL compatibility
- Ensure `data` column has a valid JSON structure for older MySQL versions
- Prevent issues with NULL or empty TEXT column values
- Improve database schema robustness and compatibility

### DIFF
--- a/src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
+++ b/src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
@@ -22,7 +22,7 @@ class Ticket_Groups extends Table {
 	/**
 	 * {@inheritdoc}
 	 */
-	public const SCHEMA_VERSION = '1.1.0';
+	public const SCHEMA_VERSION = '1.2.0';
 
 	/**
 	 * {@inheritdoc}
@@ -71,6 +71,7 @@ class Ticket_Groups extends Table {
 	 * {@inheritdoc}
 	 *
 	 * @since 5.24.1 Handle hydrating new columns from `data` JSON for Ticket Presets use, if needed.
+	 * @since TBD    Handle MySQL compatibility fix for TEXT column DEFAULT value removal.
 	 */
 	protected function after_update( array $results = [] ) {
 		$results = parent::after_update( $results );
@@ -78,6 +79,10 @@ class Ticket_Groups extends Table {
 		// Run version-specific migrations.
 		if ( self::SCHEMA_VERSION === '1.1.0' ) {
 			$results = $this->migrate_to_1_1_0( $results );
+		}
+
+		if ( self::SCHEMA_VERSION === '1.2.0' ) {
+			$results = $this->migrate_to_1_2_0( $results );
 		}
 
 		return $results;
@@ -159,6 +164,58 @@ class Ticket_Groups extends Table {
 			$results[ $table_name . '.migration' ] = sprintf(
 				'Migrated %d rows in the %s table.',
 				$migrated,
+				$table_name
+			);
+		}
+
+		return $results;
+	}
+
+	/**
+	 * Handles MySQL compatibility migration for schema version 1.2.0.
+	 *
+	 * Ensures all `data` column values are properly set since we removed
+	 * the DEFAULT ('') clause for compatibility with older MySQL versions.
+	 *
+	 * @since TBD
+	 *
+	 * @param array $results The results array to update.
+	 *
+	 * @return array The updated results array.
+	 */
+	protected function migrate_to_1_2_0( array $results = [] ) {
+		global $wpdb;
+		$table_name = self::table_name();
+
+		// Check if any rows have NULL or problematic data values.
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.NoCaching,WordPress.DB.DirectDatabaseQuery.DirectQuery,WordPress.DB.DirectDatabaseQuery.DirectQuerySchemaChange
+		$rows_with_null_data = $wpdb->get_var(
+			$wpdb->prepare(
+				"SELECT COUNT(*) FROM %i WHERE data IS NULL OR data = ''",
+				$table_name
+			)
+		);
+
+		if ( ! $rows_with_null_data ) {
+			$results[ $table_name . '.compatibility_migration' ] = "No rows needed MySQL compatibility migration for {$table_name} table.";
+			return $results;
+		}
+
+		// Update any NULL or empty data values with a proper JSON structure.
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.NoCaching,WordPress.DB.DirectDatabaseQuery.DirectQuery,WordPress.DB.DirectDatabaseQuery.DirectQuerySchemaChange
+		$updated = $wpdb->query(
+			$wpdb->prepare(
+				"UPDATE %i SET data = '{}' WHERE data IS NULL OR data = ''",
+				$table_name
+			)
+		);
+
+		if ( false === $updated ) {
+			$results[ $table_name . '.compatibility_migration' ] = "Failed to update rows for MySQL compatibility in the {$table_name} table.";
+		} else {
+			$results[ $table_name . '.compatibility_migration' ] = sprintf(
+				'Updated %d rows for MySQL compatibility in the %s table (removed dependency on TEXT DEFAULT value).',
+				$updated,
 				$table_name
 			);
 		}
--- a/src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
+++ b/src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
@@ -58,7 +58,7 @@ class Ticket_Groups extends Table {
 			CREATE TABLE `$table_name` (
 				`id` bigint(20) UNSIGNED NOT NULL AUTO_INCREMENT,
 				`slug` varchar(255) DEFAULT '' NOT NULL,
-				`data` text DEFAULT ('') NOT NULL,
+				`data` text NOT NULL,
 				`capacity` int(11) DEFAULT 0 NOT NULL,
 				`cost` decimal(10,2) DEFAULT 0 NOT NULL,
 				`name` varchar(255) DEFAULT '' NOT NULL,


Camwyn (approved):

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -80,6 +81,10 @@ protected function after_update( array $results = [] ) {
 			$results = $this->migrate_to_1_1_0( $results );
 		}
 
+		if ( self::SCHEMA_VERSION === '1.2.0' ) {
```
Camwyn: Would this be easier to add to in the future via a switch statement? Then we could just add a case 🤔

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -71,14 +71,19 @@ protected function get_definition() {
...
+	 * @since TBD    Handle MySQL compatibility fix for TEXT column DEFAULT value removal.
 	 */
 	protected function after_update( array $results = [] ) {
 		$results = parent::after_update( $results );
 
 		// Run version-specific migrations.
-		if ( self::SCHEMA_VERSION === '1.1.0' ) {
-			$results = $this->migrate_to_1_1_0( $results );
-		}
+		switch ( self::SCHEMA_VERSION ) {
```
JPry: Just curious: what if someone updates from `1.0.0` directly to a later version, say `1.3.0`? Do these still need to run?

When updating from `1.0.0` to `1.2.0`, does the `1.1.0` logic still need to run?

tests/integration/Settings/__snapshots__/Payments_Tab_Test__should_match_payment_tab_snapshots__stripe enabled__0.snapshot.html
```diff
@@ -18,7 +18,7 @@ <h2 class="tec-tickets__admin-settings-tickets-commerce-gateway-title">
...
 		<div class="tec-tickets__admin-settings-tickets-commerce-gateway-signup-settings">
 	<div class="tec-tickets__admin-settings-tickets-commerce-gateway-connect-button">
-		<a data-gateway-onboard-complete="tecTicketsCommerceGatewayStripeSignupCallback" href="https://whodat.theeventscalendar.com/commerce/v1/stripe/connect?token=https://wordpress.test/index.php?rest_route=/tribe/tickets/v1/commerce/stripe/return&#038;return_url=https://whodat.theeventscalendar.com/commerce/v1/stripe/connected&#038;version={version}&#038;mode=live&#038;" data-gateway-button="true" id="connect_to_stripe" class="tec-tickets__admin-settings-tickets-commerce-gateway-connect-button-link">
+		<a data-gateway-onboard-complete="tecTicketsCommerceGatewayStripeSignupCallback" href="https://whodat.theeventscalendar.com/commerce/v1/stripe/connect?token=https://wordpress.test/index.php?rest_route=/tribe/tickets/v1/commerce/stripe/return&#038;return_url=https://whodat.theeventscalendar.com/commerce/v1/stripe/connected&#038;version={version}.1&#038;mode=live&#038;" data-gateway-button="true" id="connect_to_stripe" class="tec-tickets__admin-settings-tickets-commerce-gateway-connect-button-link">
```
JPry, line 21: Shouldn't the added `.1` be part of the `{version}` replacement? Or is it actually something different?

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
Camwyn: Hm, good point - not in this case, but potentially. 🤔

tests/integration/Settings/__snapshots__/Payments_Tab_Test__should_match_payment_tab_snapshots__stripe enabled__0.snapshot.html
Camwyn: Yeah but I'm not messing with that right now 😉

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
Camwyn: Fixed

JPry (approved):

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -44,6 +44,19 @@ class Ticket_Groups extends Table {
...
 	protected static $uid_column = 'id';
 
+	/**
+	 * Internal way to track prior versions.
+	 *
+	 * @since TBD
+	 *
+	 * @var array<string>
+	 */
+	protected static $versions = [
```
Camwyn, line 56: This is not used now but it could be useful if we need another migration later - since the "previous previous version" gets wiped on schema update

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -71,15 +86,40 @@ protected function get_definition() {
...
 	 *
 	 * @since 5.24.1 Handle hydrating new columns from `data` JSON for Ticket Presets use, if needed.
+	 * @since TBD    Handle MySQL compatibility fix for TEXT column DEFAULT value removal.
 	 */
 	protected function after_update( array $results = [] ) {
-		$results = parent::after_update( $results );
+		$results          = parent::after_update( $results );
+		$previous_version = $this->get_stored_previous_version();
+
+		// Run version-specific migrations. Add new migrations at the end of the switch statement so they run in the correct order.
```
JPry: Minor: we can drop the reference to "switch statement"

```suggestion
		// Run version-specific migrations.
```

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -88,81 +128,221 @@ protected function after_update( array $results = [] ) {
...
+				} else {
+					++$failed;
+					// Break on first failure.
+					break;
+				}
 			}
-		}
+
+			$remaining -= count( $rows );
+		} // end while
```
JPry: This might be a pet peeve, but with modern editors I don't think these kinds of comments add any value.

```suggestion
		}
```

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -88,81 +128,221 @@ protected function after_update( array $results = [] ) {
...
+		}
+
+		$results[ $table_name . '.migration' ] = sprintf(
+			// Translators: %d: number of rows migrated, %s: table name.
+			__( 'Migrated %d rows in the %s table.', 'event-tickets' ),
+			$migrated,
+			$table_name
+		);
+
+		return $wpdb->query( 'COMMIT' ) === false ? false : true;
```
JPry: I find the ternary that swaps `true` and `false` to be confusing. I think it is easier to understand a reverse in the condition being checked, and then the ternary isn't needed.

```suggestion
		return $wpdb->query( 'COMMIT' ) !== false;
```

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -88,81 +128,219 @@ protected function after_update( array $results = [] ) {
...
-					'capacity' => $capacity,
-					'cost'     => $cost,
-				],
-				[ 'id' => $row->id ],
-				[ '%s', '%d', '%s' ],
-				[ '%d' ]
-			);
+			if ( empty( $rows ) ) {
+				$results[ $table_name . '.migration' ] = "No rows needed migration for {$table_name} table.";
+				return $wpdb->query( 'COMMIT' ) === false ? false : true;
```
JPry:
```suggestion
				return $wpdb->query( 'COMMIT' ) !== false;
```

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -88,81 +128,219 @@ protected function after_update( array $results = [] ) {
...
+
+				if ( empty( $data ) ) {
+					++$failed;
+					break;
+				}
+
+				// Extract values from data JSON.
+				$name     = isset( $data['name'] ) ? sanitize_text_field( $data['name'] ) : '';
+				$capacity = isset( $data['capacity'] ) ? absint( $data['capacity'] ) : 0;
+				$cost     = isset( $data['cost'] ) ? (string) $data['cost'] : '0.000000';
```
JPry: Do we need sanitization for `$data['cost']`?

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -88,81 +128,219 @@ protected function after_update( array $results = [] ) {
...
 		if ( $failed > 0 ) {
 			$results[ $table_name . '.migration' ] = sprintf(
-				'Migrated %d rows and failed to migrate %d rows in the %s table.',
-				$migrated,
-				$failed,
+				__( 'Migration failed, refresh the page to re-run.', 'event-tickets' ),
+			);
+
+			// Rollback data transaction.
+			$wpdb->query( 'ROLLBACK' );
```
JPry, line 245: I don't know how likely it is for some rows to succeed, and some to fail, but if that were to happen, would we still want to rollback everything? Rolling back everything is safest, but I could see an edge case where one bad row prevents all the other good rows from being migrated.

Maybe it's not likely enough to be a concern, but I thought it was worth asking.

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -88,81 +128,219 @@ protected function after_update( array $results = [] ) {
...
+		$rows_with_null_data = $wpdb->get_var(
+			$wpdb->prepare(
+				"SELECT COUNT(*) FROM %i WHERE data IS NULL OR data = '' LIMIT 1000",
 				$table_name
+			)
+		);
+
+		if ( ! $rows_with_null_data ) {
+			$results[ $table_name . '.compatibility_migration' ] = "No rows needed MySQL compatibility migration for {$table_name} table.";
+			return $wpdb->query( 'COMMIT' ) === false ? false : true;
```
JPry:
```suggestion
			return $wpdb->query( 'COMMIT' ) !== false;
```

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -88,81 +128,219 @@ protected function after_update( array $results = [] ) {
...
+					++$migrated;
+				} else {
+					++$failed;
+					// Break on first failure.
+					break;
+				}
+			}
+
+			$rows_with_null_data -= count( $rows );
+		} // end while
```
JPry:
```suggestion
		}
```

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -88,81 +128,219 @@ protected function after_update( array $results = [] ) {
...
 
-		return $results;
+		$results[ $table_name . '.migration' ] = sprintf(
+			// Translators: %1$d: number of rows migrated, %2$s: table name.
+			__( 'Migrated %1$d rows in the %2$s table.', 'event-tickets' ),
+			$migrated,
+			$table_name
+		);
+
+		return $wpdb->query( 'COMMIT' ) === false ? false : true;
```
JPry:
```suggestion
		return $wpdb->query( 'COMMIT' ) !== false;
```

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
Camwyn: No, I was using them while I was coding, but they can go now.

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
Camwyn: Yes, all or nothing.

JPry: That's fair, I just wanted to be certain 👍🏻

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
Camwyn: Yes.

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -88,81 +128,219 @@ protected function after_update( array $results = [] ) {
...
+
+				if ( empty( $data ) ) {
+					++$failed;
+					break;
+				}
+
+				// Extract values from data JSON.
+				$name     = isset( $data['name'] ) ? sanitize_text_field( $data['name'] ) : '';
+				$capacity = isset( $data['capacity'] ) ? absint( $data['capacity'] ) : 0;
+				$cost     = isset( $data['cost'] ) ? sanitize_text_field( (string) $data['cost'] ): '0.000000';
```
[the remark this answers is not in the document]
Camwyn:
```suggestion
				$cost     = isset( $data['cost'] ) ? sanitize_text_field( (string) $data['cost'] ) : '0.000000';
```

PR #3916, opened 2025-06-12
(Classy) Scaffold Classy app for ET

JPry:
### 🎫 Ticket

[TICKET_ID]

### 🗒️ Description

Scaffold the Classy app for ET.

### ✔️ Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,8 @@
     "@redux-devtools/extension": "^3.3.0",
     "@stellarwp/tyson": "*",
     "@types/node": "^20.10.6",
+    "@wordpress/components": "^29.11.0",
+    "@wordpress/data": "^10.25.0",
     "@wordpress/hooks": "^4.18.0",
     "@wordpress/scripts": "^30.14.1",
     "cross-env": "^5.2.0",
--- a/src/Tickets/Classy/Controller.php
+++ b/src/Tickets/Classy/Controller.php
@@ -1,0 +1,87 @@
+<?php
+/**
+ * The main Classy feature controller for Event Tickets.
+ *
+ * @since TBD
+ *
+ * @package TEC\Tickets\Classy;
+ */
+
+declare( strict_types=1 );
+
+namespace TEC\Tickets\Classy;
+
+use TEC\Common\Classy\Controller as Common_Controller;
+use TEC\Common\Contracts\Provider\Controller as Controller_Contract;
+use TEC\Common\StellarWP\Assets\Asset;
+use Tribe__Tickets__Main as ET;
+
+/**
+ * Class Controller
+ *
+ * @since TBD
+ */
+class Controller extends Controller_Contract {
+	/**
+	 * Registers the filters and actions hooks added by the controller.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	protected function do_register(): void {
+		// Register the main assets entry point.
+		if ( did_action( 'tec_common_assets_loaded' ) ) {
+			$this->register_assets();
+		} else {
+			add_action( 'tec_common_assets_loaded', [ $this, 'register_assets' ] );
+		}
+	}
+
+	/**
+	 * Removes the filters and actions hooks added by the controller.
+	 *
+	 * Bound implementations should not be removed in this method!
+	 *
+	 * @since TBD
+	 *
+	 * @return void Filters and actions hooks added by the controller are be removed.
+	 */
+	public function unregister(): void {
+		remove_action( 'tec_common_assets_loaded', [ $this, 'register_assets' ] );
+	}
+
+	/**
+	 * Registers the assets required to extend the Classy application with TEC functionality.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function register_assets() {
+		$post_uses_classy = fn() => $this
+			->container
+			->get( Common_Controller::class )
+			->post_uses_classy( get_post_type() );
+
+		Asset::add(
+			'tec-classy-tickets',
+			'classy.js'
+		)->add_to_group_path( "{$this->get_et_class()}-packages" )
+			// @todo this should be dynamic depending on the loading context.
+			->enqueue_on( 'enqueue_block_editor_assets' )
+			->set_condition( $post_uses_classy )
+			->add_dependency( 'tec-classy' )
+			->add_to_group( 'tec-classy' )
+			->register();
+	}
+
+	/**
+	 * Get the ET class name.
+	 *
+	 * @return class-string
+	 */
+	private function get_et_class(): string {
+		return ET::class;
+	}
+}
--- a/src/resources/packages/classy/index.ts
+++ b/src/resources/packages/classy/index.ts
@@ -1,0 +1,34 @@
+import { addFilter, addAction, didAction, doAction } from '@wordpress/hooks';
+import renderFields from './functions/renderFields';
+import { STORE_NAME, storeConfig } from './store';
+import { WPDataRegistry } from '@wordpress/data/build-types/registry';
+import { getRegistry } from '@tec/common/classy/store';
+
+/**
+ * Hook on the Classy application initialization to add Tickets store to the Classy registry.
+ *
+ * @since TBD
+ *
+ * @return {void} The Tickets store is registered.
+ */
+const registerTicketsStore = (): void => {
+	( getRegistry() as WPDataRegistry ).registerStore( STORE_NAME, storeConfig );
+
+	/**
+	 * Fires after the Tickets store is registered and the Tickets Classy application is initialized.
+	 *
+	 * @since TBD
+	 *
+	 * @return {void} The Tickets store is registered.
+	 */
+	doAction( 'tec.classy.tickets.initialized' );
+};
+
+if ( didAction( 'tec.classy.initialized' ) ) {
+	registerTicketsStore();
+} else {
+	addAction( 'tec.classy.initialized', 'tec.classy.tickets', registerTicketsStore );
+}
+
+// Hook on the Classy fields rendering logic to render the fields.
+addFilter( 'tec.classy.render', 'tec.classy.tickets', renderFields );
--- a/src/resources/packages/classy/store/constants.ts
+++ b/src/resources/packages/classy/store/constants.ts
@@ -1,0 +1,1 @@
+export const STORE_NAME = 'tribe/tickets/classy'; 
--- a/src/resources/packages/classy/store/index.ts
+++ b/src/resources/packages/classy/store/index.ts
@@ -1,0 +1,11 @@
+import { reducer } from './reducer';
+import * as actions from './actions';
+import * as selectors from './selectors';
+
+export const STORE_NAME = 'tec/classy/tickets';
+
+export const storeConfig = {
+	reducer,
+	actions,
+	selectors,
+};
--- a/src/resources/packages/classy/store/selectors.ts
+++ b/src/resources/packages/classy/store/selectors.ts
@@ -1,0 +1,33 @@
+import { createSelector } from '@wordpress/data';
+import { STORE_NAME } from './constants';
+import { StoreState, StoreSelectors } from '../types/store';
+
+export const getTicketPrice = createSelector(
+	(state: StoreState) => [state.ticket?.price],
+	(state: StoreState) => state.ticket?.price || 0
+);
+
+export const getTicketStock = createSelector(
+	(state: StoreState) => [state.ticket?.stock],
+	(state: StoreState) => state.ticket?.stock || 0
+);
+
+export const getTicketStartDate = createSelector(
+	(state: StoreState) => [state.ticket?.startDate],
+	(state: StoreState) => state.ticket?.startDate || ''
+);
+
+export const getTicketEndDate = createSelector(
+	(state: StoreState) => [state.ticket?.endDate],
+	(state: StoreState) => state.ticket?.endDate || ''
+);
+
+export const getTicketIsFree = createSelector(
+	(state: StoreState) => [state.ticket?.isFree],
+	(state: StoreState) => state.ticket?.isFree || false
+);
+
+export const getTicketQuantity = createSelector(
+	(state: StoreState) => [state.ticket?.quantity],
+	(state: StoreState) => state.ticket?.quantity || 0
+); 
--- a/src/resources/packages/classy/types/store.ts
+++ b/src/resources/packages/classy/types/store.ts
@@ -1,0 +1,25 @@
+export interface TicketState {
+	price: number;
+	stock: number;
+	startDate: string;
+	endDate: string;
+	isFree: boolean;
+	quantity: number;
+}
+
+export interface StoreState {
+	ticket: TicketState;
+}
+
+export interface StoreSelectors {
+	getTicketPrice: () => number;
+	getTicketStock: () => number;
+	getTicketStartDate: () => string;
+	getTicketEndDate: () => string;
+	getTicketIsFree: () => boolean;
+	getTicketQuantity: () => number;
+}
+
+export interface StoreActions {
+	// Add actions here when needed
+} 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "event-tickets",
-  "version": "5.24.0",
+  "version": "6.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "event-tickets",
-      "version": "5.24.0",
+      "version": "6.0.0",
       "dependencies": {
         "@babel/runtime": "^7.15.3",
         "@moderntribe/common": "file:common/src/modules",
@@ -55,6 +55,8 @@
         "@redux-devtools/extension": "^3.3.0",
         "@stellarwp/tyson": "*",
         "@types/node": "^20.10.6",
+        "@wordpress/components": "^29.11.0",
+        "@wordpress/data": "^10.25.0",
         "@wordpress/hooks": "^4.18.0",
         "@wordpress/scripts": "^30.14.1",
         "cross-env": "^5.2.0",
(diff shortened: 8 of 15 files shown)


src/Tickets/Classy/Controller.php
```diff
@@ -0,0 +1,87 @@
...
+			->add_to_group( 'tec-classy' )
+			->register();
+	}
+
+	/**
+	 * Get the ET class name.
+	 *
+	 * @return class-string
+	 */
+	private function get_et_class(): string {
```
lucatume, line 103: Why use a method to get a string value and not just use the string value directly? I'm curious about the reasons.

src/resources/packages/classy/store/constants.ts
```diff
@@ -0,0 +1 @@
+export const STORE_NAME = 'tribe/tickets/classy';
```
lucatume: Following the convention, the names should be:
* Common -> `tec/classy`
* TEC -> `tec/classy/events`
* ECP -> `tec/classy/events-pro`
* ET -> `tec/classy/tickets`
* ET+ -> `tec/classy/tickets-plus`

src/resources/packages/classy/store/index.ts
```diff
@@ -0,0 +1,11 @@
+import { reducer } from './reducer';
+import * as actions from './actions';
+import * as selectors from './selectors';
+
+export const STORE_NAME = 'tec/classy/tickets';
```
lucatume: This is likely a WIP mistake, since the `STORE_NAME` is defined in two files.
The WordPress convention here is not consistent (e.g. [here](https://github.com/WordPress/gutenberg/blob/8586e6357ff00fb46132624e65a721fc197af58b/packages/core-data/src/index.js#L21) is exported from a name file and then from the index) so we can make up our own way.

The options I see are:
* define in `store/name.ts` and export from `store/index.ts` (like the WP example I've posted does).
* define and export from `store/index.ts`
* definitely in `store/constants.ts`  (like any other constant would be defined), and export from `store/index.ts`

I'm neutral about this; let's agree on a convention and stick to that.

src/resources/packages/classy/types/store.ts
```diff
@@ -0,0 +1,25 @@
+export interface TicketState {
```
lucatume: Again about conventions: I've gone with one-type-one-file convention so far in _most_ of the cases, and stuck with `CamelCase.d.ts` file names.
I'm neutral toward file naming and grouping multiple types in one type definition file; let's agree on a convention and stick with it in all stores and in all Classy packages.

src/resources/packages/classy/store/selectors.ts
```diff
@@ -0,0 +1,33 @@
+import { createSelector } from '@wordpress/data';
+import { STORE_NAME } from './constants';
+import { StoreState, StoreSelectors } from '../types/store';
+
+export const getTicketPrice = createSelector(
```
lucatume: This applies to this selector and others: I know `createSelector` is a convention, but to memoize it will use the same value that would be pulled from the state.
If my understanding is correct, it will check whether `state.ticket.price` changed (in the current state), if it did not it will return the memoized value (from a cache that is a KV object); if it changed it will pull the value from the state (from the current state).
Since there are no costly interpolations of the value to be fetched, memoization here is adding an additional check on the value that will always be done with the same cost of just fetching the current value.

I think memoization makes sense in these cases:
* if the fetching of the current value costs more than the memoization check
* if the values to select depend on other values than the selected ones

src/Tickets/Classy/Controller.php
JPry: I just prefer to be able to use interpolation in strings. Since constants cannot be interpolated, this is just a simple helper to allow for interpolation.

src/resources/packages/classy/store/constants.ts
JPry: I agree, the store was left in a basic state after AI created it, and I still ned to clean it up.

src/resources/packages/classy/store/index.ts
JPry: > This is likely a WIP mistake, since the STORE_NAME is defined in two files.

Yes, this will be updated.

> The WordPress convention here is not consistent (e.g. [here](https://github.com/WordPress/gutenberg/blob/8586e6357ff00fb46132624e65a721fc197af58b/packages/core-data/src/index.js#L21) is exported from a name file and then from the index) so we can make up our own way.
>
> The options I see are:
>
> * define in store/name.ts and export from store/index.ts (like the WP example I've posted does).
> * define and export from store/index.ts
> * definitely in store/constants.ts (like any other constant would be defined), and export from store/index.ts
>
> I'm neutral about this; let's agree on a convention and stick to that.

I think all of these are fine, and I agree with the need for a consistent convention.

I think my preference would be for a single `constants.ts` file. I like the idea of keeping things of like-type together.

Would you want to consider having a `constants.ts` file in the root of the `classy/` package, for all constants? Or do you think we should segregate them based on component?

src/resources/packages/classy/types/store.ts
JPry: I like the current convention:

> I've gone with one-type-one-file convention so far in most of the cases, and stuck with `CamelCase.d.ts` file names.

I think this is easy to understand where to find types, and the camel case helps to distinguish types when using them in the code. So I think we should keep using this same convention.

(This file will change as part of cleaning up the store)

src/resources/packages/classy/store/index.ts
```diff
@@ -0,0 +1,12 @@
+import { reducer } from './reducer';
+import * as actions from './actions';
+import * as selectors from './selectors';
+import { STORE_NAME } from './constants';
+
+export { STORE_NAME };
+
+export const storeConfig = {
```
lucatume, line 14: The `storeConfig` constant is exported from this file and from the `store/store.ts` one; this should probably just be a re-export of the one one exported from the `store/store.ts` one.

src/resources/packages/classy/types/store.ts
lucatume: Deal, let's go with this (as you've already done in a refactoring of this code).

src/resources/packages/classy/store/index.ts
lucatume: I think one root file would be enough.
Following the WordPress pattern the constants' names would be lie `STORE_NAME`; if constants are namespaced anyway, one file will suffice.
We can revisit the decision, in case.

src/resources/packages/classy/components/CurrencyInput/CurrencyInput.tsx
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Renders a currency input field in the Classy editor.
+ *
+ * @since TBD
+ *
+ * @param {TicketComponentProps} props
+ * @return {JSX.Element} The rendered ticket price field.
+ */
+export default function CurrencyInput( props: CurrencyInputProps ): JSX.Element{
+
```
lucatume: This empty line ...

src/resources/packages/classy/components/CurrencyInput/CurrencyInput.tsx
```diff
@@ -0,0 +1,84 @@
...
+	const [ hasFocus, setHasFocus ] = useState< boolean >( false );
+
+	/*
+	 * Todo: Rework this to use imask instead of a custom renderValue function.
+	 *
+	 * When I tried using the imask library, it didn't work as expected. Attempting to
+	 * use any of the components from the library resulted in the entire ET Classy
+	 * editor failing to load, so I reverted to a custom renderValue function.
+	 */
+	const renderValue = ( value: string ): string => {
```
lucatume: Here and elsewhere you're using values that comes from state, effect or props:

* if the function has a limited number of parameters (it's not the case with this one) then move it outside of the component (i.e. define it in the same file, but outside the component function) to avoid the function being re-defined on each render
* ... or use `useCallback` and set up dependecies accordingly.

Personally I pick the first approach if the number of parameters is three or less; else I will use `useCallback`.
In this specific case, it would make sense to use `useCallback`.

package.json
```diff
@@ -127,6 +129,7 @@
     "querystringify": "^2.0.0",
     "react-18-input-autosize": "^3.0.0",
     "react-day-picker": "^9.3.0",
+    "react-imask": "^7.6.1",
```
lucatume: Can you elaborate on why you added this library? Each dependency (that is not already a WordPress or a `@wordpress/scripts` dependency) is a potential update issue in the future, so I would like to understand what is the tradeoff here.

src/resources/packages/classy/components/CurrencyInput/CurrencyInput.tsx
lucatume: Reading the rest of the code I understand you're aware of the above and this is likely work-in-progress code, but now the comment is there it might be useful.

src/resources/packages/classy/fields/Capacity/Capacity.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const defaultValue: CapacityType = 'general-admission';
+
+const capacityOptions: { label: string, value: CapacityType }[] = [
+	{
+		label: _x( 'General Admission', 'Label for general admission capacity type', 'event-tickets' ),
+		value: 'general-admission',
+	},
+	{
+		label: _x( 'Assigned Seating', 'Label for reserved seating capacity type', 'event-tickets' ),
```
lucatume, line 31: Here the component should get a flag from the backend to know whether Seating is active or not. The Seating feature can be deactivated using the `TEC_TICKETS_SEATING_DISABLED` constant or environment variable.

src/resources/packages/classy/fields/SaleDuration/EndSelector.tsx
```diff
@@ -0,0 +1,85 @@
+import * as React from 'react';
+import { MouseEventHandler } from 'react';
+import { StartOfWeek } from '@tec/common/classy/types/StartOfWeek';
+import { RefObject, useRef } from '@wordpress/element';
+import { format } from '@wordpress/date';
+import { _x } from '@wordpress/i18n';
+import { DatePicker, TimePicker } from '@tec/common/classy/components';
+import { useSelect } from '@wordpress/data';
+
+export default function EndSelector( props: {
```
lucatume, line 10: There's a very similar component in TEC for the start/end date/times, rather that duplicating it here, I would extract something we can re-use in both places and move it in Common.

I will look into this and get back to you about the move. The same thing applies to the `StartSelector`.

src/resources/packages/classy/api/tickets.ts
```diff
@@ -0,0 +1,113 @@
+import { Ticket } from '../types/Ticket';
```
lucatume: Comment on the file and the organization: not a bad idea. I will do the same in the existing API fetches in TEC to move API fetching functionality to dedicated functions.

src/resources/packages/classy/index.ts
```diff
@@ -0,0 +1,37 @@
...
+	 *
+	 * @return {void} The Tickets store is registered.
+	 */
+	doAction( 'tec.classy.tickets.initialized' );
+};
+
+if ( didAction( 'tec.classy.initialized' ) ) {
+	registerTicketsStore();
+} else {
+	addAction( 'tec.classy.initialized', HOOKS_NAMESPACE, registerTicketsStore );
```
lucatume: I've got mixed feelings about using the `HOOK_NAMESPACE` constant. On one side it keeps the code clean and DRY, on the other it makes statically skimming the code more difficult since it requires looking up another file and running code in your head to get the information. This includes IDEs running static analysis.

I'm willing to be convinced using a constant is a better approach, but my instinct would be to keep it a plain string until it's used five times or less.

src/resources/packages/classy/store/resolver.ts
```diff
@@ -0,0 +1,40 @@
...
+			async ( { dispatch } ): Promise<void> => {
+				if ( ! postId ) {
+					return;
+				}
+
+				fetchTicketsForPost( postId )
+					.then( ( tickets: Ticket[] ) => {
+						dispatch.setTicketsForPost( postId, tickets.map( mapApiTicketToTicket ) );
+					} )
+					.catch( ( error ) => {
```
lucatume: I had drafted a comment about the `fetchTicketsForPost` function not catching the error: I see here you're using the error as a signal that something has gone wrong and acting on that information. This does make sense, good idea.

Is pagination taken into account here?

src/resources/packages/classy/api/tickets.ts
```diff
@@ -0,0 +1,113 @@
...
+		searchParams.set( 'per_page', params.per_page.toString() );
+	}
+
+	if ( params.page ) {
+		searchParams.set( 'page', params.page.toString() );
+	}
+
+	const url = `/tribe/tickets/v1/tickets/${ searchParams.toString() ? `?${ searchParams.toString() }` : '' }`;
+
+	const response = await fetch( url, {
```
lucatume: I had drafted a comment here about a possible error thrown by the `fetch` function not being caught.
Seeing how you're using the function I think this makes sense, but document the choice in the function doc-block: the reason you're not catching the error is good, but not obvious.

src/resources/packages/classy/style.pcss
```diff
@@ -0,0 +1,62 @@
+:root {
```
lucatume, line 1: Pleas make sure that only things that are really about tickets will live here.
Ideally I would like to have the Classy style in Common contain all the presentational classes required to add new components and fields.
Do not be shy about refactoring Common components to have more general-use classes and their styles.

This said: if something is only applicable to a tiny piece of Tickets' component/field, it should live here.

src/resources/packages/classy/types/Actions.ts
```diff
@@ -0,0 +1,31 @@
...
+	tickets: Ticket[];
+}
+
+export type Actions = {
+	// Tickets CRUD actions
+	fetchTickets: (postId: number) => Promise<void>;
+	createTicket: (ticketData: Partial<Ticket>) => Promise<void>;
+	updateTicket: (ticketId: number, ticketData: Partial<Ticket>) => Promise<void>;
+	deleteTicket: (ticketId: number) => Promise<void>;
+	setCurrentPostId: (postId: number) => void;
```
lucatume: With access to the `select('core/editor').getCurrentPostId()` selector, why have this? Duplicated state will have to be maintained in two places: the Core stores, and the Classy stores. What is already in the Core stores, should be read and written there.

src/resources/packages/classy/types/StoreSelectors.d.ts
```diff
@@ -0,0 +1,12 @@
+import { Ticket } from './Ticket';
+
+export type StoreSelectors = {
```
lucatume: As discussed, I think it's fine to have a type for the `Store.d.ts` that contains all store-related types.

src/resources/packages/classy/types/FeeSubType.d.ts
```diff
@@ -0,0 +1 @@
+export type FeeSubType = 'flat' | 'percentage';
```
lucatume: Should this be moved to `Ticket.d.ts`?

src/resources/packages/classy/types/TicketsApiParams.d.ts
```diff
@@ -0,0 +1,5 @@
+export type TicketsApiParams = {
```
lucatume: Conflate this file and the `TicketsApiResponse.d.ts` file into an `Api.d.ts` one?

src/resources/packages/classy/types/TicketDate.d.ts
```diff
@@ -0,0 +1,15 @@
+import { Days } from '@tec/common/classy/types/Days';
+import { Hours } from '@tec/common/classy/types/Hours';
+import { Minutes } from '@tec/common/classy/types/Minutes';
+import { Months } from '@tec/common/classy/types/Months';
+
+type Seconds = Minutes;
+
+export type TicketDate = {
```
lucatume: This could be moved to a `Ticket.d.ts` file.

src/resources/packages/classy/types/TicketComponentProps.d.ts
```diff
@@ -0,0 +1,5 @@
+export type TicketComponentProps = {
```
lucatume, line 1: If this type is used only in one place (this is what I got from searching the PR); then it could be inlined in the component using it, or in a `Components.d.ts` or `Ticket.d.ts` file.

lucatume (commented):
No major issues, good work so far.
I've left comments mainly about readability and to control file proliferation.

src/resources/packages/classy/store/resolver.ts
JPry: > Is pagination taken into account here?

Not yet, but full API functionality (including pagination) is something I'll revisit once I get the basics working.

src/resources/packages/classy/types/TicketComponentProps.d.ts
JPry: It is used across multiple components. But as I work this feature I'll evaluate whether to keep it or inline it.

src/Tickets/Classy/Controller.php
```diff
@@ -0,0 +1,139 @@
...
+ */
+
+declare( strict_types=1 );
+
+namespace TEC\Tickets\Classy;
+
+use TEC\Common\Classy\Controller as Common_Controller;
+use TEC\Common\Contracts\Provider\Controller as Controller_Contract;
+use TEC\Common\StellarWP\Assets\Asset;
+use TEC\Tickets\Classy\REST\Controller as RESTController;
```
lucatume: Change the alias to snake_case, `Rest_Controller`

lucatume: Or use `REST\Controller::class` where you need to register it. I personally prefer this approach.

src/Tickets/Classy/REST/Endpoints/Tickets.php
```diff
@@ -0,0 +1,951 @@
...
+	 * Add ticket callback executed to update / add a new ticket.
+	 *
+	 * @since TBD
+	 *
+	 * @param Request $request The request object.
+	 *
+	 * @return Response
+	 * @throws RESTException
+	 */
+	public function upsert_ticket( Request $request ) {
```
lucatume, line 198: I'm sure you've put thoughts into it, so I'm asking.
Why another REST API to perform CRUD operations? What was not working correctly in the one already used by blocks to create tickets?

src/Tribe/REST/V1/Post_Repository.php
```diff
@@ -915,7 +915,6 @@ protected function clean_ticket_data( array &$data ) {
 			'ID',
 			'name',
 			'show_description',
-			'price',
```
lucatume, line 918: Why this removal?

src/resources/packages/classy/components/Icons/ChevronDown.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+
+/**
+ * ChevronDown component for rendering a chevron down icon in the Classy editor.
+ *
+ * @since TBD
+ *
+ * @return {JSX.Element} The rendered chevron down icon.
+ */
+export default function ChevronDown(): JSX.Element {
```
lucatume, line 10: Here and for other icons, make sure you're not duplicating icons already existing in [the WordPress set](https://wordpress.github.io/gutenberg/?path=/docs/icons-icon--docs)

src/resources/packages/classy/types/LocalizedData.d.ts
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @since TBD
+ */
+export type CurrencySettings = {
+	/**
+	 * The currency code, e.g., 'USD', 'EUR'.
+	 *
+	 * @since TBD
+	 */
+	code: string;
```
lucatume, line 32: Fancy! And a good idea!

src/resources/packages/classy/types/Ticket.d.ts
```diff
@@ -0,0 +1,102 @@
...
+	day: Days;
+	hour?: Hours;
+	minute?: Minutes;
+	second?: Seconds;
+}
+
+export type TicketId = number;
+
+export type Ticket = {
+	// API response fields
```
lucatume:
```suggestion
	// API response fields.
```

Full-stop at the end of lines, here and elsewhere.

lucatume (approved):
Tests, for both PHP and JS code, will be object of another PR.

src/Tickets/Classy/Controller.php
```diff
@@ -0,0 +1,139 @@
...
+	 *
+	 * @return array The data to be localized in the Classy script.
+	 */
+	private function get_data(): array {
+		$code = Currency::get_currency_code();
+
+		/** @var ET $et_main */
+		$et_main = $this->container->get( ET::class );
+
+		return [
```
rafsuntaskin, line 120: We need to add a filter in this method to allow WooCommerce currency options from ETP. Or maybe we need to store both when availble as we need to allow users to switch provider.

PR #3923, opened 2025-06-17
[SL-291] Fix issues with Seating timer getting expired for WooCommerce Cart block page and TicketsCommerce coupon application.

rafsuntaskin:
By importing the frontend/session module by file path in two different bundles (session, ticketsBlock) the file is included twice and its side-effect fires twice. By using a named import the file will be included only once, from the window object.

### 🎫 Ticket

[TICKET_ID]

### 🗒️ Description
* refactor(Seating) import by name to avoid loading same file twice.

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Seating/app/frontend/session/index.js
+++ b/src/Tickets/Seating/app/frontend/session/index.js
@@ -750,6 +750,7 @@ export function setTargetDom( targetDocument ) {
  * @return {void} The timer is synced.
  */
 export async function syncOnLoad() {
+	console.log('syncOnLoad called');
 	const syncTimerElements = Array.from( getTimerElements() ).filter( ( syncTimerElement ) => {
 		return 'syncOnLoad' in syncTimerElement.dataset;
 	} );
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -99,6 +99,27 @@ customEntryPoints['wizard/wizard'] = exposeEntry('tec.tickets.wizard', __dirname
  */
 doNotPrefixSVGIdsClasses(defaultConfig);
 
+defaultConfig.externals = [
+	// TEC legacy
+	( { context, request }, callback ) => {
+		if ( /^@moderntribe\//.test( request ) ) {
+			const path = request.replace( /\//g, '.' ).replace( '@moderntribe', 'tec' ).replace( /-/g, '_' );
+			return callback( null, `var ${ path }` );
+		}
+
+		return callback();
+	},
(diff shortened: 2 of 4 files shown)


src/Tickets/Seating/app/frontend/session/index.js
```diff
@@ -750,6 +750,7 @@ export function setTargetDom( targetDocument ) {
  * @return {void} The timer is synced.
  */
 export async function syncOnLoad() {
+	console.log('syncOnLoad called');
```
JPry: This needs to be removed for production.

rafsuntaskin: Yes. I am adding some more changes in This PR and will update this

webpack.config.js
```diff
@@ -99,6 +99,27 @@ customEntryPoints['wizard/wizard'] = exposeEntry('tec.tickets.wizard', __dirname
  */
 doNotPrefixSVGIdsClasses(defaultConfig);
 
+defaultConfig.externals = [
```
JPry, line 102: I'm curious: why are the externals added this way, instead of to the `module.exports` object below?

rafsuntaskin: I think it's just keeping them consitent with what we did for [ETP](https://github.com/the-events-calendar/event-tickets-plus/blob/f4363b5721064ad674dee785b635c664df9ae984/webpack.config.js#L90).

JPry (approved):

webpack.config.js
```diff
@@ -99,6 +99,18 @@ customEntryPoints['wizard/wizard'] = exposeEntry('tec.tickets.wizard', __dirname
...
+	// TEC modern
+	( { context, request }, callback ) => {
```
JPry, line 104: @rafsuntaskin Looking at this again, we have a method in Tyson to handle this for us: `resolveExternalToGlobal()`. We should use that here.

Also, in Classy we have this directly in the `module.exports` section:

https://github.com/the-events-calendar/the-events-calendar/blob/4775be8f16f12035207c073845b59164840d1c24/webpack.config.js#L137-L141

dpanta94 (approved):

webpack.config.js
JPry: I think what I described will be the approach in the long run, but for now the current version works without extensive testing of all the globals. So I'll withdraw my suggestion for now.

JPry (approved):

PR #3950, opened 2025-07-08
Fix collation and date issues.

Camwyn:
Attempting to fix the lingering MySQL collation and datetime issues.
[skip-changelog]


dpanta94:
Has been included in #3943

PR #3951, opened 2025-07-08
Update check-in to not allow when still processing/pending 

Camwyn:
`Row #16` in [the sheet](https://docs.google.com/spreadsheets/d/1NzXAjMpzE7Cum9rOix2ctIjEsdpR_YgFLLMKvnGtzkk/edit?gid=1765726281#gid=1765726281).
Created some tests but started hitting the existing test issues from the slic update and cannot get them passing yet.


dpanta94:
If we rebase the base branch from gambit the suites should be passing

PR #3954, opened 2025-07-09
Remove EDD references from tests.

Camwyn:


Camwyn:
Have not discovered what is causing the issue in `ft_integration` yet.

bordoni (approved):

PR #3957, opened 2025-07-14
Corrects Stripe payment element display logic

dpanta94:
Reverts a conditional that was incorrectly hiding the Stripe payment element button.

The button will now always be visible, ensuring a consistent checkout experience.

### DIFF
--- a/src/views/v2/commerce/gateway/stripe/payment-element.php
+++ b/src/views/v2/commerce/gateway/stripe/payment-element.php
@@ -25,10 +25,9 @@ if ( $must_login || ! $payment_element ) {
 	return;
 }
 
-$hidden = $shortcode->should_display_billing_info() ? 'tribe-common-a11y-hidden' : '';
 ?>
 <div id="tec-tc-gateway-stripe-payment-element" class="tribe-tickets__commerce-checkout-stripe-payment-element"></div>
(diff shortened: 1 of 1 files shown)


src/views/v2/commerce/gateway/stripe/payment-element.php
```diff
@@ -25,10 +25,9 @@
 	return;
 }
 
-$hidden = $shortcode->should_display_billing_info() ? 'tribe-common-a11y-hidden' : '';
 ?>
 <div id="tec-tc-gateway-stripe-payment-element" class="tribe-tickets__commerce-checkout-stripe-payment-element"></div>
-<button id="tec-tc-gateway-stripe-checkout-button" class="tribe-common-c-btn tribe-tickets__commerce-checkout-form-submit-button <?php echo esc_attr( $hidden ); ?>">
+<button id="tec-tc-gateway-stripe-checkout-button" class="tribe-common-c-btn tribe-tickets__commerce-checkout-form-submit-button">
```
rafsuntaskin, line 30: If we want to revert, then we should revert to the production [state](https://github.com/the-events-calendar/event-tickets/pull/3937/files#diff-befe98003b29869e7d5c2b3e6a896a8ba7ddad9e29a94ea93ccd1e776dad75a1L24).

Unless, that one is flawed and it should always display.

dpanta94: In the production state, the variable `$show_address` is always null resulting in the button always showing

PR #3965, opened 2025-07-19
ETP-1021 - pass 2

lucatume:
Ticket: [ETP-1021](https://stellarwp.atlassian.net/browse/ETP-1021)

This PR builds on the discover of #3944 and [Common PR](https://github.com/the-events-calendar/tribe-common/pull/2681) to leverage the KV-Cache for the Ticket model.

WIP

[ETP-1021]: https://stellarwp.atlassian.net/browse/ETP-1021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Events/Views/V2/Hooks.php
+++ b/src/Tribe/Events/Views/V2/Hooks.php
@@ -77,15 +77,15 @@ class Hooks extends \TEC\Common\Contracts\Service_Provider {
 	 *
 	 * @since 4.11.2
 	 *
-	 * @param array            $namespace_map Indexed array containing the namespace as the key and path to `strpos`.
-	 * @param string           $path          Path we will do the `strpos` to validate a given namespace.
-	 * @param Tribe__Template  $template      Current instance of the template class.
+	 * @param array           $namespace_map Indexed array containing the namespace as the key and path to `strpos`.
+	 * @param string          $path          Path we will do the `strpos` to validate a given namespace.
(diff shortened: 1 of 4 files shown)


src/Tribe/Events/Views/V2/Hooks.php
```diff
@@ -126,4 +127,22 @@ protected function add_filters() {
...
+	 * @param int $post_id The post ID. It could be any post type, not just events.
+	 *
+	 * @return void
+	 */
+	public function regenerate_post_kv_caches( $post_id ): void {
+		if ( ! is_int( $post_id ) ) {
+			return;
+		}
+
+		Tickets::regenerate_caches( $post_id );
```
dpanta94: Why not cast it to int and let the `regenerate_caches` handle it ?

lucatume: Defensive coding.
Most of the non-numeric strings, when cast to `int`, will return `0`.
Some will return non-intuitive values, e.g. the `true` string returns `0`.
Then float strings like `23,89` will cast to their number counterpart.
But then: would we want to run the show when the value is `0`? Or less than `0`? No. Especially due to the side-effects of the cache regeneration.
I've added the casting, but making sure the value is a positive integer first.

PR #3982, opened 2025-08-11
Classy - Integrate ET VE fields

lucatume:
This PR scaffolds missing build files and scripts (e.g. `classy:format` in Composer and node) and implements support for Virtual Location fields in ET.

In the previous code, the ET support is done in VE (i.e. in ECP), but this implementation moves the burden to ET using the Slot/Fill API and filters.

See the supporting [ECP PR](https://github.com/the-events-calendar/events-pro/pull/2821).

WIP

### DIFF
--- a/src/Tickets/Classy/ECP_Editor_Meta.php
+++ b/src/Tickets/Classy/ECP_Editor_Meta.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * Handles Event Tickets integration with Classy and Events Pro.
+ *
+ * @since   TBD
+ *
+ * @package TEC\Tickets\Classy;
+ */
+
+namespace TEC\Tickets\Classy;
+
+use Tribe\Events\Virtual\Compatibility\Event_Tickets\Event_Meta;
+use Tribe__Editor__Meta as Editor_Meta_Contract;
+
+/**
+ * Class ECP_Editor_Meta.
+ *
+ * @since   TBD
+ *
+ * @package TEC\Tickets\Classy;
+ */
+class ECP_Editor_Meta extends Editor_Meta_Contract {
+	/**
+	 * Registers the meta fields supported by Event Tickets when Events Pro is active.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function register() {
+		register_meta(
+			'post',
+			Event_Meta::$key_rsvp_email_link,
+			$this->text()
+		);
+
+		register_meta(
+			'post',
+			Event_Meta::$key_ticket_email_link,
+			$this->text()
+		);
+	}
+}
--- a/src/resources/packages/classy/fields/VirtualLocation/AdditionalSettings.tsx
+++ b/src/resources/packages/classy/fields/VirtualLocation/AdditionalSettings.tsx
@@ -1,0 +1,83 @@
+import * as React from 'react';
+import { _x } from '@wordpress/i18n';
+import { ToggleControl } from '@wordpress/components';
+import { useState } from '@wordpress/element';
+import { Settings } from '../../types/VirtualLocation';
+import { useSelect } from '@wordpress/data';
+import {
+	METADATA_EVENTS_VIRTUAL_RSVP_EMAIL_LINK,
+	METADATA_EVENTS_VIRTUAL_TICKET_EMAIL_LINK,
+} from '../../constants.tsx';
+
+const includeVideoLinkOptions: {
+	label: string;
+	value: 'rsvp' | 'ticket';
+}[] = [
+	{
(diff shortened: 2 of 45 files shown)


JPry (approved):
I know it's WIP, but looking good so far to me

src/resources/packages/classy/fields/VirtualLocation/useMetaFiltering.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @since TBD
+ *
+ * @param {string} hookName A unique name to identify the callback.
+ * @param {Function} onMetaUpdate  Callback function to filter meta data on update. Receives and returns meta object.
+ * @param {Function} onMetaDelete  Callback function to filter meta data on deletion. Receives and returns meta object.
+ * @param {Array} dependencies  React dependency array that triggers filter re-registration when values change.
+ *
+ * @return void
+ */
+export default function useMetaFiltering(
```
JPry, line 18: I love this idea for handling the meta filters.

src/resources/packages/classy/fields/VirtualLocation/ViewingPermissions.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { useSelect } from '@wordpress/data';
+import { METADATA_EVENT_VIRTUAL_SHOW_EMBED_TO } from '../../constants.tsx';
+import useMetaFiltering from './useMetaFiltering.ts';
+
+export default function ViewingPermissions(): JSX.Element {
+	const meta: {
+		showAtRsvpAttendees: boolean;
+		showAtTicketAttendees: boolean;
+	} = useSelect( ( select ) => {
+		const store: {
```
JPry: For the type here, there is `CoreEditorSelect` in the `Store.d.ts` file.

src/resources/packages/classy/fields/VirtualLocation/AdditionalSettings.tsx
```diff
@@ -0,0 +1,75 @@
...
+	METADATA_EVENTS_VIRTUAL_TICKET_EMAIL_LINK,
+} from '../../constants.tsx';
+import useMetaFiltering from './useMetaFiltering.ts';
+
+export default function AdditionalSettings(): JSX.Element {
+	const meta: {
+		includeVideoLinkInRsvpEmails: boolean;
+		includeVideoLinkInTicketEmails: boolean;
+	} = useSelect( ( select ) => {
+		const store: {
```
JPry: Same as the other file, there's a `CoreEditorSelect` type that can be used for typing in the `Store.d.ts` file.

src/Tickets/Classy/ECP_Editor_Meta.php
```diff
@@ -0,0 +1,43 @@
...
+class ECP_Editor_Meta extends Editor_Meta_Contract {
+	/**
+	 * Registers the meta fields supported by Event Tickets when Events Pro is active.
+	 *
+	 * @since TBD
+	 *
+	 * @return void
+	 */
+	public function register() {
+		register_meta(
```
JPry, line 31: Looking at this again, I was thinking: do we want to add any protection within this class for the `Event_Meta` class not existing? I know that this method is hooked to the action that triggers in Events Pro, but it might be prudent to add a protection here too.

src/resources/packages/classy/fields/VirtualLocation/ViewingPermissions.tsx
```diff
@@ -0,0 +1,104 @@
...
+				metaValue.push( 'ticket' );
+			} else {
+				metaValue = metaValue.filter( ( value: string ) => value !== 'ticket' );
+			}
+
+			meta[ METADATA_EVENT_VIRTUAL_SHOW_EMBED_TO ] = metaValue;
+
+			return meta;
+		},
+		( meta: Object ): Object => {
```
JPry, line 50: What would you think of allowing for the `onMetaDelete` parameter to be optional, and if not provided it handles this logic automatically?

src/Tickets/Classy/ECP_Editor_Meta.php
lucatume: The class will not be loaded if the ECP Classy controller has not registered.
Adding a check on the class, a `class_exist`, would defeat the purpose of using `register_on_action`.
What would break this code is a removal of the `Event_Meta` class, but that _should_ happen taking care of code like this one since the point of using the static property is to make it easy to find usages.

src/resources/packages/classy/fields/VirtualLocation/ViewingPermissions.tsx
lucatume: It would make the ergonomics of the hook not easy since it would require passing the dependencies before the function.

PR #3990, opened 2025-08-14
Set up Codeception tests for Classy

JPry:
## Notable Changes

* WIP


tests/classy_integration/Controller_Test.php
```diff
@@ -2,85 +2,242 @@
...
+		);
+	}
+
+	/**
+	 * @covers Controller::do_register
+	 */
+	public function test_do_register_registers_assets_when_tec_common_assets_loaded_action_did_run(): void {
+		$controller = $this->make_controller();
+
+		// Mock the action to have already run
```
lucatume: Here and elsewhere, add full-stop at end of comments.

tests/classy_integration/Controller_Test.php
```diff
@@ -2,85 +2,242 @@
...
+	 * @covers Controller::register_assets
+	 */
+	public function test_register_assets_registers_script_and_styles(): void {
 		/** @var Controller $controller */
 		$controller = $this->make_controller();
 
-		$post_types = $controller->add_supported_post_types( [] );
+		// Mock Common_Controller
+		$common_controller = $this->createMock( Common_Controller::class );
+		$common_controller->method( 'post_uses_classy' )->willReturn( true );
```
lucatume: You can use the filter `tec_classy_post_types` to control the supported post types.
Avoid mocking if you can just set up a fixture with a real post and add the required filters.

tests/classy_integration/Controller_Test.php
```diff
@@ -2,85 +2,242 @@
...
+		$this->assertEquals( 'tec_events_pro_classy_registered', $added_actions[1]['hook'] );
+		$this->assertEquals( [ $controller, 'register_ecp_editor_meta' ], $added_actions[1]['callback'] );
+	}
+
 	/**
-	 * @covers \TEC\Events\Classy\Controller::add_supported_post_types
+	 * @covers Controller::unregister
 	 */
-	public function test_add_supported_post_types(): void {
+	public function test_unregisters_rest_controller(): void {
```
lucatume: Do not mock controllers.

And this method is not required, the `Controller_Test_Case` will already check that all the filters/actions added by the controller, and any sub-controller it registers, are removed.
The REST controller registers on `rest_api_init` to call the `register_routes` method; that will be tested and checked by the built-in test method of the `Controller_Test_Case`

tests/classy_integration/Controller_Test.php
```diff
@@ -2,85 +2,242 @@
...
+
+	/**
+	 * @covers Controller::get_data
+	 */
+	public function test_get_data_returns_expected_structure(): void {
+		/** @var Controller $controller */
+		$controller = $this->make_controller();
+
+		// Mock the Currency class.
+		$this->set_class_fn_return( Currency::class, 'get_currency_code', 'USD' );
```
lucatume: Do not mock it, why?
Just use the real one and look for its default values.

tests/classy_integration/REST/Controller_Test.php
```diff
@@ -0,0 +1,376 @@
...
+
+class Controller_Test extends Controller_Test_Case {
+	use With_Uopz;
+
+	protected $controller_class = Controller::class;
+
+	/**
+	 * @covers Controller::register_routes
+	 */
+	public function test_register_routes_registers_all_ticket_endpoints(): void {
```
lucatume: This test case should be rewritten to be simplified: do not test the implementation of the endpoints.

Just create a data provider exercising the endpoints will all the possible arguments, then:
```
// ... set up request and fixture

$response = rest_get_server()->dispatch($request);

// ... assertions on response and side-effects.
```

This will cover the endpoint registration, including arguments sanitization and validation, and will test the code the way it's used.
[Example](https://github.com/the-events-calendar/events-pro/blob/2c4190c228f760c5eb1cfb0dbacd1691bad9582a/tests/classy_integration/Virtual_Location/REST/Zoom_Meetings_Test.php#L371)

tests/classy_integration/Controller_Test.php
```diff
@@ -2,85 +2,229 @@
...
+	 * @covers Controller::do_register
 	 */
-	public function test_add_supported_post_types(): void {
-		/** @var Controller $controller */
+	public function test_do_register_registers_assets_when_tec_common_assets_loaded_action_did_run(): void {
 		$controller = $this->make_controller();
 
-		$post_types = $controller->add_supported_post_types( [] );
+		// Mock the action to have already run.
+		$this->set_fn_return(
```
lucatume: One way you can avoid using `uopz` to mock an action is to remove all filters attached to a filter/action and then apply the filters or do the action:

```php
// Remove all filters to avoid side-effects from other code hooked to it.
remove_all_filters( 'tec_common_assets_loaded' );

// Do the action.
do_action( 'tec_common_assets_loaded' );
```

tests/classy_integration/Controller_Test.php
```diff
@@ -2,85 +2,229 @@
...
+	public function test_do_register_adds_action_when_tec_common_assets_loaded_action_has_not_run(): void {
+		$controller = $this->make_controller();
+
+		// Mock the action to have not run, and are not running.
+		$this->set_fn_return( 'did_action', static fn( $action ) => false, true );
+		$this->set_fn_return( 'doing_action', static fn( $action ) => false, true );
+
+		// Mock add_action to capture calls.
+		$added_actions = [];
+		$this->set_fn_return(
```
lucatume: No need to mock `add_action`, just check with `has_action`.

tests/classy_integration/Controller_Test.php
```diff
@@ -2,85 +2,229 @@
...
 	/**
-	 * @covers \TEC\Events\Classy\Controller::filter_data
+	 * @covers Controller::do_register
 	 */
-	public function test_filter_data(): void {
+	public function test_do_register_adds_action_when_tec_common_assets_loaded_action_has_not_run(): void {
+		$controller = $this->make_controller();
+
+		// Mock the action to have not run, and are not running.
+		$this->set_fn_return( 'did_action', static fn( $action ) => false, true );
```
lucatume: Here you're trying to say "the `tec_common_assets_loaded` action was not done and is not being done".
You can do this without using `uopz`:

```php
// Unset the actions counter.
global $wp_actions;
unset($wp_actions['tribe_common_assets_loaded'];

// 0
var_dump(did_action('tribe_common_assets_loaded'));
```

tests/classy_integration/Controller_Test.php
```diff
@@ -2,85 +2,229 @@
...
+		$this->set_fn_return( 'doing_action', static fn( $action ) => false, true );
```
lucatume: You should not need to mock `doing_action` since you're not doing it.

tests/classy_integration/Controller_Test.php
```diff
@@ -2,85 +2,229 @@
...
-			$data
+	/**
+	 * @covers Controller::get_et_class
+	 */
+	public function test_get_et_class_returns_et_class_name(): void {
+		/** @var Controller $controller */
+		$controller = $this->make_controller();
+
+		// Use reflection to access private method.
+		$method = Closure::bind(
```
lucatume: If it's private it should not be tested. You can remove this test and cover the method implicitly in other tests.

tests/classy_integration/Controller_Test.php
```diff
@@ -2,85 +2,229 @@
...
+		add_filter(
+			'tec_classy_post_types',
+			static fn( $types ) => array_unique( array_merge( $types, $post_types ) )
+		);
+
+		// Mock wp_create_nonce.
+		$this->set_fn_return( 'wp_create_nonce', static fn( $action ) => "nonce_{$action}", true );
+
+		// Use a closure to access the private method.
+		$method = Closure::bind(
```
lucatume: If the method is private, then it should not be tested.
Either make the method public, which I think would be fine in this case, or test it through another side-effect.

lucatume (changes requested):
There is still some mocking, including partial mocking and private access, that I think could be removed.

lucatume (approved):

PR #4003, opened 2025-08-22
Classy: Use new TEC API and update ticket fields

JPry:
### TEC API

* Adds `menu_order` as a parameter for the tickets API
* Updates unit tests to match the new parameter behavior, and add tests for its behavior.
* Removes the previously-added tickets API that was unique to Classy

### Classy Tickets API

* The tickets API in Classy has been updated to use the new TEC API 
* Tests have been updated to match the behavior. 
* The whole feature has been update to match changes made in the-events-calendar/the-events-calendar#5319

### Ticket update/create modal

* The `Ticket price` field now behaves correctly.
* The Sale Price section now behaves *mostly* correctly. There is an issue with the TEC API not returning the `on_sale` value, so it is always marked as NOT having a sale price/date range. Will follow up in another PR
* Styling for the Sale Price section has been updated to match the designs
* Ticket Capacity now works *mostly* correctly. There's still an issue when leaving the value blank, and the API will reject it. I'll follow that up in another PR.
* Menu order has been added to the API, but it needs more thorough testing


src/Tickets/REST/TEC/V1/Traits/With_Filtered_Ticket_Params.php
```diff
@@ -161,6 +162,11 @@ public function filter_upsert_params( array $params ): array {
 			throw new InvalidRestArgumentException( __( 'The ticket price must be a number.', 'event-tickets' ) );
 		}
 
+		// Ensure menu_order value is an integer.
```
dpanta94: Not needed since this is done for you by the sanitizer of the integer parameter

dpanta94: `menu_order` is being handled by the `$ticket_params` and not the `$post_params`.

you need to set it inside the `$new_params` further above and then unset it from the `$post_params`

```php
$new_params['ticket_menu_order'] = $params['menu_order'] ?? null;
unset( $params['menu_order'] );
```

dpanta94 (changes requested):
The ticket ORM can't handle fully all the ticket params. THat's why its the only entity that is being handled in a "special" way by the REST API. We are going to fix that in the close future, but for now i have pointed out why the current tests fail and how to fix them :D

src/Tickets/REST/TEC/V1/Traits/With_Filtered_Ticket_Params.php
JPry: > `menu_order` is being handled by the `$ticket_params` and not the `$post_params`.

@dpanta94 This doesn't make sense to me. The `menu_order` parameter should belong to the post object, so it should go with the other `$post_params`.

Why would it be handled differently?

dpanta94: It's not something i decided. It's a weakness of how the ticket_add method is designed.

https://github.com/the-events-calendar/event-tickets/blob/f9e4e7d6fea81c8cc67b54cf6588e5cc10dddc4c/src/Tribe/Tickets.php#L4006

Actually that there even a ticket_add method exists and is preferred over using the ORM is a bad choice we need to fix.

dpanta94: Also i could be wrong, i haven't debugged this to actually figure it out. Thats what i believe is going on here.

JPry: > Actually that there even a ticket_add method exists and is preferred over using the ORM is a bad choice we need to fix.

Agreed

JPry:
@dpanta94 I made the changes you requested to the API

dpanta94 (approved):
This is good @JPry ! I am approving REST API related changes, since i'm not familiar with Classy. Please look for a more suitable reviewer for that.

src/Tickets/REST/TEC/V1/Traits/With_Filtered_Ticket_Params.php
```diff
@@ -43,6 +43,7 @@ public function filter_upsert_params( array $params ): array {
 
 		$orm = $this->get_orm();
 
+		// todo: investigate deleting this logic to allow moving between events.
```
lucatume, line 46: If this feature is not currently required, then I would remove this todo to leave the developer(s) in charge of implementing the ticket move feature the freedom to explore. They will likely not look into the code for this kind of comments when investigating the ticket move feature and this annotation would likely end up being abandoned here.

src/resources/packages/classy/api/tickets.ts
```diff
@@ -179,89 +197,71 @@ export const deleteTicket = async ( ticketId: number ): Promise< void > => {
...
-			body.ticket.sale_price.end_date = salePriceData.endDate;
+			body.sale_price_end_date = salePriceData.endDate;
 		}
 	}
 
-	// Menu order.
-	body.menu_order = ticketData.menuOrder?.toString() || '0';
+	// Map additional fields
+	if ( ticketData.iac ) {
+		// todo: add iac to the API.
```
lucatume: If this feature, or the fees one, are not directly supported by the ET plugin, then these should be actions/filters that allow the ET+ plugin to extend the data.

src/resources/packages/classy/api/tickets.ts
```diff
@@ -295,44 +295,59 @@ const mapTicketSettingsToApiRequest = ( ticketData: TicketSettings, isUpdate: bo
  * @return {TicketSettings} The mapped ticket settings.
  */
 const mapApiResponseToTicketSettings = ( apiResponse: GetTicketApiResponse ): TicketSettings => {
-	const stockMode = apiResponse.capacity_details?.global_stock_mode || 'own';
+	// Map capacity settings based on stock management
 	const capacitySettings: CapacitySettings = {
-		enteredCapacity: apiResponse.capacity_details?.max || -1,
-		isShared: stockMode === 'capped' || stockMode === 'global',
+		enteredCapacity: apiResponse.stock || 0,
```
lucatume, line 295: Shared capacity is an ET+ concern: should this be an extension point (in the form of a filter or action) in place of being implemented here?

src/resources/packages/classy/types/LocalizedData.d.ts
```diff
@@ -74,6 +53,7 @@ export type CurrencySettings = {
  */
 export type Settings = {
 	currency: CurrencySettings;
+	startOfWeek: number;
```
lucatume: Should this be made into an interval of intervals from `0` to `6`?

tests/classy_jest/_support/mockApiResponse.ts
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * @return {GetTicketApiResponse} The mock API response.
+ */
+export function makeMockApiResponse(): GetTicketApiResponse {
+	return {
+		id: 1,
+		date: '2024-01-01T12:00:00+00:00',
+		date_gmt: '2024-01-01T12:00:00Z',
+		guid: {
+			rendered: 'https://example.com/?p=1',
```
lucatume: Here and in the `permalink_template` I would update the fields to look more in line with what a ticket would look like, e.g. `https://example.com/tickets/ticket-1` or something along those lines.

tests/classy_jest/_support/mockApiTickets.ts
```diff
@@ -0,0 +1,62 @@
+import { GetTicketApiResponse } from '@tec/tickets/classy/types/Api';
+
+/**
+ * Mock API tickets data for testing.
+ *
+ * @return {GetTicketApiResponse[]} The mock API tickets array.
+ */
+export function makeMockApiTickets(): GetTicketApiResponse[] {
```
lucatume: This look like a repetition of the data from the `makeMockApiResponse` function from the `mockApiResponse.ts` file.
I would update the function in `mockApiResponse.ts` to offer customization parameters and use it to build this array.

tests/classy_jest/_support/mockMappedTickets.ts
```diff
@@ -0,0 +1,49 @@
+import { TicketSettings } from '@tec/tickets/classy/types/Ticket';
+
+/**
+ * Mock mapped tickets data for testing.
+ *
+ * @return {TicketSettings[]} The mock mapped tickets array.
+ */
+export function makeMockMappedTickets(): TicketSettings[] {
```
lucatume: This seems to be an array of data  like the one produced by the `makeMockTicketDataForCreate` function from the `mockTicketData.ts` file.
I would update the function in that file to offer customization parameters and use that to build the single entry.

tests/classy_jest/_support/mockApiResponse.ts
```diff
@@ -0,0 +1,106 @@
...
+		menu_order: 0,
+	};
+}
+
+/**
+ * Mock expected result for ticket operations.
+ *
+ * @return {object} The mock expected result.
+ */
+export function makeMockExpectedResult() {
```
lucatume: I would have the expected result live in the file that is making the assertion, else scanning the test file while debugging a failure would required jumping across files to get the correct information increasing cognitive load.
It's fine for tests to be verbose if this removes cognitive load.

tests/rest_tec_v1_integration/Endpoints/Ticket_Test.php
```diff
@@ -75,37 +75,45 @@ protected function create_test_data(): array {
...
 
 		$ticket_3 = $this->create_tc_ticket( $page_1, '15.00' );
 		update_post_meta( $ticket_3, '_name', 'Workshop Registration' );
 		update_post_meta( $ticket_3, '_description', 'Basic workshop registration' );
+		wp_update_post( [ 'ID' => $ticket_3, 'menu_order' => 3 ] );
 
 		// Create ticket for private post
 		$private_ticket = $this->create_tc_ticket( $private_post, '50.00' );
 		update_post_meta( $private_ticket, '_name', 'Private Event Ticket' );
+		wp_update_post( [ 'ID' => $private_ticket, 'menu_order' => 4 ] );
```
lucatume: Since the post ID is always part of the natural ordering, having a `menu_order` replicate the natural post ID ordering is not telling you whether the `menu_order` based-ordering is working or not. I would assign, with a comment explaining why, a `menu_order` that does not follow the natural post ID ordering.
E.g. th first ticket created ha a `menu_order` of `2`, the second `1` and so on.

lucatume (dismissed):
Good work!
Some smaller changes, but overall this looks good to me.

src/Tickets/REST/TEC/V1/Traits/With_Filtered_Ticket_Params.php
JPry: > If this feature is not currently required

It's supposed to already be working, but it was missed when the API was developed. I added this note for myself after a discussion with @Camwyn.

src/resources/packages/classy/api/tickets.ts
JPry: That makes sense to me. I'll update this with actions/filters.

src/resources/packages/classy/types/LocalizedData.d.ts
JPry: Yes, good catch. I think we have a type in Common for that. I'll switch to that type.

src/resources/packages/classy/api/tickets.ts
JPry: > Shared capacity is an ET+ concern

Unless I'm mistaken, shared capacity is also used to facilitate Seating. That's part of ET, no need for ET+.

tests/classy_jest/_support/mockApiResponse.ts
JPry: The `guid` field is usually the non-pretty form; this is how WordPress does it. But I can update other fields.

PR #4035, opened 2025-09-11
Supporting work for Purchase Rules - Compatibility updates for schema/models library updates

dpanta94:
### 🎫 Ticket

[TICKET_ID]

### 🗒️ Description

Updates to support the major upgrades of schema and models.
Supporting work for implementing Purchase Rules in ETP.

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Commerce/Gateways/Stripe/Hooks.php
+++ b/src/Tickets/Commerce/Gateways/Stripe/Hooks.php
@@ -393,7 +393,7 @@ class Hooks extends \TEC\Common\Contracts\Service_Provider {
 	 */
 	public function maybe_create_stripe_payment_intent() {
 
-		if ( ! tribe( Module::class )->is_checkout_page() || ! tribe( Gateway::class )->is_enabled() || ! tribe( Merchant::class )->is_connected() ) {
+		if ( ! ( tribe( Gateway::class )->is_enabled() && tribe( Merchant::class )->is_connected() ) ) {
 			return;
 		}
 


src/admin-views/editor/metabox.php
```diff
@@ -21,15 +21,13 @@
...
-	<?php wp_nonce_field( 'tribe-tickets-meta-box', 'tribe-tickets-post-settings' ); ?>
-
-	<?php $admin_views->template( [ 'editor', 'panel', 'list' ], get_defined_vars() ); ?>
-
-	<?php $admin_views->template( [ 'editor', 'panel', 'ticket' ], get_defined_vars() ); ?>
+	<?php
+	wp_nonce_field( 'tribe-tickets-meta-box', 'tribe-tickets-post-settings' );
 
-	<?php $admin_views->template( [ 'editor', 'panel', 'settings' ], get_defined_vars() ); ?>
+	foreach ( $panels as $panel ) {
```
JPry, line 29: A couple things:

1. How does this relate to the schema updates? I don't see it mentioned in the PR description.
2. `$panels` and `$ticket_id` aren't defined, or at least not that we know about at this point. They either need to be defined, or else declared at the top of the file with the other vars. (`$ticket_id` wasn't changed in this PR, but if we're adding to the file vars anyway then we should include it now)

src/resources/js/tickets.js
```diff
@@ -361,6 +363,16 @@ let ticketHeaderImage = window.ticketHeaderImage || {};
 
 		// Trigger dependency.
 		$( '.tribe-dependency' ).trigger( 'verify.dependency' );
+
+		/**
+		 * Fires after the panels are refreshed.
+		 *
+		 * @since TBD
+		 *
+		 * @param {Object} panels The panels object.
```
JPry, line 372: These parameters don't match what is actually passed to the action.

src/Tickets/Commerce/Gateways/Square/Syncs/Integrity_Controller.php
```diff
@@ -284,28 +284,28 @@ public function check_data_integrity(): void {
 		$to_local_delete = [];
 		$to_be_checked   = [];
 
-		foreach ( Integrity_Table::fetch_all( 100, OBJECT, "WHERE last_checked < '{$a_few_minutes_ago}' AND mode = " . (int) self::$is_prod_mode ) as $item ) {
-			$post_object = get_post( $item->wp_object_id );
+		foreach ( Integrity_Table::get_all( 100, "WHERE last_checked < '{$a_few_minutes_ago}' AND mode = " . (int) self::$is_prod_mode ) as $item ) {
```
JPry: Not strictly part of this PR, but just a note that `$a_few_minutes_ago` is actually `-30 seconds` above. Now would be a good time to give the variable a better name

src/Tickets/Commerce/Gateways/Square/Syncs/Integrity_Table.php
```diff
@@ -67,20 +73,26 @@ class Integrity_Table extends Table {
...
 	 *
-	 * @since 5.24.0
+	 * @since TBD
 	 *
-	 * @var string[]
+	 * @return array<string, callable>
 	 */
-	public static function get_columns(): array {
+	public static function get_schema_history(): array {
+		$table_name = self::table_name( true );
```
JPry:
```suggestion
		$table_name = self::table_name();
```

src/Tickets/Commerce/Gateways/Square/Syncs/Integrity_Controller.php
```diff
@@ -569,13 +569,13 @@ public function sync_items(): void {
 		$skipped_tickets = [];
 		$batch           = [];
 
-		foreach ( Integrity_Table::fetch_all( 100, OBJECT, 'WHERE id IN (' . implode( ',', $ids ) . ')' ) as $item ) {
-			$is_ticket = in_array( get_post_type( $item->wp_object_id ), tribe_tickets()->ticket_types(), true );
+		foreach ( Integrity_Table::get_all( 100, 'WHERE id IN (' . implode( ',', $ids ) . ')' ) as $item ) {
+			$is_ticket = in_array( get_post_type( $item['wp_object_id'] ), tribe_tickets()->ticket_types(), true );
```
JPry: This isn't part of this PR, but while we're here it would be nice to pull `tribe_tickets()->ticket_types()` out of the `foreach` loop, since we don't expect that to change between iterations.

We could also get a small performance boost by using array key comparison instead of `in_array()`.

Here's what the whole section could look like:

```php
		$ticket_types = array_flip( tribe_tickets()->ticket_types() );
		foreach ( Integrity_Table::get_all( 100, 'WHERE id IN (' . implode( ',', $ids ) . ')' ) as $item ) {
			$is_ticket = isset( $ticket_types[ get_post_type( $item['wp_object_id'] ) ] );

```

src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Order_Modifier_Relationships.php
```diff
@@ -55,66 +61,23 @@ class Order_Modifier_Relationships extends Table {
...
-	public static function get_columns(): array {
-		return [
-			'id',
-			'modifier_id',
-			'post_id',
-			'post_type',
-		];
-	}
+	public static function get_schema_history(): array {
+		$table_name = self::table_name( true );
```
JPry:
```suggestion
		$table_name = self::table_name();
```

src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Order_Modifier_Relationships.php
```diff
@@ -55,66 +61,23 @@ class Order_Modifier_Relationships extends Table {
 	/**
 	 * An array of all the columns in the table.
 	 *
-	 * @since 5.20.0
+	 * @since TBD
 	 *
 	 * @var string[]
```
JPry: While we're here, let's fix the docblock to match reality.

src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Order_Modifier_Relationships.php
```diff
@@ -55,66 +61,23 @@ class Order_Modifier_Relationships extends Table {
...
-		if ( ! count( $results ) ) {
-			return $results;
-		}
-
-		// Helper method to check and add indexes.
-		$results = $this->check_and_add_index( $results, 'tec_order_modifier_relationship_index_modifier_id', 'modifier_id' );
-		$results = $this->check_and_add_index( $results, 'tec_order_modifier_relationship_index_post_id', 'post_id' );
+		return [
+			self::SCHEMA_VERSION => function () use ( $table_name ) {
+				$columns   = new Column_Collection();
```
JPry, line 72: This is a little out of scope for this PR, but after looking at a few of these I can't help but think that it would be nice for `Column_Collection` (or event the base `Collection` class) to have a static method that accepts the collection items and returns the new collection. E.g. something like this:

```php
$columns = Column_Collection::create(
	[
		new ID( 'id' ),
		new Referenced_ID( 'modifier_id' ),
		new Referenced_ID( 'post_id' ),
		( new String_Column( 'post_type' ) )->set_length( 20 ),
	]
);
```

Not strictly needed, but IMO it would be a bit nicer. That said, I can understand that this but be too much out of scope for now.

src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Order_Modifiers.php
```diff
@@ -56,85 +66,31 @@ class Order_Modifiers extends Table {
...
 	 *
-	 * @since 5.20.0
-	 * @since 5.25.0 Removed the `updated_at` column.
+	 * @since TBD
 	 *
 	 * @var string[]
 	 */
-	public static function get_columns(): array {
+	public static function get_schema_history(): array {
+		$table_name = self::table_name( true );
```
JPry:
```suggestion
		$table_name = self::table_name();
```

src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Order_Modifiers_Meta.php
```diff
@@ -60,72 +68,25 @@ class Order_Modifiers_Meta extends Table {
...
-	 *
-	 * @return string The table creation SQL, in the format supported
-	 *                by the `dbDelta` function.
-	 */
-	protected function get_definition() {
-		global $wpdb;
-		$table_name      = self::table_name( true );
-		$charset_collate = $wpdb->get_charset_collate();
+	public static function get_schema_history(): array {
+		$table_name = self::table_name( true );
```
JPry:
```suggestion
		$table_name = self::table_name();
```

src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Order_Modifiers_Meta.php
```diff
@@ -60,72 +68,25 @@ class Order_Modifiers_Meta extends Table {
 	/**
 	 * An array of all the columns in the table.
 	 *
-	 * @since 5.20.0
-	 * @since 5.25.0 Removed the `updated_at` column.
+	 * @since TBD
```
JPry, line 71: Should this be `5.18.0` instead of `TBD`?

src/Tickets/Commerce/Order_Modifiers/Models/Coupon.php
```diff
@@ -42,12 +43,12 @@ class Coupon extends Order_Modifier {
 	public function get_discount_amount( float $subtotal ): float {
 		if ( 'flat' === $this->sub_type ) {
 			/** @var Float_Value $amount */
-			$amount = $this->attributes['raw_amount'];
+			$amount = Float_Value::from_number( $this->getAttribute( 'raw_amount' ) );
```
JPry, line 45: With this change, you can remove the `@var` comment above.

src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier.php
```diff
@@ -71,26 +69,37 @@ class Order_Modifier extends Model implements ModelCrud, ModelFromQueryBuilderOb
...
+	 * Acts as the constructor for the model.
 	 *
 	 * @since 1.0.0
 	 *
 	 * @param array<string,mixed> $attributes Attributes.
 	 */
-	public function __construct( array $attributes = [] ) {
-		parent::__construct( $attributes );
-
+	protected function afterConstruct(): void {
```
JPry: The docblock needs updated to reflect that there is no parameter. Also, update `@since`.

src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier.php
```diff
@@ -71,26 +69,37 @@ class Order_Modifier extends Model implements ModelCrud, ModelFromQueryBuilderOb
...
 		}
 
 		// If we don't have the raw amount set, nothing else to do.
-		if ( ! array_key_exists( 'raw_amount', $this->attributes ) ) {
+		if ( ! $this->isSet( 'raw_amount' ) ) {
 			return;
 		}
 
-		if ( $this->attributes['raw_amount'] instanceof Percent_Value ) {
+		if ( $this->getAttribute( 'raw_amount' ) instanceof Percent_Value ) {
```
JPry: The original purpose of this statement was to ensure the value internal to the object was of the correct type, e.g. a `Percent_Value` object. When we call `$this->getAttribute()`, it converts the internal type to an external type (see the innards of `getAttribute()`).

I'm not entirely sure what the equivalent would be now that the attributes have been moved into `$this->propertyCollection`. When doing step debugging, I can see at this point that the `raw_amount` item within `$this->propertyCollection->properties` is an instance of `Percent_Value`, which is the desired outcome.

If we wanted to be sure, then I suppose the right thing to do would be to check `$this->propertyCollection->get( 'raw_amount' )->getValue() instanceof Percent_Value`. Since you're more aware of what changes are going on, if you think there's something better to do, or a better place to do it, I'm open to it. It might even be feasible to remove this method entirely.

src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier.php
```diff
@@ -253,19 +224,12 @@ public function setAttribute( string $key, $value ): ModelInterface {
 					$value = Float_Value::from_number( $value );
 				}
 				break;
-
-			case 'id':
-				if ( ! $value instanceof Positive_Integer_Value ) {
-					$value = Positive_Integer_Value::from_number( $value );
```
JPry, line 259: With this removal, you can remove the import for `Positive_Integer_Value` at the top of the file.

src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier.php
```diff
@@ -71,26 +69,37 @@ class Order_Modifier extends Model implements ModelCrud, ModelFromQueryBuilderOb
...
-	 * Constructor.
+	 * Returns the properties definition for this model.
+	 *
+	 * @since 5.18.0
+	 *
+	 * @return array<string, ModelPropertyDefinition>
+	 */
+	protected static function properties(): array {
+		return [
+			'raw_amount' => ( new ModelPropertyDefinition() )->type( 'float', 'object' )->castWith( fn( $value ): Float_Value => Float_Value::from_number( $value ) ),
```
JPry: Is it possible to be more specific than `object` for the type? For instance, the `raw_amount` property should be represented internally as either a `Float_Value` or a `Percent_Value`, so it would be nice if we could let the `->type()` call accept those values.

Building on that, the type for this property depends on the `sub_type` property. So a `percent` sub-type would get a `Percent_Value` and a `flat` sub-type would get a `Float_Value`.

Speaking of `sub_type`, we should probably define that here also, because it has limited options for its value: `percent` and `flat`.

src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier.php
```diff
@@ -201,36 +197,13 @@ public function to_array(): array {
...
-	 * Validates an attribute to a PHP type.
-	 *
-	 * @since 5.18.0
-	 *
-	 * @param string $key   Property name.
-	 * @param mixed  $value Property value.
-	 *
-	 * @return bool
-	 */
-	public function isPropertyTypeValid( string $key, $value ): bool {
```
JPry, line 221: Just to confirm, this is handled by the `propertyCollection` object now, right?

src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier_Relationships.php
```diff
@@ -140,6 +125,6 @@ public static function fromQueryBuilderObject( $object ) {
 	 * @return array The object properties as an array.
 	 */
 	public function to_array(): array {
-		return $this->attributes;
+		return $this->toArray();
```
JPry, line 131: Might as well mark this method as deprecated. Same for the `Order_Modifier` class in this directory. I think `Order_Modifier` is the only one that actually calls `->to_array()` somewhere, though I could be wrong. We can just update the call to use `->toArray()`.

src/Tickets/Commerce/Order_Modifiers/Traits/Coupons.php
```diff
@@ -273,7 +273,7 @@ protected function remove_coupon_use( int $coupon_id, int $quantity = 1 ) {
 			$existing_uses = $this->get_uses_meta( $coupon_id );
 
 			// Use max() to ensure we don't set the usage to a negative number.
-			$existing_uses->meta_value = max( (int) $existing_uses->meta_value - $quantity, 0 );
+			$existing_uses->meta_value = (int) max( (int) $existing_uses->meta_value - $quantity, 0 );
```
JPry: This isn't technically needed. The `max()` function will return the original value type, so since all of the values provided to `max()` are already type `int`, `max()` can only ever return type `int`.

src/Tickets/Commerce/Tables/Webhooks.php
```diff
@@ -70,18 +77,25 @@ class Webhooks extends Table {
...
 	 * An array of all the columns in the table.
 	 *
-	 * @since 5.24.0
+	 * @since TBD
 	 *
 	 * @var string[]
 	 */
-	public static function get_columns(): array {
+	public static function get_schema_history(): array {
+		$table_name = self::table_name( true );
```
JPry:
```suggestion
		$table_name = self::table_name();
```

src/Tickets/Commerce/Ticket.php
```diff
@@ -895,7 +895,7 @@ public function delete( $event_id, $ticket_id ) {
 		}
 
 		// Delete the ticket/attendee post.
-		$delete = wp_trash_post( $ticket_id );
+		$delete = wp_delete_post( $ticket_id, true );
```
JPry, line 898: Why delete instead of trash?

src/Tickets/Flexible_Tickets/Custom_Tables/Posts_And_Ticket_Groups.php
```diff
@@ -47,7 +52,26 @@ class Posts_And_Ticket_Groups extends Table {
 	/**
 	 * {@inheritdoc}
 	 */
-	protected function get_definition() {
+	public static function get_schema_history(): array {
+		$table_name = self::table_name( true );
```
JPry:
```suggestion
		$table_name = self::table_name();
```

src/Tickets/Flexible_Tickets/Custom_Tables/Ticket_Groups.php
```diff
@@ -59,12 +66,33 @@ class Ticket_Groups extends Table {
 		'1.2.0',
 	];
 
+	/**
+	 * {@inheritdoc}
+	 */
+	public static function get_schema_history(): array {
+		$table_name = self::table_name( true );
```
JPry:
```suggestion
		$table_name = self::table_name();
```

src/Tickets/REST/TEC/V1/Traits/With_Tickets_ORM.php
```diff
@@ -26,9 +26,9 @@ trait With_Tickets_ORM {
 	 *
 	 * @since 5.26.0
 	 *
-	 * @return Tribe__Repository__Interface The repository instance.
+	 * @return Repository_Interface The repository instance.
```
JPry, line 30: The `@since` tag should be updated here too, reflecting the new return type.

src/Tickets/Seating/Tables/Layouts.php
```diff
@@ -67,18 +73,25 @@ class Layouts extends Table {
...
 	 * An array of all the columns in the table.
 	 *
-	 * @since 5.20.0
+	 * @since TBD
 	 *
 	 * @var string[]
 	 */
-	public static function get_columns(): array {
+	public static function get_schema_history(): array {
+		$table_name = self::table_name( true );
```
JPry:
```suggestion
		$table_name = self::table_name();
```

src/Tickets/Seating/Tables/Maps.php
```diff
@@ -67,16 +72,23 @@ class Maps extends Table {
...
 	 * An array of all the columns in the table.
 	 *
-	 * @since 5.20.0
+	 * @since TBD
 	 *
 	 * @var string[]
 	 */
-	public static function get_columns(): array {
+	public static function get_schema_history(): array {
+		$table_name = self::table_name( true );
```
JPry:
```suggestion
		$table_name = self::table_name();
```

src/Tickets/Seating/Tables/Seat_Types.php
```diff
@@ -68,37 +72,25 @@ class Seat_Types extends Table {
...
 	 * An array of all the columns in the table.
 	 *
-	 * @since 5.20.0
+	 * @since TBD
 	 *
 	 * @var string[]
 	 */
-	public static function get_columns(): array {
+	public static function get_schema_history(): array {
+		$table_name = self::table_name( true );
```
JPry:
```suggestion
		$table_name = self::table_name();
```

src/Tickets/Seating/Tables/Seat_Types.php
```diff
@@ -126,4 +118,23 @@ protected function get_definition() {
...
+	 *
+	 * @param string $seat_type The seat type UUID to return the number of seats for.
+	 *
+	 * @return int The number of seats for the given seat type. If the seat type does not exist, returns `0`.
+	 */
+	public function get_seats( string $seat_type ): int {
+		return (int) DB::get_var(
+			DB::prepare(
+				'SELECT seats FROM %i WHERE id = %s',
+				self::table_name( true ),
```
JPry:
```suggestion
				self::table_name(),
```

src/Tribe/Assets.php
```diff
@@ -202,7 +202,7 @@ public function enqueue_scripts() {
 	public function admin_enqueue_scripts() {
 		/** @var Tribe__Tickets__Main $tickets_main */
 		$tickets_main = tribe( 'tickets.main' );
-		$ticket_js_deps = [ 'jquery-ui-datepicker', 'tribe-bumpdown', 'tribe-attrchange', 'underscore', 'tribe-validation', 'event-tickets-admin-accordion-js', 'tribe-timepicker' ];
+		$ticket_js_deps = [ 'jquery-ui-datepicker', 'tribe-bumpdown', 'tribe-attrchange', 'underscore', 'tribe-validation', 'event-tickets-admin-accordion-js', 'tribe-timepicker', 'wp-hooks' ];
```
JPry, line 205: This is long enough now that it should span multiple lines.

src/Tribe/Metabox.php
```diff
@@ -253,7 +262,19 @@ public function get_panels( $post, $ticket_id = null, ?string $ticket_type = nul
...
+		 * @since TBD
+		 *
+		 * @param array       $panels      The panels data.
+		 * @param int|WP_Post $post        The post object or ID context of the panel rendering.
+		 * @param int|null    $ticket_id   The ID of the ticket being rendered, `null` if a new ticket.
+		 * @param string      $ticket_type The ticket type being rendered, `default` if not specified.
+		 *
+		 * @return array The panels data.
+		 */
+		return apply_filters( 'tec_tickets_panels', $panels, $post, $ticket_id, $ticket_type );
```
JPry: To be safe, we should type-case the return value.

```suggestion
		return (array) apply_filters( 'tec_tickets_panels', $panels, $post, $ticket_id, $ticket_type );
```

tests/commerce_integration/TEC/Tickets/Commerce/Gateways/Stripe/Hooks_Test.php
```diff
@@ -178,34 +178,34 @@ public function test_it_should_bail_async_stripe_webhooks_when_end_result_is_don
...
-					]
-				);
-				$ticket_id_1 = $this->create_tc_ticket( $post, 10 );
-				$order = $this->create_order_through_stripe( [ $ticket_id_1 => 1 ], [ 'order_status' => Pending::SLUG ] );
-				$_REQUEST['payment_intent'] = 'pi_123';
-				$_REQUEST['payment_intent_client_secret'] = 'pi_123_secret';
-				return [ $order->ID, false ];
-			},
-		];
+		// yield 'empty request' => [
```
JPry: Should we just remove these cases? Or is there a reason to leave them commented out?

tests/order_modifiers_integration/Checkout/Coupons_Test.php
```diff
@@ -355,7 +355,7 @@ public function it_should_update_coupon_usage() {
 				[
 					'order_modifier_id' => $coupon->id,
 					'meta_key'          => 'coupons_available',
-					'meta_value'        => $limit,
+					'meta_value'        => (string) $limit,
```
JPry: Why cast to `string` when both `string` and `int` are allowable types?

tests/order_modifiers_integration/Checkout/Coupons_Test.php
```diff
@@ -428,7 +428,7 @@ public function it_should_handle_status_transitions_without_duplicate_usages() {
...
```
JPry: Same question as above.

tests/order_modifiers_integration/Coupons/Create_Coupon_Modifiers_Test.php
```diff
@@ -61,7 +61,7 @@ public function does_table_render_correctly_with_uses() {
 						[
 							'order_modifier_id' => $modifier->id,
 							'meta_key'          => 'usage_limit',
-							'meta_value'        => 20,
+							'meta_value'        => '20',
```
JPry: Same question as the other test file

JPry:
In addition to the other comments, a few general things:

1. All imports that have been modified should be alphabetized.
2. For all of the `get_schema_history()` methods, the docblocks need updating to ensure proper return types and parameters (or lack thereof)
3. In model classes, `self::table_name( true )` should be replaced with `self::table_name()`, as the parameter is not necessary.

src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Order_Modifier_Relationships.php
dpanta94: We could introduce it in the schema library.

src/Tickets/Commerce/Order_Modifiers/Custom_Tables/Order_Modifiers_Meta.php
dpanta94: I dont think so. Its a new method.

src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier.php
dpanta94: it c an be removed completely. The model definition will now take care of that.

src/Tickets/Commerce/Order_Modifiers/Models/Order_Modifier.php
dpanta94: Yes

src/Tickets/Commerce/Ticket.php
dpanta94: This is the correct approach. Previouslyh we were trashing tickets without offering the ability to be restored. Creating orphans without any reason. Since i noticed, i fixed it.

src/Tribe/Assets.php
dpanta94: I'm not against your comment, but i am against having everyone review about their personal preferences about coding style. If this is not a phpcs violation, then its fine.

Code preferences are ok. If we are not ok with something as a team, we should create a phpcs rule and having it be enforced and ideally phpcbf should be automatically fixing those so we dont waste any energy on them.

tests/commerce_integration/TEC/Tickets/Commerce/Gateways/Stripe/Hooks_Test.php
dpanta94: Good catch. No i forgot to uncomment them

tests/order_modifiers_integration/Checkout/Coupons_Test.php
dpanta94: Good catch. forgot to update them.

src/Tickets/Commerce/Gateways/Stripe/Hooks.php
```diff
@@ -41,7 +41,7 @@ public function register() {
 	 */
 	protected function add_actions() {
 		add_action( 'rest_api_init', [ $this, 'register_endpoints' ] );
-		add_action( 'wp', [ $this, 'maybe_create_stripe_payment_intent' ] );
+		add_action( 'tec_tickets_commerce_checkout_page_parse_request', [ $this, 'maybe_create_stripe_payment_intent' ], 10000 );
```
lucatume, line 44: This looks like a change that would only affect a PHP request. This is fine if detecting, or otherwise setting up the payment intent is not required in the context of any other type of request (AJAX, REST API...)

lucatume (commented):
I've only one question about some filter move, but otherwise fine.

src/Tickets/Commerce/Gateways/Stripe/Hooks.php
dpanta94: So if you look at the callback - the first thing that is being checked is if the current page is the checkout page...

So yea this is completely safe.

PR #4037, opened 2025-09-16
TEC-5587 Add upsell message filter for seating

sdokus:
### 🎫 Ticket

[TEC-5587]

### 🗒️ Description

This PR builds on [this one](https://github.com/the-events-calendar/tribe-common/pull/2761) to add an upsell message for Seating. Seating uses StellarWP Uplink and therefore requires some changes in `/event-tickets/src/Tickets/Seating/Uplink.php` to change the `no_license_tooltip` message.

### 🎥 Artifacts 

TEC-5587-pmt-and-seating

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Seating/Uplink.php
+++ b/src/Tickets/Seating/Uplink.php
@@ -76,6 +76,7 @@ class Uplink extends Controller_Contract {
 		);
 		add_action( 'stellarwp/uplink/tec/license_field_before_input', [ $this, 'render_legend_before_input' ] );
 		add_action( 'stellarwp/uplink/tec/tec-seating/connected', [ $this, 'reset_data_on_new_connection' ] );
+		add_filter( 'stellarwp_uplink_tec_tec-seating_field_html', [ $this, 'customize_field_html' ], 10, 2 );
 	}
 
 	/**
@@ -103,6 +104,7 @@ class Uplink extends Controller_Contract {
 		);
 		remove_action( 'stellarwp/uplink/tec/license_field_before_input', [ $this, 'render_legend_before_input' ] );
 		remove_action( 'stellarwp/uplink/tec/tec-seating/connected', [ $this, 'reset_data_on_new_connection' ] );
+		remove_filter( 'stellarwp_uplink_tec_tec-seating_field_html', [ $this, 'customize_field_html' ] );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


src/Tickets/Seating/Uplink.php
```diff
@@ -174,4 +176,35 @@ public function reset_data_on_new_connection() {
...
+	 * Customize the field HTML to include upsell link for seating service.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $field_html The original field HTML.
+	 * @param object $plugin The plugin resource object.
+	 *
+	 * @return string The customized field HTML.
+	 */
+	public function customize_field_html( string $field_html, $plugin ): string {
```
redscar, line 190: Could we add a quick snapshot test for this?

src/Tickets/Seating/Uplink.php
```diff
@@ -174,4 +176,35 @@ public function reset_data_on_new_connection() {
...
+		$upsell_tooltip = sprintf(
+			/* Translators: %1$s and %2$s are opening and closing <a> tags, respectively. */
+			esc_html__( '%1$sBuy a license%2$s for the Seating service to access seating management features.', 'event-tickets' ),
+			'<a href="https://evnt.is/1bed" target="_blank">',
+			'</a>'
+		);
+
+		// Replace the default tooltip text with the upsell version.
+		$default_tooltip = esc_html__( 'A valid license key is required for support and updates', 'event-tickets' );
+		$field_html      = str_replace( $default_tooltip, $upsell_tooltip, $field_html );
```
redscar: I feel like this is a bit fragile. For example, if the text in Uplink ever changes this would break. The snapshot test could help us notice that.

We could also switch this from a str_replace to use a regex like this (untested)

```
// Build the new tooltip with translation and placeholders for links.
		$upsell_tooltip = sprintf(
			/* Translators: %1$s and %2$s are opening and closing <a> tags, respectively. */
			esc_html__(
				'%1$sBuy a license%2$s for the Seating service to access seating management features.',
				'event-tickets'
			),
			'<a href="https://evnt.is/1bed" target="_blank">',
			'</a>'
		);

		// Use regex to replace only the tooltip paragraph, not touching other HTML.
		$html = preg_replace(
			'/<p class="tooltip description">.*?<\/p>/',
			'<p class="tooltip description">' . $upsell_tooltip . '</p>',
			$html
		);

```

Camwyn: I have a different idea - put a filter in TCMN PUE. Hook into it and change the message rather than having to regex the HTML. This is the third "custom" message (Aggregator, Promoter, Seating) - we should stop doing this "the hard way" - add a filter to make this easier and slip a function into the controller (here) to hook in and inject the one for Seating. The other two can be changed later.

sdokus: @Camwyn I think I added a filter like you're talking about in [this PR](https://github.com/the-events-calendar/tribe-common/pull/2761/files#diff-ceec76148b59d82a11bde344420a50178a3beb4f17c458a83f45e707b88e44e1R107) - are you suggesting we change where this filter is so we're just tweaking the message and not the whole HTML?

Camwyn: I'm suggesting you filter just `$no_license_tooltip` [here](https://github.com/the-events-calendar/tribe-common/blob/main/src/Tribe/PUE/Checker.php) instead of the entire HTML

redscar (commented):
I see there are a few methods were we changed them to inline upsells. We need to update the `@since` for those.

PR #4157, opened 2026-03-03
ET2629 fix: enhance ticket stock management and display logic for shared capacity

larodiel:
### 🎫 Ticket

[ET-2629](https://stellarwp.atlassian.net/browse/ET-2629)

### 🗒️ Description

After sell a ticket on an event the number of the available tickets on the list was showing the total of items without subtract the tickets that were sold.
Now it has been fixed and when using "Share capacity with other tickets" it will display just the number of tickets that the user can buy.

### 🎥 Artifacts 
Screenshot 2026-03-02 at 22 18 54

### ✔️ Checklist

[ET-2629]: https://stellarwp.atlassian.net/browse/ET-2629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Tribe/Events/Views/V2/Models/Tickets.php
+++ b/src/Tribe/Events/Views/V2/Models/Tickets.php
@@ -532,12 +532,13 @@ class Tickets implements ArrayAccess, Serializable {
 			$text                           = _n( '%1$s spot left', '%1$s spots left', $stock, 'event-tickets' );
 			$this->data['stock']->available = esc_html( sprintf( $text, $number ) );
 		} else {
-		$ticket_label_singular = tribe_get_ticket_label_singular_lowercase( 'event-tickets' );
-		$ticket_label_plural   = tribe_get_ticket_label_plural_lowercase( 'event-tickets' );
-		$ticket_label          = ( 1 === $stock ) ? $ticket_label_singular : $ticket_label_plural;
-		/* translators: %1$s: Number of stock, %2$s: Ticket label */
-		$text                           = _n( '%1$s %2$s left', '%1$s %2$s left', $stock, 'event-tickets' );
-		$this->data['stock']->available = esc_html( sprintf( $text, $number, $ticket_label ) );
+			$ticket_label_singular = tribe_get_ticket_label_singular_lowercase( 'event-tickets' );
+			$ticket_label_plural   = tribe_get_ticket_label_plural_lowercase( 'event-tickets' );
+			$ticket_label          = ( 1 === $stock ) ? $ticket_label_singular : $ticket_label_plural;
+			/* translators: %1$s: Number of stock, %2$s: Ticket label */
+			$text                           = _n( '%1$s %2$s left', '%1$s %2$s left', $stock, 'event-tickets' );
+			$this->data['stock']->available = esc_html( sprintf( $text, $number, $ticket_label ) );
+		}
 	}
 
 	/**
--- a/src/Tribe/Events/Views/V2/Models/Tickets.php
+++ b/src/Tribe/Events/Views/V2/Models/Tickets.php
@@ -515,10 +515,10 @@ class Tickets implements ArrayAccess, Serializable {
 			return;
 		}
 
-		$stock = (int) $stock;
+		$stock            = (int) $stock;
 		$settings_manager = tribe( 'settings.manager' );
-		$threshold = $settings_manager::get_option( 'ticket-display-tickets-left-threshold', 0 );
-		$threshold = absint( apply_filters( 'tribe_display_tickets_left_threshold', $threshold, $data, $this->post_id ) );
+		$threshold        = $settings_manager::get_option( 'ticket-display-tickets-left-threshold', 0 );
+		$threshold        = absint( apply_filters( 'tribe_display_tickets_left_threshold', $threshold, $data, $this->post_id ) );
 
 		if ( $threshold && $stock > $threshold ) {
 			$this->data['stock']->available = '';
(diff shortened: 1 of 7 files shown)


src/Tribe/Events/Views/V2/Models/Tickets.php
redscar, line 221: Let's add another `@since` since we changed this method.

redscar:
Let's make sure the new code has tests to confirm functionality.

src/Tribe/Events/Views/V2/Models/Tickets.php
```diff
@@ -313,13 +317,14 @@ public function fetch_data() {
...
 						} else {
-							// phpcs:disable -- to suppress WordPress.WP.I18n.MismatchedPlaceholders incorrect warning.
-							/* translators: %1$s: Number of stock, %2$s: Ticket label, %3$s: Tickets label */
-							$text = _n( '%1$s %2$s left', '%1$s %3$s left', $stock, 'event-tickets' );
-							// phpcs:enable
+							// Respecting the phpcs warning for WordPress.WP.I18n.MismatchedPlaceholders.
+
+							$ticket_label = ( 1 === $stock ) ? $ticket_label_singular : $ticket_label_plural;
+							/* translators: %1$s: Number of stock, %2$s: Ticket label */
+							$text = _n( '%1$s %2$s left', '%1$s %2$s left', $stock, 'event-tickets' );
```
redscar: Passing identical strings to `_n()` is incorrect — it prevents translators from providing different singular/plural forms for the ticket label (needed in many languages). The original approach was right here; the phpcs `MismatchedPlaceholders` warning is a false positive for positional args. Restore the original strings and keep the suppression:

```php
// phpcs:disable -- to suppress WordPress.WP.I18n.MismatchedPlaceholders incorrect warning.
/* translators: %1$s: Number of stock, %2$s: Ticket label, %3$s: Tickets label */
$text = _n( '%1$s %2$s left', '%1$s %3$s left', $stock, 'event-tickets' );
// phpcs:enable
$stock_html = esc_html( sprintf( $text, $number, $ticket_label_singular, $ticket_label_plural ) );
```

The same applies to the identical `_n()` call inside `refresh_cached_stock_display()`.

src/Tribe/Events/Views/V2/Models/Tickets.php
```diff
@@ -313,13 +317,14 @@ public function fetch_data() {
...
 						}
 
-						$stock_html = esc_html( sprintf( $text, $number, $ticket_label_singular, $ticket_label_plural ) );
+						$stock_html = esc_html( sprintf( $text, $number, $ticket_label ) );
```
redscar: `$ticket_label` is only assigned inside the `else` branch, but it's passed to `sprintf()` unconditionally after the block. When `$type === 'rsvp'`, this triggers a PHP notice for an undefined variable. The original code avoided this by always passing both labels explicitly to `sprintf()`.

src/Tribe/Tickets.php
```diff
@@ -2140,13 +2140,16 @@ public static function get_ticket_counts( $post_id ) {
...
+				// For both modes we rely on the event's global stock level (remaining) added in the loop below,
+				// not each ticket's cap/total, so we skip adding here to avoid showing total capacity instead of remaining.
+				if ( Tribe__Tickets__Global_Stock::GLOBAL_STOCK_MODE === $global_stock_mode
+					|| Tribe__Tickets__Global_Stock::CAPPED_STOCK_MODE === $global_stock_mode ) {
 					$types['tickets']['global'] = 1;
 					continue;
 				}
 
-				$stock_level = Tribe__Tickets__Global_Stock::CAPPED_STOCK_MODE === $global_stock_mode ? $ticket->global_stock_cap() : $ticket->available();
+				$stock_level = $ticket->available();
```
redscar: With `CAPPED_STOCK_MODE` now hitting `continue` alongside `GLOBAL_STOCK_MODE`, all three stock modes (`OWN_STOCK_MODE`, `GLOBAL_STOCK_MODE`, `CAPPED_STOCK_MODE`) now exit the loop before reaching this block — making the `$stock_level = $ticket->available()` code unreachable. Before removing it, please confirm there are no other stock modes (e.g. from extensions or filters) that could fall through to this point. If it is truly dead, it should be removed.

src/Tribe/Events/Views/V2/Models/Tickets.php
```diff
@@ -474,6 +479,68 @@ public function exist() {
...
 	}
 
+	/**
+	 * Updates cached data's stock display (X tickets/spots left) from current counts.
+	 * Uses get_ticket_counts() so shared capacity shows the pool remaining (one number), not the sum of per-ticket.
+	 * Called when serving from cache so the number stays correct despite tribe_get_event or model cache.
+	 *
+	 * @since TBD
+	 */
+	protected function refresh_cached_stock_display(): void {
```
redscar, line 455: The threshold check, number formatting, and label assembly here duplicate the same logic in `fetch_data()`. Consider extracting it into a shared private helper (e.g., `build_stock_html( int $stock, string $type ): string`) to avoid the two implementations drifting apart.

tests/wpunit/Tribe/Tickets/Partials/V2/Tickets/__snapshots__/ItemTest__test_should_render_wit_whatever_data_attribute__1.php
```diff
@@ -1,6 +1,6 @@
 <?php return '<div
-	id="tribe-block-tickets-item-{{TICKET_ID}}"
-	 class="tribe-tickets__tickets-item post-{{TICKET_ID}} tribe_tpp_tickets type-tribe_tpp_tickets status-publish hentry" 	 data-ticket-id="{{TICKET_ID}}" data-available="false" data-has-shared-cap="false" data-ticket-price="99" data-whatever="value" >
+	id="tribe-block-tickets-item-1303{{POST_ID}}"
```
redscar: Looks like the ID slipped through?

larodiel: The `tests/wpunit/Tribe/Tickets/Partials/V2/Tickets/ItemTest.php` had an issue replacing the ID's, I had to invert the order and below why:

both str_replace calls in ItemTest.php had their search/replace arrays in the wrong order. Since auto-incremented ticket IDs often contain the post ID as a numeric substring (e.g., post 500 → ticket 1303500), replacing 500 first corrupted the ticket ID to 1303{{POST_ID}} before the 1303500 → {{TICKET_ID}} replacement could run. Swapping the order so the ticket ID is replaced first eliminates the substring collision entirely.

Replacing ticket->ID → {{TICKET_ID}} first eliminates all its occurrences from the HTML before post_id is searched, so no overlap is possible in either direction.

redscar (approved):

PR #4213, opened 2026-06-04
Prime cache when RSVP is made to keep availability accurate

sdokus:
### 🎫 Ticket

[SOFT-3428](https://linear.app/nexcess/issue/SOFT-3428/availability-amount-delayed-after-successful-rsvp)

### 🗒️ Description

After a successful RSVP, the number of available RSVPs always just shows -1 from the previous amount, regardless of how many RSVPs were just reserved. Refreshing the page then shows the correct amount. 

Upon investigation, this seems to be because `TEC\Tickets\Commerce\Attendee::get_attendees_by_ticket_id()` cached the attendee list without an expiration trigger:
```php
$cache->get( 'tec_tickets_attendees_by_ticket_id' );          // no save_post trigger
$cache->set( 'tec_tickets_attendees_by_ticket_id', $list );   // no save_post trigger
```
Every other counter (inventory(), available(), capacity()) caches with `Tribe__Cache_Listener::TRIGGER_SAVE_POST`, so they invalidate when any attendee post is saved. This method didn't — so once warmed it stayed frozen for the rest of the request, and inventory() (which counts that list) reported the stale value.

> The same untriggered cache is also primed in `Admin/Tickets/List_Table.php`; it's updated to use the same trigger so the admin priming keeps matching the read key.

### 🪑 Note for Seating tests

Making `get_attendees_by_ticket_id()` refresh within the request surfaced one pre-existing assumption in `slr_integration`: `Ajax_Test::test_remove_event_layout_success` was only passing because the stale cache hid attendees.

TLDR; Before our fix: the attendee count came back stale as 0, so inventory was `1 − 0 = 1`. After: it correctly sees the 5 attendees, so `1 − 5 = 0` (or -4, but then clamped back to 0) .

- The test reads `inventory()` **before** creating attendees (warming an empty cache), creates 10 attendees, then removes the layout.
- `remove_event_layout()` resets each ticket to **capacity 1** while leaving its 5 attendees attached, so the ticket is **oversold**. With accurate counting, `inventory()`/`available()` now correctly return **0** (previously `1`, i.e. `capacity 1 − 0 stale attendees`). The `get_ticket_counts` snapshot moves from `stock/available: 2` → `0` for the same reason.

I updated the assertions + snapshot to the truthful values (capacity & raw `stock` stay `1`; `available`/`inventory` → `0`), consistent with how any oversold ticket reports. 

### 🎥 Artifacts 

Before:
image
image

After:
CleanShot 2026-06-04 at 16 20 13@2x
(No refresh needed for accurate numbers) 

### ✔️ Checklist
- [x] We can [skip-changelog] since this is an internal bug before being released

### DIFF
--- a/src/Tickets/Admin/Tickets/List_Table.php
+++ b/src/Tickets/Admin/Tickets/List_Table.php
@@ -9,6 +9,7 @@
 
 namespace TEC\Tickets\Admin\Tickets;
 
+use Tribe__Cache_Listener;
 use Tribe__Tickets__Commerce__Currency;
 use Tribe__Tickets__Ticket_Object;
 use WP_List_Table;
@@ -1039,7 +1040,7 @@ class List_Table extends WP_List_Table {
 		// @todo @codingmusician - Add query priming solutions for other providers.
 		$ticket_ids             = wp_list_pluck( $items, 'ID' );
 		$cache                  = tribe_cache();
-		$attendees_by_ticket_id = $cache->get( 'tec_tickets_attendees_by_ticket_id' );
+		$attendees_by_ticket_id = $cache->get( 'tec_tickets_attendees_by_ticket_id', Tribe__Cache_Listener::TRIGGER_SAVE_POST );
 		if ( ! is_array( $attendees_by_ticket_id ) ) {
 			$attendees_by_ticket_id = [];
 		}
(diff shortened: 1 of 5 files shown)


dpanta94 (approved):
Very nice!

PR #4216, opened 2026-06-05
Order RSVP metabox below Tickets in classic editor [SOFT-3430]

sdokus:
Ticket: [SOFT-3430](https://linear.app/nexcess/issue/SOFT-3430/classic-editor-order-rsvp-metabox-below-tickets-metabox)

### Description

In the classic editor, the new RSVP V2 metabox rendered in the wrong place relative to the Tickets metabox. This was incidental, not intentional: RSVP V1 used a single combined metabox, and V2 splits RSVP into its own box (`tec-tickets-commerce-rsvp`).

The Tickets (`tribetickets`), RSVP, and — when Event Tickets Plus is active — Waitlist (`tec-tickets-plus-waitlist-metabox`) boxes all register into the **same `normal`/`high` group**, where WordPress orders boxes purely by registration order. Tickets and Waitlist both register at hook priority `10`, so no `add_meta_boxes` hook priority can place RSVP *between* them — anything `≤10` risks landing above Tickets, anything `>10` lands below Waitlist.

This adds an explicit reorder (`Metabox::reorder_after_tickets_metabox()`) hooked to `add_meta_boxes` at priority `100` — after all boxes are registered — that moves the RSVP box to sit **directly after the Tickets box**. The result is `Tickets → RSVP → Waitlist`. The approach is independent of the Waitlist box's own registration, so it won't break if ETP changes.

Users who have manually dragged their metaboxes are unaffected — WordPress applies their saved per-user order (`meta-box-order_*` user meta) separately, and it takes precedence over this default. This only changes the default for fresh users.

## Artifacts
Before:
CleanShot 2026-06-05 at 11 39 27

After:
CleanShot 2026-06-05 at 12 00 47 2

### Testing steps

1. With Event Tickets Plus active (so the Waitlist metabox is present), ensure RSVP V2 is active (`tickets_rsvp_version` = `v2`) and Tickets Commerce is enabled.
2. Open an event (or other ticket-enabled post type) in the **classic editor**, using an account that has never reordered its editor metaboxes.
3. Confirm the order in the main column is **Tickets → RSVP → Waitlist**.
4. With ETP inactive (no Waitlist box), confirm the order is **Tickets → RSVP**.
5. Drag to reorder the boxes, reload — confirm your custom order persists (per-user order still takes precedence).

[skip-changelog]

### DIFF
--- a/src/Tickets/RSVP/V2/Metabox.php
+++ b/src/Tickets/RSVP/V2/Metabox.php
@@ -56,6 +56,60 @@ class Metabox {
 	}
 
 	/**
+	 * Repositions the RSVP metabox to render immediately after the Tickets metabox.
+	 *
+	 * The Tickets, RSVP, and (when Event Tickets Plus is active) Waitlist metaboxes all
+	 * register into the `normal`/`high` group, where WordPress orders boxes by
+	 * registration order. That can leave another box (e.g. Waitlist) between the Tickets
+	 * and RSVP boxes. This moves the RSVP box to sit directly after the Tickets box.
+	 *
+	 * Users who have manually reordered their metaboxes are unaffected: WordPress applies
+	 * their saved per-user order separately, which takes precedence over this default.
+	 *
+	 * @since TBD
+	 *
+	 * @param string|null $post_type The post type the metaboxes are being added for.
+	 *
+	 * @return void
+	 */
+	public function reorder_after_tickets_metabox( $post_type = null ): void {
+		if ( ! in_array( $post_type, Tribe__Tickets__Main::instance()->post_types(), true ) ) {
+			return;
+		}
+
+		global $wp_meta_boxes;
+
+		if ( empty( $wp_meta_boxes[ $post_type ]['normal']['high'] ) ) {
+			return;
+		}
+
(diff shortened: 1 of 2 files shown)


claudiosanches (approved):

src/Tickets/RSVP/V2/Metabox.php
```diff
@@ -55,6 +55,60 @@ public function add( $post_type = null ) {
...
+	 * Users who have manually reordered their metaboxes are unaffected: WordPress applies
+	 * their saved per-user order separately, which takes precedence over this default.
+	 *
+	 * @since TBD
+	 *
+	 * @param string|null $post_type The post type the metaboxes are being added for.
+	 *
+	 * @return void
+	 */
+	public function reorder_after_tickets_metabox( $post_type = null ): void {
```
dpanta94, line 75: Does that respect user's preferences @sdokus ? If someone forcefully changes where he likes their metaboxes ?

sdokus: It does respect user preferences! I had to test with a completely new user to make sure it just makes the default right below the tickets metabox.

PR #4188, opened 2026-05-27
[SOFT-3342] RSVP V2: Not Going attendees still decrement ticket inventory

sdokus:
## Summary

In the RSVP Merge project, attendees marked **Not Going** are counted toward reserved spots across the user-facing surfaces — even though [Strategy Doc Flow](https://docs.google.com/document/d/1prnq6XJrwSjIppM6Oda3q6h7HXYQPCtX8ygXh2-c2qA/edit?tab=t.edf19i359mxq) specifies that a Not Going RSVP creates a hidden order (so the user is tracked) but does **not** hold a seat.

Linear: [SOFT-3342](https://linear.app/nexcess/issue/SOFT-3342) (this PR also folds in [SOFT-3345](https://linear.app/nexcess/issue/SOFT-3345/rsvp-v2-stock-and-available-still-decrement-for-not-going-rsvps))

## Artifacts 
Before:

https://github.com/user-attachments/assets/feb2aaf7-9ff2-468a-bca4-c069870dbb49

SOFT-3342-Before-Attendee-Page

After: 

https://github.com/user-attachments/assets/bd46dcaa-a818-4021-874a-48089912f484

## Root cause

PR #4132 added the `$type` parameter and Not Going short-circuit to `Tribe__Tickets__Tickets::attendee_decreases_inventory()` — but only `inventory()` ever calls that filter. The other three surfaces above read `stock()`, `qty_sold()`, or `available()`, which all reflect the underlying `_stock` / qty_sold post meta decremented at TC order creation.

Compounding that, the Tickets Commerce `Module::attendee_decreases_inventory()` override accepts `$type` and silently drops it. Since RSVP V2 is served by Tickets Commerce, even `inventory()` was being computed against the un-filtered code path — Dimi's #4132 intent never reached V2 RSVPs.

### 1. Module override honors `$type`

`src/Tickets/Commerce/Module.php` — adds an import for the RSVP V2 Constants and a `tc-rsvp` short-circuit before delegating to `Attendee::decreases_inventory()`. Mirrors the abstract base method's logic.

Fixes `inventory()` and the "X remaining" number in the RSVP block.

### 2. `stock()` and `qty_sold()` adjustments for `tc-rsvp`

- `src/Tribe/Ticket_Object.php` — adds `tribe_tickets_ticket_stock` and `tribe_tickets_ticket_qty_sold` filters at the end of the two getter methods. Only fired on getter calls (setter pattern preserved). Both default to no-op for non-RSVP tickets.
[...]

PR #4221, opened 2026-06-09
[SOFT-3342] Follow-up: centralize get_tickets cache invalidation + revert stray Stripe change

sdokus:
Linear: [SOFT-3342](https://linear.app/nexcess/issue/SOFT-3342)

Follow-up to #4188 addressing [Dimi's review comments](https://github.com/the-events-calendar/event-tickets/pull/4188#discussion).

## Description

**Cache invalidation** (Dimi's comment on `Models/Tickets.php` — "we have an issue in our cache invalidation… this only fixes the issue once instead of for all"):

- The `get_tickets()` cache key was being reconstructed by hand in `regenerate_caches()`, duplicating the format defined privately inside `get_tickets()`. Added `Tribe__Tickets__Tickets::get_tickets_cache_key()` as the single source of truth and routed `get_tickets()`, `clear_ticket_cache_for_post()`, and the new invalidator through it, so the key can no longer drift.
- Added `Tribe__Tickets__Tickets::clean_tickets_cache( $event_id )` — a cross-provider, lightweight invalidator. `regenerate_caches()` now calls it instead of poking the cache key directly.
- **Root of the "only fixes it once" problem:** stock/sales live as ticket post meta updated via `update_post_meta()`, which does **not** fire `clean_post_cache` — the only signal `regenerate_caches()` was wired to. So after an order/RSVP, nothing invalidated the cache for other readers. Hooked `tec_tickets_ticket_stock_changed` / `tec_tickets_ticket_stock_added` to regenerate the connected Event caches, so counts stay fresh for **every** reader, not just the path that happens to clean the post cache first.

Remove unintended changes in Stripe Endpoint:

- Reverted the user-0 nonce swap (and its test) to the `main` branch version. It was unrelated to the RSVP Not Going fix and leaked in because I was switching branches too much 🤪 

## Testing

1. Create an event with a ticket (capacity 10).
2. Prime the tickets cache (load the event / call `get_tickets()`).
3. Place an order or change `_stock`.
4. Without clearing any cache by hand, confirm the "remaining"/"available" count is fresh on the next read (block, attendees admin, calendar/list view).
5. Confirm the Stripe return endpoint still verifies the state nonce as on `main`.

[skip-changelog]

### DIFF
--- a/src/Tickets/Commerce/Gateways/Stripe/REST/Return_Endpoint.php
+++ b/src/Tickets/Commerce/Gateways/Stripe/REST/Return_Endpoint.php
@@ -49,14 +49,7 @@ class Return_Endpoint extends Abstract_REST_Endpoint {
 			return false;
 		}
 
-		// Nonces are created as user 0 in Signup; verify the same way even when the
-		// admin returns with WordPress auth cookies on this REST callback.
-		$user_id = get_current_user_id();
-		wp_set_current_user( 0 );
-		$verified = wp_verify_nonce( $response->nonce, tribe( WhoDat::class )->get_state_nonce_action() );
-		wp_set_current_user( $user_id );
-
-		return $verified;
+		return wp_verify_nonce( $response->nonce, tribe( WhoDat::class )->get_state_nonce_action() );
 	}
 
 	/**
--- a/src/Tribe/Events/Views/V2/Models/Tickets.php
+++ b/src/Tribe/Events/Views/V2/Models/Tickets.php
@@ -171,14 +171,13 @@ class Tickets implements ArrayAccess, Serializable {
 			}
 			/** @var array<int> $connected_event_ids */
 			$connected_event_ids = array_merge( ...$connected_event_ids );
+			$tribe_cache         = tribe_cache();
 
 			foreach ( $connected_event_ids as $connected_event_id ) {
 				// Reset the `Tribe__Tickets__Tickets::get_tickets` method cache to get the last version of them.
 				$provider = Tickets_Tickets::get_event_ticket_provider_object( $connected_event_id );
 
 				if ( $provider ) {
-					$tribe_cache = tribe_cache();
-
 					$tribe_cache[ Tickets_Tickets::get_tickets_cache_key( $provider->orm_provider, $connected_event_id ) ] = null;
 				}
 
--- a/src/Tribe/Events/Views/V2/Models/Tickets.php
+++ b/src/Tribe/Events/Views/V2/Models/Tickets.php
@@ -125,11 +125,9 @@ class Tickets implements ArrayAccess, Serializable {
 			$provider = Tickets_Tickets::get_event_ticket_provider_object( $post->ID );
 
 			if ( $provider ) {
-				$tribe_cache       = tribe_cache();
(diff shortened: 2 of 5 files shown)


src/Tribe/Events/Views/V2/Models/Tickets.php
```diff
@@ -173,17 +171,15 @@ public static function regenerate_caches( int $post_id ): void {
...
 
 			foreach ( $connected_event_ids as $connected_event_id ) {
 				// Reset the `Tribe__Tickets__Tickets::get_tickets` method cache to get the last version of them.
 				$provider = Tickets_Tickets::get_event_ticket_provider_object( $connected_event_id );
 
 				if ( $provider ) {
-					$orm_provider                      = $provider->orm_provider;
-					$tickets_cache_key                 = "{$tickets_class}::get_tickets-{$orm_provider}-{$connected_event_id}";
-					$tribe_cache[ $tickets_cache_key ] = null;
+					$tribe_cache = tribe_cache();
```
rahulacaleffi: @sdokus , overall looks good to me.

Just one small question/nit, since it's a cache related situation:
Since `tribe_cache()` is a singleton, do we need to have the `$tribe_cache = tribe_cache();` here or could it be outside (right above) the loop/foreach?

src/Tickets/Commerce/Gateways/Stripe/REST/Return_Endpoint.php
```diff
@@ -49,14 +49,7 @@ public function has_permission( WP_REST_Request $request ) {
...
 
-		// Nonces are created as user 0 in Signup; verify the same way even when the
-		// admin returns with WordPress auth cookies on this REST callback.
-		$user_id = get_current_user_id();
-		wp_set_current_user( 0 );
-		$verified = wp_verify_nonce( $response->nonce, tribe( WhoDat::class )->get_state_nonce_action() );
-		wp_set_current_user( $user_id );
-
-		return $verified;
+		return wp_verify_nonce( $response->nonce, tribe( WhoDat::class )->get_state_nonce_action() );
```
dpanta94, line 52: This seems like an out of scope change ?

tests/commerce_integration/TEC/Tickets/Commerce/Gateways/Stripe/REST/Return_Endpoint_Test.php
```diff
@@ -181,42 +181,6 @@ public function should_reject_unauthorized_credential_overwrite(): void {
 		}
 	}
 
-	/**
-	 * @test
-	 */
-	public function should_accept_user_zero_nonce_when_authenticated_admin_returns_from_stripe(): void {
```
dpanta94, line 187: Also probably out of scope change ?

dpanta94 (changes requested):

src/Tickets/Commerce/Gateways/Stripe/REST/Return_Endpoint.php
sdokus: This (and the tests below) is reverting the changes I was proposing for that SVUL that accidentally leaked in here because I was switching between branches and didn't mean to include it 😞

dpanta94 (approved):

PR #4240, opened 2026-06-11
fix(SMTNC-1279): speed up ticketed/unticketed admin list counts

almirsantos99:
### 🎫 Ticket

[SMTNC-1279](https://linear.app/nexcess/issue/SMTNC-1279/slow-queries-get-ticketed-count-and-get-unticketed-count)

### 🎥 Artifacts
https://www.loom.com/share/940e67d3a64b4f4682482506d6ffb0ee

### 🗒️ Description

**Type of change:** Bug fix / performance (non-breaking).

The **Ticketed** and **Unticketed** filter-link counts at the top of the Events
admin list (and other ticket-enabled post-type lists) were generating slow
queries that scaled with the total number of posts. On a client site with ~27K
posts the admin list took **~96s** to load (`get_unticketed_count()` ~88s,
`get_ticketed_count()` ~4s); with TEC/ET disabled it loaded in ~2.7s.

**Root cause**

- `get_ticketed_count()` joined on `pm.meta_value = CONCAT( p.ID, '' )`, which
  wraps the indexed `wp_posts.ID` in a function. MySQL could not use the
  primary-key index, so it fell back to a hash join / scan of `wp_posts` —
  cost growing with the post count.
- `get_unticketed_count()` used `WHERE p.ID NOT IN ( <ticketed subquery> )`, a
  materialized anti-join that also scaled with the number of posts.

**Changes**

- `get_ticketed_count()`: join on `p.ID = pm.meta_value` so the query resolves
  via the `wp_posts` **PRIMARY** key (`eq_ref`, 1 row per ticket meta). The
  result is memoized per post type, since it's requested twice per list render.
- `get_unticketed_count()`: derive it as `total − ticketed` using a single
  index-served `COUNT(*)` (the total) plus the already-memoized ticketed count,
  replacing the `NOT IN` materialized subquery entirely.

The displayed counts are unchanged; only the execution path is faster.

**For reviewers**

This cannot be validated by page-load time on a small site — with few posts both
the old and new queries return in well under a millisecond (the dataset fits in
the buffer pool). The reliable signal at any size is the query's `EXPLAIN` plan
(exposed per-query by Query Monitor): after the fix, `wp_posts` is accessed via
`eq_ref` on `PRIMARY` (1 row) for ticketed, and unticketed is a single
index-served `COUNT(*)` with no `NOT IN` / materialized subquery. See the
Testing Instructions for the step-by-step `EXPLAIN` comparison.

### ✔️ Checklist
- [X] Are all the **required** tests passing?
- [X] Automated code review comments are addressed.
- [X] Check the base branch for your PR.
- [X] Add your PR to the project board for the release.

### DIFF
--- a/changelog/fix-SMTNC-1279-slow_queries
+++ b/changelog/fix-SMTNC-1279-slow_queries
@@ -1,4 +1,4 @@
 Significance: patch
 Type: performance
 
-Improve the performance of the Ticketed and Unticketed filter counts on admin post lists, which could take tens of seconds to load on sites with a large number of posts.
+Improve the performance of the Ticketed and Unticketed filter counts on admin post lists. [SMTNC-1279]
--- a/changelog/fix-SMTNC-1279-slow_queries
+++ b/changelog/fix-SMTNC-1279-slow_queries
@@ -1,0 +1,4 @@
+Significance: patch
+Type: performance
+
+Improve the performance of the Ticketed and Unticketed filter counts on admin post lists, which could take tens of seconds to load on sites with a large number of posts.
--- a/src/Tribe/Query.php
+++ b/src/Tribe/Query.php
@@ -13,6 +13,18 @@ class Tribe__Tickets__Query {
 	public static $has_tickets = 'tribe-has-tickets';
 
 	/**
+	 * Per-request memoization of the ticketed count, keyed by post type.
+	 *
+	 * The ticketed count is requested twice per list render (directly, and again to derive the unticketed
+	 * count), so caching it avoids running the same query twice on a single page load.
+	 *
+	 * @since TBD
+	 *
+	 * @var array<string,int>
+	 */
+	protected $ticketed_count_cache = [];
+
+	/**
 	 *  Hooks to add query vars and filter the post query.
 	 */
 	public function hook() {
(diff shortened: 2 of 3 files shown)


WayneRocha:
@almirsantos99 This invalidates https://github.com/the-events-calendar/event-tickets/pull/4215. Because this is not only a fix for both tickets, but also an upgrade in performance.
Maybe we should keep only your ticket.

WayneRocha:
@almirsantos99 In case keep yours also add an additional changelog please. Can be that https://github.com/the-events-calendar/event-tickets/pull/4215/changes#diff-1002af0c683ee15852b8c7cf3f3e8a73f889d050bf5e1ab1c13e7f7538368acc

src/Tribe/Query.php
redscar, line 137: Add an `@since`.

redscar, line 188: Update `@since`.

src/Tribe/Query.php
```diff
@@ -188,31 +194,26 @@ public function get_unticketed_count( string $post_type ): int {
...
-					WHERE p.ID NOT IN (
-						SELECT p.ID FROM $wpdb->posts p
-									JOIN $wpdb->postmeta pm ON ( $meta_keys_in ) AND pm.meta_value = CONCAT( p.ID, '' )
-									WHERE p.post_type = %s
-					) AND p.post_type = %s AND p.post_status NOT IN ('auto-draft', 'trash')",
-				$post_type,
-				$post_type
+			$total = (int) $wpdb->get_var(
+				$wpdb->prepare(
+					"SELECT COUNT(*) FROM $wpdb->posts WHERE post_type = %s AND post_status NOT IN ('auto-draft', 'trash')",
```
redscar, line 209: Going from the old query to the new query is a pretty big change. What was your method of making sure we didn't create any regressions with different scenarios?

nikolaystrikhar (changes requested):
Provide an artifact

changelog/fix-SMTNC-1279-slow_queries
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: performance
+
+Improve the performance of the Ticketed and Unticketed filter counts on admin post lists, which could take tens of seconds to load on sites with a large number of posts.
```
redscar:
```suggestion
Improve the performance of the Ticketed and Unticketed filter counts on admin post lists. [SMTNC-1279]
```
Your original changelog entry wasn't bad. I think we could go simpler though. We also have to add the ticket number to the end of the changelog entry.

redscar (changes requested):

src/Tribe/Query.php
almirsantos99: @redscar as we talked in Slack:
"The old query was overbuilt because it is just used in admin > events page so it didn't need no meta_key checking and 3 join queries, it just need to get all the posts that are not draft and reduce from the $this->get_ticketed_count, actually it improve the workflow and with a simpler query, it does reduces the chances of something happen"

src/Tribe/Query.php
```diff
@@ -148,20 +148,26 @@ public function get_ticketed_count( string $post_type ): int {
...
+			 * (one row per ticket), so resolving each to its event via the `wp_posts` primary key (`p.ID = pm.meta_value`)
+			 * is fast. The previous `pm.meta_value = CONCAT( p.ID, '' )` form wrapped the indexed `p.ID`, forcing a
+			 * full scan of `wp_posts` and making this query slow on sites with many posts.
 			 */
 			// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
 			$query = $wpdb->prepare(
-				"SELECT COUNT(DISTINCT(p.ID)) FROM $wpdb->posts p
-				  JOIN $wpdb->postmeta pm ON ( $meta_keys_in ) AND pm.meta_value = CONCAT( p.ID, '' )
-				  WHERE p.post_type = %s AND p.post_status NOT IN ('auto-draft', 'trash')",
+				"SELECT COUNT(DISTINCT pm.meta_value) FROM $wpdb->postmeta pm
```
dpanta94: Why count the pm.meta_value and not the p.ID ?

dpanta94: And probably this should become an INNER JOIN so we avoid counting orphaned meta

dpanta94: Please do create a test where you create orhpaned meta and assert that they are not counted for.

src/Tribe/Query.php
dpanta94: This is a clever switch. Nice.

dpanta94 (changes requested):

src/Tribe/Query.php
almirsantos99: the JOIN will bring only parent that have child or child that have parent, no orphans data, but either way I changed to Left join so we can see it clearly, same with the pm.meta_value to p.ID
Thanks @dpanta94

dpanta94 (approved):

src/Tribe/Query.php
dpanta94: You mean INNER, thats fine :D

redscar (approved):

nikolaystrikhar (approved):

PR #4218, opened 2026-06-05
Show Going/Not Going status + hide check-in for declined RSVPs on attendee report [SOFT-3432]

sdokus:
**Ticket:** [SOFT-3432](https://linear.app/nexcess/issue/SOFT-3432/rsvp-v2-attendee-report-doesnt-show-goingnot-going-status-or-hide)

## Description

On the **Attendees report** for an event with a TC RSVP, attendees who selected "Not going" were indistinguishable from those going, and they were still offered a **Check In**:

- The **Status** column showed **"Completed"** for everyone (all RSVP attendees share a "Completed" order status; the going/not-going answer lives in the `_tec_tickets_commerce_rsvp_status` attendee meta, which the table never read).
- The **Check In** control appeared for all attendees, including those who declined.

The top **Attendance Overview** panel already shows the correct RSVPs Total/Going/Not Going breakdown — this PR only fixes the per-attendee table rows.

## Artifacts
Before:
CleanShot 2026-06-05 at 13 45 27@2x

After:
CleanShot 2026-06-05 at 13 30 44@2x

## Testing steps

1. Open the Attendees report for an event with a TC RSVP that has both "going" and "not going" responses.
2. **Status** column shows **"Going"** / **"Not Going"** per attendee (not "Completed").
3. **Check In** control is hidden for "not going" attendees and present for "going".
4. Hover row actions for "not going" attendees no longer include Check In / Undo Check In.
5. Non-RSVP (paid ticket) attendees are unchanged.

[skip-changelog]

PR #4243, opened 2026-06-12
Backfill RSVP V2 test coverage (metabox order, responses info, attendee report)

sdokus:
Follow-up test coverage requested on three merged RSVP V2 PRs:
- [SOFT-3430](https://linear.app/nexcess/issue/SOFT-3430) — metabox order (#4216)
- [SOFT-3431](https://linear.app/nexcess/issue/SOFT-3431) — responses count + View Attendees link (#4217)
- [SOFT-3432](https://linear.app/nexcess/issue/SOFT-3432) — Going/Not Going status + hide check-in (#4218)

## Description

Test-only follow-up — no production code changes. Backfills coverage for three RSVP V2 features that merged into `bucket/rsvp-merge` without dedicated tests.

- **`Metabox::reorder_after_tickets_metabox` (#4216):** verifies the RSVP box is moved to sit directly after the Tickets box, is a no-op when it already follows Tickets, and bails for non-ticketable post types or when a box is missing. (Reviewer asked whether reordering respects user preferences — the function only adjusts the default registration order; WordPress applies each user's saved order separately at render. These tests lock in the default-order behavior.)
- **`Metabox::display_responses_info` (#4217):** verifies it renders the response count + View Attendees link, shows the "not going" tooltip only when `show_not_going` is enabled, and renders nothing for a non-RSVP ticket type, a missing RSVP id, or zero responses.
- **Attendees report modifiers (#4218):** `modify_status_display` (Going / Not Going pill), `modify_checkin_display` (check-in hidden for not-going), and `modify_row_actions` (check-in action dropped for not-going), plus attendee-id resolution via the `ID` fallback and non-RSVP passthrough.

## Testing steps

1. `slic run rsvp_v2_integration RSVP/V2/Metabox_Test.php`
2. `slic run rsvp_v2_integration RSVP/V2/Attendees_Test.php`
3. Both suites should pass.

[skip-changelog]

### DIFF
--- a/tests/rsvp_v2_integration/RSVP/V2/Attendees_Test.php
+++ b/tests/rsvp_v2_integration/RSVP/V2/Attendees_Test.php
@@ -136,89 +136,104 @@ class Attendees_Test extends WPTestCase {
 		}
 	}
 
-	public function test_exclude_rsvp_tickets_from_tickets_view_data_link_count_returns_args_unchanged_when_context_is_not_get_my_tickets_link_data(): void {
-		$post_id = static::factory()->post->create();
-		$this->create_tc_rsvp_ticket( $post_id );
+	/**
+	 * Each fixture returns [ $args, $post_id, $user_id, $context ]; the flag is whether the
+	 * RSVP-exclusion filter should be added for that combination.
+	 */
+	public static function exclude_rsvp_tickets_data_provider(): array {
+		return [
+			'unchanged for non-matching context' => [
+				function () {
+					$post_id = static::factory()->post->create();
+					$this->create_tc_rsvp_ticket( $post_id );
 
-		$attendees = tribe( Attendees::class );
-		$args      = [ 'by' => [ 'event' => $post_id ] ];
+					return [ [ 'by' => [ 'event' => $post_id ] ], $post_id, null, 'other_context' ];
+				},
+				false,
+			],
 
-		$result = $attendees->exclude_rsvp_tickets_from_tickets_view_data_link_count( $args, $post_id, null, 'other_context' );
+			'unchanged for null context' => [
+				function () {
+					$post_id = static::factory()->post->create();
+					$this->create_tc_rsvp_ticket( $post_id );
 
-		$this->assertEquals( $args, $result );
-		$this->assertArrayNotHasKey( 'meta_not_equals', $result['by'] );
-	}
+					return [ [ 'by' => [ 'event' => $post_id ] ], $post_id, null, null ];
+				},
+				false,
+			],
 
-	public function test_exclude_rsvp_tickets_from_tickets_view_data_link_count_returns_args_unchanged_when_context_is_null(): void {
-		$post_id = static::factory()->post->create();
-		$this->create_tc_rsvp_ticket( $post_id );
+			'adds filter for tc provider' => [
+				function () {
+					$post_id = static::factory()->post->create();
+					$this->create_tc_rsvp_ticket( $post_id );
 
-		$attendees = tribe( Attendees::class );
-		$args      = [ 'by' => [ 'event' => $post_id ] ];
+					return [ [ 'by' => [ 'event' => $post_id ] ], $post_id, null, 'get_my_tickets_link_data' ];
+				},
+				true,
+			],
 
-		$result = $attendees->exclude_rsvp_tickets_from_tickets_view_data_link_count( $args, $post_id, null, null );
+			'preserves existing args' => [
+				function () {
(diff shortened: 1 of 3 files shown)


tests/rsvp_v2_integration/RSVP/V2/Attendees_Test.php
```diff
@@ -233,4 +233,130 @@ public function test_get_rsvp_attendees_by_id_bails_when_attendees_already_filte
...
+			? $this->create_not_going_tc_rsvp_attendees( 1, $ticket_id, $post_id )[0]
+			: $this->create_going_tc_rsvp_attendees( 1, $ticket_id, $post_id )[0];
+
+		return [
+			'ticket_type' => Constants::TC_RSVP_TYPE,
+			$id_key       => $attendee_id,
+		];
+	}
+
+	public function test_modify_status_display_returns_label_unchanged_for_non_rsvp_item(): void {
```
jonwaldstein: @sdokus have you ever used a testing data provider?  you could probably condense alot of these tests with a data provider 😄

jonwaldstein (approved):
@sdokus looks good, I left a suggestion but no need to hold up a merge

tests/rsvp_v2_integration/RSVP/V2/Metabox_Test.php
```diff
@@ -10,13 +10,37 @@
...
+		$this->wp_meta_boxes_backup = $wp_meta_boxes;
+	}
+
+	public function tearDown(): void {
+		global $wp_meta_boxes;
+		$wp_meta_boxes = $this->wp_meta_boxes_backup;
+
+		parent::tearDown();
+	}
+
```
dpanta94: Have you tested if you actually need this ? I think you dont.

tests/rsvp_v2_integration/RSVP/V2/Attendees_Test.php
```diff
@@ -136,101 +136,310 @@ public function test_get_rsvp_attendees_by_id_with_order_id( Closure $fixture ):
...
 
-	public function test_get_rsvp_attendees_by_id_bails_when_attendees_already_filtered(): void {
-		$post_id   = static::factory()->post->create();
-		$ticket_id = $this->create_tc_rsvp_ticket( $post_id );
-		$this->create_order( [ $ticket_id => 3 ] );
+	/**
+	 * @dataProvider modify_row_actions_data_provider
+	 */
+	public function test_modify_row_actions( Closure $fixture, bool $keeps_checkin ): void {
+		$item      = Closure::bind( $fixture, $this, self::class )();
```
dpanta94:
```suggestion
		$item      = $fixture();
```

I think that should work just fine as well ?

dpanta94 (changes requested):

tests/rsvp_v2_integration/RSVP/V2/Attendees_Test.php
sdokus: Good suggestion, I refactored to do that and made a note to try to do that more in the future!

tests/rsvp_v2_integration/RSVP/V2/Attendees_Test.php
sdokus: $fixture() on its own would actually fatal here — these provider closures call instance helpers like $this->make_rsvp_item() / $this->create_tc_rsvp_ticket(), and since the providers are static, the closures have no $this bound. Closure::bind($fixture, $this, self::class) is what gives them $this (and private-method scope). All the bound providers in this file touch $this, so I think this one has to stay, but I'm happy to switch if you've got a cleaner pattern 👀

tests/rsvp_v2_integration/RSVP/V2/Metabox_Test.php
sdokus: You're right, I tested and it didn't make a difference if it was there or not so I removed it 🙏

tests/rsvp_v2_integration/RSVP/V2/Attendees_Test.php
dpanta94: So why not make the providers non-static ?

sdokus: Okay for my own learning: while using `Closure:bind` technically functions and would work, the simpler approach is to just remove the `static` before each function so that the scope of `$this` does not have to be rebound to each function.

dpanta94 (approved):
Exactly! What i meant! Nice.

PR #4247, opened 2026-06-17
[SOFT-3334] Frontend: IAC-aware RSVP flow and TC-RSVP plumbing

larodiel:
## Summary

- IAC-aware frontend RSVP: route going/not-going through ARI when IAC is active
- Output `data-iac` on RSVP wrapper for JS detection
- TC-RSVP ticket discovery via repository; fallback admin list label

Part 1 of 2 in the SOFT-3334 ET stack. Pair with ET+ #2074 for IAC save QA.

**Downstream:** `split/soft-3334-et-02-block-iac` (stacked on this branch)

🎥 Artifacts
https://www.loom.com/share/e1b2566ae5a0443f831b2fcbd24b0956

PR #4248, opened 2026-06-17
[SOFT-3334] Block editor: RSVP IAC state and attendee collection UI

larodiel:
## Summary

- Redux `iac` state with SET_RSVP_IAC action, selector, and REST thunks
- Attendee collection accordion + IAC radio in RSVP block settings (ET+ gated)
- ARF iframe postMessage sync and localStorage bridge for pre-create saves
- TC-RSVP ticket discovery in block editor

Part 2 of 2 in the SOFT-3334 ET stack. Full E2E requires ET+ #2076 merged.

**Upstream:** #4247
**ET+ companion:** event-tickets-plus #2076

🎥 Artifacts
https://www.loom.com/share/e1b2566ae5a0443f831b2fcbd24b0956

## Test plan

- [ ] `npm run test:jest -- rsvp`
- [ ] `vendor/bin/codecept run rsvp_v2_integration Rsvp_Block_Test`
- [ ] Block editor: inline IAC accordion saves via REST
- [ ] Block editor: ARF modal sets IAC before ticket exists, persists on first save
- [ ] Frontend IAC flow still works (from #4247)


src/modules/blocks/rsvp-v2/action-dashboard/container.js
```diff
@@ -51,6 +51,21 @@ const onCancelClick = ( state, dispatch ) => () => {
...
 const onConfirmClick = ( state, dispatch ) => () => {
+	const postId = select( 'core/editor' ).getCurrentPostId();
+
+	// The IAC admin modal (iac-admin.js) writes to localStorage when in block editor
+	// context since the ticket may not yet exist. Read it here to include the selected
+	// IAC value in the REST call, then the thunk clears it after a successful save.
+	let iac = selectors.getRSVPIAC( state );
+	if ( postId ) {
+		try {
+			const stored = JSON.parse( localStorage.getItem( 'tec_rsvp_attendee_info_' + postId ) || 'null' );
```
dpanta94: Why do we need localStorage though ? As i understrand it we are still in the same screen. We don't need to maintain a value accross page refreshes (thats what localStorage does). Actually maintaining this storage across refreshes could result in weird things where we include data from the user's older sessions for the same post ?

i would go with appropriately exposing the variable, since its available and calculated already.

dpanta94 (changes requested):

dpanta94 (approved):

PR #4250, opened 2026-06-17
[SMTNC-1519] fix: Enhance license management

larodiel:
### 🎫 Ticket

[SMTNC-1519](https://linear.app/nexcess/issue/SMTNC-1519/2percent-tec-fee-still-applies-on-stripe-after-etp-activation)

### 🗒️ Description

Fixes the 2% Tickets Commerce application fee continuing to apply on Stripe (and Square) after Event Tickets Plus is activated with a valid license.

The issue was in Settings::is_licensed_plugin(): when $revalidate was true, it still returned a stale cached false from before ETP was licensed, so Application_Fee::calculate() never saw the updated license status.

This PR bypasses the cache when revalidating, clears the cache when ETP is activated or its license changes, and stops forcing a revalidation API call on every checkout.

### 🎥 Artifacts 
https://www.loom.com/share/0bbd7a4d24034f12b5d465c1fdb23458

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Commerce/Gateways/Square/Application_Fee.php
+++ b/src/Tickets/Commerce/Gateways/Square/Application_Fee.php
@@ -50,7 +50,7 @@ class Application_Fee {
 			return Value::create();
 		}
 
-		if ( Settings::is_licensed_plugin( true ) ) {
+		if ( Settings::is_licensed_plugin() ) {
 			return Value::create();
 		}
 
--- a/src/Tickets/Commerce/Gateways/Square/Application_Fee.php
+++ b/src/Tickets/Commerce/Gateways/Square/Application_Fee.php
@@ -50,7 +50,7 @@ class Application_Fee {
 			return Value::create();
 		}
 
-		if ( Settings::is_licensed_plugin() ) {
+		if ( Settings::is_licensed_plugin( true ) ) {
 			return Value::create();
 		}
 
--- a/src/Tickets/Commerce/Hooks.php
+++ b/src/Tickets/Commerce/Hooks.php
@@ -105,6 +105,12 @@ class Hooks extends Service_Provider {
 		add_action( 'admin_menu', tribe_callback( Orders_Page::class, 'add_orders_page' ), 15 );
 
 		add_action( 'tec_tickets_commerce_async_webhook_process', [ $this, 'route_async_webhook_process' ], 10, 2 );
+
+		add_action( 'tribe_settings_save', [ $this, 'clear_licensed_plugin_cache' ] );
+		add_action( 'update_option_pue_install_key_event_tickets_plus', [ $this, 'clear_licensed_plugin_cache' ] );
+		add_action( 'stellarwp/uplink/tec/connected', [ $this, 'maybe_clear_licensed_plugin_cache_on_uplink_change' ] );
+		add_action( 'stellarwp/uplink/tec/disconnected', [ $this, 'maybe_clear_licensed_plugin_cache_on_uplink_change' ] );
+		add_action( 'activated_plugin', [ $this, 'maybe_clear_licensed_plugin_cache_on_plugin_activation' ], 10, 2 );
 	}
 
 	/**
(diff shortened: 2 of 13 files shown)


src/Tickets/Commerce/Gateways/Square/Application_Fee.php
```diff
@@ -50,7 +50,7 @@ public static function calculate( Value $value ) {
 			return Value::create();
 		}
 
-		if ( Settings::is_licensed_plugin( true ) ) {
+		if ( Settings::is_licensed_plugin() ) {
```
redscar, line 53: Pulling revalidate=true out of Application_Fee::calculate() means we lean on PUE's own revalidation now - do we know how often that runs, and are we comfortable that a lapsed key keeps waiving the 2% fee until it does? The same applies on the Stripe side.

src/Tickets/Commerce/Hooks.php
```diff
@@ -105,6 +105,13 @@ protected function add_actions() {
 		add_action( 'admin_menu', tribe_callback( Orders_Page::class, 'add_orders_page' ), 15 );
 
 		add_action( 'tec_tickets_commerce_async_webhook_process', [ $this, 'route_async_webhook_process' ], 10, 2 );
+
+		add_action( 'tribe_settings_save', [ $this, 'clear_licensed_plugin_cache' ] );
+		add_action( 'update_option_pue_install_key_event_tickets_plus', [ $this, 'clear_licensed_plugin_cache' ] );
+		add_action( 'stellarwp/uplink/tec/connected', [ $this, 'maybe_clear_licensed_plugin_cache_on_uplink_change' ] );
+		add_action( 'stellarwp/uplink/tec/disconnected', [ $this, 'maybe_clear_licensed_plugin_cache_on_uplink_change' ] );
+		add_action( 'activated_plugin', [ $this, 'maybe_clear_licensed_plugin_cache_on_plugin_change' ], 10, 2 );
+		add_action( 'deactivated_plugin', [ $this, 'maybe_clear_licensed_plugin_cache_on_plugin_change' ], 10, 2 );
```
redscar, line 114: Can we get some additional tests on the different hooks to make sure they work properly?

src/Tickets/Commerce/Hooks.php
```diff
@@ -1187,4 +1194,56 @@ public function route_async_webhook_process( $order_id, $retry = 0 ): void {
...
+	public function maybe_clear_licensed_plugin_cache_on_uplink_change( $plugin ): void {
+		if ( ! is_object( $plugin ) || ! method_exists( $plugin, 'get_slug' ) ) {
+			return;
+		}
+
+		if ( 'event-tickets-plus' !== $plugin->get_slug() ) {
+			return;
+		}
+
+		Settings::clear_licensed_plugin_cache();
```
redscar: I think I would prefer we use `did_action( 'tec_tickets_plus_fully_loaded' )` to check if ET+ is activated. Is there a specific reason you went this way?

src/Tickets/Commerce/Settings.php
```diff
@@ -636,32 +640,55 @@ public static function is_licensed_plugin( $revalidate = false ) {
 			return false;
 		}
 
-		$pue = tribe( \Tribe__Tickets_Plus__PUE::class );
+		$pue       = tribe( \Tribe__Tickets_Plus__PUE::class );
+		$cache_key = self::get_licensed_plugin_cache_key();
+
+		if ( $pue->is_current_license_valid( $revalidate ) ) {
+			set_transient( $cache_key, wp_json_encode( true ), HOUR_IN_SECONDS );
```
redscar: Is the transient doing anything now? Since nothing passes `revalidate=true` anymore, the live check runs before we read the cache and a cached true never gets read back - is there a case I'm missing, or could this collapse down to return $pue->is_current_license_valid( $revalidate );?

redscar:
Can you also check that the `@since` params are set in each docblock that you changed? Make sure they are either set to `TBD` or `5.28.4.1` since we know what release this is going out in.

src/Tickets/Commerce/Hooks.php
```diff
@@ -1187,4 +1194,56 @@ public function route_async_webhook_process( $order_id, $retry = 0 ): void {
...
+	 * Clears the license cache when Event Tickets Plus Uplink status changes.
+	 *
+	 * @since 5.28.5
+	 *
+	 * @param object $plugin The Uplink plugin resource.
+	 *
+	 * @return void
+	 */
+	public function maybe_clear_licensed_plugin_cache_on_uplink_change( $plugin ): void {
+		if ( ! is_object( $plugin ) || ! method_exists( $plugin, 'get_slug' ) ) {
```
redscar: This bails on `! is_object( $plugin )` , but the disconnected action passes the slug as a string, not the resource object - so on disconnect we never reach clear_licensed_plugin_cache(). Am I reading that right? If so, the fee keeps getting waived for the hour until the transient expires.

redscar (approved):

PR #4257, opened 2026-06-23
fix: render RSVP V2 attendee name on confirmation [SOFT-3632]

sdokus:
## Ticket
[SOFT-3632](https://linear.app/nexcess/issue/SOFT-3632)

## Description

After a successful RSVP, the "Your RSVPs" confirmation list showed each attendee row as just the ticket label **"RSVP"** instead of the guest's name (e.g. "RSVP / RSVP" for two guests).

Each row renders two lines — `attendee/name.php` (guest name) and `attendee/rsvp.php` (ticket title, literally "RSVP"). RSVP V2 creates Tickets Commerce attendees (`tec_tc_attendee`) and stores the holder name under `_tec_tickets_commerce_full_name`, but `name.php` only read the legacy `_tribe_rsvp_full_name` key. That key is empty on V2 attendees, so `name.php` bailed early and only the "RSVP" label rendered.

This was reproducible **without** Event Tickets Plus, so it is not an IAC/individual-attendee-collection gap — the name is captured, it just wasn't being read from the right place.

The fix reads the Tickets Commerce meta key when the legacy key is empty, with a final fallback to the attendee post title.

## Artifacts
Before:
CleanShot 2026-06-23 at 14 07 41@2x

After:
CleanShot 2026-06-23 at 14 37 47@2x

## Testing

1. Activate Event Tickets.
2. Add an RSVP ticket to an event.
3. On the front end, RSVP as "Going" for 2 guests and submit.
4. In the "Your RSVPs" confirmation list, confirm each row now shows the guest name above the "RSVP" ticket label (instead of only "RSVP").

[skip-changelog]

### DIFF
--- a/src/views/v2/commerce/rsvp/attendees/attendee/name.php
+++ b/src/views/v2/commerce/rsvp/attendees/attendee/name.php
@@ -16,9 +16,12 @@
  * @var int $attendee_id The attendee ID.
  *
  * @since 5.7.1
+ * @since TBD Read the attendee name from the Tickets Commerce meta key used by RSVP V2, with fallbacks.
  *
- * @version 5.7.1
+ * @version TBD
  */
+
+use TEC\Tickets\Commerce\Attendee;
 
 defined( 'ABSPATH' ) || die();
 if ( empty( $attendees ) || empty( $attendee_id ) ) {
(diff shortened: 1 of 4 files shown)


dpanta94 (changes requested):
Shouldn't we accompany the change along with a test case ? it could be a snapshot of course. Or even better a provider returning closures and the test case itself doing snapshot assertions

larodiel (approved):

dpanta94 (approved):

PR #4260, opened 2026-06-24
[SMTNC-1646] PHP Compatibility improvements

dpanta94:
### 🎫 Ticket

SMTNC-1646

### 🗒️ Description

Improves php compatibility for php version from 7.4 up to 8.4.

Breaking contract by adding explicitly nullable `?` is not a breaking change.

PHPCompatibility and PHPCompatibilityWP are included via the `sh` script because the latest stable version of PHPCompatibilityWP does not cover 8+ compatibility concerns.

### 🎥 Artifacts 
Tests continue passing are part of the artifact.

Test ensuring that the cotract being broken is not considered a breaking change.

CleanShot 2026-06-25 at 00 33 40

### ✔️ Checklist

### DIFF
--- a/.github/workflows/php-compatibility.yml
+++ b/.github/workflows/php-compatibility.yml
@@ -1,0 +1,39 @@
+name: 'PHP Compatibility'
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: php-compat-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  php-compat:
+    # Allow opting out on a specific PR with [skip-php-compat] in the description.
+    if: "!contains(github.event.pull_request.body, '[skip-php-compat]')"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.3'
+          coverage: none
+          tools: composer
+
+      # Cache the downloaded phpcs phar + PHPCompatibility/PHPCSUtils standards.
+      # Key is bound to bin/php-compat.sh, so bumping the pinned versions there
+      # automatically invalidates the cache.
+      - name: Cache PHP compatibility toolchain
+        uses: actions/cache@v4
+        with:
+          path: dev/php-compat
+          key: php-compat-${{ runner.os }}-${{ hashFiles('bin/php-compat.sh') }}
+
+      # Runs `bash bin/php-compat.sh`, which scans the paths in phpcompat.xml.dist
+      # against testVersion 8.0-8.4 and exits non-zero on any finding (hard gate).
+      - name: Run PHP 8 compatibility scan
+        run: composer compat
--- a/.github/workflows/php-compatibility.yml
+++ b/.github/workflows/php-compatibility.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout the repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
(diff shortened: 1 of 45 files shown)


src/Tickets/Commerce/Gateways/Stripe/Gateway.php
```diff
@@ -254,13 +254,14 @@ public function filter_stripe_currency_precision( $currency_data, $currency_code
 	 * Get the appropriate precision for Stripe based on their currency requirements.
 	 *
 	 * @since 5.26.7
+	 * @since TBD Made $default_precision explicity nullable.
```
defunctl, line 257:
```suggestion
	 * @since TBD Made $default_precision explicitly nullable.
```

src/Tickets/Commerce/Reports/Attendees.php
```diff
@@ -433,7 +433,7 @@ public function maybe_generate_csv() {
 
 			// Output the lines into the file.
 			foreach ( $items as $item ) {
-				fputcsv( $output, $item, $delimiter ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fputcsv
+				fputcsv( $output, $item, $delimiter, '"', '\\' ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fputcsv
```
defunctl, line 436: As mentioned in the repo, I think this should actually be an empty string?

There are a few other places in the PR this is happening that would need to be changed as well.

https://www.php.net/manual/en/function.fputcsv.php

src/Tickets/Flexible_Tickets/Series_Passes/Base.php
```diff
@@ -570,7 +570,7 @@ public function filter_out_series_type_tickets_from_order_report( $include, $tic
 	 *
 	 * @return null|bool The filtered HTML, or `false` to hide the option.
 	 */
-	public function hide_header_image_option_from_ticket_settings( string $html = null, string $file, array $name, Template $template, array $context ): ?bool {
+	public function hide_header_image_option_from_ticket_settings( ?string $html = null, string $file = '', array $name = [], ?Template $template = null, array $context = [] ): ?bool {
```
defunctl, line 574: Just flagging these as it's a signature change, we're 100% sure all subclasses have been updated?

src/Tribe/Events/Views/V2/Models/Tickets.php
```diff
@@ -28,7 +27,7 @@
  *
  * @package Tribe\Tickets\Events\Views\V2\Models
  */
-class Tickets implements ArrayAccess, Serializable {
+class Tickets implements ArrayAccess {
```
defunctl, line 30: My only concern with this is if anything is doing a `$whatever instance of Serializable` conditional check, it would now fail. This class seems to properly delegate to the magic methods, though, so that's good.

defunctl (changes requested):

src/Tribe/Events/Views/V2/Models/Tickets.php
dpanta94: We are not doing that in our codebase, so i think this is an okay possibly breaking change since the chance of actually happening is very very low.

src/Tickets/Flexible_Tickets/Series_Passes/Base.php
dpanta94: CleanShot 2026-06-26 at 16 58 30

made it a not breaking change, good catch!

dpanta94: So, i had to revert the change because of php compat failures. We are not extending that class ever, and i highly doubt any integration externally is.

.github/workflows/php-compatibility.yml
```diff
@@ -0,0 +1,39 @@
...
+  cancel-in-progress: true
+
+jobs:
+  php-compat:
+    # Allow opting out on a specific PR with [skip-php-compat] in the description.
+    if: "!contains(github.event.pull_request.body, '[skip-php-compat]')"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v4
```
estevao90:
```suggestion
        uses: actions/checkout@v6
```

defunctl (approved):

PR #4267, opened 2026-06-26
fix: prevent infinite WP-Cron loop in Glance_Items attendee count [SMTNC-1631]

WayneRocha:
### 🎫 Ticket

[SMTNC-1631](https://linear.app/nexcess/issue/SMTNC-1631/tec-tickets-update-glance-item-attendee-counts-causes-infinite-wp-cron)

### PR related

https://github.com/the-events-calendar/event-tickets-plus/pull/2079

### 🗒️ Description

_Speedup: ~2,500×_

**Bug:** `tec_tickets_update_glance_item_attendee_counts` WP-Cron event reschedules itself on every dashboard page load on high-volume sites, causing PHP timeouts and server instability.

**Root cause (two bugs):**

1. **Performance:** `update_attendee_count()` called `get_attendees_by_args([])` which hydrated every attendee as a full PHP object — 94 MB / 17 s for 12,983 attendees. PHP `max_execution_time` was hit before `set_transient()` could be called.
2. **Zero-count loop:** The old `if (empty($total)) return;` guard skipped `set_transient()` when the count was zero, leaving the transient permanently unset on fresh installs — same infinite cron loop.

**Fixes:**

- Replaced `get_attendees_by_args([])` with `tribe_attendees()->per_page(1)->found()` — issues a single `SQL_CALC_FOUND_ROWS LIMIT 1` query; MySQL counts server-side, PHP receives one integer. Benchmark: 0.008 s / 0 MB extra memory.
- Always call `set_transient()` with `(int) $total`, even when zero.
- Added `tec_tickets_glance_item_attendee_count_enabled` filter so high-volume sites can disable the glance item entirely.

### 🎥 🗒️ Benchmark / Artifacts
**Benchmark (12,983 attendees, local):**

image

image

image

benchmarks can run via `wp smtnc-1631 benchmark` QA helper plugin (included in QA instructions on Linear ticket).

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Admin/Glance_Items.php
+++ b/src/Tickets/Admin/Glance_Items.php
@@ -102,13 +102,14 @@ class Glance_Items {
 	 * @since 5.6.0
 	 * @since TBD Replace full object hydration with a single COUNT query via the Repository.
 	 * @since TBD Always persist the transient (even when count is zero) to prevent infinite cron rescheduling.
+	 * @since TBD Exclude RSVP "not going" attendees via the `rsvp_status__or_none` Repository filter.
 	 */
 	public function update_attendee_count() {
 		if ( ! $this->is_enabled() ) {
 			return;
 		}
 
-		$total = tribe_attendees()->per_page( 1 )->found();
+		$total = tribe_attendees()->where( 'rsvp_status__or_none', 'yes' )->per_page( 1 )->found();
 
 		set_transient( static::$attendee_count_key, (int) $total, DAY_IN_SECONDS );
 	}
--- a/src/Tickets/Admin/Glance_Items.php
+++ b/src/Tickets/Admin/Glance_Items.php
@@ -37,7 +37,7 @@ class Glance_Items {
 	 *
 	 * @return bool Whether the glance item attendee count is enabled. Default true.
 	 */
-	protected function is_enabled(): bool {
+	private function is_enabled(): bool {
 		/**
 		 * Filters whether the attendee count glance item is enabled.
 		 *
--- a/src/Tickets/Admin/Glance_Items.php
+++ b/src/Tickets/Admin/Glance_Items.php
@@ -102,9 +102,9 @@ class Glance_Items {
 		 * @since TBD
 		 *
 		 * @param bool $enabled Whether the glance item attendee count is enabled. Default true.
-	 	 * @param array $items The array of items to be displayed. Default empty array.
-
-	 	 * @return bool $enabled Whether the glance item attendee count is enabled.
+		 * @param array $items The array of items to be displayed. Default empty array.
+		 *
+		 * @return bool $enabled Whether the glance item attendee count is enabled.
 		 */
 		if ( ! apply_filters( 'tec_tickets_glance_item_attendee_count_enabled', true ) ) {
 			return;
--- a/src/Tickets/Admin/Glance_Items.php
+++ b/src/Tickets/Admin/Glance_Items.php
@@ -28,6 +28,29 @@ class Glance_Items {
 	protected static string $attendee_count_key = 'tec_tickets_glance_item_attendees_count';
 
 	/**
+	 * Check if the attendee count glance item is enabled.
+	 *
+	 * Return false to disable the count entirely (no cron scheduled, no display).
+	 * Useful on high-volume sites where even the transient-backed display is unwanted.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool Whether the glance item attendee count is enabled. Default true.
+	 */
+	protected function is_enabled(): bool {
+		/**
+		 * Filters whether the attendee count glance item is enabled.
+		 *
+		 * @since TBD
+		 *
+		 * @param bool $enabled Whether the glance item attendee count is enabled. Default true.
+		 *
+		 * @return bool
+		 */
+		return apply_filters( 'tec_tickets_glance_item_attendee_count_enabled', true );
+	}
+
+	/**
 	 * Method to register glance items related hooks.
 	 *
 	 * @since 5.5.10
@@ -47,19 +70,7 @@ class Glance_Items {
 	 * @return array $items The maybe modified array of items to be displayed.
 	 */
 	public function custom_glance_items_attendees( $items = [] ): array {
-		/**
-		 * Filters whether the attendee count glance item is enabled.
-		 *
-		 * Return false to disable the count entirely (no cron scheduled, no display).
-		 * Useful on high-volume sites where even the transient-backed display is unwanted.
-		 *
-		 * @since TBD
-		 *
-		 * @param bool $enabled Whether the glance item attendee count is enabled. Default true.
-		 *
-		 * @return bool
-		 */
-		if ( ! apply_filters( 'tec_tickets_glance_item_attendee_count_enabled', true ) ) {
+		if ( ! $this->is_enabled() ) {
 			return (array) $items;
 		}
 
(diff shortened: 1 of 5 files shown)


src/Tickets/Admin/Glance_Items.php
```diff
@@ -67,15 +89,29 @@ public function custom_glance_items_attendees( $items = [] ): array {
...
+		 * Useful on high-volume sites where even the transient-backed display is unwanted.
+		 *
+		 * @since TBD
+		 *
+		 * @param bool $enabled Whether the glance item attendee count is enabled. Default true.
+		 * @param array $items The array of items to be displayed. Default empty array.
+		 *
+		 * @return bool $enabled Whether the glance item attendee count is enabled.
+		 */
+		if ( ! apply_filters( 'tec_tickets_glance_item_attendee_count_enabled', true ) ) {
```
dpanta94: Instead of using the filter twice maybe wrap it in a method ?

src/Tickets/Admin/Glance_Items.php
```diff
@@ -67,15 +89,29 @@ public function custom_glance_items_attendees( $items = [] ): array {
...
 			return;
 		}
 
-		set_transient( static::$attendee_count_key, $total, DAY_IN_SECONDS );
+		$total = tribe_attendees()->per_page( 1 )->found();
```
dpanta94: Is that working well with different providers ? aka RSVP or WooCommerce ?

dpanta94: Probably add some test cases for WC provider in ETP - It's ok to open a PR just to add test coverage even when no production changes are required.

dpanta94 (changes requested):

src/Tickets/Admin/Glance_Items.php
WayneRocha: Yes, now that you mentioned, I searched and this is also made in `TEC\Tickets\Admin\Tickets\List_Table` and `TEC\Tickets\Admin\Tickets\Page` for example.
Addressed in aa54d4009c0aa1f7b4da0be5002aebe0a756a410

src/Tickets/Admin/Glance_Items.php
WayneRocha: `tribe_attendees()` involves all attendee post types (RSVP and Commerce), except if I pass a param to it, so it works normal for all by default.
As recommendation of yours I added additional test coverage for ETP here in https://github.com/the-events-calendar/event-tickets-plus/pull/2079

src/Tickets/Admin/Glance_Items.php
```diff
@@ -22,6 +27,29 @@ class Glance_Items {
...
+	 * Check if the attendee count glance item is enabled.
+	 *
+	 * Return false to disable the count entirely (no cron scheduled, no display).
+	 * Useful on high-volume sites where even the transient-backed display is unwanted.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool Whether the glance item attendee count is enabled. Default true.
+	 */
+	protected function is_enabled(): bool {
```
dpanta94: Why not private ?

tests/integration/TEC/Tickets/Admin/Glance_ItemsTest.php
```diff
@@ -0,0 +1,149 @@
...
+	 */
+	public function setUp(): void {
+		parent::setUp();
+
+		delete_transient( self::TRANSIENT_KEY );
+		wp_clear_scheduled_hook( self::CRON_HOOK );
+
+		add_filter( 'tribe_tickets_post_types', static function () {
+			return [ 'post' ];
+		} );
```
dpanta94: Adding a filter here will make it leak in other tests.

Either add it inside the testcase so the framework can reset the state, or manually remove it.

tests/integration/TEC/Tickets/Admin/Glance_ItemsTest.php
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	/**
+	 * Clean up after each test.
+	 */
+	public function tearDown(): void {
+		// your tear down methods here
+
+		// then
+		parent::tearDown();
```
dpanta94: If you dont need the method don't use it.

tests/integration/TEC/Tickets/Admin/Glance_ItemsTest.php
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	/**
+	 * update_attendee_count() stores the correct attendee total in the transient.
+	 *
+	 * @test
+	 */
+	public function update_attendee_count_sets_transient_to_real_count_when_attendees_exist(): void {
```
dpanta94, line 68: I'd like to see a test case where you create attendees for RSVPs and Tickets over not a single event events, but lets say at least 3.

Also utilize the RSVP - not going flag.

What happens then ?

If oyu delete an event that has attendees are those being counted ?

dpanta94 (changes requested):

tests/integration/TEC/Tickets/Admin/Glance_ItemsTest.php
WayneRocha: IA probably copied from this [pattern](https://github.com/search?q=repo%3Athe-events-calendar%2Fevent-tickets+%2F%2F+your+tear+down+methods+here++%2F%2F+then&type=code). Removed it: 30306afc12f652e13a118d7009193594aa759850

src/Tickets/Admin/Glance_Items.php
WayneRocha: So maybe it could be overrided if needed was my thought. But thinking that the filter already allows to extend or modify it can be private (is private the first prefered option[question]?)
Updated 2eae67d55a04431355cd83fa6d63ad9bc7c318e8

tests/integration/TEC/Tickets/Admin/Glance_ItemsTest.php
WayneRocha: Added additional tests to cover these in 45cb4d3352f6c5988824676cf70beb274db9ae9b

src/Tickets/Admin/Glance_Items.php
dpanta94: Yes, private is the preferred. We should only expose APIs when its required that includes protected method that can be overwritten when someone extends our class.

tests/integration/TEC/Tickets/Admin/Glance_ItemsTest.php
```diff
@@ -0,0 +1,239 @@
...
+		// One "going" and one "not going" attendee.
+		$this->create_attendee_for_ticket( $ticket_id, $post_id, [ 'rsvp_status' => 'yes' ] );
+		$this->create_attendee_for_ticket( $ticket_id, $post_id, [ 'rsvp_status' => 'no' ] );
+
+		tribe( Glance_Items::class )->update_attendee_count();
+
+		$stored = get_transient( self::TRANSIENT_KEY );
+
+		$this->assertNotFalse( $stored, 'Transient should be set when attendees exist.' );
+		$this->assertSame( 2, (int) $stored, 'Both "going" and "not going" RSVP attendees should be included in the count.' );
```
dpanta94: This seems like a logical result to me, but i would discuss it with @leahkoerper.

if she disagrees and she thinks that `not-going` attendees should not be counted, we should create a ticket to fix in a future maintenance

dpanta94 (approved):

WayneRocha:
Leah think is better to not count the "Not Going" ones.
Performance degraded a little bit because of the filter on `postmeta`, but nothing we can do about it, it's way more acceptable than the performance we had before.

I updated the benchmark:

image

[436ae7a](https://github.com/the-events-calendar/event-tickets/pull/4267/commits/436ae7a8fb386fd4183bbbd183298baf19eeb3d1)

dpanta94 (approved):

PR #4271, opened 2026-06-26
Fix fatal when deleting an attendee

sdokus:
## Ticket 
None right now, but I can add one if necessary. 

## Description

Deleting an attendee from the Attendees table fatals with an uncaught `TypeError`.

`Attendees_Table::do_delete()` calls `delete_ticket( null, $attendee_id )` — `null` post ID by design. That `null` flows into `clear_ticket_cache_for_post()` → `get_tickets_cache_key()`, whose strict `int $post_id` hint throws.

I introduced this new fatal behavior when I tried centralizing the cache key. The old inline string concatenation tolerated `null`; the new typed method did not. This PR should adjust the code so that it can tolerate null again.

### DIFF
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -760,6 +760,11 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 		 * @param int $post_id The post ID.
 		 */
 		public function clear_ticket_cache_for_post( $post_id ) {
+			// No post context (e.g. deleting an attendee) means there is nothing to clear.
+			if ( empty( $post_id ) ) {
+				return;
+			}
+
 			/** @var Tribe__Cache $cache */
 			$cache = tribe( 'cache' );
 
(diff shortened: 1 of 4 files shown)


src/Tribe/Tickets.php
```diff
@@ -854,13 +854,14 @@ public function get_tickets( $post_id, string $context = null ) {
...
+		 * @param string   $orm_provider The provider ORM slug (e.g. `tribe-commerce`, `rsvp`).
+		 * @param int|null $post_id      The post (event) ID the tickets belong to. May be null when the
+		 *                               caller has no post context (e.g. deleting an attendee).
 		 *
 		 * @return string The cache key.
 		 */
-		public static function get_tickets_cache_key( string $orm_provider, int $post_id ): string {
-			return self::class . '::get_tickets-' . $orm_provider . '-' . $post_id;
+		public static function get_tickets_cache_key( string $orm_provider, ?int $post_id ): string {
+			return self::class . '::get_tickets-' . $orm_provider . '-' . (int) $post_id;
```
larodiel: @sdokus , as `$post_id` is optional, shouldn't we check it?

larodiel (changes requested):

src/Tribe/Tickets.php
sdokus: Good call, we should address the root of the problem not just a band aid 🙃

larodiel (approved):

src/Tribe/Tickets.php
```diff
@@ -760,6 +760,11 @@ public function post_has_tickets( $post ) {
 		 * @param int $post_id The post ID.
 		 */
 		public function clear_ticket_cache_for_post( $post_id ) {
```
nikolaystrikhar, line 762: Such things should be covered with tests. You probably caught it manually? That's nice, but would be better to cover this exact cases (or all cases if there is nothing for this function) in tests.

nikolaystrikhar (changes requested):
Explain the situation with tests please

nikolaystrikhar (approved):

nikolaystrikhar:
@sdokus The code looks good, but there is a failed test and I'm not sure if it's related. Maybe makes sense to fix separately.

sdokus:
> @sdokus The code looks good, but there is a failed test and I'm not sure if it's related. Maybe makes sense to fix separately.

I think you're right, it looks like the same test is failing in the sanity check - I'll merge this and then look into that test failure in a new PR 🫡

PR #4281, opened 2026-06-30
Merge Forward RSVP-Bucket

sdokus:
Primary goal here is to merge the `bucket/rsvp-merge` forward with the changes merged into `main` yesterday. 

Additionally, I've added changes in `src/Tribe/RSVP.php` to fix the failing `wpunit` tests caused by this root cause: 

- A test commit (cd268c0b70, May 29) changed these cases to expect false. They passed at the time by accident: htmlentities() ran with PHP 8.1's default ENT_QUOTES, which encoded the script payload's single quotes into &#039;; the ; made is_email() reject it.
- A later "php compatibility" commit (d20e98fb54, Jun 25) added explicit ENT_COMPAT to that call. ENT_COMPAT leaves single quotes alone, so the sanitized payload scriptalert'hello'/script@test.com passed is_email() and the method returned an array — breaking the security test.
- The V2 REST endpoint had already been hardened with an explicit "reject markup" guard (e39994f683), but the legacy Tribe/RSVP.php method never received it. The merge ported the test expectation without the production hardening.

I also added some additional tests to cover this better moving forward. 

[skip-changelog][skip-phpcs]

### DIFF
--- a/src/Tickets/Commerce/Gateways/Stripe/Payment_Intent.php
+++ b/src/Tickets/Commerce/Gateways/Stripe/Payment_Intent.php
@@ -245,6 +245,40 @@ class Payment_Intent {
 	}
 
 	/**
+	 * Determines whether a Payment Intent's amount matches a local order's total.
+	 *
+	 * The Payment Intent ID is supplied by the client, so this guards against binding or
+	 * completing an order with a Payment Intent that was created for a different cart and a
+	 * different (e.g. lower) amount. Only the charge amount is compared: the application fee
+	 * is the platform's commission and is not part of the buyer's payment authorization, and
+	 * intermediate statuses legitimately report a zero captured amount.
+	 *
+	 * @since 5.28.5.1
+	 *
+	 * @param array    $payment_intent Payment intent data from Stripe.
+	 * @param \WP_Post $order          The local Tickets Commerce order.
+	 *
+	 * @return bool
+	 */
+	public static function is_valid_for_order( array $payment_intent, \WP_Post $order ): bool {
+		if ( empty( $payment_intent['id'] ) || ! empty( $payment_intent['errors'] ) ) {
+			return false;
+		}
+
+		if ( empty( $order->total_value ) || ! $order->total_value instanceof Value ) {
+			return false;
+		}
+
+		$formatter = new Gateway_Value_Formatter( tribe( Gateway::class ) );
+		$value     = $formatter->format( $order->total_value );
+
+		$expected_amount = (string) $value->get_integer();
+		$actual_amount   = isset( $payment_intent['amount'] ) ? (string) $payment_intent['amount'] : '';
+
+		return $expected_amount === $actual_amount;
+	}
+
+	/**
 	 * Calls the Stripe API and returns an existing Payment Intent.
 	 *
 	 * @since 5.3.0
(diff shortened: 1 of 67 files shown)


src/Tickets/Commerce/Gateways/Stripe/Payment_Intent.php
```diff
@@ -244,6 +244,40 @@ public static function is_valid_for_cart( array $payment_intent, Cart $cart ): b
...
+	 * @param \WP_Post $order          The local Tickets Commerce order.
+	 *
+	 * @return bool
+	 */
+	public static function is_valid_for_order( array $payment_intent, \WP_Post $order ): bool {
+		if ( empty( $payment_intent['id'] ) || ! empty( $payment_intent['errors'] ) ) {
+			return false;
+		}
+
+		if ( empty( $order->total_value ) || ! $order->total_value instanceof Value ) {
```
larodiel, line 268: @sdokus is `$order->total_value` always created?, or worth we to check `isset` first?
Another thing in this condition, maybe we should parse `empty( $order->total_value )` to `int`, because if for some reason we receive a string such as `0,00` it will return false. Maybe I'm overengineering, but I just want to make sure.

src/Tribe/RSVP.php
```diff
@@ -3170,8 +3170,16 @@ public function parse_attendee_details() {
 		}
 		// phpcs:enable WordPress.Security.NonceVerification.Missing
 
-		$attendee_email        = empty( $first_attendee['email'] ) ? null : htmlentities( sanitize_email( html_entity_decode( $first_attendee['email'], ENT_COMPAT ) ), ENT_COMPAT );
-		$attendee_email        = is_email( $attendee_email ) ? $attendee_email : null;
+		$attendee_email = null;
+		if ( ! empty( $first_attendee['email'] ) ) {
```
larodiel, line 3174: how about to change it to `filter_var($first_attendee['email'], FILTER_VALIDATE_EMAIL)`, is this change worth?
Then we just will enter on the condition if it's a real email.

larodiel (dismissed):

src/Tickets/Commerce/Gateways/Stripe/Payment_Intent.php
sdokus: Good questions, both are actually covered here:

On `isset` — we don't need it because `empty()` (same as `isset()`) doesn't throw an "`undefined property`" warning when the property isn't set; it just returns `true`. So `empty( $order->total_value )` already safely handles the "not created" case on its own.

On the `0,00` string — that can't reach this comparison: `total_value` is handled earlier on to always be a `Value` object, so I think it's safe to assume it will never be a string 😄

src/Tribe/RSVP.php
sdokus: I'd lean toward keeping it as is — `filter_var( FILTER_VALIDATE_EMAIL )` would actually lose two behaviors this block needs:

1. Entity decoding — we `html_entity_decode()` first so an entity-encoded but valid address (e.g. `foo&#64;bar.com`) still passes. `filter_var` on the raw input would reject it.
2. Markup rejection — that's the whole reason for this change. The `$decoded_email === wp_strip_all_tags( $decoded_email )` check deliberately rejects input with HTML/script markup instead of silently sanitizing it into a valid-looking email. `filter_var` alone wouldn't reproduce that, and `is_email()` already handles final validation — so the swap wouldn't add coverage, just change (and weaken) the flow 🤔

larodiel: Thank you for the explanation @sdokus :)

dpanta94 (approved):

PR #4286, opened 2026-07-01
[RSVP] Part 5/22 - Shared Data Store: Utility Functions

larodiel:
## Part 5/19

These files add **utility functions** to the `rsvp-shared` data store:

- `hydrate-rsvp-attendance-counts.js` — Derive going/not-going counts from attendee list
- `hydrate-rsvp-from-ticket.js` — Convert ticket data to RSVP store state
- `normalize-rsvp-response.js` — Normalize API responses to store shape
- `dispatch-rsvp-details.js` — Dispatch detail payloads to Redux
- `select-latest-rsvp-ticket.js` — Selector for latest RSVP ticket from state
- Tests for hydration and selection

---

> **Previous:** [split/soft-3335-04-data-shared-redux](https://github.com/the-events-calendar/event-tickets/pull/4285)  
> **Next:** [split/soft-3335-06-data-rsvp-migrate](https://github.com/the-events-calendar/event-tickets/pull/4287)


src/modules/data/blocks/rsvp-shared/utils/hydrate-rsvp-from-ticket.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {Object}   ticket   Ticket REST response object.
+ * @param {Object}   options  Normalization options passed to normalizeRSVPResponseFromV2Ticket.
+ * @return {boolean} Whether the ticket was hydrated.
+ */
+export const hydrateRsvpFromTicket = ( dispatch, actions, ticket, options = {} ) => {
+	if ( ! ticket?.id ) {
+		return false;
+	}
+
+	const normalized = normalizeRSVPResponseFromV2Ticket( ticket, {
```
sdokus, line 26: Will this overwrite the title/description if one does exist? Should we add a check for "if there's a title/description, use those, if not use these defaults"?

src/modules/data/blocks/rsvp-shared/utils/normalize-rsvp-response.js
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * Resolves RSVP capacity from a V2 ticket REST payload.
+ *
+ * @param {Object} rsvp Ticket REST response object.
+ * @return {number|string} Capacity value or empty string for unlimited.
+ */
+const resolveV2Capacity = ( rsvp ) => {
+	if ( rsvp.stock_mode === 'unlimited' ) {
+		return '';
+	}
```
sdokus: I'm seeing `stock_mode` in `/wp-json` but not in the actual `/wp-json/tec/v1/tickets` endpoint - are we sure that that's a viable endpoint to use for this check? Should we also rely on a check for if `stock === -1` or something like that?

src/modules/data/blocks/rsvp-shared/utils/hydrate-rsvp-from-ticket.js
```diff
@@ -0,0 +1,44 @@
...
+		title: 'RSVP',
+		description: '',
+		...options,
+	} );
+
+	dispatch( actions.createRSVP() );
```
sdokus, line 32: Tickets and RSVP V1's reducer/hydration path includes IAC handling, but we seem to drop that for the V2/shared path - is this intentional? 🤔

larodiel: Yes @sdokus we still having the IAC in V2
https://linear.app/nexcess/issue/SOFT-3334/update-iacarf-modal-and-support-iac-block-editor
We still given the support to that on block editor.
Is that your question or I get it wrong? 🤔

src/modules/data/blocks/rsvp-shared/utils/hydrate-rsvp-from-ticket.js
larodiel: on [Miro](https://miro.com/app/board/uXjVIRz-XSE=/?moveToWidget=3458764627986418152&cot=10) the block doesn't have any description or title

sdokus: YUP okay that makes sense, I was thinking about Tickets, but RSVPs never have titles or descriptions.

src/modules/data/blocks/rsvp-shared/utils/hydrate-rsvp-from-ticket.js
sdokus: I think we might have lost the IAC support somewhere between SOFT-3334 and this PR 👀 I'm struggling to phrase things correctly, but this is what my Claude is finding: 

> SOFT-3334 built working IAC support into the original unified reducer → SOFT-3335 split that reducer into V1's (rsvp/reducer.js) and V2's shared piece (rsvp-shared/reducer.js) → V1 kept its copy, but the case never got carried over into the shared one, so V2 lost it while V1 didn't.

larodiel:
@sdokus thank you, I did some modifications on it

sdokus:
> @sdokus thank you, I did some modifications on it

Looking good! The IAC looks solid, though I think the capacity is still off — `Number('')` evaluates to `0` which passes the `>= 0` check, so an unlimited ticket (real payload: `capacity: '', stock: -1`) still resolves to `0` instead of falling through to the stock check. Might need something like `rsvp.capacity !== '' && Number(rsvp.capacity) >= 0` instead.

sdokus (approved):

PR #4287, opened 2026-07-01
[RSVP] Part 6/22 - Data Layer: Refactor V1 Store to Use Shared

larodiel:
## Part 6/19

These files refactor the **RSVP V1 data store** to consume the shared `rsvp-shared` store:

- `index.js` — Re-export types, actions, selectors, sagas from shared
- `reducer.js` — Compose shared reducer with V1-only header-image slice
- `sagas.js` — Fork shared watchers + V1-specific sagas
- `thunks.js` — Use shared actions and hydration utilities
- `header-image.js` — Import action types from shared store
- `header-image-sagas.js` — V1-only header image saga
- `post-save-sagas.js` — V1-only post-save saga
- Test updates for changed imports

> Old `data/rsvp/` files (actions, types, selectors, etc.) are NOT yet deleted — they will be removed in a follow-up PR once all consumers are migrated.

---

> **Previous:** [split/soft-3335-05-data-shared-utils](https://github.com/the-events-calendar/event-tickets/pull/4286)  
> **Next:** [split/soft-3335-07-data-rsvp-v2](https://github.com/the-events-calendar/event-tickets/pull/4288)


src/modules/data/blocks/rsvp/reducer.js
```diff
@@ -1,140 +1,31 @@
...
 	iac: 'none',
-	details: DETAILS_DEFAULT_STATE,
-	tempDetails: DETAILS_DEFAULT_STATE,
 	headerImage: HEADER_IMAGE_DEFAULT_STATE,
 };
 
 export default ( state = DEFAULT_STATE, action ) => {
+	const nextState = sharedReducer( state, action );
+
 	switch ( action.type ) {
```
sdokus, line 17: We might want to add another special `case` here (similar to [here](https://github.com/the-events-calendar/event-tickets/pull/4286/changes#diff-89afbaf2f3c8d8451120c699f8e5d1814a23ef1d6700fd6de30ee5236a6cb5b1R92-R96) in PR 5 for `SET_RSVP_IAC`), but this time for `DELETE_RSVP`

`DELETE_RSVP` isn't currently special-cased the way `SET_RSVP_IAC`/`SET_RSVP_HEADER_IMAGE` are, so on delete it falls through to `nextState` (from the shared reducer), whose default state doesn't include `headerImage`. 

Could we add one more case here, same pattern as the other two:
 ```
case types.DELETE_RSVP:
     return DEFAULT_STATE;
```
That resets to the full local default (which already includes `headerImage` and `iac`) instead of the shared-only one. Might also be worth a quick test alongside it 🤔

File: src/modules/data/blocks/rsvp/index.js [hunk 2]
```diff
@@ -1,12 +1,9 @@
 /**
  * Internal dependencies
  */
-import * as types from './types';
-import * as actions from './actions';
-import * as selectors from './selectors';
-import * as thunks from './thunks';
+export { types, actions, selectors, sagas } from '../rsvp-shared';
```
sdokus: This re-exports `sagas` straight from `rsvp-shared` instead of the local `./sagas.js` (which forks `sharedWatchers` + `headerImageWatchers` + `postSaveWatchers` together). Since `data/sagas.js` only ever imports sagas from this file, doesn't that mean the header-image and post-save watchers never actually get registered? 

In other words, it looks like the local `sagas.js` might never have gotten wired up. Am I missing another place this gets picked up, or is this a genuine gap?

sdokus (approved):

PR #4293, opened 2026-07-01
[RSVP] Part 12/22 - V1 Blocks: Container Refactors

larodiel:
## Part 12/19

These files refactor **RSVP V1 block containers** to consume shared components:

- `container.js` — Use shared initial-state and close-overlay factories
- `action-dashboard/`, `capacity/`, `duration-picker/`, `inactive-block/` — Import from shared templates
- `container-header/` — Use shared counters and action buttons
- `container-content/` — Updated imports
- `index.js` + `filters.js` — Block lifecycle filter hooks

---

> **Previous:** [split/soft-3335-11-blocks-shared-b](https://github.com/the-events-calendar/event-tickets/pull/4292)  
> **Next:** [split/soft-3335-13-blocks-rsvp-b](https://github.com/the-events-calendar/event-tickets/pull/4294)


src/modules/blocks/rsvp/container.js
```diff
@@ -18,68 +12,50 @@ import { actions, selectors, thunks } from '../../data/blocks/rsvp';
...
+		goingCount: String( selectors.getRSVPGoingCount( state ) ?? '' ),
+		notGoingCount: String( selectors.getRSVPNotGoingCount( state ) ?? '' ),
 	};
 };
 
 const mapDispatchToProps = ( dispatch, ownProps ) => ( {
 	initializeRSVP: () => dispatch( actions.initializeRSVP() ),
-	onBlockRemoved: () => dispatch( actions.deleteRSVP() ),
-	setInitialState: setInitialState( dispatch, ownProps ),
-	setAddEditClosed: () => dispatch( actions.setRSVPIsAddEditOpen( false ) ),
```
sdokus, line 82: `setAddEditClosed` is removed here with no replacement wired up, but `rsvp/template.js` (not touched by this PR) still requires it: it's `PropTypes.func.isRequired` and gets called directly in a document-level outside-click handler (`template.js` ~L106) that's active any time the RSVP block has been interacted with. That'll throw `setAddEditClosed` is not a function on the next click anywhere outside the block.

src/modules/blocks/rsvp/container.js
```diff
@@ -18,68 +12,50 @@ import { actions, selectors, thunks } from '../../data/blocks/rsvp';
...
+	onBlockRemoved: ( props ) => {
+		if ( props.created && props.rsvpId ) {
+			dispatch( thunks.deleteRSVP( props.rsvpId ) );
+		}
```
sdokus, line 48: Previously `onBlockRemoved` only reset local Redux state — now it also calls `thunks.deleteRSVP`, which trashes the underlying RSVP ticket over REST (confirmed this specific thunk in `data/blocks/rsvp/thunks.js` doesn't pass force, so it's a soft trash, not a permanent delete, which is good). 

But this fires from `componentWillUnmount` in `with-save-data.js`, which runs on any unmount of this component, not just an explicit block deletion — e.g. toggling to Code Editor and back unmounts/remounts every block. Is it confirmed that `componentWillUnmount` here only ever fires on genuine removal in practice? If there's any unmount/remount path that isn't a real deletion, this would trash a live RSVP ticket unintentionally 🤔

src/modules/blocks/rsvp/container.js
sdokus: 😅 nevermind again, lol this seems to be addressed in the next PR - `closeBlockOverlays` replaces `setAddEditClosed`

sdokus (approved):

PR #4296, opened 2026-07-01
[RSVP] Part 15/22 - V2 Panel, Content, Header & Dashboard

larodiel:
## Part 15/22

Update V2 panel, content, header, action-dashboard, and inactive-block containers.

> Previous: [split/soft-3335-14-blocks-v2-core](https://github.com/the-events-calendar/event-tickets/pull/4295) · Next: [split/soft-3335-16-blocks-v2-comps-a](https://github.com/the-events-calendar/event-tickets/pull/4297)


sdokus:
Similar to the last PR, there are files being introduced as imports that haven't actually been added yet:

> .`./create-form/template`, `../saved-summary/container`, `../attendee-information-section/container` (all land PR 17), and `../remove-rsvp/container` (lands PR 18).

We should probably re-evaluate how we're splitting the PRs from PR 14-20

src/modules/blocks/rsvp-v2/container-panel/template.js
```diff
@@ -10,25 +10,46 @@ import classNames from 'classnames';
  */
 import { ContainerPanel } from '../../../elements';
 import { LAYOUT } from '../../../elements/container-panel';
-import RSVPContainerHeader from '../container-header/container';
```
sdokus, line 13: This PR drops the last usage of `RSVPContainerHeader` here, and I don't see it get picked up anywhere else later in the stack — looks like `RSVPSavedSummary/RSVPFrontendMirror` (a couple PRs out) end up covering the same "collapsed view" job instead. If that's right, `rsvp-v2/container-header/container.js` and `template.js` are dead weight at that point — is it worth deleting them, or is there a reason to keep them around?

larodiel:
> Similar to the last PR, there are files being introduced as imports that haven't actually been added yet:
> 
> > .`./create-form/template`, `../saved-summary/container`, `../attendee-information-section/container` (all land PR 17), and `../remove-rsvp/container` (lands PR 18).
> 
> We should probably re-evaluate how we're splitting the PRs from PR 14-20

I've added a comment on it @sdokus to we be aware of the further changes, it might help, but, if you think that it still messy, we can think in something together.

sdokus (approved):
Reviewed Part 15/22 (container-panel/content/header refactor). Confirmed the removed inactive-block/style.pcss wiring is intentionally deferred and lands in Part 22 (#4303), so the stack converges correctly. LGTM.

PR #4297, opened 2026-07-01
[RSVP] Part 16/22 - V2 Capacity, Duration & AR Link

larodiel:
## Part 16/22

Add V2 capacity, duration-fields, duration-picker, and attendee-registration-link components.

> Previous: [split/soft-3335-15-blocks-v2-panel](https://github.com/the-events-calendar/event-tickets/pull/4296) · Next: [split/soft-3335-17-blocks-v2-comps-b](https://github.com/the-events-calendar/event-tickets/pull/4298)


src/modules/blocks/rsvp-v2/attendee-registration-link/container.js
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * WordPress dependencies
+ */
+import { select } from '@wordpress/data';
+
+/**
+ * Internal dependencies
+ */
+import RSVPAttendeeRegistrationLink from './template';
```
sdokus: This rebuilds the whole attendee-info iframe logic from scratch instead of reusing the shared version — capacity and duration-fields both just wrap the shared component here, so this one stands out. Since that iframe logic has needed several fixes already, could this wrap the shared container the same way, so future fixes only need to happen once?

sdokus (approved):
Reviewed Part 16/22 (capacity, duration-fields, duration-picker, attendee-registration-link). Backward-compatible refactor of the shared attendee-registration template; V1 usage unaffected. New components aren't wired into the render tree yet so nothing breaks. LGTM.

PR #4298, opened 2026-07-01
[RSVP] Part 17/22 - V2 Create Form, Attendee Info & Saved Summary

larodiel:
## Part 17/22

Add V2 create-form, attendee-information-section, and saved-summary components.

> Previous: [split/soft-3335-16-blocks-v2-comps-a](https://github.com/the-events-calendar/event-tickets/pull/4297) · Next: [split/soft-3335-18-blocks-v2-comps-c](https://github.com/the-events-calendar/event-tickets/pull/4299)


src/modules/blocks/rsvp-v2/attendee-information-section/container.js
```diff
@@ -0,0 +1,100 @@
...
+	const Gate = compose(
+		withStore(),
+		connect( ( state ) => ( {
+			created: selectors.getRSVPCreated( state ),
+			hasTicketsPlus: plugins.selectors.hasPlugin( state )( plugins.constants.TICKETS_PLUS ),
+		} ) )
+	)( ( { created, hasTicketsPlus } ) => {
+		if ( ! hasTicketsPlus || ! created || ! isSelected ) {
+			return null;
+		}
```
sdokus: `Gate` gets rebuilt on every render of `RSVPAttendeeInformationSectionContainer` since it's defined inside the function body — React treats each new copy as a different component, so it remounts the whole subtree (including the `attendee-registration` modal if it's open) every time this re-renders, rather than just re-rendering it. 

Can `Gate` be moved outside, defined once, the way `ConnectedAttendeeInformation` above it already is?

tests/wpunit/Tribe/Tickets/Commerce/RSVP/__snapshots__/RSVPBlockV2Test__test_should_render_ticket_block_after_update with data set 0__1.php
```diff
@@ -5,7 +5,7 @@ class="tribe-common event-tickets"
 >
 			<div
 			class="tribe-tickets__rsvp-wrapper"
-			data-rsvp-id="[TICKET_ID]"
+			data-rsvp-id="10176"
```
sdokus: All of these snapshots seem to be indicative of a flaky test - this is the analysis from Claude:
> Notice this snapshot has a literal ticket ID baked in here (data-rsvp-id="10176") instead of the [TICKET_ID] placeholder used everywhere else — traced it back to RSVPBlock_TestCase::test_should_render_ticket_block_after_update (in tests/_support/Testcases/RSVPBlock_TestCase.php). The "update" step there calls setup_block_ticket() a second time with 'ticket_id' => $ticket_id' as an override, but that's not a real WordPress post field — the factory needs 'ID', not 'ticket_id'. So instead of updating the existing ticket, it silently creates a second, brand-new one on the same event. The test never captures that second ticket's ID, so it can't be swapped for the placeholder, and it ends up hardcoded in the snapshot.

> There's already an update_rsvp_ticket() helper right next to create_rsvp_ticket() in the same file that looks like it was built for exactly this. Worth swapping the second call to use that instead — should fix both the snapshot fragility and make the test actually exercise "render after an update" rather than "render with two tickets on one event." Not something this PR broke, just something the snapshot regen surfaced.

larodiel:
For some reason the PHP tests has leaked to this PR :/

sdokus (approved):
Reviewed Part 17/22. PHP: nice fix to the after-update test harness (update_rsvp_ticket instead of the force-ID recreate hack) — confirmed the regenerated snapshots now correctly show a single rendered block instead of two. JS: create-form wires up Part 16's capacity/duration/AR-link components; saved-summary/attendee-information-section correctly forward-reference Parts 18-20 modules but aren't wired into the render tree yet, and the new tests properly mock those children so nothing breaks. LGTM.

PR #4301, opened 2026-07-01
[RSVP] Part 20/22 - V2 Edit Popovers & Utilities

larodiel:
## Part 20/22

Add V2 edit-popovers, utility functions, and remaining frontend-mirror partials.

> Previous: [split/soft-3335-19-blocks-v2-comps-d](https://github.com/the-events-calendar/event-tickets/pull/4300) · Next: [split/soft-3335-21-e2e-resources](https://github.com/the-events-calendar/event-tickets/pull/4302)

### DIFF
--- a/src/modules/blocks/rsvp-v2/utils/format-rsvp-window.js
+++ b/src/modules/blocks/rsvp-v2/utils/format-rsvp-window.js
@@ -1,0 +1,67 @@
+import { globals } from '@moderntribe/common/utils';
+
+const PHP_TO_MOMENT_FORMAT = {
+	d: 'DD',
+	D: 'ddd',
+	j: 'D',
+	l: 'dddd',
+	N: 'E',
+	S: 'o',
+	w: 'e',
+	z: 'DDD',
+	W: 'W',
+	F: 'MMMM',
+	m: 'MM',
+	M: 'MMM',
+	n: 'M',
+	t: '',
+	L: '',
+	o: 'YYYY',
+	Y: 'YYYY',
+	y: 'YY',
+	a: 'a',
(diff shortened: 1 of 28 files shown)


src/modules/blocks/rsvp-v2/utils/format-rsvp-window.js
```diff
@@ -0,0 +1,67 @@
+import { globals } from '@moderntribe/common/utils';
+
+const PHP_TO_MOMENT_FORMAT = {
```
sdokus: This table is a copy of `momentUtil.toFormat()` from `common/src/modules/utils/moment.js` — same mapping, already used elsewhere in this codebase (e.g. the duration-picker container). Could this call that instead of maintaining a second copy? Any future date-format edge case fix would otherwise need to happen in both places.

sdokus (approved):
Reviewed Part 20/22. All the utils that Parts 16-19 forward-referenced (block-state, format-rsvp-window, schedule-persist-rsvp, is-nested-date-picker-interaction) land here with signatures matching every existing call site. Verified the shared date-time-range-picker changes (fromLabel/toLabel, conditional separators) are backward-compatible for V1 - defaults preserve the old output. edit-popovers props line up exactly with saved-summary/container.js from Part 17.

One drive-by, unrelated to this PR: src/resources/postcss/rsvp/_rsvp.pcss has a pre-existing typo, `var(-tec-grid-gutter-half)` (missing a dash) inside the @media (--viewport-medium) block. This PR only reformatted whitespace around that line; the typo predates it. Worth a quick follow-up fix, not blocking here.

LGTM.

PR #4302, opened 2026-07-01
[RSVP] Part 21/22 - V2 Util Tests & E2E Tests

larodiel:
## Part 21/22

Add V2 unit tests and E2E Playwright tests for RSVP V2 admin flows.

> Previous: [split/soft-3335-20-blocks-v2-utils](https://github.com/the-events-calendar/event-tickets/pull/4301) · Next: [split/soft-3335-22-resources](https://github.com/the-events-calendar/event-tickets/pull/4303)


tests/end-to-end/rsvp-v2/end-to-end-flow.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+		);
+		await rsvp.deselectBlock();
+		await expect(rsvp.savedSummary).toBeVisible({ timeout: 15000 });
+
+		// Attendee Information section should exist in the saved summary.
+		const attendeeSection = page.locator('.tribe-editor__rsvp-attendee-information');
+		const count = await attendeeSection.count();
+		// The section may or may not be present depending on ET+ state.
+		// When ET+ is active, it should be visible.
+		expect(count).toBeGreaterThanOrEqual(0);
```
sdokus: `expect(count).toBeGreaterThanOrEqual(0)` can never fail — `count()` is never negative, so this doesn't actually check anything 😆 

Since the comment says it should show up when ET+ is active, could this check the actual ET+ state and assert accordingly (visible when active, absent when not), instead of accepting either outcome?

src/Tickets/RSVP/V2/REST/Ticket_Endpoint.php
```diff
@@ -0,0 +1,348 @@
...
+		if ( ! $ticket_post instanceof WP_Post ) {
+			return new WP_REST_Response(
+				[
+					'success' => false,
+					'message' => __( 'Ticket not found or does not belong to this event.', 'event-tickets' ),
+				],
+				404
+			);
+		}
+
```
sdokus, line 243:
```suggestion

$ticket_event_id = (int) get_post_meta( $ticket_id, Ticket::$event_relation_meta_key, true );
if ( $ticket_event_id !== $post_id ) {
    return new WP_REST_Response(
        [
            'success' => false,
            'message' => __( 'Ticket not found or does not belong to this event.', 'event-tickets' ),
        ],
        404
    );
}

```
We should add an ownership check here to avoid a future security vulnerability. The check above verifies the ticket exists, but not that it belongs to this event. 

*Note:* This needs `use TEC\Tickets\Commerce\Ticket;` added to the file's use block — it isn't currently imported here

src/Tickets/RSVP/V2/REST/Ticket_Endpoint.php
```diff
@@ -0,0 +1,348 @@
...
```
sdokus, line 325:
```suggestion

$ticket_event_id = (int) get_post_meta( $ticket_id, Ticket::$event_relation_meta_key, true );
if ( $ticket_event_id !== $post_id ) {
    return new WP_REST_Response(
        [
            'success' => false,
            'message' => __( 'Ticket not found or does not belong to this event.', 'event-tickets' ),
        ],
        404
    );
}

```
Same thing as the suggestion above, we only check that the ticket exists, but not that it belongs to this event.

src/Tickets/RSVP/V2/REST/Ticket_Meta_Endpoint.php
```diff
@@ -0,0 +1,165 @@
...
+		if ( ! $ticket_post instanceof WP_Post ) {
+			return new WP_REST_Response(
+				[
+					'success' => false,
+					'message' => __( 'Ticket not found.', 'event-tickets' ),
+				],
+				404
+			);
+		}
+
```
sdokus, line 136:
```suggestion

$ticket_event_id = (int) get_post_meta( $ticket_id, Ticket::$event_relation_meta_key, true );
if ( $ticket_event_id !== $post_id ) {
    return new WP_REST_Response(
        [
            'success' => false,
            'message' => __( 'Ticket not found.', 'event-tickets' ),
        ],
        404
    );
}

```
One more place where we should actually check that the ticket belongs to the event. 

**Note: ** This file also needs `use TEC\Tickets\Commerce\Ticket;` added at the top of the file.

sdokus (approved):
All three ownership-check concerns are fixed and verified (ticket-belongs-to-event check now present in both Ticket_Endpoint.php spots and Ticket_Meta_Endpoint.php, with the missing `use` import added). PHPCS is clean on HEAD.

PR #4310, opened 2026-07-04
[SOFT-3351] Stop RSVPs from triggering Purchase Receipt and Order Completed emails

larodiel:
### 🎫 Ticket

[SOFT-3351](https://linear.app/nexcess/issue/SOFT-3351/rsvps-trigger-purchase-receipt-and-order-completed-emails)

### 🗒️ Description

The new (unreleased) TC-RSVP order flow (`Order_Endpoint::process_rsvp_step()`) creates a real Tickets Commerce order and transitions it to `Completed`, which fires every "Completed" status flag generically — including the purchase-oriented `send_email_purchase_receipt` and `send_email_completed_order` flags, and the paid-ticket `send_email` (ticket/QR) flag. None of those had any RSVP awareness, so RSVP submissions were sending three wrong emails and no RSVP confirmation at all.

**Fix:**

* Added a shared `is_rsvp_order()` guard on `Flag_Action_Abstract` (an order is RSVP-only when every item is of type `tc-rsvp`), mirroring the existing precedent in `Decrease_Stock`.
* `Send_Email_Purchase_Receipt`, `Send_Email_Completed_Order`, and `Send_Email` now bail early for RSVP orders.
* New `Send_Email_RSVP` flag action sends the correct `Email\RSVP` (going) or `Email\RSVP_Not_Going` (not going) confirmation, registered alongside the other `Send_Email_*` flag actions.
* Fixed a related root cause in `Order::create_from_cart()` / `Abstract_Cart::add_full_item_params()`: the generic cart path was hardcoding every item's `type` to `ticket`, which could silently drop RSVP items from an order regardless of the ticket's real type. `create_from_cart()` now reads from the dedicated `RSVP_Cart` repository when building a TC-RSVP order, and the type is only defaulted (not overwritten) when missing.

### 🎥 Artifacts

https://github.com/user-attachments/assets/b579930c-836a-4777-a848-6822614196f0

(sorry to don't be on the Loom, my video quota is over and already asked to the account upgrade)

### ✔️ Checklist

- [X] Code is covered by **NEW** tests.
- [ ] Code is covered by **EXISTING** tests.
- [ ] Changelog item has been added. If not required for this change, the `No Changelog Needed` label should be added.
- [ ] All CI checks have passed, or any expected failures are explained in the Description section.
- [ ] Have you added Testing Instructions for QA?


src/Tickets/Commerce/Order.php
```diff
@@ -619,7 +621,10 @@ static function ( $item ) {
 	 * @throws \Tribe__Repository__Usage_Error When there is a repository usage error.
 	 */
 	public function create_from_cart( Gateway_Interface $gateway, $purchaser = null, $ticket_type = 'ticket' ) {
-		$cart = tribe( Cart::class );
+		$cart        = tribe( Cart::class );
+		$cart_reader = RSVP_V2_Constants::TC_RSVP_TYPE === $ticket_type
```
dpanta94: Instaed of touching code here, i would prefer you to utilize the filter `tec_tickets_commerce_cart_repository` and change the returned repository to the RSVP_Cart from there. Inside the RSVP/V2 directory.

src/Tickets/Commerce/Flag_Actions/Send_Email.php
```diff
@@ -37,6 +37,10 @@ class Send_Email extends Flag_Action_Abstract {
 	 * {@inheritDoc}
 	 */
 	public function handle( Status_Interface $new_status, $old_status, \WP_Post $order ) {
+		// Bail if the order is a TC-RSVP order; RSVP confirmations are handled by Send_Email_RSVP.
+		if ( $this->is_rsvp_order( $order ) ) {
+			return;
+		}
```
dpanta94: I dont love this approach. Can we better handle which Flag actions are for Tickets Only and which are for RSVP's V2 only in a higher level ?

This can easily produce issues later if someone forgets to check for which the action is for.

Registering actions per order type in a higher level is a better option IMO

dpanta94 (changes requested):
This PR will also needs more integration test coverage. Creating various orders with different statuses and asserting which flag actions are being fired.

src/Tickets/Commerce/Emails/Order_Email_Sender_Registry.php
```diff
@@ -0,0 +1,105 @@
...
+		 * Filters the order email senders used when the `send_email` flag action fires.
+		 *
+		 * Prefer registering senders on the container tag
+		 * `Order_Email_Sender_Registry::CONTAINER_TAG` via `tribe()->tag()`.
+		 *
+		 * @since TBD
+		 *
+		 * @param Order_Email_Sender_Interface[] $senders Registered senders.
+		 */
+		return (array) apply_filters( 'tec_tickets_commerce_order_email_senders', $senders );
```
dpanta94:
```suggestion
		/**
		 * Filters the order email senders used when the `send_email` flag action fires.
		 *
		 * Prefer registering senders on the container tag
		 * `Order_Email_Sender_Registry::CONTAINER_TAG` via `tribe()->tag()`.
		 *
		 * @since TBD
		 *
		 * @param Order_Email_Sender_Interface[] $senders Registered senders.
		 */
		return (array) apply_filters( 'tec_tickets_commerce_order_email_senders', $this->container->tagged( self::CONTAINER_TAG ) );
```

src/Tickets/Commerce/Flag_Actions/Flag_Action_Handler.php
```diff
@@ -25,25 +27,44 @@ class Flag_Action_Handler extends \TEC\Common\Contracts\Service_Provider {
 	protected $flag_actions = [];
 
 	/**
-	 * Which classes we will load for order flag actions by default.
+	 * Flag actions that apply to all Tickets Commerce order types.
 	 *
 	 * @since 5.1.9
```
sdokus:
```suggestion
	 * @since TBD
```
I could be off here, but did `$default_flag_actions` get moved below so this should be treated as a new variable?

larodiel: humm, I don't think so @sdokus, the variable continue existing, it just moved its place

src/Tickets/Commerce/Flag_Actions/Flag_Action_Handler.php
```diff
@@ -25,25 +27,45 @@ class Flag_Action_Handler extends \TEC\Common\Contracts\Service_Provider {
...
 		Send_Email_Purchase_Receipt::class,
 		Send_Email_Completed_Order::class,
-		End_Duplicated_Pending_Orders::class,
 	];
 
+	/**
+	 * Which classes we will load for order flag actions by default.
+	 *
+	 * @since 5.1.9
+	 * @since TBD Merges the shared and ticket-only flag action lists, registers the order email
```
sdokus: Did this get cut off? Is it supposed to match the doc block below?

larodiel: Yes, it's explaining the use of the array, I removed the `@since 5.1.9`

src/Tickets/Commerce/Emails/Order_Email_Sender_Registry.php
```diff
@@ -0,0 +1,99 @@
...
+	/**
+	 * Order_Email_Sender_Registry constructor.
+	 *
+	 * @since TBD
+	 *
+	 * @param ContainerInterface $container The plugin DI container.
+	 */
+	public function __construct( ContainerInterface $container ) {
+		$this->container = $container;
+	}
```
dpanta94: There are many arguments of why you shouldn't be passing the container instance around. I can also think of arguments for passing it around.

I do have to say that i am a developer that will agree mostly with the arguments that support NOT passing the container instance around.

What do we need from the container ? I see only `$this->container->tagged( self::CONTAINER_TAG )`.

So this is what i would pass, BUT with lazy resolution.

The contructor signature would need to change to `__construct( array $senders )`

Then the bidning of this service would be something like this:

```php
$this->container->singleton( Order_Email_Sender_Registry:class );
$this->container->when( Order_Email_Sender_Registry:class )->needs( '$senders' )
->give( fn(): array => $this->container->tagged( self::CONTAINER_TAG ) );
```

src/Tickets/Commerce/Emails/Order_Email_Sender_Registry.php
```diff
@@ -0,0 +1,99 @@
...
+		 *
+		 * Prefer registering senders on the container tag
+		 * `Order_Email_Sender_Registry::CONTAINER_TAG` via `tribe()->tag()`.
+		 *
+		 * @since TBD
+		 *
+		 * @param Order_Email_Sender_Interface[] $senders Registered senders.
+		 */
+		return (array) apply_filters( 'tec_tickets_commerce_order_email_senders', $this->container->tagged( self::CONTAINER_TAG ) );
+	}
```
dpanta94: I would apply the filter inside the container definiton (aka the ->give() method) and remove this private method from here.

src/Tickets/Commerce/Emails/Order_Email_Sender_Registry.php
```diff
@@ -0,0 +1,99 @@
...
+			if ( ! $sender instanceof Order_Email_Sender_Interface ) {
+				continue;
+			}
+
+			if ( $sender->supports( $order ) ) {
+				$sender->send( $order );
+
+				return;
+			}
+		}
```
dpanta94, line 84: Isn't this a weird choice ? I have an `Order_Email_Sender_Registry` but only the first sender that supports the order is being sent ?

src/Tickets/Commerce/Emails/RSVP_Email_Sender.php
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function send( WP_Post $order ): void {
+		if ( ! tec_tickets_emails_is_enabled() ) {
+			return;
+		}
+
```
dpanta94: I would put this check a level higher - The send method should not be aware of this setting, if we have reached this point the setting should be on.

src/Tickets/Commerce/Flag_Actions/Flag_Action_Abstract.php
```diff
@@ -94,8 +105,32 @@ public function get_post_types() {
...
+
+		/**
+		 * Allows modifications of which order types will trigger this action.
+		 *
+		 * @since TBD
+		 *
+		 * @param string[] $contexts    Which order types will trigger this action.
+		 * @param static   $action_flag Instance of action flag we are triggering.
+		 */
+		return apply_filters( 'tec_tickets_commerce_flag_actions_get_order_contexts', $contexts, $this );
```
dpanta94: This isn't a good filter name for an abstract class. it should at least have one more versioned with a unique identifier in the filter name

src/Tickets/Commerce/Flag_Actions/Flag_Action_Abstract.php
```diff
@@ -125,19 +164,58 @@ public function is_correct_post_type( \WP_Post $post ) {
...
+
+		if ( $wants_rsvp ) {
+			return $is_rsvp;
+		}
+
+		if ( $wants_ticket ) {
+			return ! $is_rsvp;
+		}
+
+		return false;
```
dpanta94: Let's simplify this ?


```suggestion
return $wants_rsvp ? $is_rsvp : ! $is_rsvp;
```

src/Tickets/Commerce/Flag_Actions/Flag_Action_Handler.php
```diff
@@ -25,25 +27,44 @@ class Flag_Action_Handler extends \TEC\Common\Contracts\Service_Provider {
...
+	 * Flag actions that apply only to standard ticket orders.
+	 *
+	 * @since TBD
+	 *
+	 * @var string[]
+	 */
+	protected $ticket_only_flag_actions = [
 		Send_Email_Purchase_Receipt::class,
 		Send_Email_Completed_Order::class,
-		End_Duplicated_Pending_Orders::class,
```
dpanta94, line 44: This move of the `End_Duplicated_Pending_Orders` smells like backwards incompatible. Remember one codebase for both V1 and V2

src/Tickets/Commerce/Flag_Actions/Flag_Action_Handler.php
```diff
@@ -59,22 +80,56 @@ public function get_all() {
...
+	 *            senders, and resolves each flag action via the container instead of `new`.
 	 */
 	public function register() {
-		foreach ( $this->default_flag_actions as $flag_action_class ) {
-			// Spawn the new instance.
-			$flag_action = new $flag_action_class;
+		$this->default_flag_actions = array_merge(
+			$this->shared_flag_actions,
+			$this->ticket_only_flag_actions
+		);
```
dpanta94: Similar here

src/Tickets/Commerce/Flag_Actions/Flag_Action_Handler.php
```diff
@@ -59,22 +80,56 @@ public function get_all() {
...
+		$this->container->singleton( RSVP_Email_Sender::class );
+		$this->container->singleton( Ticket_Email_Sender::class );
+
+		$this->container->tag(
+			[
+				RSVP_Email_Sender::class,
+				Ticket_Email_Sender::class,
+			],
+			Order_Email_Sender_Registry::CONTAINER_TAG
+		);
```
dpanta94, line 128: Tag has a wow factor to it. But do we really need it ? Right below you can bind this array as the first argument of `Order_Email_Sender_Registry` directly.

src/Tickets/Commerce/Flag_Actions/Order_Context.php
```diff
@@ -0,0 +1,46 @@
...
+namespace TEC\Tickets\Commerce\Flag_Actions;
+
+/**
+ * Class Order_Context.
+ *
+ * @since TBD
+ *
+ * @package TEC\Tickets\Commerce\Flag_Actions
+ */
+class Order_Context {
```
dpanta94: Lets make it final - prefer final in general. throughout the PR and the code you are writing. We would remove `final` only when we specifically want to extend a class.

src/Tickets/Commerce/Order.php
```diff
@@ -621,6 +623,14 @@ static function ( $item ) {
 	public function create_from_cart( Gateway_Interface $gateway, $purchaser = null, $ticket_type = 'ticket' ) {
 		$cart = tribe( Cart::class );
 
+		tribe_set_var( Cart::ORDER_FROM_CART_TICKET_TYPE_VAR, $ticket_type );
+
+		try {
+			$cart_reader = $cart->get_repository();
+		} finally {
+			tribe_unset_var( Cart::ORDER_FROM_CART_TICKET_TYPE_VAR );
+		}
```
dpanta94: I'm not a big fan of this. Why not pass the $ticket_type into the get_repository method ? Which would be option and `ticket` by default ?

src/Tickets/RSVP/V2/Cart/Repository_Filter.php
```diff
@@ -0,0 +1,43 @@
...
+	 * @return Cart_Interface
+	 */
+	public function use_rsvp_cart_when_needed( Cart_Interface $cart ): Cart_Interface {
+		$ticket_type = tribe_get_var( Cart::ORDER_FROM_CART_TICKET_TYPE_VAR, 'ticket' );
+
+		if ( Constants::TC_RSVP_TYPE !== $ticket_type ) {
+			return $cart;
+		}
+
+		return tribe( RSVP_Cart::class );
```
dpanta94: Instead of using the container bound param, i would also pass the $ticket_type param through the filter.

As a result the method could become something like:

`return $ticket_type === 'ticket' ? $cart : tribe( RSVP_Cart::class );`

src/Tickets/RSVP/V2/REST/Order_Endpoint.php
```diff
@@ -156,9 +142,13 @@ public function handle_steps( WP_REST_Request $request ): WP_REST_Response {
 		$ticket_id = absint( $request->get_param( 'ticket_id' ) ?: 0 );
 		$step      = $request->get_param( 'step' );
 
-		add_filter( 'tec_tickets_commerce_cart_repository', [ $this, 'setup_cart' ] );
+		tribe_set_var( Cart::ORDER_FROM_CART_TICKET_TYPE_VAR, Constants::TC_RSVP_TYPE );
```
dpanta94: Similar here! Not a fan. This is a hidden mutation. Very hard to debug and it requires knowledge about this container var.

dpanta94 (changes requested):

src/Tickets/Commerce/Emails/Order_Email_Sender_Registry.php
larodiel: it's on purpose — `RSVP_Email_Sender and Ticket_Email_Sender` can never both say "yes" to the same order (one checks `is_rsvp_order()`, the other checks the opposite), so only one sender is ever a real match anyway. First-match-wins is just how you'd write "pick the one that applies."
Added a docblock on `send()` so it's clear this is intentional and not something someone forgot to finish. If we ever add a sender that could overlap with another, that's a `supports()` problem more than a registry problem, happy to talk through it if that comes up.

src/Tickets/Commerce/Flag_Actions/Flag_Action_Handler.php
larodiel: I think the actual issue was `$default_flag_actions sitting there` empty `([])` until `register()` ran, so anything reading it before that would get nothing. That's fixed now, it's just a local var inside `register()`.
 As for moving `End_Duplicated_Pending_Orders`, I don't think it actually changes anything at runtime. Every flag action fires on its own hook name (`send_email`, `send_email_purchase_receipt`, etc. are all different strings), so where it sits in this array doesn't affect when it runs relative to the others. Also double-checked: V1 RSVP never creates a TC order, so it never even hits this class either way.
If you know of something outside this class that reads `$shared_flag_actions or $default_flag_actions directly`, let me know and I'll fix that up.

src/Tickets/Commerce/Flag_Actions/Flag_Action_Handler.php
larodiel: Kept the tag, but fixed the bigger issue that was hiding next to it, the registry doesn't take the container anymore. Constructor's just array `$senders` now, and the filter moved out of the class into the binding:
 ```php
$this->container->when( Order_Email_Sender_Registry::class )
     ->needs( '$senders' )
     ->give( fn() => (array) apply_filters( 'tec_tickets_commerce_order_email_senders', $this->container->tagged( self::CONTAINER_TAG ) ) );
```
Registry's easy to unit test now, just pass it an array of fake senders, no container needed.
On the tag itself, kept it since it's the "add a sender without touching this file" path for other code. If you'd rather I just bind the array directly, that's an easy change, just say the word.

File: src/Tickets/Commerce/Flag_Actions/Flag_Action_Handler.php [hunk 19]
```diff
@@ -56,25 +68,89 @@ public function get_all() {
...
+					 * @param Order_Email_Sender_Interface[] $senders Registered senders.
+					 */
+					return (array) apply_filters(
+						'tec_tickets_commerce_order_email_senders',
+						$this->container->tagged( Order_Email_Sender_Registry::CONTAINER_TAG )
+					);
+				}
+			);
+
+		$this->container->singleton( Order_Email_Sender_Registry::class );
```
dpanta94, line 151: Looked very weird to me that we bind after we explain to the container how to build it. Had to look on the source, and this is fine though.

dpanta94 (approved):

PR #4324, opened 2026-07-13
Fix Promoter triggered emails failing with 403 unauthorized token

sdokus:
Ticket: [SMTNC-1632](https://linear.app/nexcess/issue/SMTNC-1632/promoter-triggered-emails-fail-with-403-unauthorized-token)

I need to still reproduce the bug locally and then confirm that this fix works to resolve it, but in theory this should work 🤔 

## Summary
- Triggered emails (RSVP going, ticket purchased) were failing silently with a `403 Forbidden` / `"Unauthorized token."` response from Promoter's connector, even though attendee sync succeeded.
- Root cause: the trigger JWT payload built in `Dispatcher::get_payload()` was missing the `domain` claim that Promoter's connector requires. The other three JWT payloads this plugin builds (`connect`, `connect/auth`, `connect/notify`) all include `domain`; the trigger payload for `connect/trigger` never had it added.
- Because `Connector::make_call()` swallows non-2xx responses into a debug log and returns `false` without throwing, the failure was invisible outside of `tribe_log` debug output — matching the "no Mailgun logs, sync succeeds" symptom reported by customers.

## Fix
Add the same domain derivation used elsewhere (`strtolower( untrailingslashit( preg_replace( '#^https?://#i', '', home_url( '/' ) ) ) )`) to the trigger payload in `src/Tribe/Promoter/Triggers/Dispatcher.php`.

I tested with this snippet:
```php
add_filter( 'pre_http_request', function ( $preempt, $parsed_args, $url ) {
	if ( strpos( $url, 'connect/trigger' ) === false ) {
			return $preempt;
	}

	$token   = $parsed_args['body']['token'] ?? '';
	$parts   = explode( '.', $token );
	$payload = isset( $parts[1] ) ? json_decode( base64_decode( strtr( $parts[1], '-_', '+/' ) ), true ) : null;

	error_log( 'PROMOTER TRIGGER PAYLOAD: ' . print_r( $payload, true ) );

	return $preempt; // returning false (the default) lets the real request go out too
}, 10, 3 );
```

When an attendee registered for an RSVP, this was the response with production code:
```php
PROMOTER TRIGGER PAYLOAD: Array
(
    [license] => test-license
    [type] => rsvp_going
    [event] => Array
        (
            [id] => 3747
        )

    [ticket] => Array
        (
            [id] => 3748
            [name] => RSVP V1
        )

    [attendee] => Array
        (
            [id] => 3751
            [email] => clown@email.com
        )

)
```

And this was the response with the fix in place (note that there is now the `[domain]` key):
```php
(
    [license] => test-license
    [domain] => stabledev.lndo.site
    [type] => rsvp_going
    [event] => Array
        (
            [id] => 3747
        )

    [ticket] => Array
        (
            [id] => 3748
            [name] => RSVP V1
        )

    [attendee] => Array
        (
            [id] => 3750
            [email] => sam.dokus@theeventscalendar.com
        )

)
```

### DIFF
--- a/changelog/fix-smtnc-1632-promoter-trigger-domain
+++ b/changelog/fix-smtnc-1632-promoter-trigger-domain
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixed triggered emails (RSVP going, ticket purchased) failing silently with a 403 "Unauthorized token" response from Promoter due to a missing `domain` claim in the trigger JWT payload.
--- a/src/Tribe/Promoter/Triggers/Dispatcher.php
+++ b/src/Tribe/Promoter/Triggers/Dispatcher.php
@@ -95,6 +95,7 @@ class Dispatcher {
 	 * Creat ea payload using the trigger object.
 	 *
 	 * @since 4.12.3
+	 * @since TBD Added the `domain` key, required by the Promoter connector to authorize the request.
 	 *
 	 * @param Triggered $trigger The trigger object creating this action.
 	 *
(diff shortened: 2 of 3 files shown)


redscar:
Overall the code is looking good. I think once you can replicate it this will be ready to go :+1:

redscar (approved):

PR #4333, opened 2026-07-20
SOFT-3910: Fix Attendees table shows stale check-in state until object cache flush

estevao90:
### 🎫 Ticket

[SOFT-3910]

### 🗒️ Description

On Tickets → Attendees (tec-tickets-attendees), check-in / undo check-in updates correctly in the AJAX UI, but a full page reload still shows the previous status until Redis/object cache is flushed.

#### Cause
Two caches kept hydrated attendee rows with outdated check_in:

- `tec_tc_get_attendee()` — week-long cache keyed without check-in meta (check-in only updates post meta, not post_modified / save_post)
- `Attendees_Table::prepare_items()` — 60s persistent Redis cache of table rows, which can short-circuit before (1)

#### Fix
- Included the check-in meta in the `tec_tc_get_attendee cache` key
- Used `Tribe__Cache::NON_PERSISTENT` in `prepare_items()` (same-request dedupe only)

### 🎥 Artifacts 

https://www.loom.com/share/895dc7c799b34689a76bfa04e972c740

### ✔️ Checklist

### DIFF
--- a/src/Tribe/Attendees_Table.php
+++ b/src/Tribe/Attendees_Table.php
@@ -1126,7 +1126,7 @@ class Tribe__Tickets__Attendees_Table extends WP_List_Table {
 		}
 
 		$this->items = $items;
-		$cache->set( $cache_key, $items, 60 );
+		$cache->set( $cache_key, $items, Tribe__Cache::NON_PERSISTENT );
 
 		$this->set_pagination_args( $pagination_args );
 	}
--- a/changelog/fix-SOFT-3910-attendees-listing-page-cache
+++ b/changelog/fix-SOFT-3910-attendees-listing-page-cache
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+The Attendees page table could show stale check-in data in some cases.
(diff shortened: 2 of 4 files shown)


src/Tribe/Attendees_Table.php
```diff
@@ -1126,7 +1126,7 @@ public function prepare_items() {
 		}
 
 		$this->items = $items;
-		$cache->set( $cache_key, $items, 60 );
+		$cache->set( $cache_key, $items, Tribe__Cache::NON_PERSISTENT );
```
dpanta94, line 1129: Why are we switching to runtime cache only over here through ?

common
dpanta94, line 1: Changes the common should come in their own PR

dpanta94 (changes requested):

src/Tribe/Attendees_Table.php
estevao90: That's a good question. 

We have two layers of cache. The first is in `tec_tc_get_attendee()`,  which has a per-attendee data cache that I've fixed in the file above. This one is long (about one week and it's the most important fix).

However, the Attendees table also caches the full row set for 60s, keyed only on query args + event ID. So, if we have a cache, we return all the table data, without checking anything. 

So, I've changed it to `NON_PERSISTENT` to avoid cache invalidation everywhere. If it's critical, I can identify all the places and invalidate the cache, but I think it can be complex. 

The point is that, with the current cache, any data is the same for ~60s, so when we click on the "Check in" button, for example, and refresh the page, we can't see the change. 

We can also accept it (as it can also be an edge case?) and keep the cache here for 60s without any invalidation. That's why I'd like a review from a TEC expert, so we can decide the right approach.

dpanta94 (approved):

PR #4341, opened 2026-07-23
[SMTNC-1637] fix: Add attendee check-in validation

larodiel:
### 🎫 Ticket

[SMTNC-1637](https://linear.app/nexcess/issue/SMTNC-1637/tickets-commerce-refunded-tickets-should-remove-the-ability-for)

### 🗒️ Description

Fixed an issue that we were allowing attendee to check-in after the refund.
As the refund action relay on a queue, the user can still perform the check-in right after the refund be made.
To revert it, once the attendee is moved to trash/archive we run a method to unchecking them.

### 🎥 Artifacts 

https://www.loom.com/share/055e22b1e4814c638f7dd6d5c87069c6

### ✔️ Checklist


changelog/fix-smtnc-1637-tickets-commerce-refunded-tickets-should-remove-the-ability
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Resolved an issue where an attendee could still be checked in after their order was refunded.
```
sdokus: Can you re-add this changelog with the new stellarwp/changelog? If you re-pull the release branch it should have it! The new format looks lsomething like this and the file will have the `.yaml` extension:
```
significance: patch
type: fix
entry: Resolved an issue where an attendee could still be checked in after their order was refunded.
```

larodiel: done!
Thank you @sdokus

src/Tickets/Hooks.php
```diff
@@ -112,8 +136,71 @@ public function generate_payments_pages() {
...
+			return false;
+		}
+
+		/*
+		 * The order's live status can still read as completed while a gateway webhook that would move it
+		 * away from that status (e.g. a refund) is deferred and not yet applied - see
+		 * Order::has_on_checkout_screen_hold(). Gateways that defer webhooks this way hook into this
+		 * filter to report that an in-flight, not-yet-applied change exists for the order.
+		 */
+		if ( ! empty( $attendee['order_id'] ) && apply_filters( 'tec_tickets_commerce_order_has_pending_non_completed_transition', false, (int) $attendee['order_id'] ) ) {
```
coutinhocouto: here should not have docblocks for the `apply_filters`?

src/Tickets/Hooks.php
```diff
@@ -112,8 +136,71 @@ public function generate_payments_pages() {
...
+		}
+
+		$attendee = $ticket_provider->get_attendees_by_id( $attendee_id );
+		$attendee = reset( $attendee );
+
+		if ( ! is_array( $attendee ) || ! isset( $attendee['order_status'] ) ) {
+			return $checkin;
+		}
+
+		$completed_statuses = (array) tribe( 'tickets.status' )->get_completed_status_by_provider_name( $ticket_provider );
```
coutinhocouto, line 202: `get_completed_status_by_provider_name()` returns `[]` whenever `get_provider_slug()` can't resolve the class. it only knows the five hard-coded entries in [Manager::$module_slugs](https://github.com/the-events-calendar/event-tickets/blob/db698889124fd8702b5dd54904474dac5a62bc5c/src/Tribe/Status/Manager.php#L21-L29). With an empty array, `in_array( …, [], true )` is false and the guard returns false, so every check-in for any third-party ticket provider is blocked. Meanwhile line 184 returns `$checkin` (allow) when attendee data is missing. Worth an explicit `if ( empty( $completed_statuses ) ) { return $checkin; }` so an unknown provider can't be silently locked out.

src/Tickets/Commerce/Gateways/Stripe/Webhooks/Handler.php
```diff
@@ -143,6 +143,20 @@ public static function update_order_status( \WP_Post $order, Commerce_Status\Sta
 			return false;
 		}
 
-		return tribe( Order::class )->modify_status( $order->ID, $status->get_slug(), $metadata );
+		$updated = tribe( Order::class )->modify_status( $order->ID, $status->get_slug(), $metadata );
+
+		if ( ! $updated ) {
```
coutinhocouto, line 149: `modify_status()` returns `false` on two normal conditions, not just failures: order-lock contention ([Order.php#L488-L492](https://github.com/the-events-calendar/event-tickets/blob/db698889124fd8702b5dd54904474dac5a62bc5c/src/Tickets/Commerce/Order.php#L487-L493)) and can_transition_to() returning false ([Order.php#L494-L500](https://github.com/the-events-calendar/event-tickets/blob/db698889124fd8702b5dd54904474dac5a62bc5c/src/Tickets/Commerce/Order.php#L493-L501)), which is what happens when Stripe re-delivers a webhook for an order already in the target status. The file's own comment 10 lines above acknowledges Stripe resends. So every duplicate delivery writes an ERROR-level entry. log_warning, or distinguishing "already in target status" from a genuine failure, would fit better.

src/Tickets/Hooks.php
```diff
@@ -48,6 +49,29 @@ public function register() {
...
+	 * An Attendee can legitimately check in before a deferred webhook that would move the order away
+	 * from a completed status is applied - see Order::has_on_checkout_screen_hold(). Once that webhook
+	 * does apply and the order's Attendees are archived, any check-in that happened in that window is
+	 * stale and must be revoked.
+	 *
+	 * @since TBD
+	 *
+	 * @param int $attendee_id The Attendee post ID that was just archived.
+	 */
+	public function uncheckin_attendee_on_archive( int $attendee_id ): void {
```
coutinhocouto, line 73: The docblock justifies this for the deferred-webhook race, but the action fires from [Attendee::archive()](https://github.com/the-events-calendar/event-tickets/blob/db698889124fd8702b5dd54904474dac5a62bc5c/src/Tickets/Commerce/Attendee.php#L261-L263) for any status carrying the archive_attendees flag ([Flag_Actions/Archive_Attendees.php#L85](https://github.com/the-events-calendar/event-tickets/blob/db698889124fd8702b5dd54904474dac5a62bc5c/src/Tickets/Commerce/Flag_Actions/Archive_Attendees.php#L84-L86)), and uncheckin() deletes checkin_key, checkin_key . '_details' and _tribe_qr_status irreversibly ([Tribe/Tickets.php#L1176-L1178](https://github.com/the-events-calendar/event-tickets/blob/db698889124fd8702b5dd54904474dac5a62bc5c/src/Tribe/Tickets.php#L1175-L1179)). Refunding after the event has run silently wipes the record that the person attended. The ticket asks to prevent check-in, not to erase history, consider scoping this to the hold window.

coutinhocouto (changes requested):

src/Tickets/Hooks.php
```diff
@@ -112,8 +150,87 @@ public function generate_payments_pages() {
...
+		}
+
+		/**
+		 * Filters whether the order has a pending non-completed transition, i.e. a deferred webhook that
+		 * would move it away from a completed status (e.g. a refund) that is not yet applied. The order's
+		 * live status can still read as completed while a gateway webhook that would move it
+		 * away from that status (e.g. a refund) is deferred and not yet applied - see
+		 * Order::has_on_checkout_screen_hold(). Gateways that defer webhooks this way hook into this
+		 * filter to report that an in-flight, not-yet-applied change exists for the order.
+		 */
```
redscar: Is this docblock on purpose? Can it be combined with the below?

src/Tickets/Commerce/Gateways/Stripe/Hooks.php
```diff
@@ -171,6 +211,12 @@ public function process_async_stripe_webhook( int $order_id, int $retry = 0 ): v
 		// On multiple checkout completes, make sure we dont process the same webhook twice.
 		$webhooks->delete_pending_webhooks( $order->ID );
 
+		// Flag the order so that attendee archive triggered by the async webhook resolution
+		// (via modify_status below) can be distinguished from post-event refund archives.
+		// The flag is checked by TEC\Tickets\Hooks::uncheckin_attendee_on_archive() and deleted
+		// after all pending webhooks are processed.
+		update_post_meta( $order->ID, '_tec_tickets_commerce_webhook_resolving_archive', 1 );
```
redscar, line 218: If modify_status() throws mid-loop (it can raise Tribe__Repository__Usage_Error), we never reach the delete_post_meta() and the `_webhook_resolving_archiveflag` sticks to the order permanently. Then any later archive (e.g. a post-event refund) would wipe check-in history, which is what this flag exists to prevent. Should the loop be wrapped in try/finally so the flag is always cleaned up? Applies to both the Stripe and Square processors.

larodiel: hey @redscar could you take a look and see if get it right?

coutinhocouto (approved):
Leaving my approuval since my requests are fixed, but wait @redscar approuval also :)

redscar (approved):

sdokus (approved):

PR #4368, opened 2026-08-06
[SMTNC-1957] Fatal error on WooCommerce order edit screen by Event Tickets 5.29.2

coutinhocouto:
### 🎫 Ticket

[SMTNC-1957](https://linear.app/nexcess/issue/SMTNC-1957/fatal-error-on-woocommerce-order-edit-screen-by-event-tickets-5292)

### 🗒️ Description

Resolved an issue where opening a WooCommerce order containing tickets threw `Cannot traverse an already closed generator` and made the order edit screen inaccessible. The attendee cache warm-up added in 5.29.2 introduced a second pass over the attendee set, which failed whenever a caller supplied a generator, so `Tribe__Tickets__Tickets::get_attendees_from_module()` now normalizes traversable input before reading it.

### 🎥 Artifacts
https://www.loom.com/share/38e2cf93aa9b487e95a1591da4354daa

### ✔️ Checklist
- [ x Check the base branch for your PR.

### DIFF
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -1698,13 +1698,22 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 		 *
 		 * @since 4.10.6
 		 * @since 5.29.2 Added cache to handle the attendees data from the current module.
+		 * @since TBD Normalized traversable input, as the attendees are read more than once here.
 		 *
-		 * @param array $attendees Attendee objects or IDs.
-		 * @param int   $post_id   Parent post ID.
+		 * @param array|Traversable $attendees Attendee objects or IDs.
+		 * @param int               $post_id   Parent post ID.
 		 *
 		 * @return array The attendee data for attendees.
 		 */
 		public function get_attendees_from_module( $attendees, $post_id = 0 ) {
+			/*
+			 * Callers may hand over a generator, e.g. from `Tribe__Repository::all( true )`, and this method
+			 * reads the set twice: once to warm the post caches and once to build the data.
+			 */
+			if ( $attendees instanceof Traversable ) {
+				$attendees = iterator_to_array( $attendees, false );
+			}
+
 			/** @var Tribe__Cache $cache */
 			$cache = tribe( 'cache' );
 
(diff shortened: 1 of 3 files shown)


src/Tribe/Tickets.php
```diff
@@ -1698,13 +1698,22 @@ public static function get_attendees_from_modules( $attendees, $post_id = 0 ) {
...
+		 * @param int               $post_id   Parent post ID.
 		 *
 		 * @return array The attendee data for attendees.
 		 */
 		public function get_attendees_from_module( $attendees, $post_id = 0 ) {
+			/*
+			 * Callers may hand over a generator, e.g. from `Tribe__Repository::all( true )`, and this method
+			 * reads the set twice: once to warm the post caches and once to build the data.
+			 */
+			if ( $attendees instanceof Traversable ) {
```
larodiel, line 1713: I'm just wondering why there's an iterator being passed to this method 🤔 
It doesn't seems right, maybe we should add an array guard or something.
Maybe @redscar or @dpanta94 can help us with that.

larodiel (changes requested):

src/Tribe/Tickets.php
redscar: Looks like the reason for this piece of code is because of  ET+, Enhanced_Templates/Hooks.php:392, the only all( true ) in that repo.

The approach looks right to me. I checked whether any provider subclass overrides get_attendees_from_module(), which would mean the guard never runs on the Woo screen, and there's only the one definition, so everything inherits it. preserve_keys: false is the right choice too — the repository generator can repeat keys, so preserving them would drop attendees.

One ask before this goes in: none of the three tests hit the caching branch, they either omit $post_id or pass an empty set. So the 5.29.2 code that caused the bug still has no coverage. A call with $post_id, then a second asserting the cached return matches, would close it.

coutinhocouto: @redscar good catch, added a test for the cache path. You're right that it was the one branch with no coverage, and since it's the 5.29.2 code that caused this, it's the part that most needed it.

redscar (approved):
Approved, there was one failing test when I looked, but I think it has to do with Github outages.

larodiel (approved):

PR #4374, opened 2026-08-07
Fix: RSVP orders leak into Orders admin screen

larodiel:
### Description

**Fix** — Adds a `pre_get_posts` hook (priority 20, after `pre_filter_admin_order_table`) that excludes RSVP orders from the `edit.php?post_type=tec_tc_order` screen via `NOT LIKE` on `_tec_tc_order_items` meta. RSVP orders use the Free gateway with all items typed `tc-rsvp`.


sdokus (approved):

larodiel:
It's not a leak, it's expected to be this way

PR #4385, opened 2026-08-12
[SMTNC-1899] RSVPs made by the same user should update the same attendee record and not create a new one

larodiel:
### 🎫 Ticket

[SMTNC-1899](https://linear.app/nexcess/issue/SMTNC-1899/rsvps-made-by-the-same-user-should-update-the-same-attendee-record-and)

### 🗒️ Description

**Fix** (bug fix - duplicate attendee records)

Resolved an issue where a logged-in user clicking Go or Not Going on an RSVP repeatedly created a new attendee record each time instead of updating their existing one. `Tribe__Tickets__RSVP::generate_tickets_for()` now looks up the user's existing attendee for the ticket and updates it in place, adjusting sales/stock accordingly, rather than always inserting a duplicate.

### 🎥 Artifacts 
https://www.loom.com/share/64626db5c14b466597109b1409103501

**Testing IAC**

https://www.loom.com/share/3b6b39efab1942a983d460a5f422c4ee

### ✔️ Checklist

### DIFF
--- a/src/Tribe/RSVP.php
+++ b/src/Tribe/RSVP.php
@@ -2151,6 +2151,23 @@ class Tribe__Tickets__RSVP extends Tribe__Tickets__Tickets {
 	}
 
 	/**
+	 * Get the IDs of the attendee record(s) a user already holds for a specific RSVP ticket.
+	 *
+	 * @since TBD
+	 *
+	 * @param int $product_id The RSVP ticket (product) ID.
+	 * @param int $user_id    The user ID.
+	 *
+	 * @return int[] List of attendee IDs.
+	 */
+	public function get_attendee_ids_by_user_for_ticket( $product_id, $user_id ) {
+		/** @var Tribe__Tickets__Attendee_Repository $repository */
+		$repository = tribe_attendees( $this->orm_provider );
+
+		return $repository->by( 'ticket', $product_id )->by( 'user', $user_id )->get_ids();
+	}
+
+	/**
 	 * {@inheritdoc}
 	 *
 	 * Get all the attendees for post type. It returns an array with the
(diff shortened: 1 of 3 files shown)


larodiel:
the branch `release/M26.machamp` wasn't SYNCED with main, I've opened a PR do it https://github.com/the-events-calendar/event-tickets/pull/4386

Once it merge just will remain the files for this PR, that's why it's on draft for now.

src/Tribe/RSVP.php
```diff
@@ -2818,7 +2843,15 @@ public function generate_tickets_for( $product_id, $ticket_qty, $attendee_detail
 					'user_id'           => is_user_logged_in() ? get_current_user_id() : 0,
 				];
 
-				$attendee_ids[] = $this->create_attendee_for_ticket( $ticket_type, $attendee_data );
+				$existing_attendee_id = array_shift( $existing_attendee_ids );
+
+				if ( $existing_attendee_id ) {
+					$this->update_sales_and_stock_by_order_status( $existing_attendee_id, $attendee_order_status, $product_id );
+					$this->update_attendee( $existing_attendee_id, $attendee_data );
```
sdokus, line 2865: Will this overwrite any IAC for guests with more than one RSVP in their "order"?

sdokus:
I think this is why we have the "My RSVP" page, so users can edit their RSVPs there 🤔 I think these edits might open a can of worms if users did want to submit multiple RSVPs for some reason under the same account. I'm curious if this is reported as an issue by users?

src/Tribe/RSVP.php
larodiel: @sdokus when you click in "can't go" there's a field to select the number of guests that will not go, so, this number will be updated accordingly with the selected number.
I believe that it's more a business rule that need to be discussed, like, if the main guest is not going all guests will not go, or, when select the number, how do we know which guests will not go, or it doesn't matter as a guest is a guest.

https://www.loom.com/share/3b6b39efab1942a983d460a5f422c4ee

larodiel:
> I think this is why we have the "My RSVP" page, so users can edit their RSVPs there 🤔 I think these edits might open a can of worms if users did want to submit multiple RSVPs for some reason under the same account. I'm curious if this is reported as an issue by users?

@sdokus if you think on the logic of it, the attendee must be able to answer it once, even if this attendee has guest, the guest still being 1 attendee.
So, yes, we can have multiples RSVP, but each user can just answer the RSVP once, even a guest.
Does it makes sense?

sdokus:
This might be a heavier lift, but what do we think about changing the UI to something like this:
CleanShot 2026-08-12 at 16 31 24

Like it could say "You already have ## tickets - edit them or add a new RSVP"

larodiel:
@coderabbitai review

larodiel:
As we are waiting answer about this one, we will move it to `M26.octillery`

coutinhocouto (approved):

larodiel:
this is not a bug, and no action should be taken to correct it.

PR #4387, opened 2026-08-12
Handle lost cookies with Stripe

sdokus:
### 🎫 Ticket

[SMTNC-1983](https://linear.app/nexcess/issue/SMTNC-1983/event-tickets-stripe-checkout-never-completes-order-update-rest-route)

### 🗒️ Description

We strengthened a part of Tickets Commerce for security purposes, but ended up locking out legit purchasers who have their cart cookie get lost in the roundtrip authorization journey to Stripe, which can happen based on different site configurations. 

This PR completes Stripe orders whose cart cookie was lost in transit by authorizing the order-update request with the Payment Intent secret this site issued for that order, and replaces the endless checkout spinner with a real error message when the update still fails.

### 🎥 Artifacts 

https://www.loom.com/share/f3556dd607a04ad183fdfe2846654aa2

### ✔️ Checklist

### DIFF
--- a/src/resources/js/commerce/gateway/stripe/checkout.js
+++ b/src/resources/js/commerce/gateway/stripe/checkout.js
@@ -261,6 +261,7 @@ tribe.tickets.commerce.gateway.stripe.checkout = {};
 	 * When a successful request is completed to our Approval endpoint.
 	 *
 	 * @since 5.3.0
+	 * @since TBD Updated to handle the case where the payment intent is not present.
 	 *
 	 * @param {Object} data Data returning from our endpoint.
 	 *
@@ -269,13 +270,7 @@ tribe.tickets.commerce.gateway.stripe.checkout = {};
 	obj.handlePaymentSuccess = async ( data ) => {
 		tribe.tickets.debug.log( 'stripe', 'handlePaymentSuccess', data );
 
-		const response = await obj.handleUpdateOrder( data.paymentIntent );
-
-		// Redirect the user to the success page.
-		if ( response.redirect_url && URL.canParse( response.redirect_url ) ) {
-			window.location = response.redirect_url;
-		}
-		return true;
+		return obj.handleUpdateOrderResponse( await obj.handleUpdateOrder( data.paymentIntent ) );
 	};
 
 	/**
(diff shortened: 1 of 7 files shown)


rahulacaleffi (approved):
Looks good, @sdokus !

src/resources/js/commerce/gateway/stripe/checkout.js
```diff
@@ -288,14 +283,44 @@ tribe.tickets.commerce.gateway.stripe.checkout = {};
...
+	 *
+	 * @param {Object} response The parsed response from the order update endpoint.
+	 *
+	 * @return {boolean} Whether the order update succeeded.
+	 */
+	obj.handleUpdateOrderResponse = ( response ) => {
+		try {
+			// Not URL.canParse(): it is missing on older browsers, where calling it throws and leaves
+			// the checkout hanging on the loader. new URL() is everywhere and throws on bad input.
+			window.location = new URL( response.redirect_url ).href;
```
larodiel, line 307:
```suggestion
			window.location = response.redirect_url ? new URL( response.redirect_url ).href : '/';
```

larodiel: I would just add a guard here for an edge case of this object is empty/null/undefined

larodiel (changes requested):

src/resources/js/commerce/gateway/stripe/checkout.js
sdokus: Being in a `try` block _should_ already guard against `null`/`undefined` to throw an error instead of redirecting to the same page, which is the main goal of the PR - we want people to see the error message being thrown instead of thinking their payment didn't go through and starting another order.

Should I update the comment in the code to clarify that?

sdokus:
@coderabbitai review

src/resources/js/commerce/gateway/stripe/checkout.js
larodiel: @sdokus would be good :)
Thank you :)

sdokus:
Took the `assertNotWPError` one — `handle_create_order()` has six `WP_Error` return paths and `WP_Error` has no `get_data()`, so a regression would fatal instead of failing readably.

Skipping the other two:

**`order_id` type guard** — the premise doesn't hold. Core runs `has_valid_params()` / `sanitize_params()` in `dispatch()` (`class-wp-rest-server.php:1109-1116`) *before* `respond_to_request()` invokes the `permission_callback` (`:1253`). All three routes using this callback already declare `order_id` with an `is_string()` `validate_callback`, so an array-valued `order_id` gets a 400 before `current_user_can_edit_order()` ever runs.

**Gateway-filter test** — `get_stored_client_secret()` already requires a payload whose own `id` matches the requested Payment Intent, which a non-Stripe order never has. Dropping the `gateway` arg wouldn't open a hole, so the test would only pin defense-in-depth.

larodiel (approved):

PR #4388, opened 2026-08-13
[SOFT-4176] fix(rsvp-block): RSVPInactiveBlock: Dispatch persistRSVP thunk on set…

larodiel:
### 🎫 Ticket

[SOFT-4176](https://linear.app/nexcess/issue/SOFT-4176/rsvp-should-enable-based-on-block-editor-defaults)

### 🗒️ Description

**Fix** (bug fix - RSVP stayed inactive in the block editor)

Resolved an issue where an RSVP added via the block editor only activated after a default value (e.g. capacity or dates) was changed. The inactive block now dispatches the `persistRSVP` thunk when the add/edit panel opens, so the RSVP persists and enables with the block editor defaults immediately.

### 🎥 Artifacts 
https://www.loom.com/share/1908c6f2547c441e8b5c38bbe594465b

### ✔️ Checklist

### DIFF
--- a/src/modules/blocks/rsvp-v2/inactive-block/container.js
+++ b/src/modules/blocks/rsvp-v2/inactive-block/container.js
@@ -14,7 +14,7 @@ import { compose } from 'redux';
  * Internal dependencies
  */
 import RSVPInactiveBlock from '../../rsvp-shared/templates/inactive-block/template';
-import { actions, selectors } from '../../../data/blocks/rsvp-v2';
+import { actions, selectors, thunks } from '../../../data/blocks/rsvp-v2';
 import { withStore } from '@moderntribe/common/hoc';
 
 const mapStateToProps = ( state ) => ( {
@@ -22,7 +22,10 @@ const mapStateToProps = ( state ) => ( {
 } );
 
 const mapDispatchToProps = ( dispatch ) => ( {
-	setAddEditOpen: () => dispatch( actions.setRSVPIsAddEditOpen( true ) ),
+	setAddEditOpen: () => {
+		dispatch( actions.setRSVPIsAddEditOpen( true ) );
+		dispatch( thunks.persistRSVP() );
+	},
 } );
 
 export default compose( withStore(), connect( mapStateToProps, mapDispatchToProps ) )( RSVPInactiveBlock );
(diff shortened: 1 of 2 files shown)


src/modules/blocks/rsvp-v2/inactive-block/container.js
```diff
@@ -14,15 +14,18 @@ import { compose } from 'redux';
...
 
 const mapStateToProps = ( state ) => ( {
 	created: selectors.getRSVPCreated( state ),
 } );
 
 const mapDispatchToProps = ( dispatch ) => ( {
-	setAddEditOpen: () => dispatch( actions.setRSVPIsAddEditOpen( true ) ),
+	setAddEditOpen: () => {
+		dispatch( actions.setRSVPIsAddEditOpen( true ) );
+		dispatch( thunks.persistRSVP() );
```
sdokus, line 27: We should look into if an event has already started if this would create issues - this is what my Claude is flagging:
>   This dispatch quietly does nothing on events that have already started. `initializeRSVP` sets the RSVP window to start now and end when the event starts ([`sagas.js#L147-L171`]). On an event that's already underway, that's a window closing before it opens, so `handleRSVPDurationError `flags it — and `persistRSVP` bails at the top before it ever reaches `createRSVP` ([`thunks.js#L290`]):
 ```php
  if ( selectors.getRSVPHasDurationError( state ) ) {
      return;
  }
```
> Result: click "Add RSVP" on an in-progress or past event and the panel opens looking normal, but nothing is created.

larodiel: **What happens today on a started event**

`initializeRSVP` seeds the RSVP window as start = post date, end = event start (`sagas.js`), then runs `handleRSVPDurationError`. Two cases:

1. **Post date < event start** (e.g. event published 3 days ago, started 2 hours ago) → no duration error → `persistRSVP` creates the RSVP with the default window. This is the same payload the pre-PR "Confirm" path produced (`onConfirmClick` → `createRSVP`), so no behavior change beyond what SOFT-4176 intends.
2. **Post date ≥ event start** (backdated / in-progress event) → `handleRSVPDurationError` flags it → `persistRSVP` bails on the existing `getRSVPHasDurationError` guard (`thunks.js`) before reaching `createRSVP`. The panel still opens, but the duration-fields UI already renders the existing duration error and Confirm is disabled — identical to the pre-PR flow. No broken RSVP is silently created.

So the dispatch is safe: the guard that already protects the Confirm path protects this new auto-create path too. I did not change the container — the guard is the correct protection.

sdokus (approved):

PR #4396, opened 2026-08-17
[SMTNC-2017] Ticket_Object::inventory() N+1 and uncached attendee counts slow admin and frontend

WayneRocha:
### 🎫 Ticket

[SMTNC-2017](https://linear.app/nexcess/issue/SMTNC-2017/etetp-ticket-objectinventory-n1-and-uncached-attendee-counts-slow)

### 📹Artifact

BEFORE

image

AFTER

image

### 🗒️ Description

Reduced the database queries behind `Ticket_Object::inventory()` and `get_event_checkedin_attendees_count()`:

- Primed the post/meta caches before formatting attendee results so `get_post()` no longer issues one query per attendee (801 queries → a handful on an 800-attendee ticket).
- Cached the checked-in attendees count per request so repeated calls no longer re-query.

### ⚠️ Blockers

None.

### ✔️ Checklist
- [x] Are all the **required** tests passing? (slic unavailable locally — not run)

### DIFF
--- a/changelog/smtnc-2017-ticket-inventory-n1-and-uncached-attendee-counts
+++ b/changelog/smtnc-2017-ticket-inventory-n1-and-uncached-attendee-counts
@@ -1,4 +1,4 @@
 Significance: patch
 Type: performance
 
-Reduced the number of database queries made when calculating a ticket inventory and the checked-in attendees count, and cached the checked-in attendees count to avoid redundant lookups on admin and frontend screens.
+Cached the checked-in attendees count to avoid redundant lookups on admin and frontend screens.
--- a/changelog/smtnc-2017-ticket-inventory-n1-and-uncached-attendee-counts
+++ b/changelog/smtnc-2017-ticket-inventory-n1-and-uncached-attendee-counts
@@ -1,0 +1,4 @@
+Significance: patch
+Type: performance
+
+Reduced the number of database queries made when calculating a ticket inventory and the checked-in attendees count, and cached the checked-in attendees count to avoid redundant lookups on admin and frontend screens.
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -1955,7 +1955,7 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 		 * @static
 		 *
 		 * @param int $post_id ID of parent "event" post
-		 * @return mixed
+		 * @return int
 		 */
 		final public static function get_event_checkedin_attendees_count( $post_id ) {
 			// Post ID is required.
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -872,7 +872,16 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 			/** @var Tribe__Tickets__Attendee_Repository $repository */
 			$repository = tribe_attendees( $this->orm_provider );
 
-			return $this->get_attendees_from_module( $repository->by( 'event', $event_id )->all(), $event_id );
+			$repository = $repository->by( 'event', $event_id );
+
+			// Prime the post caches before formatting so get_post() does not hit the DB once per attendee.
+			$ids = $repository->get_ids();
+
+			if ( $ids ) {
+				tribe( 'cache' )->warmup_post_caches( $ids, true );
+			}
+
+			return $this->get_attendees_from_module( $repository->all(), $event_id );
 		}
 
 		/**
@@ -888,7 +897,16 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 			/** @var Tribe__Tickets__Attendee_Repository $repository */
 			$repository = tribe_attendees( $this->orm_provider );
 
-			return $this->get_attendees_from_module( $repository->by( 'ticket', $ticket_id )->all() );
+			$repository = $repository->by( 'ticket', $ticket_id );
+
+			// Prime the post caches before formatting so get_post() does not hit the DB once per attendee.
+			$ids = $repository->get_ids();
+
+			if ( $ids ) {
+				tribe( 'cache' )->warmup_post_caches( $ids, true );
+			}
+
+			return $this->get_attendees_from_module( $repository->all() );
 		}
 
 		/**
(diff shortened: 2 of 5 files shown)


WayneRocha:
@coderabbitai review

claudiosanches (approved):

src/Tribe/Tickets.php
```diff
@@ -872,7 +872,16 @@ protected function get_attendees_by_post_id( $event_id ) {
...
+			$repository = $repository->by( 'event', $event_id );
+
+			// Prime the post caches before formatting so get_post() does not hit the DB once per attendee.
+			$ids = $repository->get_ids();
+
+			if ( $ids ) {
+				tribe( 'cache' )->warmup_post_caches( $ids, true );
+			}
+
+			return $this->get_attendees_from_module( $repository->all(), $event_id );
```
coutinhocouto: `get_attendees_from_module()` already does this warmup — L1750-1760, added in [ff8a4bea3c](https://github.com/the-events-calendar/event-tickets/commit/ff8a4bea3c) which is on base. So we're doing it twice.

The part that worries me more is what `get_ids()` does to `all()`. It executes the memoised query, so when `all()` calls `build_query()` it gets that same executed query back and bails out early at `Repository.php:690`:

```php
return array_map( [ $this, 'format_item' ], $query->posts );
```

That path skips the `array_filter()` on L725. So any attendee where `get_post()` returns null now comes back as a literal `null` in the array instead of being dropped.

Couple of smaller side effects too: `no_found_rows` never gets set to `skip_found_rows` before the query runs, and `pre_get_posts` no longer fires for these lookups — `pre_get_ids_for_posts` fires instead.

And on the path we actually care about this is slower, not faster: `get_attendees_from_module()` returns from its own cache at L1741-1745 *before* it reaches the warmup, so we're paying for a full query plus a warmup in front of a cache hit.

Can we just drop this block? The downstream warmup already covers it.

src/Tribe/Tickets.php
```diff
@@ -888,7 +897,16 @@ protected function get_attendees_by_ticket_id( $ticket_id ) {
...
+			$repository = $repository->by( 'ticket', $ticket_id );
+
+			// Prime the post caches before formatting so get_post() does not hit the DB once per attendee.
+			$ids = $repository->get_ids();
+
+			if ( $ids ) {
+				tribe( 'cache' )->warmup_post_caches( $ids, true );
+			}
+
+			return $this->get_attendees_from_module( $repository->all() );
```
coutinhocouto: Same thing here as the event path.

One difference worth noting: this one calls `get_attendees_from_module()` without a `$post_id`, so it never hits the formatter's cache short-circuit. But the duplicated warmup and the lost `array_filter()` still apply.

changelog/smtnc-2017-ticket-inventory-n1-and-uncached-attendee-counts
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: performance
+
+Reduced the number of database queries made when calculating a ticket inventory and the checked-in attendees count, and cached the checked-in attendees count to avoid redundant lookups on admin and frontend screens.
```
coutinhocouto: This credits an inventory improvement the diff doesn't make — `Ticket_Object.php` isn't touched, and the attendee getters don't reduce the query count either.

```suggestion
Cached the checked-in attendees count so repeated lookups on admin and frontend screens no longer re-query.
```

tests/wpunit/Tribe/Tickets/TicketsTest.php
```diff
@@ -230,6 +230,29 @@ public function should_allow_fetching_attendees_checkedin_count() {
...
+		update_post_meta( $attendee_ids[0], $rsvp_main->checkin_key, 1 );
+
+		$this->assertEquals( 1, Tickets::get_event_checkedin_attendees_count( $post_id ) );
+
+		// Check in a second attendee; the cached count should still be returned for the rest of the request.
+		update_post_meta( $attendee_ids[1], $rsvp_main->checkin_key, 1 );
+
+		$this->assertEquals( 1, Tickets::get_event_checkedin_attendees_count( $post_id ) );
+	}
+
```
coutinhocouto, line 261: This passes if the cache works, but it'd also pass if the method were broken into always returning its first result — the assertion can't tell those apart.

Could we assert a `$wpdb->num_queries` delta across the two calls instead? That actually proves the second one didn't hit the DB.

src/Tribe/Tickets.php
```diff
@@ -1940,10 +1958,27 @@ public static function find_matching_event( $possible_ticket ) {
 		 * @return mixed
```
coutinhocouto: Both paths return an int now — `0` from the new guard, `found()` otherwise. Docblock-only so no BC risk on the `final public static`.

```suggestion
	 * @return int
```

coutinhocouto (changes requested):

src/Tribe/Tickets.php
WayneRocha: Addresses in [ba8e18e](https://github.com/the-events-calendar/event-tickets/pull/4396/commits/ba8e18e462f7f8ef66d76b7979bb9ff939146f68)

tests/wpunit/Tribe/Tickets/TicketsTest.php
WayneRocha: Addressed in [1507a84](https://github.com/the-events-calendar/event-tickets/pull/4396/commits/1507a84b09e6165733799c3bc621b6928551dafa)

changelog/smtnc-2017-ticket-inventory-n1-and-uncached-attendee-counts
WayneRocha: Addressed in [9045642](https://github.com/the-events-calendar/event-tickets/pull/4396/commits/90456423035c74d4096864e4afcfa21fdb5753ae)

src/Tribe/Tickets.php
WayneRocha: [1507a84](https://github.com/the-events-calendar/event-tickets/pull/4396/commits/1507a84b09e6165733799c3bc621b6928551dafa)

src/Tribe/Tickets.php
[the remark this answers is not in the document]
WayneRocha: Resolved after @coutinhocouto suggestions applied. Re-check for more points

coutinhocouto (approved):

PR #4408, opened 2026-08-18
[SMTNC-1468] fix: Auto-refresh the Square OAuth access token

claudiosanches:
### 🎫 Ticket

[SMTNC-1468]

### 🗒️ Description

Square access tokens expire about 30 days after they are minted, and nothing renewed them. Checkout started failing at Square's authorization layer while the admin still reported the gateway as connected, and the only recovery was disconnect and reconnect. `WhoDat::refresh_token()` already existed with zero callers, and the `expires_at` the OAuth callback stores was never read.

**The fix.** `Square\Requests::request()` renews the token before signing a request whose expiration is due, and retries once when Square rejects it anyway. That path was chosen over a scheduled action because it still runs on a site whose cron queue is stalled, which is where the reports come from.

**What only surfaced against the live service.** Running this against WhoDat with a Square sandbox account turned up two things:

- The refresh endpoint answers `GET` with **405 Method Not Allowed** and only accepts `POST`. The existing method sent a GET, so it could never have worked.
- A rejected refresh token comes back as an **HTML 500**, not a machine readable error, so the response body cannot tell a dead connection apart from a WhoDat outage. Failures are judged by status code instead. A 401 or 403 is final on its own; anything else has to be corroborated by `oauth/token/status`. That endpoint reports on the *access* token, which on this path is expired by definition, so its refusal only carries information while the access token is still inside its own lifetime. Otherwise a rejection is final only once five failures have piled up over more than a day. A transport failure never marks the connection unavailable, it only feeds an exponential backoff.

**When renewal fails.** The connection is marked unavailable: `Merchant::is_connected()` returns false, the settings screen shows the connect panel rather than a false "Connected", the gateway drops off checkout, and an admin notice explains it with a reconnect link. That is a long backoff rather than a dead end. It is rechecked every 12 hours, with an `admin_init` trigger so a site with no traffic heals without anyone touching the settings.

**Notes for review.**

- Square's status is stored in a new mode-scoped option rather than through `Merchant::set_merchant_unauthorized()`. Those option names are byte-identical to Stripe's and are not mode-scoped, so writing them from Square would flip Stripe's state, which `Stripe\Hooks::handle_stripe_errors()` reads.
- `check_account_status()` was reading an `error` key the status endpoint does not return, so even an explicit `is_connected( true )` recheck reported a revoked token as connected. It now reads the `type` and `message` shape the endpoint actually sends.
- The refresh takes a cross-process lock so parallel requests do not stampede the endpoint. Confirmed against the live service that the refresh token does not rotate, so a lost race is harmless.
- Connections made before `expires_at` was tracked are never treated as expired. They retry at most twice a day until a refresh teaches them a real expiration, and the 401 path covers them meanwhile.

### 🎥 Artifacts

https://www.loom.com/share/5bb28e2887f7457b8cf04f8a68d9fb5d

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Commerce/Gateways/Square/Merchant.php
+++ b/src/Tickets/Commerce/Gateways/Square/Merchant.php
@@ -247,7 +247,7 @@ class Merchant extends Abstract_Merchant {
 	public function save_signup_data( array $signup_data ): bool {
 		unset( $signup_data['state'] );
 
-		$this->delete_token_status();
+		$this->delete_refresh_status();
 
 		return update_option( $this->get_signup_data_key(), $signup_data );
 	}
@@ -261,7 +261,7 @@ class Merchant extends Abstract_Merchant {
 	 *
 	 * @return string
 	 */
-	public function get_token_status_option_key(): string {
+	public function get_refresh_status_option_key(): string {
 		$gateway_key = Gateway::get_key();
 		$mode        = $this->get_mode();
 
@@ -271,12 +271,15 @@ class Merchant extends Abstract_Merchant {
 	/**
 	 * Returns the outcome of the last access token refresh attempt.
 	 *
+	 * The timestamps are UTC, so that they compare correctly against strtotime(), which WordPress pins
+	 * to UTC whatever the site timezone is.
+	 *
 	 * @since TBD
 	 *
-	 * @return array{invalid_at: string, error: string, failures: int, last_attempt_at: string}
+	 * @return array{invalid_at: string, error: string, failures: int, last_attempt_at: string, first_failure_at: string}
 	 */
-	public function get_token_status(): array {
-		$status = get_option( $this->get_token_status_option_key(), [] );
+	public function get_refresh_status(): array {
+		$status = get_option( $this->get_refresh_status_option_key(), [] );
 
 		if ( ! is_array( $status ) ) {
 			$status = [];
@@ -284,13 +287,25 @@ class Merchant extends Abstract_Merchant {
 
 		return array_merge(
 			[
-				'invalid_at'      => '',
-				'error'           => '',
-				'failures'        => 0,
-				'last_attempt_at' => '',
+				'invalid_at'       => '',
+				'error'            => '',
+				'failures'         => 0,
+				'last_attempt_at'  => '',
+				'first_failure_at' => '',
 			],
 			$status
 		);
+	}
+
+	/**
+	 * How many consecutive refresh attempts have failed.
+	 *
+	 * @since TBD
+	 *
+	 * @return int
+	 */
+	public function get_refresh_failure_count(): int {
+		return (int) $this->get_refresh_status()['failures'];
 	}
 
 	/**
@@ -302,8 +317,8 @@ class Merchant extends Abstract_Merchant {
 	 *
 	 * @return bool
 	 */
-	public function update_token_status( array $data ): bool {
-		return update_option( $this->get_token_status_option_key(), array_merge( $this->get_token_status(), $data ), false );
+	public function update_refresh_status( array $data ): bool {
+		return update_option( $this->get_refresh_status_option_key(), array_merge( $this->get_refresh_status(), $data ), false );
 	}
 
 	/**
@@ -313,8 +328,8 @@ class Merchant extends Abstract_Merchant {
 	 *
 	 * @return bool
 	 */
-	public function delete_token_status(): bool {
-		return delete_option( $this->get_token_status_option_key() );
+	public function delete_refresh_status(): bool {
+		return delete_option( $this->get_refresh_status_option_key() );
 	}
 
 	/**
@@ -378,7 +393,7 @@ class Merchant extends Abstract_Merchant {
 	 * @return bool
 	 */
 	public function is_token_invalid(): bool {
-		return ! empty( $this->get_token_status()['invalid_at'] );
+		return ! empty( $this->get_refresh_status()['invalid_at'] );
 	}
 
 	/**
@@ -397,7 +412,7 @@ class Merchant extends Abstract_Merchant {
 
 		$update = [
 			'expires_at'   => '',
-			'refreshed_at' => Dates::build_date_object()->format( Dates::DBDATETIMEFORMAT ),
+			'refreshed_at' => Dates::build_date_object( 'now', 'UTC' )->format( Dates::DBDATETIMEFORMAT ),
 		];
 
 		// Only overwrite what the response actually carried; omitted keys keep their current value.
@@ -426,7 +441,7 @@ class Merchant extends Abstract_Merchant {
 	 */
 	public function flush_option_cache(): void {
 		wp_cache_delete( $this->get_signup_data_key(), 'options' );
-		wp_cache_delete( $this->get_token_status_option_key(), 'options' );
+		wp_cache_delete( $this->get_refresh_status_option_key(), 'options' );
 		wp_cache_delete( 'alloptions', 'options' );
 	}
 
@@ -455,7 +470,8 @@ class Merchant extends Abstract_Merchant {
 			return $return;
 		}
 
-		$accepted = tribe( WhoDat::class )->is_token_accepted();
+		// Forced: this is the explicit recheck, and a token refreshed moments ago outdates the cache.
+		$accepted = tribe( WhoDat::class )->is_token_accepted( true );
 
 		if ( null === $accepted ) {
 			$return['errors'][] = __( 'Unable to connect to Square.', 'event-tickets' );
@@ -463,7 +479,7 @@ class Merchant extends Abstract_Merchant {
 		}
 
 		if ( ! $accepted ) {
-			$status = tribe( WhoDat::class )->get_token_status();
+			$status = tribe( WhoDat::class )->get_token_status( true );
 
 			$return['errors'][] = $status['message'] ?? $status['error_description'] ?? __( 'Square no longer accepts the stored credentials.', 'event-tickets' );
 			return $return;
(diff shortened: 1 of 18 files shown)


src/Tickets/Commerce/Gateways/Square/Token_Refresher.php
```diff
@@ -0,0 +1,689 @@
...
+	protected const LOCK_TIMEOUT = 60;
+
+	/**
+	 * How long to keep polling for the winner's result when the lock is contended, in microseconds.
+	 *
+	 * @since TBD
+	 *
+	 * @var int
+	 */
+	protected const LOCK_WAIT = 1000000;
```
coutinhocouto: One second isn't long enough for the job this is doing. The holder can spend 3s on the POST and another 5s inside `is_rejection_final()`, that wp_remote_get passes no timeout, so it takes WP's default, which is ~8s worst case, and every waiter has already walked away at 1s. When they do, `do_refresh()` returns false, `Requests::request()` bails at line 128, and the shopper gets the 401 anyway. Which is the thing the docblock on `wait_for_lock_holder()` says this is here to stop. Either bump `LOCK_WAIT `past the holder's worst case, or put a lid on the holder.

src/Tickets/Commerce/Gateways/Square/Token_Refresher.php
```diff
@@ -0,0 +1,689 @@
...
+	 * either way. It only carries independent information while the access token is still inside its
+	 * own lifetime. When it does not, the connection is written off only after failures that keep
+	 * repeating over more than a day, which an outage does not do.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	protected function is_rejection_final(): bool {
+		if ( false !== $this->who_dat->is_token_accepted( true ) ) {
```
coutinhocouto: This lands in Abstract_WhoDat::get(), which is a plain wp_remote_get( $url ), so WP's 5s default. You'd already thought about this for the refresh POST ("This runs inside a checkout request, so it may not hold it open for WordPress's default") and capped it at 3s, but this call is on the same checkout request and didn't get the same treatment. Worst case the shopper sits through 3s + 5s. Worth passing a timeout down here too.

src/Tickets/Commerce/Gateways/Square/Merchant.php
```diff
@@ -268,15 +470,18 @@ public function check_account_status( array $client_data = [] ): array {
 			return $return;
 		}
 
-		$status = tribe( WhoDat::class )->get_token_status();
+		// Forced: this is the explicit recheck, and a token refreshed moments ago outdates the cache.
+		$accepted = tribe( WhoDat::class )->is_token_accepted( true );
```
coutinhocouto: We're paying for this status twice, is_token_accepted( true ) fetches it and drops the body, then 482 fetches the same response again just to get at message. Both skip the cache, so a failed recheck is two round trips. Could the first call hand the status back so the verdict and the message both come off it? CodeRabbit flagged this one as well and the thread's still open.

src/Tickets/Commerce/Gateways/Square/Token_Refresher.php
```diff
@@ -0,0 +1,689 @@
...
+
+	/**
+	 * Refreshes the access token when it is close enough to expiring.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool Whether the stored credentials are known to be current.
+	 */
+	public function refresh_if_needed(): bool {
+		if ( ! $this->should_refresh() ) {
```
coutinhocouto, line 153: Everything below this: get_api_url, process_response, get_environment_url, get_headers, is public and was here already, and the file was all-public before this PR. Mind dropping the helper underneath them?

src/Tickets/Commerce/Gateways/Square/Token_Refresher.php
```diff
@@ -0,0 +1,689 @@
...
+	protected function do_refresh( string $reason, bool $forced = false ): bool {
+		if ( $this->refreshing ) {
+			return false;
+		}
+
+		$previous_token = $this->merchant->get_access_token();
+
+		if ( ! $this->create_lock() ) {
+			// Only a caller whose request already failed waits; a proactive refresh has nothing to gain
+			// from holding a page load open while somebody else does the work.
```
coutinhocouto:
```suggestion
			/**
			* Only a caller whose request already failed waits; a proactive refresh has nothing to gain
			* from holding a page load open while somebody else does the work.
			*/
```

src/Tickets/Commerce/Gateways/Square/Token_Refresher.php
```diff
@@ -0,0 +1,689 @@
...
+	 */
+	protected function is_rejection_final(): bool {
+		if ( false !== $this->who_dat->is_token_accepted( true ) ) {
+			return false;
+		}
+
+		$expiration = $this->merchant->get_token_expiration();
+
+		// Square refusing a token that is still inside its own lifetime means the grant itself is gone.
+		// An expiration we never recorded proves nothing either way, so it does not count here.
```
coutinhocouto:
```suggestion
		/**
		* Square refusing a token that is still inside its own lifetime means the grant itself is gone.
		* An expiration we never recorded proves nothing either way, so it does not count here.
		*/
```

tests/square_integration/Requests_Test.php
```diff
@@ -0,0 +1,350 @@
...
+			'expires_at'    => gmdate( 'Y-m-d\TH:i:s\Z', time() + 30 * DAY_IN_SECONDS ),
+		];
+
+		$this->square_response_override = null;
+
+		self::$current = $this;
+		add_filter( 'pre_http_request', [ __CLASS__, 'route_request' ], 10, 3 );
+
+		// The suite bootstrap turns any error or warning log into an exception; these tests assert on
+		// failure paths, which log on purpose. Lifted by name so the rest of $wp_filter stays untouched.
```
coutinhocouto:
```suggestion
         /**
		* The suite bootstrap turns any error or warning log into an exception; these tests assert on
		* failure paths, which log on purpose. Lifted by name so the rest of $wp_filter stays untouched.
		*/
```

tests/square_integration/_bootstrap.php
```diff
@@ -40,6 +40,9 @@ function tec_tickets_tests_get_fake_merchant_data(): array {
 		'merchant_country'  => 'US',
 		'merchant_currency' => 'USD',
 		'whodat_signature'  => 'whodat-signature',
+		// Fixed, not relative: this function is re-called inside assertions. Far enough out that no
+		// suite drifts into the token refresh window and starts making outbound requests.
```
coutinhocouto:
```suggestion
         /**
  		 * Fixed, not relative: this function is re-called inside assertions. Far enough out that no
  		 * suite drifts into the token refresh window and starts making outbound requests.
		*/
```

tests/square_integration/_bootstrap.php
```diff
@@ -57,14 +60,18 @@ function tec_tickets_tests_get_fake_merchant_data(): array {
 // Enable pretty permalinks.
 update_option( 'permalink_structure', '/%postname%/' );
 
-// When we have logs of level error, critical, warning, throw an exception.
-add_action( 'tribe_log', static function ( $level, $message, $context ) {
+// When we have logs of level error, critical, warning, throw an exception. Kept in a global so that the
+// tests asserting on failure paths can lift it by name: emptying $wp_filter['tribe_log'] instead risks
+// being caught in WordPress's one-time hook snapshot, which would silently disarm this for a whole run.
```
coutinhocouto:
```suggestion
/**
* When we have logs of level error, critical, warning, throw an exception. Kept in a global so that the
* tests asserting on failure paths can lift it by name: emptying $wp_filter['tribe_log'] instead risks
* being caught in WordPress's one-time hook snapshot, which would silently disarm this for a whole run.
*/
```

src/Tickets/Commerce/Gateways/Square/Notices_Controller.php
```diff
@@ -445,6 +487,92 @@ public function render_currency_mismatch_notice(): string {
...
+		if ( ! $this->gateway->is_enabled() ) {
+			return false;
+		}
+
+		// The rejected notice covers this case with a clearer message.
+		if ( $this->merchant->is_token_invalid() ) {
+			return false;
+		}
+
+		if ( $this->merchant->get_refresh_failure_count() < 3 ) {
```
coutinhocouto: Where does the 3 come from? The disconnect threshold next door is `Token_Refresher::PERSISTENT_FAILURES = 5` with a docblock explaining itself, so this one being a bare literal stands out a bit. Give it a name?

tests/square_integration/Token_Refresher_Test.php
```diff
@@ -0,0 +1,755 @@
...
+	 * The query args WhoDat was called with, one entry per call.
+	 *
+	 * @var array
+	 */
+	protected array $whodat_calls = [];
+
+	/**
+	 * What the token status endpoint reports about the stored access token, decoded or as a raw body.
+	 *
+	 * @var mixed
```
coutinhocouto: The docblock right above says it's either the decoded array or the raw JSON string, and the tests set both, so `array|string` rather than `mixed`?

coutinhocouto (changes requested):
Some comments

> [!TIP]
> I'm seeing that the PR is quite big, I would recommend that next time it could be plotted on stacked ones, since this one has 2776 LOC and if it hits 1000 it should be splitted

coutinhocouto (approved):
@claudiosanches seems okay for me now 😸

WayneRocha (approved):

PR #4420, opened 2026-08-21
Stop the PayPal connect button from linking to a 404 when signup fails

claudiosanches:
### 🎫 Ticket

[SMTNC-1915]

### 🗒️ Description

`Signup::generate_url()` returns `false` when the signup request fails, and `esc_url( false )` emits an empty string, so the template rendered:

```html
<a href="&displayMode=minibrowser" id="connect_to_paypal" ...>
```

No scheme, no leading `/`, no `?`, so it resolves against the current directory and "Connect Automatically with PayPal" navigates to `/wp-admin/&displayMode=minibrowser`. Nothing told the admin it had failed.

**What changed**

- The template renders an error notice in place of the anchor when there is no URL.
- `displayMode` is folded in with `add_query_arg()`, in the template and in `ajax_refresh_connect_url()`.
- The country selector updates the Connect button again, and reveals the same notice when the refresh fails.
- A failed generation drops the cached signup URL.

**Notes for review**

The signup transient has no country in its key and is only written on success, so after a failed country change the previous country's URL was still cached. Reloading served it and rendered a healthy-looking button for the wrong country. Dropping it on failure is what makes "reload this page to try again" true instead of misleading.

The country selector has been inert since 624d1b187e, which renamed the template classes from `paypal` to `gateway` and left `obj.selectors.button` on the old name. Nothing in that handler ran, which is why the error and disabled handling reads as new code rather than a tweak.

The snapshot diff shows `&` becoming `&#038;`. The whole URL now passes through `esc_url()`, where before `&displayMode=minibrowser` was concatenated after it. Browsers decode both identically.

### 🎥 Artifacts 

https://www.loom.com/share/05f87aa9322c4a64b8c75072772b6e8b

### ✔️ Checklist


changelog/fix-smtnc-1915-paypal-connect-link
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixed the Tickets Commerce PayPal "Connect Automatically" button linking to a 404 when the connection link could not be generated. An error notice is shown in its place.
```
coutinhocouto: Fixed word here is wrong because on the changelog will appear `Fix - Fixed ...`

I recomment to change to `Resolved`

changelog/fix-smtnc-1915-paypal-connect-link
```diff
@@ -0,0 +1,4 @@
+Significance: patch
```
coutinhocouto: This changelog entry and next one is wrong, it must use the chagelogger and generate a yaml file

changelog/fix-smtnc-1915-paypal-country-change
```diff
@@ -0,0 +1,4 @@
+Significance: patch
```
coutinhocouto: Here the same file issue and the reword needed

tests/commerce_integration/TEC/Tickets/Commerce/Gateways/PayPal/SignupTest.php
```diff
@@ -0,0 +1,208 @@
...
+		if ( ! preg_match( '/\bhref="([^"]*)"/', $anchor[0], $href ) ) {
+			return null;
+		}
+
+		return html_entity_decode( $href[1] );
+	}
+
+	/**
+	 * A signup URL that fails to generate used to render `href="&displayMode=minibrowser"`, which the
+	 * browser resolves against /wp-admin/.
```
coutinhocouto:
```suggestion
	 * A bare `&displayMode=minibrowser` href has no scheme and no leading `/`, so the browser resolves
	 * it against /wp-admin/.
```

tests/commerce_integration/TEC/Tickets/Commerce/Gateways/PayPal/SignupTest.php
```diff
@@ -0,0 +1,208 @@
...
+		$this->assertSame( 'error', $this->logs[0][0] );
+		$this->assertStringContainsString( 'PayPal signup URL could not be generated', $this->logs[0][1] );
+	}
+
+	/**
+	 * The cached URL carries no country, so a failed refresh must not leave the previous country's URL
+	 * behind for the next page load to serve as if nothing went wrong.
+	 */
+	public function test_failed_refresh_discards_the_cached_url(): void {
+		$this->stub_whodat( 200, $this->whodat_signup_body( 'https://www.paypal.com/onboard?token=US' ) );
```
coutinhocouto, line 127: > [!NOTE]  
> **Question not a blocker:**
> delete_transient_data() drops the whole payload, but only links[1].href (the signup URL) is country-scoped. links[0].href goes with it, and that's what get_referral_data_link() (line 157) hands to On_Boarding_Endpoint::handle_signup_redirect(). On the $force = true path the cache can hold a complete valid payload, so if a seller changes country and that refresh fails while they have PayPal onboarding open, they return to get_referral_data_link() === false. Worth scoping the invalidation to the signup link, or confirming the case isn't reachable.

tests/commerce_integration/TEC/Tickets/Commerce/Gateways/PayPal/SignupTest.php
```diff
@@ -0,0 +1,208 @@
...
+		// The notice ships hidden so the refresh handler can reveal it without a page load.
+		$this->assertStringContainsString( 'PayPal connection unavailable', $html );
+		$this->assertStringContainsString( 'display: none;', $html );
+
+		$this->assertCount( 0, $this->logs, 'A successful request should log nothing.' );
+	}
+
+	/**
+	 * The concatenation this replaced produced `...entry&displayMode=...` as a single path segment when
+	 * the signup URL carried no query string of its own.
```
coutinhocouto:
```suggestion
	 * `add_query_arg()` prepends `?` when the signup URL carries no query string of its own.
```

coutinhocouto (changes requested):

tests/commerce_integration/TEC/Tickets/Commerce/Gateways/PayPal/SignupTest.php
claudiosanches: `generate_url()` rotates the signup hash before it calls WhoDat, so a country change invalidates an open onboarding session whether the refresh succeeds or fails. `handle_signup_redirect()` checks that hash first and redirects with `invalid-paypal-signup-hash`, so it never reaches `get_referral_data_link()`.

Covered by `test_failed_refresh_invalidates_an_open_onboarding_session`.

coutinhocouto (approved):

PR #4423, opened 2026-08-24
Force the updated Tickets and RSVP experience on and drop the settings

claudiosanches:
### 🎫 Ticket

[SMTNC-1227]

### 🗒️ Description

Turns the updated Tickets and RSVP experience on for every site and removes the settings that could turn it off.

**The problem.** The two toggles only rendered for sites first installed before 5.0.3, or 5.0 for the RSVP one. Old installs could still switch the new views off.

**The fix.** `tribe_tickets_new_views_is_enabled()` and its RSVP equivalent return true and no longer read the option or the install date. `Tribe__Tickets__Updater` writes both options true from its constant update callbacks, and the Display section goes away with `Tribe__Tickets__Admin__Display_Settings` moving to `src/deprecated`. A dismissible notice tells a switched site that its forms changed, keyed off the migration record so fresh installs do not see it.

**Kept on purpose.** The `TRIBE_TICKETS_NEW_VIEWS` and `TRIBE_TICKETS_RSVP_NEW_VIEWS` constants, their environment variables, and both filters. Each needs file access, so none is reachable from the settings screen, and the filters are how `TicketsBlock_TestCase` still renders V1. Now documented on both functions: assets register on `init` at priority 10 and read them, so a filter added later renders old markup against new assets.

**Notes for review.** The migration runs off the version bump rather than a version key, because a key at or below a released version is skipped when the site already recorded it, and one above the shipping version is skipped too. `tickets.admin.settings.display` is deliberately not rebound to the deprecated class, since di52 resolves a string class name eagerly and would load `src/deprecated` on every request. An org-wide search found no caller. The V1 branches at all 21 call sites are left in place to keep this diff reviewable.

### 🎥 Artifacts 

https://www.loom.com/share/c570f34e27954145969331887d070677

### ✔️ Checklist


src/Tribe/Updater.php
```diff
@@ -28,13 +30,84 @@ public function is_new_install() {
...
+		$enabled = [
+			'tickets_use_new_views'      => tribe_tickets_new_views_is_enabled(),
+			'tickets_rsvp_use_new_views' => tribe_tickets_rsvp_new_views_is_enabled(),
 		];
+
+		$switched = false;
+
+		foreach ( $enabled as $option => $is_enabled ) {
+			// A site holding the views off through the constant, the env var or a filter keeps a
+			// stored value that agrees with what it actually renders.
```
coutinhocouto:
```suggestion
			/** 
			* A site holding the views off through the constant, the env var or a filter keeps a
			* stored value that agrees with what it actually renders.
			*/
```

changelog/smtnc-1227-force-enable-updated-tickets-experience
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: fix
```
coutinhocouto:
```suggestion
Type: removed
```

coutinhocouto: The prose describes a removal plus a forced default, it doesn't read as a defect being resolved, which is what fix renders as. Compare the sibling entry here, which frames itself around "…to avoid an issue where…".

src/Tribe/Updater.php
```diff
@@ -28,13 +30,84 @@ public function is_new_install() {
...
+	 * at all, and what it rendered then depended on when it was first installed. This reproduces that
+	 * answer, and it is the only thing left that needs to.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $option The option to resolve.
+	 *
+	 * @return bool
+	 */
+	protected function was_rendering_new_views( $option ) {
```
coutinhocouto, line 128: The order in this file is now public → protected → public:

```php
 57 |	public function migrate_force_new_views() {             // public
102 |	protected function was_rendering_new_views( $option ) { // protected
118 |	public function migrate_4_12_hide_attendees_list() {    // public
```

Move the helper below `migrate_4_12_hide_attendees_list()` so it reads public → protected as usual.

tests/wpunit/Tribe/Tickets/Template_Tags/NewViewsIsEnabledTest.php
```diff
@@ -0,0 +1,231 @@
...
+			tribe_update_option( $option, $value );
+		}
+	}
+
+	/**
+	 * Puts the site in the state that used to show the settings toggles: an install old enough to
+	 * predate the new views, with both of them turned off.
+	 */
+	protected function given_an_old_install_with_the_views_turned_off(): void {
+		tribe_update_option( 'previous_event_tickets_versions', [ '4.9.2' ] );
```
coutinhocouto: `given_an_old_install_with_the_views_turned_off()` is sitting ahead of twelve public `@test` methods. Test classes follow the same member order as production code, so this belongs at the bottom. The `@before`/`@after` hooks above it are fine where they are.

tests/integration/Deprecated_Test.php
```diff
@@ -88,4 +88,43 @@ public function test_deprecated_classes_5_6_5( string $class ): void {
...
+			[ 'Tribe__Tickets__Admin__Display_Settings' ],
+		];
+	}
+
+	/**
+	 * @dataProvider deprecated_classes_tbd
+	 */
+	public function test_deprecated_classes_tbd( string $class ): void {
+		if ( class_exists( $class, false ) ) {
+			$this->markTestSkipped( $class . 'was already loaded' );
```
coutinhocouto:
```suggestion
			$this->markTestSkipped( $class . ' was already loaded' );
```

src/Tickets/Admin/Notice_New_Views_Upgrade.php
```diff
@@ -0,0 +1,108 @@
...
+
+		// The notice is about a setting that was removed, so it is only meaningful to someone who had it.
+		if ( ! current_user_can( $settings->requiredCap ) ) {
+			return false;
+		}
+
+		/** @var \Tribe__Admin__Helpers $admin_helpers */
+		$admin_helpers = tribe( 'admin.helpers' );
+
+		return $admin_helpers->is_screen() || $admin_helpers->is_post_type_screen();
```
coutinhocouto, line 77: Testing-instruction step 1 says "load any wp-admin page, expect the warning", this gate doesn't do that. `is_screen()` with no argument wants the current screen id to contain `tribe` or `tec`, and `is_post_type_screen()` wants a `Tribe__Main::get_post_types()` post type. On Dashboard, Plugins, Users or Settings → General, neither is true.

I ran the fixture from the testing instructions on a local site at this commit rather than just reading it: with `tec_tickets_new_views_forced_on` genuinely `true`, the notice was **absent** on `/wp-admin/index.php` and **present** on Tickets → Settings → Tickets. So QA bounces this on its first step.

Two honest ways out, and I don't think it's my call which:

```php
// (a) make it genuinely site-wide — drop the screen gate
return true;

// (b) keep it scoped, and reword step 1 to name the screens it really covers
return $admin_helpers->is_screen() || $admin_helpers->is_post_type_screen();
```

Worth knowing the new wpunit coverage can't catch this either way, `it_should_not_offer_the_upgrade_notice_to_a_user_who_cannot_change_settings` only ever sets `tribe_events_page_tec-tickets-settings`, so it's already scoped to the passing case.

src/Tribe/Updater.php
```diff
@@ -28,13 +30,84 @@ public function is_new_install() {
...
+	 * that already shipped never runs, because the site recorded that version when it got there, and
+	 * a key above the shipping version never runs either, so the correct key is whatever release this
+	 * turns out to go out in. Writing an option that is already true costs nothing.
+	 *
+	 * @since TBD
+	 */
+	public function migrate_force_new_views() {
+		$enabled = [
+			'tickets_use_new_views'      => tribe_tickets_new_views_is_enabled(),
+			'tickets_rsvp_use_new_views' => tribe_tickets_rsvp_new_views_is_enabled(),
```
coutinhocouto, line 73: There's a timing contradiction between this and the docblocks the same PR adds. `run_updates()` is hooked at priority **0**:

```php
// src/Tribe/Main.php:744
add_action( 'init', [ $this, 'run_updates' ], 0, 0 );
```

but `src/template-tags/tickets.php` now tells integrators the flag is read at priority **10** — "Assets are registered on `init` at priority 10 and read this, so a filter added after that point renders the old markup against the new assets."

Anything registered in that 1–10 window isn't attached yet when the migration calls these functions. So the migration sees `true`, writes both options `true`, and sets `tec_tickets_new_views_forced_on` — which fires "Your Ticket and RSVP forms have been updated" on a site that then honours the filter and renders the **old** views for the rest of the request. Wrong options, wrong notice, and it lands on exactly the population the filter was kept for.

The filter is one of the three escape hatches the description says are kept on purpose, so I'd fix this rather than document around it:

```php
// run updates after plugins and themes have had their say
add_action( 'wp_loaded', [ $this, 'run_updates' ], 0, 0 );
```

If `init:0` has to stay, then say so plainly on both functions — filters must be registered before `init` priority 0, not 10.

src/Tribe/Admin/Notices.php
```diff
@@ -25,90 +26,12 @@ public function hook() {
 			return;
 		}
 
-		add_action( 'admin_init', [ $this, 'maybe_display_rsvp_new_views_options_notice' ] );
```
coutinhocouto, line 28: `maybe_display_rsvp_new_views_options_notice()` is gone entirely, along with its `add_action`. It was a public method on a public class, so anything third-party calling it now fatals.

This file already knows how to handle that — 270 lines down it keeps the previously-removed FSE notice alive as a no-op:

```php
	/**
	 * Will be removed. Doing nothing other than triggering a deprecation notice.
	 *
	 * @since 5.17.0
	 * @deprecated 5.18.0 This method will be removed in a future release.
	 *
	 * @return void
	 */
	public function maybe_display_fse_ar_page_notice() {
		_deprecated_function( __METHOD__, '5.18.0' );
	}
```

Same treatment here:

```php
	/**
	 * Will be removed. Doing nothing other than triggering a deprecation notice.
	 *
	 * @since 4.12.3
	 * @deprecated TBD This method will be removed in a future release.
	 *
	 * @return void
	 */
	public function maybe_display_rsvp_new_views_options_notice() {
		_deprecated_function( __METHOD__, 'TBD' );
	}
```

Unhooking it was right. Deleting it is the part I'd change.

src/Tribe/Updater.php
```diff
@@ -28,13 +30,84 @@ public function is_new_install() {
...
 		];
+
+		$switched = false;
+
+		foreach ( $enabled as $option => $is_enabled ) {
+			// A site holding the views off through the constant, the env var or a filter keeps a
+			// stored value that agrees with what it actually renders.
+			if ( ! $is_enabled ) {
+				continue;
+			}
```
coutinhocouto: Take a site that had the option stored as `true` and is now held off by the constant, the env var or a filter. `$is_enabled` is `false`, we `continue`, and that stale `true` stays on disk — which is the opposite of what the comment directly above says is happening.

It also walks into the promise made twenty lines up:

> They are written all the same, for anything outside the plugin that reads them directly.

An external consumer reading the raw option on such a site gets the wrong answer, and that's the one class of site the sentence was written for. Either write the resolved value in both directions, or drop the promise — right now the code and the docblock disagree about what this method guarantees.

tests/integration/Deprecated_Test.php
```diff
@@ -88,4 +88,43 @@ public function test_deprecated_classes_5_6_5( string $class ): void {
 		$this->setExpectedDeprecated( $file_path );
 		$this->assertTrue( class_exists( $class ) || interface_exists( $class ), 'Class "' . $class . '" does not exist.' );
 	}
+
+	public function deprecated_classes_tbd() {
```
coutinhocouto: The release tooling swaps `TBD` in docblocks, not in identifiers — so unlike the `@since TBD` tags everywhere else in this PR, these two names ship as `_tbd` and stay that way.

Every other provider in this same file is named for its version:

```php
	public function deprecated_classes_4_0()   { /* … */ }
	public function deprecated_classes_3_10()  { /* … */ }
	public function deprecated_classes_5_6_5() { /* … */ }
```

Name these for whatever release this goes out in, and keep them in step with the version you end up passing to `_deprecated_file()` in the shim.

changelog/smtnc-1227-force-enable-updated-tickets-experience
coutinhocouto: The prose describes a removal plus a forced default — it doesn't read as a defect being resolved, which is what `fix` renders as. Compare the sibling entry in this directory, which frames itself around "…to avoid an issue where…". `removed` is a valid type for this repo's changelogger.

```suggestion
Type: removed
```

If you'd rather keep `fix`, the other way round works too — reword the prose so it leads with the problem (customers could still switch the new views off and break their Ticket blocks) rather than with the removal.

coutinhocouto (changes requested):

src/Tribe/Updater.php
coutinhocouto: Claude likes do use this sometimes, I think you need to add a specific rule so multiline comments comes with the correct format 😄

src/Tickets/Admin/Notice_New_Views_Upgrade.php
claudiosanches: I'm going to update the testing instructions. We should see this notice only on the plugin's pages.

src/Tribe/Updater.php
```diff
@@ -28,13 +30,69 @@ public function is_new_install() {
...
+
+		$switched = false;
+
+		foreach ( $enabled as $option => $is_enabled ) {
+			if ( $is_enabled && ! $this->was_rendering_new_views( $option ) ) {
+				$switched = true;
+			}
+
+			// A site holding the views off through the constant, the env var or a filter stores the
+			// value it actually renders, which is the whole point of still writing these.
```
coutinhocouto:
```suggestion
			/**
			* A site holding the views off through the constant, the env var or a filter stores the
			* value it actually renders, which is the whole point of still writing these.
			*/
```

coutinhocouto (approved):
Just adding a non blocking one, but please add this on the claude instructions, claude likes to do it often

PR #4424, opened 2026-08-24
[SMTNC-1507] Stop PayPal SCA card payments stranding the buyer on the checkout loader

claudiosanches:
### 🎫 Ticket

[SMTNC-1507]

### 🗒️ Description

Card payments that go through SCA/3DS, Swedish BankID among them, left the buyer on a spinning
loader with no error and no way forward. Reported against 5.28.2 on a site behind a platform
firewall and edge cache, where roughly a third of transactions failed during an on-sale.

**The problem.** Checkout took two requests. The first captured the money, the second returned where
to send the buyer. When that second request was blocked or slow, the buyer waited forever. The
handler meant to report the failure read `data.data.name` on a `WP_Error` that serializes with
`data: null`, threw, and was swallowed by a `.catch()` that only logged, so nothing surfaced at all.

**The fix.** The capture now records the payment and returns the destination itself, so checkout
finishes in one request. When PayPal's answer is unreadable the outcome is treated as unknown rather
than failed: there is no PayPal webhook to settle it later, and telling the buyer to retry risks
charging them twice.

**Also fixed.** Unsettled PayPal states such as `CREATED` and `APPROVED` were written over the
order, walking it out of pending for money that was never captured. And `timeoutRedirect` compared a
class using `is( className() )`, which never matches, so an expiring client token could reload
checkout in the middle of a payment.

### 🎥 Artifacts

https://www.loom.com/share/08f89f9d1dc548a2be98830a55aa6805

### ✔️ Checklist

### DIFF
--- a/changelog/smtnc-1507-paypal-sca-checkout-stall
+++ b/changelog/smtnc-1507-paypal-sca-checkout-stall
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fix
 
-Fixed PayPal card payments that go through SCA/3DS, such as Swedish BankID, leaving the buyer on a spinning loader with no error: every failure now clears the loader and shows a message, the capture is recorded as soon as PayPal answers it, an approved order nothing captured is captured on the recheck, and the client token no longer reloads checkout out from under a card authentication.
+Fixed PayPal card payments that go through SCA/3DS, such as Swedish BankID, leaving the buyer on a spinning loader with no error: every failure now clears the loader and shows a message, the capture is recorded as soon as PayPal answers it, an approved order with nothing captured is captured on the recheck, and the client token no longer reloads checkout from under a card authentication.
--- a/changelog/smtnc-1507-paypal-sca-checkout-stall
+++ b/changelog/smtnc-1507-paypal-sca-checkout-stall
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixed PayPal card payments that go through SCA/3DS, such as Swedish BankID, leaving the buyer on a spinning loader with no error: every failure now clears the loader and shows a message, the capture is recorded as soon as PayPal answers it, an approved order nothing captured is captured on the recheck, and the client token no longer reloads checkout out from under a card authentication.
--- a/changelog/smtnc-1507-paypal-sca-checkout-stall
+++ b/changelog/smtnc-1507-paypal-sca-checkout-stall
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fix
 
-Fixed PayPal card payments that go through SCA/3DS, such as Swedish BankID, leaving the buyer on a spinning loader with no error: every failure now clears the loader and shows a message, the capture is recorded as soon as PayPal answers it, an approved order with nothing captured is captured on the recheck, and the client token no longer reloads checkout from under a card authentication.
+Resolved an issue where PayPal card payments that go through SCA/3DS, such as Swedish BankID, left the buyer on a spinning loader with no error: every failure now clears the loader and shows a message, the capture is recorded as soon as PayPal answers it, an approved order with nothing captured is captured on the recheck, and the client token no longer reloads checkout from under a card authentication.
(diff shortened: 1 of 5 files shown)


src/Tickets/Commerce/Gateways/PayPal/REST/Order_Endpoint.php
```diff
@@ -337,90 +380,307 @@ public function handle_update_order( WP_REST_Request $request ) {
...
+			return new WP_Error(
+				'tec-tc-gateway-paypal-invalid-capture-status',
+				$messages['invalid-capture-status'],
+				[ 'status' => 500 ]
+			);
 		}
 
-		if ( in_array( $paypal_order_status, [ Status::FAILED, Status::DECLINED ], true ) ) {
-			return new WP_Error( 'tec-tc-gateway-paypal-capture-declined', $messages['capture-declined'] );
+		tribe( Order::class )->modify_status(
```
coutinhocouto: We're dropping the return value here. `modify_status()` is `: bool` and comes back false when
`lock_order()` can't get the lock, when `can_transition_to()` refuses, or when the save fails — and
in all three we still return `$status`, so the caller clears the cart and redirects to the success
page.

Order stays Pending, no attendees, buyer lands on Order Received. Same bug as the ticket, reached a
different way. And the overlapping capture/recheck is exactly when that lock is contended.

```suggestion
		$updated = tribe( Order::class )->modify_status(
			$order->ID,
			$status->get_slug(),
			[ 'gateway_payload' => $response ]
		);

		if ( ! $updated ) {
			return new WP_Error(
				'tec-tc-gateway-paypal-unconfirmed-capture',
				$messages['unconfirmed-capture'],
				[ 'status' => 502 ]
			);
		}
```

`handle_recheck_order` used to check this (`is_wp_error( $updated )`) — a6cd9d5 dropped it. Right
call since it's `bool` now, but it wanted to become `if ( ! $updated )` rather than go away.

changelog/smtnc-1507-paypal-sca-checkout-stall
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixed PayPal card payments that go through SCA/3DS, such as Swedish BankID, leaving the buyer on a spinning loader with no error: every failure now clears the loader and shows a message, the capture is recorded as soon as PayPal answers it, an approved order with nothing captured is captured on the recheck, and the client token no longer reloads checkout from under a card authentication.
```
coutinhocouto:
```suggestion
Resolved an issue where PayPal card payments that go through SCA/3DS, such as Swedish BankID, left the buyer on a spinning loader with no error: every failure now clears the loader and shows a message, the capture is recorded as soon as PayPal answers it, an approved order with nothing captured is captured on the recheck, and the client token no longer reloads checkout from under a card authentication.
```

src/Tickets/Commerce/Gateways/PayPal/Assets.php
```diff
@@ -105,8 +105,9 @@ public function register() {
 					'name' => 'tecTicketsCommerceGatewayPayPalCheckout',
 					'data' => static function () {
 						return [
-							'orderEndpoint' => tribe( Order_Endpoint::class )->get_route_url(),
-							'advancedPayments' => [
+							'orderEndpoint'        => tribe( Order_Endpoint::class )->get_route_url(),
+							'requestFailedMessage' => esc_html__( 'We could not confirm your payment. Check your PayPal account before trying again, so that you are not charged twice.', 'event-tickets' ),
```
coutinhocouto: Same string as `unconfirmed-capture` in `get_error_messages()` (Order_Endpoint.php:880) — two copies
for translators to keep in sync. We already resolve `Order_Endpoint` on the line above.

```suggestion
							'requestFailedMessage' => tribe( Order_Endpoint::class )->get_error_messages()['unconfirmed-capture'],
```

coutinhocouto (changes requested):

coutinhocouto (approved):

PR #4435, opened 2026-08-28
[SMTNC-2280] Event Ticket Sale End Date Reverts to Event Date in Block Editor

larodiel:
### 🎫 Ticket

[SMTNC-2280](https://linear.app/nexcess/issue/SMTNC-2280/event-ticket-sale-end-date-reverts-to-event-date-in-block)

### 🗒️ Description

**Fix** (bug fix - custom ticket sale end date reverted to event date in block editor)

Fixed an issue where a custom ticket sale end date set in the block editor was overwritten with the event start date when the event was saved or the page reloaded. Sale end dates that differ from the event start are now flagged as manually set when the ticket is created, so the event-date sync skips them; the block editor no longer defaults existing tickets' sale end to the event start, and the date picker re-syncs to the saved value once the ticket data loads.

Note: the day-picker change ships in the `tribe-common` submodule and is tracked on the matching `fix/smtnc-2280-...` branch in `the-events-calendar/tribe-common`.

https://github.com/the-events-calendar/tribe-common/pull/2984

### 🎥 Artifacts 
https://www.loom.com/share/7ca20b14e1af42d79aaf8c3d1a7cf855

### ✔️ Checklist

### DIFF
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -4121,7 +4121,7 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 					}
 				}
 
-				if ( ! $is_default ) {
+				if ( ! $is_default && ! $tickets_handler->has_manual_update( $ticket->ID, $tickets_handler->key_end_date ) ) {
 					add_post_meta( $ticket->ID, $tickets_handler->key_manual_updated, $tickets_handler->key_end_date );
 				}
 			}
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -4121,7 +4121,7 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 					}
 				}
 
-				if ( ! $is_default && ! $tickets_handler->has_manual_update( $ticket->ID, $tickets_handler->key_end_date ) ) {
+				if ( ! $is_default ) {
 					add_post_meta( $ticket->ID, $tickets_handler->key_manual_updated, $tickets_handler->key_end_date );
 				}
 			}
--- a/src/Tribe/Tickets.php
+++ b/src/Tribe/Tickets.php
@@ -4090,6 +4090,42 @@ if ( ! class_exists( 'Tribe__Tickets__Tickets' ) ) {
 			// Pass the control to the child object.
 			$save_ticket = $this->save_ticket( $post_id, $ticket, $data );
 
+			/*
+			 * The sale end date is added, not updated, on ticket creation, so the usual
+			 * manual-update flag does not fire. Flag explicitly supplied custom sale end
+			 * dates so later event start changes do not overwrite them.
+			 */
+			if (
+				! $update
+				&& ! empty( $ticket->ID )
+				&& ! empty( $data['ticket_end_date'] )
+				&& ! empty( $ticket->end_date )
+				&& ! $tickets_handler->has_manual_update( $ticket->ID, $tickets_handler->key_end_date )
+			) {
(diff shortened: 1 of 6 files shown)


common
dpanta94, line 1: We should not be updating common through ET or TEC PRs.

src/Tribe/Tickets.php
```diff
@@ -4090,6 +4090,42 @@ public function ticket_add( $post_id, $data ) {
...
+						$is_default = false !== $event_ts && false !== $end_date_ts && false !== $ticket_ts && $event_ts === $ticket_ts;
+					} else {
+						$is_default = false !== $event_ts && false !== $end_date_ts && $end_date === gmdate( Tribe__Date_Utils::DBDATEFORMAT, $event_ts );
+					}
+				}
+
+				if ( ! $is_default ) {
+					add_post_meta( $ticket->ID, $tickets_handler->key_manual_updated, $tickets_handler->key_end_date );
+				}
+			}
```
dpanta94: How many times could this meta be added ? Could you add a test case proving it ?

tests/wpunit/Tribe/Tickets/Tickets_Sale_End_DateTest.php
```diff
@@ -0,0 +1,158 @@
...
+
+		$ticket_id = $this->create_ticket(
+			$post_id,
+			[
+				'ticket_end_date' => '2026-09-15',
+				'ticket_end_time' => '08:00:00',
+			]
+		);
+
+		$this->assertContains( '_ticket_end_date', get_post_meta( $ticket_id, '_tribe_ticket_manual_updated' ) );
```
dpanta94, line 45: Could you also assert Ticket methods along with their meta ?

dpanta94 (changes requested):

src/Tribe/Tickets.php
```diff
@@ -4090,6 +4090,42 @@ public function ticket_add( $post_id, $data ) {
...
+
+					if ( ! empty( $ticket->end_time ) ) {
+						$ticket_ts  = strtotime( $end_date . ' ' . $ticket->end_time );
+						$is_default = false !== $event_ts && false !== $end_date_ts && false !== $ticket_ts && $event_ts === $ticket_ts;
+					} else {
+						$is_default = false !== $event_ts && false !== $end_date_ts && $end_date === gmdate( Tribe__Date_Utils::DBDATEFORMAT, $event_ts );
+					}
+				}
+
+				if ( ! $is_default && ! $tickets_handler->has_manual_update( $ticket->ID, $tickets_handler->key_end_date ) ) {
```
dpanta94: Isn;t this check already considered in line 4103 ?

larodiel: 4103 runs checking if it's not an `update`.
and this one will run just if it's not default, so, I'm checking on both to make sure that none of them has manual updates, but if you thing that it's redundant, I can remove it

dpanta94:
```
! $update
&& ! empty( $ticket->ID )
&& ! empty( $data['ticket_end_date'] )
&& ! empty( $ticket->end_date )
&& ! $tickets_handler->has_manual_update( $ticket->ID, $tickets_handler->key_end_date )
```

If here `$tickets_handler->has_manual_update( $ticket->ID, $tickets_handler->key_end_date )` evals to `true`

No matter what the rest are the conditional will always evaluate to false. So it will never enter. Thats why it is redundant. It may not be, if we have race conditions. Could its value change from line 4103 to 4124 ? Considering this is an atomic operation of inserting a new ticket, it should not.

dpanta94 (approved):

PR #4436, opened 2026-08-28
[SMTNC-1296] tec_kv_cache() returns outdated ticket status/number on calendar views (Sold out vs Remaining tickets)

coutinhocouto:
### 🎫 Ticket

[SMTNC-1296](https://linear.app/nexcess/issue/SMTNC-1296/tec-kv-cache-returns-outdated-ticket-statusnumber-on-calendar-views)

### 🎥 Artifacts

https://www.loom.com/share/805d73040b25463282f89f29e7be43d3

### 🗒️ Description

Keeps the calendar-view ticket model in step with real availability: it is dropped when tickets sell rather than rebuilt mid-order, keyed on the normalized Event ID so invalidation reaches the entry the views actually read, and cleared on plugin update so entries written by an earlier version don't linger.

### ⚠️ Blockers

None

### ✔️ Checklist

### DIFF
--- a/src/Tribe/Events/Views/V2/Models/Tickets.php
+++ b/src/Tribe/Events/Views/V2/Models/Tickets.php
@@ -174,10 +174,22 @@ class Tickets implements ArrayAccess {
 					$tribe_cache[ $tickets_cache_key ] = null;
 				}
 
+				/*
+				 * The model reads the aggregate list, not the per-provider one, and that list holds Ticket
+				 * objects carrying the stock they had when it was built.
+				 */
+				$all_tickets_cache_key                 = "{$tickets_class}::get_all_event_tickets-{$connected_event_id}";
+				$tribe_cache[ $all_tickets_cache_key ] = null;
+
+				/*
+				 * Drop the Event model entry before rebuilding it: a model built while its entry is live
+				 * restores that entry's data and stores it back, keeping the pre-purchase stock.
+				 */
+				tec_kv_cache()->delete( self::get_cache_key( $connected_event_id ) );
+
 				$model = new self( $connected_event_id );
 				// The call will trigger a priming of the model cache.
 				$model->exist();
-				$model->prime_cache();
 			}
 		}
 
--- a/src/Tribe/Events/Views/V2/Models/Tickets.php
+++ b/src/Tribe/Events/Views/V2/Models/Tickets.php
@@ -14,6 +14,7 @@ use ArrayAccess;
 use Closure;
 use InvalidArgumentException;
 use ReturnTypeWillChange;
+use TEC\Events\Custom_Tables\V1\Models\Occurrence;
 use Tribe\Utils\Lazy_Events;
 use Tribe__Events__Main as TEC;
 use Tribe__Tickets__Ticket_Object as Ticket_Object;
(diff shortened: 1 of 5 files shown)


larodiel (approved):

sdokus (approved):

WayneRocha (approved):

dpanta94 (changes requested):
Lets figure out the test failure before approving please!

Also there are multiple missed `@since` tags that need to be added.

src/Tribe/Events/Views/V2/Models/Tickets.php
```diff
@@ -608,12 +635,21 @@ private function restore_from_cache(): void {
...
 	 *
 	 * @return string The model cache key used to store it in the key-value cache.
 	 */
 	public static function get_cache_key( int $post_id ): string {
+		if ( class_exists( Occurrence::class, false ) ) {
+			/*
+			 * Views hand over the Occurrence's provisional ID while the invalidation side only ever
+			 * knows the Event post ID, so the two have to meet on the normalized one.
+			 */
+			$post_id = apply_filters( 'tec_tickets_normalize_occurrence_id', Occurrence::normalize_id( $post_id ) );
```
dpanta94: Can we provide a full docblock ?

tests/integration/Tribe/Tickets/UpdaterTest.php
```diff
@@ -0,0 +1,35 @@
...
+
+	/**
+	 * @test
+	 * @covers Tribe__Tickets__Updater::get_constant_update_callbacks
+	 */
+	public function should_flush_the_key_value_cache_on_every_version_update(): void {
+		$updater = new Tribe__Tickets__Updater( Tribe__Tickets__Main::VERSION );
+
+		$this->assertContains( 'flush_key_value_cache', array_column( $updater->get_constant_update_callbacks(), 1 ) );
+	}
```
dpanta94, line 33: I think a better test would be combining them into one, effectively testing that a fake update does trigger the flush_key_value_cache method.

Instead now you are testing:

1: flush_key_value_cache method on its own.
2; That the callback is part of the update callbacks.

If you instead triggered a fake update, effectively you could test both in one testcase.

src/Tribe/Events/Views/V2/Models/Tickets.php
```diff
@@ -119,11 +124,7 @@ public static function regenerate_caches( int $post_id ): void {
 
 		add_action( 'parse_query', $do_not_cache_results );
 
-		if ( $post->post_type === TEC::POSTTYPE ) {
```
dpanta94, line 122: Can you explain why we no longer need to handle the case for when its a tribe_event ?

dpanta94 (changes requested):

src/Tribe/Events/Views/V2/Models/Tickets.php
coutinhocouto: It is still handled, just earlier in the method. The first thing regenerate_caches() does is delete the entry for whatever post ID it was called with, so for an Event its own key is already gone by the time we get to that branch.

tests/integration/Tribe/Tickets/UpdaterTest.php
coutinhocouto: Done

src/Tribe/Events/Views/V2/Models/Tickets.php
coutinhocouto: Added

dpanta94 (approved):

PR #4440, opened 2026-09-01
[SOFT-3323] RSVP import

larodiel:
### 🎫 Ticket

[SOFT-3323](https://linear.app/nexcess/issue/SOFT-3323/rsvp-import)

### 🗒️ Description

**Fix** (CSV import compatibility with RSVP V2)

Fixed RSVP CSV import on RSVP V2 by porting the importer to create Tickets Commerce `tc-rsvp` tickets with version-switched `is_rsvp_v2()` dispatch, hook-verifiable version handling, and shared `match_existing_post` caching. Resolved an issue where V2 imports created legacy `tribe_rsvp_tickets` or `default` Commerce tickets with incorrect name, capacity and duplicate RSVPs per event.

Depends on: https://github.com/the-events-calendar/the-events-calendar/pull/5752

### 🎥 Artifacts 
https://www.loom.com/share/1eb893f46e2e44ddb4753c04ce8e51bd

### ✔️ Checklist

### DIFF
--- a/src/Tribe/CSV_Importer/RSVP_Importer.php
+++ b/src/Tribe/CSV_Importer/RSVP_Importer.php
@@ -517,6 +517,8 @@ class Tribe__Tickets__CSV_Importer__RSVP_Importer extends Tribe__Events__Importe
 	 * @return bool
 	 */
 	public function is_valid_record( array $record ) {
+		$this->row_message = false;
+		
 		$valid = parent::is_valid_record( $record );
 		if ( empty( $valid ) ) {
 			return false;
--- a/src/Tribe/CSV_Importer/RSVP_Importer.php
+++ b/src/Tribe/CSV_Importer/RSVP_Importer.php
@@ -62,15 +62,15 @@ class Tribe__Tickets__CSV_Importer__RSVP_Importer extends Tribe__Events__Importe
 	}
 
 	/**
-	 * Whether the RSVP->Tickets Commerce migration is currently running or paused, in which
-	 * case neither V1 nor V2 tickets should be imported.
+	 * Whether RSVP is disabled on this site, in which case neither V1 nor V2 ticket
+	 * should be imported.
 	 *
 	 * @since TBD
 	 *
 	 * @return bool
 	 */
 	private function is_rsvp_disabled(): bool {
-		return RSVP_Controller::is_migration_in_progress();
+		return ! tribe( RSVP_Controller::class )->is_rsvp_enabled();
 	}
 
 	/**
--- a/src/Tribe/CSV_Importer/RSVP_Importer.php
+++ b/src/Tribe/CSV_Importer/RSVP_Importer.php
@@ -203,12 +203,12 @@ class Tribe__Tickets__CSV_Importer__RSVP_Importer extends Tribe__Events__Importe
 		}
 
 		// Fallback: direct query if repository not available.
-		// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query -- One-off fallback lookup, limited to 1 result.
 		$found = ( new WP_Query(
 			[
 				'post_type'      => 'tec_tc_ticket',
 				'posts_per_page' => 1,
 				'post_status'    => 'any',
+				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query -- One-off fallback lookup, limited to 1 result.
 				'meta_query'     => [
 					[
 						'key'   => '_type',
@@ -312,11 +312,14 @@ class Tribe__Tickets__CSV_Importer__RSVP_Importer extends Tribe__Events__Importe
 	 * @return int
 	 */
 	private function create_post_v2( array $record, WP_Post $event, array $data ): int {
-		$ticket_id                                     = Commerce_Module::get_instance()->ticket_add( $event->ID, $data );
-		self::$ticket_name_cache[ 'v2-' . $event->ID ] = true;
+		$ticket_id = Commerce_Module::get_instance()->ticket_add( $event->ID, $data );
 
-		$tickets_handler = tribe( 'tickets.handler' );
-		update_post_meta( $event->ID, $tickets_handler->key_provider_field, Commerce_Module::class );
+		if ( $ticket_id ) {
+			self::$ticket_name_cache[ 'v2-' . $event->ID ] = true;
+
+			$tickets_handler = tribe( 'tickets.handler' );
+			update_post_meta( $event->ID, $tickets_handler->key_provider_field, Commerce_Module::class );
+		}
 
 		return (int) $ticket_id;
 	}
--- a/src/Tribe/CSV_Importer/RSVP_Importer.php
+++ b/src/Tribe/CSV_Importer/RSVP_Importer.php
@@ -5,6 +5,10 @@
  * @since 4.7
  */
 
+use TEC\Tickets\Commerce\Module as Commerce_Module;
+use TEC\Tickets\RSVP\Controller as RSVP_Controller;
+use TEC\Tickets\RSVP\V2\Constants as RSVP_V2_Constants;
+
 // phpcs:disable StellarWP.Classes.ValidClassName.NotSnakeCase
 
 /**
@@ -15,7 +19,7 @@ class Tribe__Tickets__CSV_Importer__RSVP_Importer extends Tribe__Events__Importe
 	/**
 	 * @var array
 	 */
-	protected $required_fields = [ 'event_name', 'ticket_name' ];
+	protected $required_fields = [ 'event_name' ];
 
 	/**
 	 * @var array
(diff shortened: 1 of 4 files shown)


src/Tribe/CSV_Importer/RSVP_Importer.php
```diff
@@ -15,7 +19,7 @@ class Tribe__Tickets__CSV_Importer__RSVP_Importer extends Tribe__Events__Importe
 	/**
 	 * @var array
 	 */
-	protected $required_fields = [ 'event_name', 'ticket_name' ];
+	protected $required_fields = [ 'event_name' ];
```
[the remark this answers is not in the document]
sdokus: Would this make V1 RSVP imports break? We drop the `ticket_name` in V2, but V1 still has it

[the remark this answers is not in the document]
sdokus: `is_valid_record()` is already overridden below (line 529), so the cheapest fix is to keep `ticket_name` out of `$required_fields` and add the check there when `!$this->is_rsvp_v2()`.

src/Tribe/CSV_Importer/RSVP_Importer.php
```diff
@@ -57,6 +61,30 @@ public static function reset_cache() {
...
+	 *
+	 * @return bool
+	 */
+	private function is_rsvp_v2(): bool {
+		if ( ! function_exists( 'tec_tickets_commerce_is_enabled' ) || ! tec_tickets_commerce_is_enabled() ) {
+			return false;
+		}
+
+		$default = RSVP_Controller::VERSION_1;
+		$version = tribe_get_option( RSVP_Controller::VERSION_OPTION_KEY, $default );
```
sdokus, line 94: `DISABLED` falls through to the V1 path while V1 is unregistered.

This reads the raw option and only compares against `VERSION_2`, so `Controller::DISABLED` returns `false` and we take the V1 branch. `DISABLED` isn't hypothetical — `get_version_from_migration_status()` returns it whenever the RSVP→Tickets Commerce migration is `RUNNING` or `PAUSED`.

In that state `Controller::register_disabled()` has already unregistered the V1 provider and post type, but `Tribe__Tickets__Main` registers `tribe_events_import_rsvp_tickets_importer` unconditionally (`src/Tribe/Main.php:777`). So an import during a paused migration writes legacy `tribe_rsvp_tickets` posts to a site that can't display them, and the in-flight migration won't pick them up.

Suggest bailing out of the import when the resolved version is `DISABLED` rather than defaulting to V1. Ideally this reuses Controller's resolver instead of re-deriving the version from the option — `Controller::get_version()` is private static today, but exposing it would keep the two from drifting.

src/Tribe/CSV_Importer/RSVP_Importer.php
```diff
@@ -108,20 +177,81 @@ public function match_existing_post( array $record ) {
...
 		}
 
-		self::$ticket_name_cache[ $cache_key ] = $match;
+		// Fallback: direct query if repository not available.
+		$found = ( new WP_Query(
+			[
+				'post_type'      => 'tec_tc_ticket',
+				'posts_per_page' => 1,
+				'post_status'    => 'any',
+				// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query -- One-off fallback lookup, limited to 1 result.
```
sdokus: The "fallback if repository not available" branch doesn't do that, and it runs on every miss.

`tribe( 'tickets.ticket-repository.rsvp' )` is always bound (`Tribe__Service_Providers__ORM` binds the V1 repo; `RSVP\V2\Controller` rebinds it to `RSVP\V2\Repositories\Ticket_Repository`), so `$repo` is never falsy and the "not available" case can't happen. If the binding genuinely were missing, di52 would throw here rather than fall through.

What actually happens: every time the repository legitimately finds no RSVP — the common case, i.e. every event that still needs one — we fall through and run a second full `WP_Query` with a two-clause meta_query that asks the same question. That's an extra query per imported row for nothing.

Suggest deleting the fallback and returning the repository result directly.

src/Tribe/CSV_Importer/RSVP_Importer.php
```diff
@@ -217,13 +390,93 @@ protected function get_event_from( array $record ) {
...
+
+		// Unlimited when blank - keep mode empty (Classic_Editor_Post_Data logic).
+		if ( '' === $capacity || '0' === $capacity ) {
+			$data['tribe-ticket'] = [ 'mode' => '' ];
+		} else {
+			$data['tribe-ticket'] = [
+				'mode'     => \Tribe__Tickets__Global_Stock::OWN_STOCK_MODE,
+				'capacity' => (int) $capacity,
+			];
+		}
```
sdokus: We should match what we're doing in the DTO instead of reinventing the wheel - we risk accidentally setting the RSVP as sold out instead of unlimited:

`src/Tickets/RSVP/V2/Data_Transfer_Objects/Classic_Editor_Post_Data.php` is the class the Classic Editor uses when someone saves an RSVP through the admin UI. It takes the submitted `rsvp_limit` field and turns it into a capacity value, at line 141:
```php
$limit          = trim( (string) ( $post_data['rsvp_limit'] ?? '' ) );
$self->capacity = ( '' !== $limit && (int) $limit > 0 ) ? (int) $limit : null;
```
Then at 179–183 it converts that into the tribe-ticket array the ticket-add path expects — null capacity means unlimited (`mode => ''`), a positive int means own-stock mode.

src/Tribe/CSV_Importer/RSVP_Importer.php
```diff
@@ -306,15 +559,15 @@ public function is_valid_record( array $record ) {
...
+		return false === $this->row_message ? parent::get_skipped_row_message( $row ) : $this->row_message;
 	}
 
 	/**
 	 * Registers the RSVP post type as a trackable activity.
 	 *
 	 * @param Tribe__Events__Aggregator__Record__Activity $activity The activity instance.
 	 */
-	public function register_rsvp_activity( $activity ) {
+	public static function register_rsvp_activity( $activity ) {
```
dpanta94, line 583: Is this method released already ? If yes: Why break the contract without a reason ? If by any chance other plugins/3rd party code is using this method its going to break for them.

If not, then its ok

src/Tribe/CSV_Importer/RSVP_Importer.php
```diff
@@ -57,6 +61,30 @@ public static function reset_cache() {
...
+	 * @return bool
+	 */
+	private function is_rsvp_v2(): bool {
+		if ( ! function_exists( 'tec_tickets_commerce_is_enabled' ) || ! tec_tickets_commerce_is_enabled() ) {
+			return false;
+		}
+
+		$default = RSVP_Controller::VERSION_1;
+		$version = tribe_get_option( RSVP_Controller::VERSION_OPTION_KEY, $default );
+		$version = apply_filters( 'tec_tickets_rsvp_version', $version );
```
dpanta94, line 97: That filter needs docs or needs to point to where its documented if its reused.

dpanta94 (changes requested):

src/Tribe/CSV_Importer/RSVP_Importer.php
larodiel: it started on RSVP V2 @dpanta94 
I've improved it docBlock to be more clear about its use

src/Tribe/CSV_Importer/RSVP_Importer.php
```diff
@@ -57,6 +61,44 @@ public static function reset_cache() {
...
+	 * Whether the RSVP->Tickets Commerce migration is currently running or paused, in which
+	 * case neither V1 nor V2 tickets should be imported.
+	 *
+	 * @since TBD
+	 *
+	 * @return bool
+	 */
+	private function is_rsvp_disabled(): bool {
+		return RSVP_Controller::is_migration_in_progress();
+	}
```
sdokus, line 74: This is a good start, but only closes one of five potential "doors" into `register_disabled()`

`is_migration_in_progress()` checks just if the version option is exactly `DISABLED` - `do_register()` bails to `register_disabled()` in four other cases:
1. The `TEC_TICKETS_RSVP_DISABLED` constant
2. the env var of the same name
3. `get_option( 'tec_tickets_rsvp_active' )` returning `false`
4. Anything hooked into `tec_tickets_rsvp_enabled`

The Controller already owns a method that checks all five possible "doors" 

```suggestion
	/**
	 * Whether RSVP is disabled on this site, in which case neither V1 nor V2 ticket
	 * should be imported.
	 *
	 * @since TBD
	 *
	 * @return bool
	 */
	private function is_rsvp_disabled(): bool {
		return ! tribe( RSVP_Controller::class )->is_rsvp_enabled();
	}
```

src/Tribe/CSV_Importer/RSVP_Importer.php
sdokus, line 522: This is not being introduced here, but the changes are exposing that this call needs to reset the row message between each call:

```suggestion
		$this->row_message = false;
		
		$valid = parent::is_valid_record( $record );
```

Here's how Claude is explaining it to me:
>  is_valid_record() runs once per CSV row and answers one question: import this row or
  skip it? When it skips, it leaves a sentence in $this->row_message, and that sentence
  is what the user sees next to the row number in the import summary.

 > Nothing wipes that sentence between rows. It's a whiteboard nobody erases. So a row
  that fails without writing its own message (the missing-event case at line 534) gets
  reported using whatever the last failing row left behind.

  > On the base branch that needed a recurring event to set it up, so it was rare enough
  nobody noticed. This PR adds two more paths that write a message, and the
  missing-ticket_name one at line 537 will fire on every unmapped row of a V1 import now
  that the column isn't caught at the mapping screen. So the setup for the bug goes
  from unusual to routine: one bad row poisons the message for every silent failure
  after it, and the user gets told to fix a column that has nothing to do with their
  actual problem.

 > There's a $this->row_message = false; at line 554 that looks like it should handle
  this, but it only runs on sites without The Events Calendar. On a real install the
  tribe_is_recurring_event branch above it returns first.

dpanta94 (approved):

PR #4452, opened 2026-09-07
[SMTNC-1468] fix: Auto-refresh the Square OAuth access token

coutinhocouto:
### 🎫 Ticket

[SMTNC-1468]

### 🗒️ Description

Square access tokens expire about 30 days after they are minted, and nothing renewed them. Checkout started failing at Square's authorization layer while the admin still reported the gateway as connected, and the only recovery was disconnect and reconnect. `WhoDat::refresh_token()` already existed with zero callers, and the `expires_at` the OAuth callback stores was never read.

**The fix.** `Square\Requests::request()` renews the token before signing a request whose expiration is due, and retries once when Square rejects it anyway. That path was chosen over a scheduled action because it still runs on a site whose cron queue is stalled, which is where the reports come from.

**What only surfaced against the live service.** Running this against WhoDat with a Square sandbox account turned up two things:

- The refresh endpoint answers `GET` with **405 Method Not Allowed** and only accepts `POST`. The existing method sent a GET, so it could never have worked.
- A rejected refresh token comes back as an **HTML 500**, not a machine readable error, so the response body cannot tell a dead connection apart from a WhoDat outage. Failures are judged by status code instead. A 401 or 403 is final on its own; anything else has to be corroborated by `oauth/token/status`. That endpoint reports on the *access* token, which on this path is expired by definition, so its refusal only carries information while the access token is still inside its own lifetime. Otherwise a rejection is final only once five failures have piled up over more than a day. A transport failure never marks the connection unavailable, it only feeds an exponential backoff.

**When renewal fails.** The connection is marked unavailable: `Merchant::is_connected()` returns false, the settings screen shows the connect panel rather than a false "Connected", the gateway drops off checkout, and an admin notice explains it with a reconnect link. That is a long backoff rather than a dead end. It is rechecked every 12 hours, with an `admin_init` trigger so a site with no traffic heals without anyone touching the settings.

**Notes for review.**

- Square's status is stored in a new mode-scoped option rather than through `Merchant::set_merchant_unauthorized()`. Those option names are byte-identical to Stripe's and are not mode-scoped, so writing them from Square would flip Stripe's state, which `Stripe\Hooks::handle_stripe_errors()` reads.
- `check_account_status()` was reading an `error` key the status endpoint does not return, so even an explicit `is_connected( true )` recheck reported a revoked token as connected. It now reads the `type` and `message` shape the endpoint actually sends.
- The refresh takes a cross-process lock so parallel requests do not stampede the endpoint. Confirmed against the live service that the refresh token does not rotate, so a lost race is harmless.
- Connections made before `expires_at` was tracked are never treated as expired. They retry at most twice a day until a refresh teaches them a real expiration, and the 401 path covers them meanwhile.

### 🎥 Artifacts

https://www.loom.com/share/5bb28e2887f7457b8cf04f8a68d9fb5d

### ✔️ Checklist

### DIFF
--- a/changelog/fix-smtnc-1468-square-oauth-token-refresh.yaml
+++ b/changelog/fix-smtnc-1468-square-oauth-token-refresh.yaml
@@ -1,0 +1,7 @@
+significance: patch
+type: fix
+entry: Refreshed the Square OAuth access token automatically before Square API calls and retried once
+  when Square rejects it, so Tickets Commerce checkout no longer stops working about 30 days after
+  connecting. When Square refuses to renew the stored credentials the gateway is now shown as
+  disconnected with an admin notice, instead of reporting a healthy connection while live charges fail.
+timestamp: 2026-08-18T21:46:14.000Z
--- a/changelog/fix-smtnc-1468-square-oauth-token-refresh.yaml
+++ b/changelog/fix-smtnc-1468-square-oauth-token-refresh.yaml
@@ -1,7 +1,8 @@
 significance: patch
 type: fix
-entry: Refreshed the Square OAuth access token automatically before Square API calls and retried once
-  when Square rejects it, so Tickets Commerce checkout no longer stops working about 30 days after
-  connecting. When Square refuses to renew the stored credentials the gateway is now shown as
-  disconnected with an admin notice, instead of reporting a healthy connection while live charges fail.
+entry: Refreshed the Square OAuth access token before Square API calls and retried once when Square
+  rejects it, so Tickets Commerce checkout no longer stops working about 30 days after connecting. The
+  refresh request was also being sent as a GET, which the endpoint answers with 405. When Square will
+  neither renew the credentials nor accept the stored token, the gateway is now shown as disconnected
+  with an admin notice instead of reporting a healthy connection while live charges fail.
 timestamp: 2026-08-18T21:46:14.000Z
--- a/changelog/fix-smtnc-1468-square-oauth-token-refresh.yaml
+++ b/changelog/fix-smtnc-1468-square-oauth-token-refresh.yaml
@@ -1,8 +1,7 @@
 significance: patch
 type: fix
-entry: Refreshed the Square OAuth access token before Square API calls and retried once when Square
-  rejects it, so Tickets Commerce checkout no longer stops working about 30 days after connecting. The
-  refresh request was also being sent as a GET, which the endpoint answers with 405. When Square will
+entry: Fixed Tickets Commerce checkout failing with Square about 30 days after connecting, by renewing
+  the Square access token before it expires and retrying once when Square rejects it. When Square will
   neither renew the credentials nor accept the stored token, the gateway is now shown as disconnected
   with an admin notice instead of reporting a healthy connection while live charges fail.
 timestamp: 2026-08-18T21:46:14.000Z
--- a/changelog/fix-smtnc-1468-square-oauth-token-refresh.yaml
+++ b/changelog/fix-smtnc-1468-square-oauth-token-refresh.yaml
@@ -1,7 +1,8 @@
 significance: patch
 type: fix
-entry: Fixed Tickets Commerce checkout failing with Square about 30 days after connecting, by renewing
-  the Square access token before it expires and retrying once when Square rejects it. When Square will
-  neither renew the credentials nor accept the stored token, the gateway is now shown as disconnected
-  with an admin notice instead of reporting a healthy connection while live charges fail.
+entry: Resolved an issue where Tickets Commerce checkout failed with Square about 30 days after
+  connecting, by renewing the Square access token before it expired and retrying once when Square
+  rejected it. When Square would neither renew the credentials nor accept the stored token, the
+  gateway was shown as disconnected with an admin notice instead of reporting a healthy connection
+  while live charges failed.
 timestamp: 2026-08-18T21:46:14.000Z
--- a/changelog/fix-smtnc-1468-square-oauth-token-refresh.yaml
+++ b/changelog/fix-smtnc-1468-square-oauth-token-refresh.yaml
@@ -1,8 +1,8 @@
 significance: patch
 type: fix
-entry: Resolved an issue where Tickets Commerce checkout failed with Square about 30 days after
-  connecting, by renewing the Square access token before it expired and retrying once when Square
-  rejected it. When Square would neither renew the credentials nor accept the stored token, the
-  gateway was shown as disconnected with an admin notice instead of reporting a healthy connection
-  while live charges failed.
+entry: Renewed the Square access token ahead of its expiration and retried a rejected request once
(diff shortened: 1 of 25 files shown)


src/Tickets/Commerce/Gateways/Square/Token/Refresh_Lock.php
```diff
@@ -0,0 +1,234 @@
...
+	 */
+	public function wait_for_holder( string $previous_token ): bool {
+		$waited = 0;
+
+		while ( $waited < self::WAIT ) {
+			usleep( self::POLL );
+
+			$waited += self::POLL;
+
+			$this->merchant->flush_option_cache();
```
sdokus, line 216: flush_option_cache() drops the entire alloptions cache :
```php
  public function flush_option_cache(): void {
      wp_cache_delete( $this->get_signup_data_key(), 'options' );
      wp_cache_delete( 'alloptions', 'options' );   // ← this one
  }
```

This is only problematic in this `while` loop, which on a site with persistent object cache (Redis, Memcached) could result in dropping the `alloptions` every 250ms for up to 4 seconds, which could be 16 times per checkout attempt, which then requires a rebuild with a full `SELECT` over `wp_options`

Fix suggestion from Claude: read the row straight from the DB in the poll loop and skip the cache entirely — you're already using DB next door in Refresh_Lock:
```php
  $token = DB::get_var( DB::prepare(
      'SELECT option_value FROM %i WHERE option_name = %s',
      DB::prefix( 'options' ),
      $this->merchant->get_signup_data_key()
  ) );
```
Keep the single `flush_option_cache()` in `do_refresh()` — that one's correct and cheap.

changelog/fix-smtnc-1468-square-oauth-token-refresh.yaml
```diff
@@ -0,0 +1,8 @@
+significance: patch
+type: fix
+entry: Resolved an issue where Tickets Commerce checkout failed with Square about 30 days after
```
sdokus: I made this comment in another PR too, but we should try to switch up the phrasing so every changelog doesn't start with "Resolved an issue where..." 😄

src/Tickets/Commerce/Gateways/Square/Merchant.php
```diff
@@ -98,6 +101,11 @@ public function is_connected( $recheck = false ): bool {
 			return false;
 		}
 
+		// A token Square refused to renew cannot be recovered without a new OAuth handshake.
+		if ( $this->is_token_invalid() ) {
+			return false;
+		}
+
```
sdokus, line 120:
```suggestion
		if ( $recheck && $this->is_token_invalid() ) {
			// An explicit recheck is a human saying something changed; let it jump the backoff
			tribe( Refresh_Status::class )->update( [ 'last_attempt_at' => '' ] );
			tribe( Token_Refresher::class )->refresh_now( 'admin_recheck' );
		}

		if ( $this->is_token_invalid() ) {
			return false;
		}

```
The order here should be swapped because the early return fires before the recheck can run. `$recheck = true` is a person asking us to go and check, but with the flag set we just answer from the flag and never look. Trying first and reading the flag second means a connection that started working again can clear itself, instead of waiting out the 12-hour backoff.

src/Tickets/Commerce/Gateways/Square/Token/Refresh_Policy.php
```diff
@@ -0,0 +1,173 @@
+<?php
```
sdokus, line 1: Claude is prompting me to comment on the heft of this PR and focused on this file as somewhere that we could trim it up a little bit:
>  On the size of this: I pulled the added src/ lines apart and it's 697 lines of code under 991 lines of docblock, so most of the LOC count is the WordPress-Docs ruleset doing its job rather than anything you overbuilt. No argument with the structure.

> There are four things in here that nothing calls, though, and each one becomes permanent surface area the moment it ships:


> 1. The three filters in Refresh_Policy (..._token_refresh_window, ..._unknown_expiration_interval, ..._invalid_recheck_interval). Nothing uses them and no test exercises them. Three private getters exist only to host them. Private consts do the same job in three lines, and we can add a filter the first time someone asks to tune one.
> 2. tec_tickets_commerce_square_access_token_refreshed and ..._refresh_failed in Token_Refresher. No listeners anywhere, no test assertions.
> 3. WhoDat::refresh_token(). Zero callers before this PR and zero after, since everything now routes through request_token_refresh(). It's kept for back compat, but it's back compat with a method that has never worked, because it sent a GET and the endpoint answers 405. If there's a policy against dropping public methods mid-cycle that's fair, but the docblock should say so, because right now it reads like a supported entry point.
> 4. Refresh_Outcome::is_transient(). Defined, never called. do_refresh() checks success, then permanent, then falls through.

> That's roughly 80 lines with no behaviour change and no tests to update.

>The bigger one is a judgment call rather than a request. The amber "refresh failing" notice is about 85 production lines plus three tests and a snapshot, and it takes Merchant::is_token_expiring_within() with it, since that's its only caller. It fires only within two days of expiry and after three failures, and it asks the admin to do the same thing the red notice will ask for shortly after. If you want this under the 1000 line mark that came up on #4408, that's the cleanest piece to pull into a follow-up.

src/Tickets/Commerce/Gateways/Square/Token/Refresh_Lock.php
coutinhocouto: Good catch. The poll now reads the row straight from the DB via a new Merchant::get_access_token_uncached() and only flushes the cache on the pass where the value actually changed.

src/Tickets/Commerce/Gateways/Square/Merchant.php
coutinhocouto: Agreed on the ordering, fixed. I went with an $ignore_backoff flag on refresh_now() rather than writing last_attempt_at => '' from Merchant , that reaches into Refresh_Status's storage and destroys a real record to steer a policy. Same effect, status stays truthful. Two tests added (recovers on success, stays rejected on a 401).

src/Tickets/Commerce/Gateways/Square/Token/Refresh_Policy.php
coutinhocouto: Took 3 of the 4. The three filters are now private consts, both do_actions are gone (tribe_log already covers the failure path), and Refresh_Outcome::is_transient() is deleted.

Kept WhoDat::refresh_token(), it's `@since` 5.24.0 and public, so removing it mid-cycle is a BC break. Went with your fallback instead: the docblock now says nothing calls it and that request_token_refresh() is the real path.

On the amber notice, agreed it's the cleanest thing to pull, but it's feature removal rather than a review fix, so I'd rather do it as a separate PR than fold it in here. Happy to cut that now if you'd prefer it before this merges.

sdokus (approved):

PR #4460, opened 2026-09-10
[SOFT-4441] Update migration screen helper text

larodiel:
### 🎫 Ticket

[SOFT-4441](https://linear.app/nexcess/issue/SOFT-4441/update-migration-screen-helper-text)

### 🗒️ Description

**Tweak** (copy update - migration screen helper text)

Expanded the RSVP migration helper text to explain the downtime and timing expectations before a migration is started, and surfaced the recommended site backup link above the migration list. The backup sentence is rendered from the Migrations tab callback rather than the migration description because the StellarWP Migrations card template escapes descriptions, which would have printed the anchor as literal text. Also moved the tab's previously inline styles into `_migrations.pcss`.

### 🎥 Artifacts 
Screenshot 2026-09-09 at 21 07 55

### ✔️ Checklist

### DIFF
--- a/src/Tickets/Migrations/Controller.php
+++ b/src/Tickets/Migrations/Controller.php
@@ -97,22 +97,19 @@ class Controller extends Controller_Contract {
 					return $filters;
 				};
 				?>
-				<?php
-				// Hide the tag search field and individual migration tags using inline CSS
-				// to avoid overriding the StellarWP migrations templates. Strategy requested
-				// these fields to be hidden from the Event Tickets migrations tab.
-				?>
-				<style>
-					.stellarwp-migrations-filters__row,
-					.stellarwp-migration-card__tags {
-						display: none !important;
-					}
-					.tickets_page_tec-tickets-settings .tec-settings-form.tec-settings-form__migrations-tab--active {
-						padding-top: 0;
-					}
-				</style>
-				<div class="tec-settings-form tec-settings-form__migrations-tab--active" style="grid-template-columns: 1fr;">
+				<div class="tec-settings-form tec-settings-form__migrations-tab--active">
 					<?php
+					// The migration card template escapes its description, so the backup link is rendered here instead.
+					printf(
+						'<div class="notice notice-info inline"><p>%s</p></div>',
+						sprintf(
+							// Translators: %1$s is the opening anchor tag for the site backup guide, %2$s is the closing anchor tag.
+							esc_html__( 'Just in case, we recommend doing a %1$ssite backup%2$s before starting migration.', 'event-tickets' ),
+							'<a href="' . esc_url( 'https://evnt.is/1bei' ) . '" target="_blank" rel="noopener noreferrer">',
+							'</a>'
+						)
+					);
+
 					add_filter( 'stellarwp_migrations_tec_filters', $add_tags );
 					$ui->render_list();
 					remove_filter( 'stellarwp_migrations_tec_filters', $add_tags );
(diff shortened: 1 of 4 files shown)


src/Tickets/Migrations/Controller.php
```diff
@@ -97,22 +97,19 @@ public function register_migrations_tab( $admin_page ): void {
...
+					printf(
+						'<div class="notice notice-info inline"><p>%s</p></div>',
+						sprintf(
+							// Translators: %1$s is the opening anchor tag for the site backup guide, %2$s is the closing anchor tag.
+							esc_html__( 'Just in case, we recommend doing a %1$ssite backup%2$s before starting migration.', 'event-tickets' ),
+							'<a href="' . esc_url( 'https://evnt.is/1bei' ) . '" target="_blank" rel="noopener noreferrer">',
+							'</a>'
+						)
+					);
+
```
sdokus, line 112: Should this be conditional to not show once the migration is completed?

src/resources/postcss/tickets-admin/settings/_migrations.pcss
```diff
@@ -1,8 +1,37 @@
...
+			gap: 0;
+			grid-template-columns: 1fr;
+
+			> .wrap {
+				grid-column: 1;
+				padding-left: var(--tec-spacer-settings-form-right-gap);
+			}
+		}
 	}
 }
```
sdokus, line 38: Two things got lost moving the inline block into the stylesheet. The old inline CSS had `.tec-settings-form.tec-settings-form__migrations-tab--active { padding-top: 0; }`, and nothing here replaces it, so the tab picks the 16px back up from `_form.pcss:39`. And the rules now hang off `.tec-settings-form`, which is the wrapper on every settings tab, not just this one, so `> .notice.notice-info.inline` will grey-box and negative-margin any inline info notice a direct child of a settings form anywhere on the page (the Help Hub views use that wrapper too). Nesting everything under the tab class fixes both. 
```suggestion
.tribe_settings {

	.tec-settings-form__migrations-tab--active {
		gap: 0;
		grid-template-columns: 1fr;
		padding-top: 0;

		/*
			The StellarWP migration templates are shared with every plugin that uses
			them, so the tag search field and the per-migration tags are hidden from
			this tab here rather than by overriding those templates.
		*/

		.stellarwp-migrations-filters__row,
		.stellarwp-migration-card__tags {
			display: none !important;
		}

		> .notice.notice-info.inline {
			align-items: center;
			background: var(--tec-admin-core-grey-0);
			display: flex;
			margin-bottom: calc(var(--tec-spacer-2) * -1);
			margin-left: 34px;
			max-width: 765px;
			padding-bottom: var(--tec-spacer-3);
		}

		> .wrap {
			grid-column: 1;
			padding-left: var(--tec-spacer-settings-form-right-gap);
		}
	}
}
```

src/Tickets/Migrations/Controller.php
larodiel: @sdokus As the user has the power to do a rollback, should be good to left this message there to remind them that they are messing with DB and a backup is always a good thing to do. (IMO)

src/resources/postcss/tickets-admin/settings/_migrations.pcss
larodiel: @sdokus I've changed it, as we are putting a higher class on the cascade just it is enough to the rule be applied and overwrite the rules from `common`, the final CSS after the compiling is...

```css
.tribe_settings .tec-settings-form__migrations-tab--active {
    gap: 0;
    grid-template-columns: 1fr;
    padding-top: 0;
}
```

and for the migration warning, we now have a specific class `.tribe_settings .tec-settings-form>.notice.notice-info.notice-info--migration`

sdokus (approved):

PR #4465, opened 2026-09-16
[SMTNC-1739] Fix Series Pass Attendee check-in in ETP App

larodiel:
### 🎫 Ticket

[SMTNC-1739](https://linear.app/nexcess/issue/SMTNC-1739/series-pass-attendees-not-checking-in-via-etp-app)

### 📋 Plan

smtnc-1739

### 🗒️ Description

**Fix** (bug fix - Series Pass Attendee check-in status not persisting)
Fixed the QR check-in endpoint and the Attendees admin table to read and report a Series Pass Attendee's checkin status from its per-Occurrence clone Attendee instead of the original Series-level Attendee, which never carries the real checkin meta. This resolved the check-in appearing to succeed but not persisting via the ETP app, the QR scanner reporting an attendee as already checked in, and the wp-admin attendees list never showing the checked-in status for Series Pass Attendees.

Depends on: https://github.com/the-events-calendar/event-tickets-plus/pull/2159

### 🎥 Artifacts 

---
**Bug**
https://www.loom.com/share/024c0245c4e44c47a4884fec52a1f312

**Fix**
https://www.loom.com/share/501483f8817f4fbea4b83f8866bc9e17

---

### ✔️ Checklist

### 🤖 AI Disclosure

- **Was AI used to author this PR?** Y
- **If yes, which model?** Fable 5.1
- **I own the work being done in this PR and I have personally reviewed it before asking another human to.** ☑️

### DIFF
--- a/changelog/fix-smtnc-1739-series-pass-attendees-not-checking-in-via-etp-app.yaml
+++ b/changelog/fix-smtnc-1739-series-pass-attendees-not-checking-in-via-etp-app.yaml
@@ -1,0 +1,4 @@
+significance: patch
+type: fix
+entry: Resolved an issue where a Series Pass Attendee checked in from the Event Tickets Plus App appeared checked in and then reverted to unchecked on the next refresh the Attendee list no longer drops the per-Occurrence Attendee that carries the check-in status, it reports the original Attendee ID alongside it so the App keeps one identity per Attendee across Occurrences, and a manual check-in from the App is no longer restricted to the Events happening now the way a QR code scan is.
+timestamp: 2026-09-12T05:33:04.121Z
--- a/src/Tribe/REST/V1/Endpoints/Attendee_Archive.php
+++ b/src/Tribe/REST/V1/Endpoints/Attendee_Archive.php
@@ -82,6 +82,7 @@ class Tribe__Tickets__REST__V1__Endpoints__Attendee_Archive
 	 * @param WP_REST_Request $request
 	 *
 	 * @since 4.12.0 Returns 401 Unauthorized if Event Tickets Plus is not loaded.
+	 * @since TBD Stop narrowing a manage-access request by the related post status.
 	 *
 	 * @return WP_Error|WP_REST_Response An array containing the data on success or a WP_Error instance on failure.
 	 */
@@ -123,9 +124,19 @@ class Tribe__Tickets__REST__V1__Endpoints__Attendee_Archive
 		}
 
 		if ( tribe( 'tickets.rest-v1.main' )->request_has_manage_access() ) {
+			/*
+			 * A request with manage access is not narrowed by the status of the post the Attendee belongs to,
+			 * so no `event_status` default is set below and the filter is skipped entirely. Defaulting it to
+			 * `any` would not do: the filter downgrades `any` to `publish` for any request without a WP user
+			 * - which is every Event Tickets Plus App request, authorized by API key alone - and then joins
+			 * `wp_posts` on the Attendee to Event ID meta value. A Series Pass Attendee cloned to an
+			 * Occurrence holds a provisional Occurrence ID there, which has no `wp_posts` row, so every
+			 * cloned Attendee, and with it the check-in status the App reads, would be dropped from the
+			 * results. A status the request asked for explicitly still applies, having been mapped into
+			 * `event_status` above.
+			 */
 			$permission                 = Tribe__Tickets__REST__V1__Attendee_Repository::PERMISSION_EDITABLE;
 			$fetch_args['post_status']  = Tribe__Utils__Array::get( $fetch_args, 'post_status', 'any' );
-			$fetch_args['event_status'] = Tribe__Utils__Array::get( $fetch_args, 'event_status', 'any' );
 			$fetch_args['order_status'] = Tribe__Utils__Array::get( $fetch_args, 'order_status', 'any' );
 		} else {
 			$permission                 = Tribe__Tickets__REST__V1__Attendee_Repository::PERMISSION_READABLE;
--- a/src/Tribe/REST/V1/Endpoints/Attendee_Archive.php
+++ b/src/Tribe/REST/V1/Endpoints/Attendee_Archive.php
@@ -84,8 +84,8 @@ class Tribe__Tickets__REST__V1__Endpoints__Attendee_Archive
 	 * @param WP_REST_Request $request
 	 *
 	 * @since 4.12.0 Returns 401 Unauthorized if Event Tickets Plus is not loaded.
-	 * @since TBD Stop narrowing a manage-access request by the related post status only when the request is
-	 *            about a Series Pass Occurrence - every other request keeps being narrowed as before.
+	 * @since TBD Stop narrowing a manage-access request by the related post status.
+	 * @since TBD Reindexes the attendees so the response always encodes them as a JSON array.
 	 *
 	 * @return WP_Error|WP_REST_Response An array containing the data on success or a WP_Error instance on failure.
 	 */
@@ -225,6 +225,11 @@ class Tribe__Tickets__REST__V1__Endpoints__Attendee_Archive
 		 * @param WP_REST_Request $request The request object for this endpoint.
 		 */
 		$data = apply_filters( 'tec_tickets_rest_attendee_archive_data', $data, $request );
+
+		/* Reindexed last so the array shape holds for whatever the filter returned. */
+		if ( isset( $data['attendees'] ) ) {
+			$data['attendees'] = Tribe__Tickets__REST__V1__Archive_List::reindex( $data['attendees'] );
+		}
 
 		return new WP_REST_Response( $data, 200, $headers );
 	}
--- a/src/Tribe/REST/V1/Endpoints/Attendee_Archive.php
+++ b/src/Tribe/REST/V1/Endpoints/Attendee_Archive.php
@@ -1,5 +1,7 @@
 <?php
 
+use TEC\Events\Custom_Tables\V1\Models\Occurrence;
+
 class Tribe__Tickets__REST__V1__Endpoints__Attendee_Archive
 	extends Tribe__Tickets__REST__V1__Endpoints__Base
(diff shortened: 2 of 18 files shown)


src/Tribe/REST/V1/Endpoints/QR.php
```diff
@@ -438,6 +464,36 @@ public function check_in( WP_REST_Request $request ) {
...
+	 * @return int The Attendee ID that carries the checkin status meta: the clone if one exists, the original otherwise.
+	 */
+	private function get_checkin_status_attendee_id( int $attendee_id, int $event_id, $ticket_provider ): int {
+		$attendee_event_key = $ticket_provider->attendee_event_key ?? '';
+
+		if ( ! $attendee_event_key ) {
+			return $attendee_id;
+		}
+
+		$clone_id = tribe_attendees()
```
sdokus: This only finds a clone when `$event_id` is an Occurrence provisional ID or an Event post ID. A real Series Pass QR code carries the Series ID as `event_id` (the attendee's `attendee_event_key` meta points at the Series), so for a normal door scan this returns the original, and both the pre-check on line 351 and the re-resolve after check-in are no-ops. The "already checked in" 403 still happens today, but only because `Attendees::checkin_attendee_using_provider()` catches it downstream, and the response's `attendee.checked_in` stays false. `QR_Series_Pass_Test` passes because it sends the single Event's post ID, not the Series ID.

I'd recommend a check for when nothing matches and `$event_id` is the Series, fall back to the candidate Occurrence the Series Pass handler would pick

changelog/smtnc-1227-force-enable-updated-tickets-experience
coutinhocouto, line 1: This is not suppoused to be here, please rebase or sync this branch

src/Tribe/REST/V1/Endpoints/Attendee_Archive.php
```diff
@@ -124,10 +127,35 @@ public function get( WP_REST_Request $request ) {
...
+			 * provisional Occurrence ID there, which has no `wp_posts` row, so every cloned Attendee, and
+			 * with it the check-in status the App reads, would be dropped from the results - whether the
+			 * status came from that downgrade or was asked for explicitly. `Occurrence::normalize_id()`
+			 * is a safe no-op for a real post ID (returns it unchanged), so this only skips the filter for
+			 * an actual Occurrence request.
+			 */
+			$requested_event_id    = $fetch_args['event'] ?? null;
+			$is_occurrence_request = class_exists( Occurrence::class )
+				&& is_numeric( $requested_event_id )
+				&& Occurrence::normalize_id( (int) $requested_event_id ) !== (int) $requested_event_id;
```
coutinhocouto: `include_post` and `post_id` both write `$fetch_args['event']` (lines 102 and 104), but `include_post` goes through `Tribe__Utils__Array::list_to_array`, so it always arrives as an array. `is_numeric()` is false for an array, so `$is_occurrence_request` stays false and this whole branch falls back to the old behaviour for any request that filters by `include_post`.

The App uses `post_id`, so nothing user-facing breaks today. It's the silence that bothers me: the same endpoint, the same Occurrence, one query parameter apart, and no way to tell from the response which path you got. Either handle the multi-value case, or say in the comment that it's knowingly out of scope.

src/Tribe/REST/V1/Endpoints/QR.php
```diff
@@ -438,6 +464,36 @@ public function check_in( WP_REST_Request $request ) {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param int     $attendee_id     The Attendee ID as provided in the check-in request.
+	 * @param int     $event_id        The ID of the ticket-able post the Attendee is being checked into.
+	 * @param Tickets $ticket_provider The Attendee ticket provider.
+	 *
+	 * @return int The Attendee ID that carries the checkin status meta: the clone if one exists, the original otherwise.
+	 */
+	private function get_checkin_status_attendee_id( int $attendee_id, int $event_id, $ticket_provider ): int {
```
coutinhocouto: `Tribe\Tickets\Plus\REST\V1\Endpoints\Bulk_Checkin` extends this class, so it would inherit this method for free, except `private` puts it out of reach. That's why ETP #2159 ends up with `resolve_checkin_attendee_id()` carrying the identical `tribe_attendees()` two-meta lookup. Two copies of the same query in two separately versioned plugins will drift, and the thing they'd drift on is which attendee holds the check-in state.

```suggestion
	protected function get_checkin_status_attendee_id( int $attendee_id, int $event_id, $ticket_provider ): int {
```

changelog/fix-smtnc-1739-series-pass-attendees-not-checking-in-via-etp-app.yaml
```diff
@@ -0,0 +1,4 @@
+significance: patch
+type: fix
+entry: Resolved an issue where Series Pass Attendees checked in from the Event Tickets Plus App appeared checked in and then reverted to unchecked on the next refresh.
```
coutinhocouto: Use other wording than `Resolved`, claude wants aways to write all changelogs as `Resolved `😄 

I had updated the skill that generate this so if you get the lastest version this will be fixed.

src/Tribe/REST/V1/Endpoints/Attendee_Archive.php
```diff
@@ -124,10 +127,35 @@ public function get( WP_REST_Request $request ) {
...
+
 			$permission                 = Tribe__Tickets__REST__V1__Attendee_Repository::PERMISSION_EDITABLE;
 			$fetch_args['post_status']  = Tribe__Utils__Array::get( $fetch_args, 'post_status', 'any' );
-			$fetch_args['event_status'] = Tribe__Utils__Array::get( $fetch_args, 'event_status', 'any' );
 			$fetch_args['order_status'] = Tribe__Utils__Array::get( $fetch_args, 'order_status', 'any' );
+
+			if ( $is_occurrence_request ) {
+				// An Occurrence has no status of its own to match against `event_status` - drop it even if
+				// the request set it explicitly (e.g. via `post_status`, mapped above), not just when it
+				// would otherwise default to one.
```
coutinhocouto:
```suggestion
				/*
				 * An Occurrence has no status of its own to match against `event_status` - drop it even if
				 * the request set it explicitly (e.g. via `post_status`, mapped above), not just when it
				 * would otherwise default to one.
				 */
```

tests/ft_integration/Tribe/Tickets/REST/V1/Endpoints/QR_Series_Pass_Test.php
```diff
@@ -0,0 +1,127 @@
...
+
+	protected string $controller_class = Attendees::class;
+
+	/**
+	 * Creates a Series Pass Attendee and a single Event, part of the Series, that is the
+	 * only checkin candidate for that Attendee.
+	 *
+	 * @return array{0: int, 1: int} The Series Pass Attendee ID and the Event ID.
+	 */
+	private function make_series_pass_attendee_with_one_occurrence(): array {
```
coutinhocouto: Both private helpers sit above the `@before` hook and every `@test`. Private members go last. Same
in `Attendee_Archive_Series_Pass_Test.php:60`.

src/Tribe/REST/V1/Endpoints/QR.php
coutinhocouto:
```php
'tribe_qr_status' => get_post_meta( $attendee_id, '_tribe_qr_status', 1 ),
```

The already-checked-in gate at line 362 now reads the clone via `$checkin_status_attendee_id`, but
this diagnostic field still reads the original. For a Series Pass Attendee `_tribe_qr_status` only
ever lives on the clone, so this is always empty in a failed check-in response for precisely the
attendees this PR is about. Worth switching to `$checkin_status_attendee_id`, and the third argument
is `1` rather than `true` while you're in there.

coutinhocouto (changes requested):

src/Tribe/REST/V1/Endpoints/Attendee_Archive.php
```diff
@@ -124,10 +127,43 @@ public function get( WP_REST_Request $request ) {
...
-			$fetch_args['event_status'] = Tribe__Utils__Array::get( $fetch_args, 'event_status', 'any' );
 			$fetch_args['order_status'] = Tribe__Utils__Array::get( $fetch_args, 'order_status', 'any' );
+
+			if ( $is_occurrence_request ) {
+				/*
+				 * An Occurrence has no status of its own to match against `event_status` - drop it even if
+				 * the request set it explicitly (e.g. via `post_status`, mapped above), not just when it
+				 * would otherwise default to one.
+				 */
+				unset( $fetch_args['event_status'] );
```
sdokus, line 169: This `unset` removes the only place the related event's `post_status` is enforced: `Attendee_Repository::filter_by_event_status` (460-489), which downgrades any to publish for API-key-only callers. Nothing in the REST repository's `build_query`, `filter_by_event`, or the Series Passes `include_series_to_fetch_attendees` re-applies a status check, and CT1 keeps occurrence rows on trash. So `GET /attendees?event=<provisional id>` with an ET API key and no WP session on a draft, pending, private or trashed recurring event now returns the cloned Series Pass attendees and the Series' own attendees with full manage data. Before, that returned nothing. Narrow (API-key holders, provisional IDs only) but it's a real widening. I get why the unset is there, the `wp_posts LEFT JOIN` on a provisional ID drops every clone, but we should make sure that we keep a door to avoid anyone with an API key from accessing a trashed event's attendees. 

Cheap fix: do the status check yourself for occurrence requests, something like this: 
```
$allowed_statuses = $has_manage_access ? [ 'publish', 'draft', 'pending', 'private' ] : [ 'publish' ];

$fetch_args['event'] = array_filter( $requested_event_ids,static fn( $event_id ) => in_array( get_post_status( Occurrence::normalize_id( (int) $event_id ) ), $allowed_statuses, true ) );

unset( $fetch_args['event_status'] );
```
Then bail with the usual empty result if nothing survives the filter.

PR #4466, opened 2026-09-16
[SMTNC-2748] PayPal orders remain stuck in Pending after successful PayPal payment capture

coutinhocouto:
### 🎫 Ticket

[SMTNC-2748](https://linear.app/nexcess/issue/SMTNC-2748/event-tickets-commerce-paypal-orders-remain-stuck-in-pending-tec-tc)

### 🎥 Artifacts

https://www.loom.com/share/5060d48fd9b64576acb094161d250fe2

### 🗒️ Description

Stops PayPal orders stranding in Pending after the buyer has approved the payment, by giving the capture an order-scoped credential to fall back on when the cart cookie is lost, serving the webhook route PayPal is already told to deliver to, and letting a lock left behind by a dead request be taken over instead of freezing the order permanently.

### ⚠️ Blockers
None

### ✔️ Checklist
- [ ] There is an OpenSpec plan for this work in `tec-plans`, and it still describes what this PR actually does.
- [x] Changelog file(s) added.
- [x] Code is covered by tests.
- [x] All the **required** tests are passing.
- [x] Artifacts added.
- [x] The linked ticket includes all the information for a QA engineer to verify the work.
- [x] Pull Request is ready to be reviewed by a Human.

### 🤖 AI Disclosure

Was AI used for this PR [**Y**/n]:

Which model(s):

In what way was AI used: 

- [x] I own the work being done in this PR and I have personally reviewed it before asking another human to.


src/Tickets/Commerce/Order.php
```diff
@@ -1366,6 +1395,103 @@ public function is_order_locked( int $order_id ): bool {
...
+	 * carry a readable timestamp, and is left alone.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $lock_id The value held in the lock column.
+	 *
+	 * @return int|null The Unix timestamp the lock was taken, or null when it cannot be read.
+	 */
+	protected function get_lock_timestamp( string $lock_id ): ?int {
+		$pattern = '/^' . preg_quote( self::LOCK_ID_PREFIX, '/' ) . '([0-9a-f]{8})[0-9a-f]{5}\./';
```
dpanta94:
```suggestion
		$pattern = '/^' . preg_quote( self::LOCK_ID_PREFIX, '/' ) . '(?<timestamp>[a-f0-9]{8})(?<rest>[a-f0-9]+)\.(?<entropy>[0-9]+)$/;
```

src/Tickets/Commerce/Order.php
```diff
@@ -1366,6 +1395,103 @@ public function is_order_locked( int $order_id ): bool {
...
+
+		if ( ! preg_match( $pattern, $lock_id, $matches ) ) {
+			return null;
+		}
+
+		return absint( hexdec( $matches[1] ) );
```
dpanta94:
```suggestion
		return absint( hexdec( $matches['timestamp'] ) );
```

src/resources/js/commerce/gateway/paypal/checkout.js
```diff
@@ -57,6 +57,18 @@ window.tribe.tickets.commerce.gateway.paypal.checkout = {};
...
+	 * Credentials issued by our Create Order endpoint, keyed by PayPal Order ID.
+	 *
+	 * They authorize the capture when the cart cookie did not survive the round trip to PayPal, so
+	 * every request that settles an order sends the one issued for it.
+	 *
+	 * @since TBD
+	 *
+	 * @type {Object}
+	 */
+	obj.orderCredentials = {};
```
dpanta94: I don't think we need a JS change. If the cart cookie is lost in the request, that's a valid reason for order operations to fail.

src/Tickets/Commerce/Gateways/PayPal/REST/Order_Endpoint.php
```diff
@@ -489,6 +498,50 @@ public function handle_recheck_order( $order_id, $order ) {
...
+	 *
+	 * @since TBD
+	 *
+	 * @param WP_REST_Request $request          The REST Request instance.
+	 * @param string          $gateway_order_id The PayPal order id.
+	 *
+	 * @return bool Whether the request carries the credential issued for this order.
+	 */
+	protected function request_carries_order_credential( WP_REST_Request $request, string $gateway_order_id ): bool {
+		$credential = $request->get_param( Order_Credential::REQUEST_PARAM );
```
dpanta94: I would go with the Pending_Order implementation instead

dpanta94 (changes requested):

changelog/smtnc-2748-paypal-orders-stuck-in-pending.yaml
```diff
@@ -0,0 +1,8 @@
+significance: patch
+type: fix
+entry: "Stopped PayPal orders stranding in Pending after a successful payment,
+  which left buyers without their attendees or ticket emails: the capture no
+  longer depends on the cart cookie surviving the round trip, PayPal's webhook
+  route is served rather than answering 404, and a lock left behind by a failed
+  request no longer freezes the order."
```
sdokus:
```suggestion
entry: "Stopped PayPal orders from getting stranding in Pending after a successful payment."
```

src/Tickets/Commerce/Gateways/PayPal/REST.php
```diff
@@ -21,8 +26,10 @@ public function register() {
 	 * Register the endpoints for handling webhooks.
 	 *
 	 * @since 5.1.6
+	 * @since TBD Serves the webhook route, which onboarding already registers with PayPal.
 	 */
 	public function register_endpoints() {
+		$this->container->make( REST\Webhook_Endpoint::class )->register();
```
sdokus, line 32: Pre-existing, but serving the route ships it: `Webhook_Endpoint::handle_request()` line 115 reads the event with `get_body_params()`. WordPress only fills that for form-encoded bodies. PayPal posts JSON, so `$event` is `[]`, `event_type` is `undefined`, and every delivery answers `500`. Swap to `$request->get_json_params()`, same as Stripe, Square, and `handle_create_order()` in this gateway. `Webhook_Route_Test` only asserts registration, so a test that posts a JSON body through the route would catch this.

src/Tickets/Commerce/Gateways/PayPal/Webhooks/Handler.php
```diff
@@ -61,9 +64,30 @@ public function process_event( $event ) {
 
 		$new_status = tribe( Events::class )->convert_to_commerce_status( $event['event_type'] );
 
-		$link = $this->get_parent_payment_link( $event['resource']['links'] );
+		$link = $this->get_parent_payment_link( Arr::get( $event, [ 'resource', 'links' ], [] ) );
```
sdokus: The `href` fix is right, but I don't think it's enough on its own. `get_parent_payment_link()` on line 30 still looks for `rel === 'parent_payment'`, which is a Payments v1 thing. The v2 `PAYMENT.CAPTURE.*` events we accept in `Events.php` come with self, refund, and up, where up points at `/v2/checkout/orders/{id}`. So `$link` is `false` and the new missing-link branch fires for every real event.

The test passes because the fixture in `Webhook_Handler_Test.php:194-209` hands it a `parent_payment` link that PayPal wouldn't send. I'd key on `up` (or `resource.supplementary_data.related_ids.order_id`), check it against a sandbox payload, and update the fixture to match what PayPal actually sends.

src/Tickets/Commerce/Gateways/PayPal/Webhooks/Handler.php
```diff
@@ -16,25 +17,85 @@
...
+
+		foreach ( [ 'up', 'parent_payment' ] as $relation ) {
+			foreach ( $links as $link ) {
+				if ( ! is_array( $link ) || $relation !== Arr::get( $link, 'rel' ) ) {
+					continue;
+				}
+
+				$href = Arr::get( $link, 'href', '' );
+
+				if ( is_string( $href ) && 0 === strpos( $href, $api_base . '/' ) ) {
```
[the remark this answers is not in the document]
sdokus: My Claude seems to agree with this one: 
> +1 to CodeRabbit here, and it's a real one. I pulled PayPal's published OpenAPI spec for Payments v2 and every sample `up` link is on `api-m.paypal.com` or `api-m.sandbox.paypal.com`, not `api.`. So this prefix check rejects the links PayPal actually sends, and the whole fallback below is dead in production. The three fixtures in the PR all use `api.paypal.com`, which is why nothing catches it. Suggest parsing the host and allowing `api`, `api-m` for the selected environment, and switching one fixture to `api-m` so it stays covered.

src/Tickets/Commerce/Gateways/PayPal/Webhooks/Handler.php
```diff
@@ -16,25 +17,85 @@
...
+			return '';
+		}
+
+		$payment = tribe( Client::class )->request( 'GET', $link );
+
+		if ( ! is_array( $payment ) || empty( $payment['id'] ) || ! is_string( $payment['id'] ) ) {
+			return '';
+		}
+
+		return $payment['id'];
```
sdokus: Should we add a check around here if this is a refund? Or some sort of check to only reach here if we confirm that the link path is `/v2/checkout/orders/`?

coutinhocouto: @sdokus nice call, fixed

src/Tickets/Commerce/Gateways/PayPal/Client.php
```diff
@@ -557,10 +558,37 @@ public function create_order( array $units = [] ) {
...
+	 * @since TBD
+	 *
+	 * @param string $operation The operation being performed, so the same subject used for different
+	 *                          calls does not share a key.
+	 * @param string $subject   What the request is about: the PayPal order id, or the payload for a
+	 *                          call that has no id yet.
+	 *
+	 * @return string The idempotency key to send PayPal.
+	 */
+	protected function get_request_id( string $operation, string $subject ): string {
```
dpanta94: Can it be private ?

src/Tickets/Commerce/Gateways/PayPal/Webhooks/Handler.php
```diff
@@ -16,25 +17,74 @@
...
 	 *
-	 * @param array $links
+	 * @since TBD
+	 *
+	 * @param array $event The PayPal payment event object.
 	 *
-	 * @return array
+	 * @return string The PayPal order id, or an empty string when it cannot be resolved.
 	 */
-	protected function get_parent_payment_link( $links ) {
```
dpanta94, line 27: Protected methods of a non final class are part of the contract. We should not just remove them :(

src/Tickets/Commerce/Gateways/PayPal/Webhooks/Handler.php
```diff
@@ -16,25 +17,74 @@
...
-		return current( array_filter( $links, static function ( $link ) {
-			return 'parent_payment' === $link['rel'];
-		} ) );
+	protected function get_gateway_order_id( array $event ): string {
```
dpanta94: Can it be private ?

src/Tickets/Commerce/Gateways/PayPal/Webhooks/Handler.php
```diff
@@ -16,25 +17,74 @@
...
+	 * host is pinned for the same reason a fetch would need it -- an event is not a trusted source of
+	 * urls -- even though nothing is requested here.
+	 *
+	 * @since TBD
+	 *
+	 * @param array $links The links on the event's resource.
+	 *
+	 * @return string The order id, or an empty string when no link addresses an order.
+	 */
+	protected function get_order_id_from_links( array $links ): string {
```
dpanta94: Similar

sdokus:
Don't hate me Filipe, I have one more request: 
> The `500` status at `Handler.php:126/145/164` is hit by every normal sale: the buyer's browser completes the order, then the webhook arrives, sees "already Completed", and gets a `500`. PayPal retries that up to 25 times over 3 days, each costing a signature verification call to PayPal. That is 25 error-log entries and 25 outbound calls per sale, starting the day this ships. Fix is adding `[ 'status' => 200 ]` style data (or returning a success response) on the "already updated" and "not found" branches.

src/Tickets/Commerce/Order.php
```diff
@@ -1366,6 +1411,112 @@ public function is_order_locked( int $order_id ): bool {
...
+	 * Unlike its neighbours this carries no try/catch of its own: it runs inside lock_order()'s, so a
+	 * query exception there already answers the acquisition with false.
+	 *
+	 * @since TBD
+	 *
+	 * @param int $order_id The order ID.
+	 *
+	 * @return bool Whether the stale lock was taken over.
+	 */
+	protected function reclaim_stale_lock( int $order_id ): bool {
```
dpanta94: Can it be private por favore ?

src/Tickets/Commerce/Order.php
```diff
@@ -1366,6 +1411,112 @@ public function is_order_locked( int $order_id ): bool {
...
+	 * date one. Anything not matching that shape is not a lock this plugin wrote, or is too old to
+	 * carry a readable timestamp, and is left alone.
+	 *
+	 * @since TBD
+	 *
+	 * @param string $lock_id The value held in the lock column.
+	 *
+	 * @return int|null The Unix timestamp the lock was taken, or null when it cannot be read.
+	 */
+	protected function get_lock_timestamp( string $lock_id ): ?int {
```
dpanta94: Can it be private please ?

dpanta94 (changes requested):

dpanta94:
I think the below part from coderabbit's review makes sense.

```php
 			return new WP_Error( 'tec-tickets-commerce-paypal-webhook-order-status-already-updated', null, [
				'status'           => 200,
 				'gateway_order_id' => $gateway_order_id,
 				'order'          => $order,
```

src/Tickets/Commerce/Order.php
```diff
@@ -1259,18 +1297,25 @@ public function lock_order( int $order_id ): bool {
...
-			$result = (bool) DB::query(
+			/*
+			 * Matched on the lock id this request holds. Without that match any caller could clear any
+			 * lock, including one another request had just taken over from a stale holder.
+			 */
+			$result = 0 < DB::query(
 				DB::prepare(
-					"UPDATE %i set $lock_key = '' where ID = $order_id",
-					DB::prefix( 'posts' )
+					"UPDATE %i SET $lock_key = '' WHERE ID = %d AND $lock_key = %s",
```
dpanta94, line 1315: I dont understand why the unlock would need the current lock key. It is then only possible to unlock an order with this method only during the same request's lifecycle.

src/Tickets/Commerce/Order.php
```diff
@@ -66,6 +66,30 @@ class Order extends Abstract_Order {
...
+	 * Prefix every generated lock id carries.
+	 *
+	 * The lock column is a general purpose WordPress column, so this is what tells a lock this plugin
+	 * wrote apart from anything else that may be stored there.
+	 *
+	 * @since TBD
+	 *
+	 * @var string
+	 */
+	protected const LOCK_ID_PREFIX = '_order_lock';
```
dpanta94, line 79: private

src/Tickets/Commerce/Order.php
```diff
@@ -66,6 +66,30 @@ class Order extends Abstract_Order {
...
+	 * How long a lock is honoured before another request may take it over, in seconds.
+	 *
+	 * A lock is held for one order mutation, which is the work of a single request, so anything still
+	 * held after this belongs to a request that died before it could release it.
+	 *
+	 * @since TBD
+	 *
+	 * @var int
+	 */
+	protected const LOCK_TTL = 15 * MINUTE_IN_SECONDS;
```
dpanta94, line 91: Similar

src/Tickets/Commerce/Order.php
```diff
@@ -1366,6 +1411,112 @@ public function is_order_locked( int $order_id ): bool {
...
+	 * Unlike its neighbours this carries no try/catch of its own: it runs inside lock_order()'s, so a
+	 * query exception there already answers the acquisition with false.
+	 *
+	 * @since TBD
+	 *
+	 * @param int $order_id The order ID.
+	 *
+	 * @return bool Whether the stale lock was taken over.
+	 */
+	private function reclaim_stale_lock( int $order_id ): bool {
```
dpanta94, line 1435: This method is good. I like how you give the option to reclaim a lock which has been never unlocked.

dpanta94 (changes requested):

src/Tickets/Commerce/Order.php
coutinhocouto: This needs to be implement, because if not the reclaim_stale_lock() will not work property

